Issue #167, opened 2015-02-20
Custom GSL error handler


Lestropie:
This is required for dwi2response, such that if the matrix for a particular voxel is not positive definite, it is simply excluded from the SF mask, rather than crashing the program.

jdtournier:
?? But there is already one, defined in lib/app.cpp, line 40, set in App:: init (), line 514. Is that not sufficient...?

Lestropie:
OK, so it is there. Was still getting an error in dwi2response because it wasn't the CSD step (that lives in a try{} clause) that was crashing, but the SH::Transform constructor with a dodgy acquisition scheme.

However, looks like the error handler isn't performing as it should. Even though `mrtrix_gsl_error_handler()` is called, the terminal only gives this:
```
terminate called after throwing an instance of 'MR::Exception'
terminate called recursively
Aborted (core dumped)
```

jdtournier:
OK, so the GSL handler isn't the issue since it clearly states that an `MR::Exception ` is being thrown. That should be handled at the very least at the top level in `main ()`, which wraps everything up in a try / catch clause. My best guess is that whatever is catching the exception is itself throwing an exception that somehow leads to a recursive invocation of the _same_ handler. Hard to see how that could happen, unless the exception handler (including all the cleanup code involved in unwrapping the stack, i.e. class destructors, etc) itself raises a GSL error...?

Lestropie:
Yeah, it's a weird one. Going to have to produce a MWE to get at this one I think.

jdtournier:
Just did a quick search for _terminate called after throwing an instance of  terminate called recursively Aborted_ - seems there are lots of reports of broken exception handling on various versions of g++, some of them quite recent. Might be worth trying this on a different compiler / distro...?

jdtournier:
Another thing you could try is to print out the Exception's `msg` in its constructor, so you can track exactly when each being created. Currently the error message is only shown by the global catch() call or some other catch() handler, not as it is thrown...

Lestropie:
It looks like the 'terminate called recursively' is just something detecting multiple instances of the same exception; which in this case is multiple processing threads throwing exceptions for the same reason. Set `-nthreads 1` and that part of the error goes away. So it's not actually 'recursive'.

[...]

PR #180, opened 2015-03-12
C++11 thread exception handling

jdtournier:
Changes to handle exceptions thrown from threads more gracefully, as discussed in issue #167. Are we happy to merge this branch...?


Lestropie:
As long as there's no performance hit, I don't have a problem with it.

jdtournier:
No, any performance hit would potentially be at thread creation and rejoin time, so I don't expect we would ever notice even if it was a big performance hit (which is unlikely)... I'll merge.

Issue #168, opened 2015-02-23
Config option to bypass VT100 codes


Lestropie:
In the terminal emulator provided with Git on Windows, was getting weird codes appearing in the terminal, probably either VT100 codes or colour codes. Should add a config file option to use 'plain' text and just '\r' for carriage return.

jdtournier:
There is one: TerminalColour - see `lib/exception.cpp`, line 60 or so. Even documented it, for once...

It still allows the line clear code though, in addition to the carriage return. Hopefully that won't cause any issues...?

Lestropie:
Setting TerminalColor to false removes most of them, but not all: there's still a '<-[2K' at the end of progress bars.

I'm thinking of changing the 'TerminalColor' conf option to something like 'SimpleTerminal', and using it to disable both terminal colours and line clears.

jdtournier:
Crap. The problem with that is that when using the ProgressBar update() or set_text() methods, you get dangling characters at the end if the string becomes shorter.

A better approach would be to figure out the correct codes for that terminal. Maybe `tput` can help us to get this right...?

Lestropie:
Well the dangling strings would be no worse than what we could have had previously without the VT100 code. Alternatively we could store the number of characters in the current message, and if SimpleTerminal is enabled and the new message is shorter than the old one, append with spaces.

jdtournier:
Strange... I just tried this on my Windows VM (Windows 7 64 bit), and I get the same as you. However, everything seems to work OK when using these codes directly on the command-line, e.g.:

```ShellSession
$ echo -e "first line to be cleared\r\033[2Knext line"
next line
```

Or the approach in the current `log_to_file` branch (clear from current position, rather than clear whole line):

```ShellSession
$ echo -e "first line to be cleared\rnext line\033[0K"
next line
```

But then if I try this in Python:
```ShellSession
$ python
>>> print 'first line to be cleared\r\033[2Knext line'
←[2Knext line to be cleared
```

Clearly bash does its own translation... 
I'll dig a little deeper and see whether we can figure something out...

jdtournier:
OK, turns out just adding `\015` in front of the `\033[0K` or `\033[2K` codes works for the ProgressBar. Still can't get the colour codes to work. Struggling to even get the config file to load in at the moment...

jdtournier:
[...]

Issue #161, opened 2015-02-09
'Quieter' command line output


davesl:
Hi,

I’ve noticed that the output logs on HPC systems can get quite long during tckgen and tcksift. One of these output logs reached almost 200MB! This is because everything that would normally be printed to the command line is saved in the log.
 
Would it be possible to have an option in addition to -quiet which would only print an update when, say, each percentile is reached? Then only 100 lines would need to be logged instead of some millions!

Cheers,
David

jdtournier:
Yes, I've noticed that output files will contain a line for every update of the progress status. This is normally OK for percentage-only output, since the code will only update the status when the percentage value changes. Still, it's probably annoying to see 100 lines of updates...

One idea might be to try to detect whether the command's stderr is connected to a terminal, and modify behaviour accordingly - or just rely on a command-line switch instead. Either way, I'm not convinced that's the best way to deal with this.

What I don't like about this approach is that by preventing status updates at the normal frequency, or worse until processing is done (which would keep your log files nice and lean), you might lose important information as to when exactly a crash or error occurred. It's often quite useful to know how far processing got before crashing out.

Maybe a better approach would be to provide a `-log` option, so the user could specify the name of the log file rather than having to redirect stderr, and we would then have the ability to *seek* within the log file and update the right bits with the expected frequency, while keeping the log file as clean as the output would appear on the command line. 

This would actually be relatively straightforward to implement, by providing alternative handlers for the report_to_user() functions. Only trick is to catch the `-log` option early enough that any other errors would not have occurred *before* these functions have been set... Basically it should be the very first option parsed.

Lestropie:
Keep in mind that `tckgen` and `tcksift` don't use `report_to_user_func()`; currently they're direct `fprintf(stderr, ...)` calls, due to requiring the line clear & carriage return characters as well as no automatic newline character.

We could wrap these into a sort of 'manual formatting' progressbar class, then have a command-line option to constrain the outputs. Or probably a simpler option would be to have the default time interval for the IntervalTimer class as a config file option; HPC users could set it to e.g. 1 second or more instead of 1/30th of a second.

None of these will solve the issue with `tcksift` though: in that case we don't actually know for sure when the algorithm will finish, so can't necessarily calculate a percentage, nor are the updates based on a timing interval - it's a line per iteration. But I suspect that this problem would be far worse for `tckgen` than `tcksift`...?

jdtournier:
[...]

PR #181, opened 2015-03-12
Log to file

jdtournier:
This attempts to detect when the output is redirected to file and adjusts the output accordingly. See discussion around issues #161 & #168. 

Are we happy to merge this in...?


Lestropie:
On Windows I'm getting a leading `<-[0K` on progress bars as they're working, and a residual one on its own line after the progress bar is done before the terminal prompt appears.

In `tckgen` the `<-[0K` overlaps the command name at the start of the line, i.e. it reads:
`<-[0Ken: [ xx%]      xxxxx generated,      xxxxx selected`

jdtournier:
@Lestropie : are you sure that was running with the `log_to_file` branch and the latest commit (ef6d06f)? I can't see any way you'd see that output if that was the case - if it wasn't working you'd see a _trailing_  `<-[0K`...

jdtournier:
OK, I was wrong. I get the same as you...
I'll simplify the windows case to a simple carriage return, and pad out the string with spaces if the `update()` shrinks it...

jdtournier:
OK, see what you make of the latest changes...

Lestropie:
Looks correct to me on both Linux and Windows.

Issue #122, opened 2014-09-20
Robust method for default lmax


jdtournier:
Using the number of entries in the DW encoding to determine the SH order is not robust when directions are duplicated. A better approach is to start from this order, compute the condition number, and if it's large, reduce lmax and repeat.

jdtournier:
Merged, closing.

Issue #178, opened 2015-03-08
mrconvert -bvalue_scaling


bjeurissen:
I am confused with the bvalue_scaling option that comes with mrconvert, e.g. as in:
mrconvert -bvalue_scaling true DICOM dwi.mif

This does not result in a dwi.mif file with scaled b-values.

I vaguely remember that the idea of the bvalue_scaling parameter was to not touch the data, but to use it to calculate the actual bvalues to be used in the processing. However, for a command like mrconvert, I kind of expect it to generate a mif file with scaled b-values.

So I think, either

1) the b-values should be scaled in the resulting mif file

2) the option should be removed from the mrconvert command

Where I would strongly prefer option 1) as option 2) makes processing multi-shell data a pain (i.e. having to remember to set the bvalue_scaling command for all DWI processing commands, instead of just once when importing from dicom).

bjeurissen:
E.g. the following patch would result in option 1):

```
diff --git a/cmd/mrconvert.cpp b/cmd/mrconvert.cpp
old mode 100644
new mode 100755
index 5d9c07f..bd8e77b
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -101,7 +101,9 @@ inline std::vector<int> set_header (Image::Header& header, const InfoType& input
   header.intensity_offset() = 0.0;
   header.intensity_scale() = 1.0;

-  if (get_options ("grad").size() || get_options ("fslgrad").size())
+  if (get_options ("bvalue_scaling").size())
+    header.DW_scheme() = DWI::get_valid_DW_scheme<float> (header);
+  else if (get_options ("grad").size() || get_options ("fslgrad").size())
     header.DW_scheme() = DWI::get_DW_scheme<float> (header);

   Options opt = get_options ("axes");
````

What do you guys think?

jdtournier:
[...]

PR #182, opened 2015-03-13
functions for robust lmax detection

jdtournier:
As per issue #122, this adds a function to determine the condition number of a matrix, and uses that to ensure the matrix mapping SH -> amplitudes is sufficiently well-conditioned for use. More specifically:

- new `Math::cond()` function taking a `Math::Matrix` as the sole argument

- overhaul of the `DWI::get_SH2amp_mapping()` functions, splitting it up into distinct parts to make the various bits available in different workflows:

     - `DWI::get_gradient_table_and_directions()` finds the DW gradient table (using `get_valid_DW_scheme()`), identifies the relevant set of DWIs (using the `DWI::Shells` interface) and b-zeros, and converts the directions to spherical coordinates, ready for use in `SH::init_transform()`. This processes the gradient import and shell selection options.

    - `DWI::compute_SH2amp_mapping()` takes the matrix of directions in spherical coordinates and returns the SH->amp matrix, up to the given _lmax_ (which can be overridden on the command-line). If the condition number is too high (default 2), it'll try a lower lmax - unless overridden on the command-line.

    - the `DWI::get_SH2amp_mapping()` functions basically wrap both of these steps into a single call.

So far, I've only added this to `amp2sh`, obviously it'll need to be added to all relevant commands. Just thought I'd get some early feedback before merging to master...


jdtournier:
OK, simplified a few things, the `DWI::get_SH2amp_mapping()` functions were probably overkill. Also changed the `GradImportOptions` interface to allow skipping the `-bvalue_scaling` option, required for `mrconvert`, as per issue #178.

I'll go ahead and merge this tomorrow - I have a feeling I'm not going to get any feedback on this one unless I foist it onto everyone else...

Lestropie:
Not sure if the `cond < 2` heuristic is adequate.

From VicTeens data w. dodgey 45-direction scheme, good data:
```
b = 3000 [ 45 directions ]

    nearest-neighbour angles: mean = 21.9693, range [ 21.6738 - 22.2778 ]
    energy: total = 4225.16, mean = 93.8924, range [ 91.9293 - 95.8401 ]

    nearest-neighbour angles: mean = 21.9693, range [ 21.6738 - 22.2778 ]
    energy: total = 2881.58, mean = 64.0351, range [ 37.2916 - 78.7913 ]
    condition numbers for lmax = 2 -> 8: 1.03488 1.13211 1.30392 3.68967
```

Same cohort, bad data:
```
b = 3000 [ 45 directions ]

    nearest-neighbour angles: mean = 13.0479, range [ 3.48923 - 24.8669 ]
    energy: total = 14238.8, mean = 316.417, range [ 59.9398 - 963.412 ]

    nearest-neighbour angles: mean = 14.0363, range [ 3.48923 - 47.6115 ]
    energy: total = 9870.4, mean = 219.342, range [ 23.3363 - 667.919 ]
    condition numbers for lmax = 2 -> 8: 3.04532 6.04692 152.772 1.10812e+06
```

The first works otherwise at lmax 8 (cond = 3.7), whereas the second seems to play ball if you reduce lmax to 6 (cond = 153).

Would it be better to simply attempt to compute the inverse in a try{}/catch{} clause, and reduce lmax until it succeeds?

jdtournier:
OK, I wasn't sure what value of the condition number would be appropriate, but it definitely sounds like 2 might be a bit too harsh... 

As to the bad DW scheme, I'd say cond=153 is still pretty bad... We're really relying on the non-negativity constraint to help out here - and the fact that the response is averaged over a lot of voxels. I'd personally prefer that the default behaviour reduces the lmax to 4 on that one (which is still pretty poor at cond=6), which at least warns the user that the DW scheme is pretty rotten. They can still force the lmax to whatever value they want - it'll only issue a warning in that case, but it will allow it. 

So my preference would be to increase the threshold to ~10. Relying on outright failure of the Cholesky decomposition is basically a lottery, since this will be dictated by floating-point errors. And allowing very large condition numbers means we'll allow potentially very poor results by default, when we now have the ability to predict from the condition number that they'll be terrible...

Lestropie:
Cool. You're the pro here, I can only comment on what has and hasn't worked with the dodgey data, and it seems to happily map to lmax=6 under any rotation (straight A2SH, no non-negativity). Is it worth having two thresholds - one for printing a warning (> 2) and another for triggering a decrease in lmax (> 10)?

jdtournier:
> Is it worth having two thresholds - one for printing a warning (> 2) and another for triggering a decrease in lmax (> 10)?

Sounds like a good plan. Although I'd say 10 for a warning, and maybe 100 for a decrease in lmax, given the data you show and what you've observed. 

To be honest, I'm surprised that a straight amp2sh on data with a condition number of 150 wouldn't look horrible - it implies a noise amplification for at least some of the coefficients of the order of 150, which ought to produce ugly results assuming your SNR is lower than that... Maybe I just need to go revise on condition numbers... 

In the meantime, I think the best thing to do is adjust based on empirical evidence of what works and what doesn't. The 45 DW scheme should be well-conditioned for lmax=8, yet the condition number is ~4. This isn't unexpected - it's impossible to produce a perfectly conditioned set of directions, and 45 is the minimum reqjuired for lmax=8. In any case, that should definitely be allowed, which is why I think thresholds of 10 & 100 might be more appropriate.

bjeurissen:
Could it be useful tot use the rank instead and check if rank(A) >= nparams(lmax)?

jdtournier:
@bjeuriss - can you check whether you're happy with the latest changes in `mrinfo` - I know it shouldn't really be part of this branch, but hey...

About using rank(A): really in essence the condition number fulfils the same role, with fewer ambiguities. The SH transform should have full column rank, since it maps nparam(lmax) SH coefficients onto some larger number of DW signals. If it didn't have full rank - i.e. if rank(A)  < nparams(lmax) - its condition number would explode to infinity: there would be at least one completely undetermined direction. In other words, one combination of SH coefficients that would have no influence on the output DW directions whatsoever, implying at least one zero eigenvalue. This is captured by the condition number, since it is defined as the ratio of the maximum to minimum eigenvalue. 

The advantage of the condition number is it tells us how bad that degeneracy is. The rank would only tell us if that minimum eigenvalue fell below machine precision - i.e. it falls below some threshold that we need to define ourselves. The problem is, a full rank transform with a huge condition number really should be considered unusable. I think we need to take it for granted that the SH matrix has full rank, and look at the condition number to verify how well that assumption is being met. 

Where the rank might be useful is as a safeguard against multiple repeats of the same set of directions. In this case, the rank would tell us directly how many unique directions were actually acquired. But as soon as these directions are not identical, this is no longer reliable - including for example correcting the b-vectors following motion correction. This would actually suggest full rank (or close to it), when in reality the chances are this would lead to a very unstable fit. The condition number would in this case provide a better indication of how well that DW scheme support a given lmax. 

My 2 cents...

jdtournier:
I'll merge this now - we won't start using this functionality until it's merged...

@Lestropie - you might want to try using this for `dwi2response` - I had a look, but I don't want to mess things up, it's your baby...

PR #185, opened 2015-03-23
DWI::Shells: bzero_threshold no longer a global variable

jdtournier:
The bzero_threshold variable used in the DWI::Shells class was declared as a global variable, and initialised from File::Config. Problem is that initialisation for a global variable happens before entering main(), and hence before Config::init() (in App::parse()). So there is no way that the Config::get() call could return anything other than the default supplied, since the Config would appear empty at that point...

Solution proposed here is to  return a static member variable in a function, which defers initialisation until the first time it's called. Hopefully that function will never be called before Config::init()...

### DIFF
--- a/src/dwi/shells.cpp
+++ b/src/dwi/shells.cpp
@@ -1,7 +1,6 @@
 
 #include "dwi/shells.h"
 
-#include "file/config.h"
 
 
 namespace MR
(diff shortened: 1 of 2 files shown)


jdtournier:
@Lestropie : wouldn't mind your OK on this one before I merge,,,

Lestropie:
Huh. Hadn't thought of that. Looks right to me.

jdtournier:
Yes, I learned the hard way on that one myself... I'll go merge.

Issue #196, opened 2015-03-27
mrconvert: Segfault


Lestropie:
Running with `-zero` and `axes 0,1,2` together in the one call on a 4D image segfaults.

On a related note: Never been a fan of the code branching in mrconvert. Could the `-zero` option be handled in `mrcalc` instead?

jdtournier:
Can't reproduce. Are you sure this is a clean build...?

And I know what you mean re the code branching, it's not pretty. I guess I wouldn't mind shunting it over to `mrcalc`, it admittedly is a better place for it...

Lestropie:
Not really sure. Was on our HPCs, have had weird installation issues there before. Either option alone was happy, but run them together and unhappiness ensued. Really more as a self reminder; should have self-assigned...

Have wanted that functionality out of `mrconvert` for a while, just never been 100% sure how to do it. Could have a `-nan2zero` operator in `mrcalc`, avoid having to do a `-finite` then `-mult`? (I;m guessing the explicit option is there currently because it's used a lot?) Or more generally, a `-replace` ternary operator that does C->A if A==B?

jdtournier:
OK, I won't worry about it then.

I don't think the option is actually used much, if at all... There was some reason I included it in the distant past (it's in the 0.2.x series), but to be honest I doubt anyone will miss it. Happy to just remove it. 

And the syntax for `mrcalc` is fine as-is. As long as people can figure out how to do it, I don't think we need to do anything about it. But it does remind me that we probably need to add `-dup` and `-exch` options in `mrcalc` to duplicate and exchange stack entries respectively. This would avoid the need to load the image twice in the call, i.e. the syntax would change from:

```ShellSession
$ mrcalc in.mif in.mif -finite -mult out.mif
```

to: 

```ShellSession
$ mrcalc in.mif -dup -finite -mult out.mif
```

Would also be useful to if statement, etc. But then, I have a strong suspicion people will probably just supply the argument twice, without looking at this possibility. So maybe that would just be a waste of time. Might be better to try to detect when the same image is being loaded twice...?

Lestropie:
Might have been added because back then there were separate commands for e.g. `mradd`, `mrmult`, no `mrcalc` ... If nothing is jogging your memory, I'd also be tempted to remove `-zero` from `mrconvert`. Potential advantage of reducing the code branching in `mrconvert` is doing the conversion in the output image datatype, rather than converting everything that goes through `mrconvert` to complex float.

[...]

PR #200, opened 2015-03-31
mrconvert: Operate using native data type

### LINKED ISSUE #3: Template buffer types in mrconvert copy
Currently `mrconvert` uses cfloat internally to perform the copy from input to output images. This may result in a loss of precision in some rare use cases. It should be possible to either template the buffer creation, or simply code branch, based on the output data type.

jdtournier:
mrconvert now performs copy operation using the data type of the output image.
On a straight copy of a floating-point image, this provides a speedup of about 30%.
It also means that image values will not lose precision due to being cast to a complex float during copy.
Discussed in #196.
Closes #3.

### DIFF
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -138,18 +138,33 @@ inline void copy_permute (Image::Header& header_in, Image::Header& header_out, c
   typedef Image::Buffer<T> buffer_type;
   typedef typename buffer_type::voxel_type voxel_type;
   typedef Image::Adapter::Extract<voxel_type> extract_type;
-  typedef Image::Adapter::PermuteAxes<extract_type> permute_type;
 
   buffer_type buffer_in (header_in);
   voxel_type in = buffer_in.voxel();
-  extract_type extract (in, pos);
-  const std::vector<int> axes = set_header (header_out, extract);
-  buffer_type buffer_out (output_filename, header_out);
-  voxel_type out = buffer_out.voxel();
-  DWI::export_grad_commandline (buffer_out);
 
-  permute_type perm (extract, axes);
-  Image::threaded_copy_with_progress (perm, out, 2);
+  if (pos.empty()) {
+
+    const std::vector<int> axes = set_header (header_out, in);
+    buffer_type buffer_out (output_filename, header_out);
+    voxel_type out = buffer_out.voxel();
+    DWI::export_grad_commandline (buffer_out);
+
+    Image::Adapter::PermuteAxes<voxel_type> perm (in, axes);
+    Image::threaded_copy_with_progress (perm, out, 2);
+
+  } else {
+
+    extract_type extract (in, pos);
+    const std::vector<int> axes = set_header (header_out, extract);
+    buffer_type buffer_out (output_filename, header_out);
+    voxel_type out = buffer_out.voxel();
+    DWI::export_grad_commandline (buffer_out);
+
+    Image::Adapter::PermuteAxes<extract_type> perm (extract, axes);
+    Image::threaded_copy_with_progress (perm, out, 2);
+
+  }
+
 }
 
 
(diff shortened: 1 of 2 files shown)


jdtournier:
You masochist...

Couldn't resist either... On `master`:

```ShellSession
$ touch cmd/mrconvert.cpp 
$ time ./build 
[CC] cmd/mrconvert.o
[LB] bin/mrconvert

real	0m7.273s
user	0m7.153s
sys	0m0.223s
```

On `mrconvert_template`:

```ShellSession
$ touch cmd/mrconvert.cpp 
$ time ./build 
[CC] cmd/mrconvert.o
[LB] bin/mrconvert

real	0m17.847s
user	0m17.827s
sys	0m0.330s
```

Looks like the compiler's getting a good workout then... ;)

Kidding aside, just one question: I notice you no longer branch with a direct copy when `-coord` is _not_ supplied. Might this not cause a (minor) slowdown for a simple conversion? Would be trivial to handle within the `copy_permute()` call the way you're refactored it, simply by checking for an empty `pos` vector in that case. I can see that the `PermuteAxes` adapter probably won't affect performance much, if at all - the compiler should be able to optimise a lot of the additional indexing away. But I doubt it would be able to do so for the `Extract` adapter...

Lestropie:
Yeah, I tend to do these things by changing whatever's necessary to remove the code branching, then adding the templating; guess I forgot to look again to see if the extract could be bypassed. I'll have another play tomorrow.

Lestropie:
Hmm, compile time seems to have further doubled. Wonder why...

jdtournier:
Hehe... That's why I wasn't all that keen on the idea initially, having witnessed the compile times for `tckmap`... ;)

Maybe we can reduce this by lumping all the integer operations into a `int32` version (signed and unsigned), all the floats into `float64`, and all the complex types into `cdouble`? That should cut down the compile time at least in half...

Lestropie:
But, but, ...

jdtournier:
Hehe... yes, sorry about your bubble...

By the way, the rationale for the extensive template action in `tckmap` is quite different from what it is here. In `tckmap`, we really need to keep overall RAM usage to a minimum for the requested data type. In mrconvert, this is only used to hold a temporary buffer for a single row of data per thread... No RAM limitations whatsoever. What the templating does is avoid unnecessary float->int conversions, and operating on complex types for no reason...

Actually, about that last point, we should verify that operating on a real input image to generate a complex output still works...

Lestropie:
Yeah I know. Do anything for a giggle. bool, 8 and 16 bits can go. I'd be tempted to keep both (u)int32 and (u)int64 though, just so that most integer ops work in 32 bit but the 64-bit version is there just in case someone's testing the limits. I'd imagine float<->double conversions would cost bugger all. But I'd rather test to see if there's any speed penalty first.

Type selection is based on the output image. Though I would have thought that kind of operation would be handled using `mrcalc`...?

jdtournier:
> Type selection is based on the output image. Though I would have thought that kind of operation would be handled using `mrcalc`...?

Yes, scratch that, getting myself in a muddle up.

jdtournier:
@Lestropie - how are you getting on with this one? You want to leave it as-is, or try to reduce the compile times as we discussed...?

Lestropie:
Was just talking about it with Thijs before I left work. Will cut some templates out.

My understanding though was that even 64-bit processors tend to prefer operating on 32-bit integers rather than 64? I'll speed test everything anyways.

jdtournier:
> My understanding though was that even 64-bit processors tend to prefer operating on 32-bit integers rather than 64?

Not sure - it sounds unnecessary for the processors to duplicate their integer pipelines for this. Where it might make a difference in purely in terms of the amount of RAM used to store each row of values. Doubling the size needed might mean if exceeds the CPU L1 cache, which could potentially slow it down. But frankly, I doubt you'd notice the difference... 

> I'll speed test everything anyways

Good idea.

jdtournier:
@Lestropie - I'm thinking we should merge this now, and leave cutting back on the number of template types for later. I'm keen to make sure this can be merged cleanly, which means not leaving it to linger too long. And also there's #211 looming...

Lestropie:
OK, finally got around to testing this.

* Boolean, int8 and int16 actually fractionally faster if anything when run as `copy_permute<uint32_t>`.

* About a 15% slowdown if copying a 32-bit image using `copy_permute<int64_t>`, so kept both 32-bit and 64-bit versions.

* Didn't see any slowdown copying to a float image using `copy_permute<double>`. So removed `copy_permute<(c)float>`.

Compile time back down to ~20s on my system, which I think is acceptable.

jdtournier:
> About a 15% slowdown if copying a 32-bit image using copy_permute<int64_t>, so kept both 32-bit and 64-bit versions.

Not sure I'd be that bothered about 15%... But I guess removing the 32 bit versions isn't going to make all that much difference to the compile times now. Let's merge.

Issue #208, opened 2015-04-09
mgh/mgz formats do not handle stride changes properly


jdtournier:
Something else that popped up out of the testing framework...

```ShellSession
mrconvert in.mif -stride 3,2,1 out.mgz
testing_diff_data out.mgz in.mif
````

reports a non-match. Same for `.mgh`.

Issue #207, opened 2015-04-09
compressed mif format not working


jdtournier:
Something that popped up while I was setting up the testing infrastructure...

Conversion to `.mif.gz` gives different values from input...

Lestropie:
Seems to be fixed by 24c0a61...?

Lestropie:
OK, so it's maybe kinda fixed but not entirely.
Looks like I'm relying on a hack of re-generating a text header after having read the header info into the Image::Header class, in order to produce a GZ lead-in. With the changes in #111, the input and output headers may be different lengths... let's see if this works.

jdtournier:
OK, didn't see that one coming...

@Lestropie - Thanks for figuring this out. Not sure I like the approach though, it involves ignoring the version information on read, which means you don't see it in mrinfo... I've had a go at what is probably a cleaner approach: basically set the version info as generic header fields, which means the previous code (i.e. before the changes in #111) would simply write it out as-is without modification. It also gets read as a normal uninterpreted field, so will show up in mrinfo. To update the information, the Image::Header::create() call updates those fields with the current information - images are written out with the current version info, images read in retain the version information they were created with. 

I've also used the same approach for track files (and track_scalar files), but in this case it's the Properties that need to be updated before write-out. Haven't tested that bit, but I guess it should work fine.

I've created a new branch called `testing_alternate` with these changes. I've also cherry-picked your fixes for the MGH/MGZ formats (glad you could incorporate the NIfTI handling in there so easily, I was expecting it to get messy...). 

Let me know if you're happy with that.

Lestropie:
Figured most commands wouldn't need access to it, but yeah didn't think of `mrinfo`. Generic header field is better then.

jdtournier:
Cool. I'll pull the previous merge request, and merge the `testing_alternate` one to master instead. These fixes are important enough that we want them in master now, and the testing stuff really shouldn't interfere with anything anyway. We can pad it out with further tests as we go.

jdtournier:
closed with 1050df8

PR #209, opened 2015-04-09
Testing

jdtournier:
Having a go at setting up some form of automated testing to check for regression. Basically, this adds a `testing` folder, into which scripts can be added that will be run (actually, sourced) by the `run_tests.sh` script. There is a `data` folder within the `testing` folder to hold any precomputed images needed for testing - these can be generated using the new `testing_gen_data` app (please keep these test data small!). 

I've added a couple of scripts to test `mrconvert` and `mrcalc`, and these have already thrown up a couple of issues (see #207 & #208)... Feel free to add more tests.

Remaining issue is how to test stochastic apps like 'tckgen'. I'll probably add a check for an environment variable from which to set the seed for the random number generator, and run the command without multi-threading. This is not ideal though, since a lot of issues will be related to multi-threading. If anyone has any thoughts about how to proceed, please voice them...


draffelt:
Awesome...we are all grown ups now.

I'm pretty happy about this. I've come across a few cases recently where changes to back end code have broken commands. I'll add some tests as well.

jdtournier:
Merged with 1cf62e9

PR #213, opened 2015-04-13
Replace MRtrix smart pointers with C++11 versions

jdtournier:
This branch replaces all the smart pointers in MRtrix3 (Ptr, RefPtr, VecPtr) with generic C++11 variants. Briefly:

- RefPtr => std::shared_ptr
- Ptr => std::unique_ptr - when not used in a copy-constructible class, i.e. for multi-threading)
- Ptr => MR::copy_ptr - when used in a copy-constructible class.
- VecPtr => std::vector<std::unique_ptr>

The new MR::copy_ptr inherits from std::unique_ptr, and overrides its copy constructor and assignment to perform a deep copy of the pointed-to object. When used as a class member, this mean the class's default copy constructor can be used as-is when the intention is for each instance of the class to have its own copy of the pointed-to object (the std::shared_ptr should be used when the intention is to have all classes share the pointed-to object). This is because std::unique_ptr does not allow copy-construction. 

These changes required a few changes where these are used, namely:

- assigning a raw pointer to a smart pointer can no longer be done using operator=, this needs to be replaced with the reset() method, assigning from ~~std::make_unique() /~~ std::make_shared(), or assigning from another smart pointer.
- these pointers don't allow implicit casting to a raw pointer of the same type. When the raw pointer is needed, you need to use the .get() method. 

Hopefully I won't have broken too many things in the process, but it might be a good idea for people to give it the once-over, and maybe also add a few tests to verify that things still work as expected... :)

[EDIT: std::make_unique was not included in C++11, for some unknown reason. It is part of C++14, but given how many issues C++11 is already causing for older systems, switching to C++14 would not be a good idea]


rtabbara:
@jdtournier After this change I had to make a few small changes to get g++ to stop complaining so please take a look to see if it's ok.

jdtournier:
I just had a look. Changing `const` to `constexpr` is great - something I'll probably need to do more systematically throughout the code. I'm surprised you had to change the sort functions though, that should be correct C++11. Could be incomplete C++11 support in the compiler, I suppose... In any case, these lambdas look pretty ugly, we should probably replace with a ptr_comp template function - there might even be one in the standard. Otherwise it's trivial to write, that's how it was done in the previous version...

jdtournier:
OK, cleaned up the sort functions a bit... Seems to work OK. Let me know if these changes still cause problems for gcc 4.8.2...

rtabbara:
Builds fine for me. Thanks!

Issue #194, opened 2015-03-26
MRView: remove single ODF display from ODF tool?


jdtournier:
I'm starting to wonder whether we need the little display of the ODF in the tool itself. When using it recently, I've always gone straight for the overlay... Does anyone think this is still useful? I rarely use it, and if I genuinely did want to inspect my ODFs close up, there's nothing to stop me zooming in anyway...

The reason I ask is because it does take up a lot of space, and we could simplify the ODF tool significantly by simply getting rid of it. It's also historically been a bit of a pain having two OpenGL viewports - used to cause glitches with older drivers, although I guess this is probably no longer relevant. It also messes with the delayed update of the main window, although we did implement a workaround.

If we did this, we could make the tool look more like the current overlay, tractography, and ROI tools, which would make the interface a little bit more consistent.

bjeurissen:
I am in favor of leaving it out.

dchristiaens:
I do think it comes in handy from time to time, especially for quickly inspecting high-res datasets, but most often I jump to the overlay as well... I guess I'm undecided :)

draffelt:
I never use it. Now that you can overlay off axis, you can always zoom in
and rotate. I've lost minutes of precious life clicking that overlay check
box, and changing the wrong 'detail' input box

On Thu, Mar 26, 2015 at 11:53 PM, Daan Christiaens <notifications@github.com

Lestropie:
I find it useful every now and then, usually assessing the results of FOD segmentation. We could definitely decrease its default size. And possibly have the overlay default to on.

I am also in favour of keeping it because I want to have the same thing once I implement a dixel overlay. These will have a lot more fine detail per voxel, so being able to examine one voxel data more closely will be useful. Then, I still like the idea of having a somewhat consistent interface for all of the orientation overlay tools, which would sensibly include a close-up view of voxel data.

jdtournier:
Man, I was hoping for a consensus...

@Lestropie - OK, not sure we can shrink the ODF panel a great deal, it's already not that big. Besides, it saves us nothing in terms of reducing the amount of bloat in the code, avoiding issues with multiple OpenGL viewports, or cleaning up the UI on the ODF tool to help @draffelt hit the right button...

[...]

PR #219, opened 2015-04-15
MRView: remove single ODF display from ODF tool

jdtournier:
This remove the ODF display pane from the ODF tool, as discussed in issue #194.
Feedback welcome.

### DIFF
--- a/icons/odf_preview.svg
+++ b/icons/odf_preview.svg
@@ -1,0 +1,442 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="48"
+   height="48"
+   id="svg6430"
+   version="1.1"
+   inkscape:version="0.91 r13725"
+   sodipodi:docname="odf_preview.svg">
+  <defs
+     id="defs6432">
+    <linearGradient
+       id="linearGradient4474">
+      <stop
+         id="stop4476"
+         offset="0"
+         style="stop-color:#000000;stop-opacity:1;" />
+      <stop
+         id="stop4478"
+         offset="1"
+         style="stop-color:#808080;stop-opacity:1" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient4253">
+      <stop
+         id="stop4255"
+         offset="0"
+         style="stop-color:#00ff00;stop-opacity:1" />
+      <stop
+         style="stop-color:#c0c000;stop-opacity:1"
+         offset="0.5"
+         id="stop4387" />
+      <stop
+         id="stop4257"
+         offset="1"
+         style="stop-color:#af0000;stop-opacity:1" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient3913">
+      <stop
+         style="stop-color:#00ff00;stop-opacity:1"
+         offset="0"
+         id="stop3915" />
+      <stop
+         style="stop-color:#ff0000;stop-opacity:1"
+         offset="1"
+         id="stop3917" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient7550">
+      <stop
+         style="stop-color:#898989;stop-opacity:1;"
+         offset="0"
+         id="stop7552" />
+      <stop
+         style="stop-color:#f5f5f5;stop-opacity:1;"
+         offset="1"
+         id="stop7554" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5863-1">
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="0"
+         id="stop5865-9" />
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="1"
+         id="stop5867-2" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5873-5">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="0"
+         id="stop5875-5" />
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="1"
+         id="stop5877-9" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5863">
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="0"
+         id="stop5865" />
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="1"
+         id="stop5867" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5873">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="0"
+         id="stop5875" />
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="1"
+         id="stop5877" />
+    </linearGradient>
+    <linearGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient5863-8"
+       id="linearGradient7238-9"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="translate(-57,1004.8621)"
+       x1="72"
+       y1="8"
+       x2="72"
+       y2="39" />
+    <linearGradient
+       id="linearGradient5863-8">
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="0"
+         id="stop5865-1" />
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="1"
+         id="stop5867-7" />
+    </linearGradient>
+    <linearGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient5873-8"
+       id="linearGradient7240-9"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="translate(-57,1004.8621)"
+       x1="76"
+       y1="8"
+       x2="76"
+       y2="39" />
+    <linearGradient
+       id="linearGradient5873-8">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="0"
+         id="stop5875-7" />
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="1"
+         id="stop5877-5" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5863-1-8">
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="0"
+         id="stop5865-9-4" />
+      <stop
+         style="stop-color:#808080;stop-opacity:1"
+         offset="1"
+         id="stop5867-2-3" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient5873-5-6">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="0"
+         id="stop5875-5-2" />
+      <stop
+         style="stop-color:#000000;stop-opacity:1;"
+         offset="1"
+         id="stop5877-9-6" />
+    </linearGradient>
+    <radialGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient5873-9"
+       id="radialGradient6252"
+       cx="182.89905"
+       cy="24.915091"
+       fx="182.89905"
+       fy="24.915091"
+       r="13.693534"
+       gradientTransform="matrix(1,0,0,0.92467276,0,1.6584141)"
+       gradientUnits="userSpaceOnUse" />
+    <linearGradient
+       id="linearGradient5873-9">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:0.26923078;"
+         offset="0"
+         id="stop5875-9" />
+      <stop
+         style="stop-color:#000000;stop-opacity:0.34615386;"
+         offset="1"
+         id="stop5877-1" />
+    </linearGradient>
+    <radialGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient5873-9"
+       id="radialGradient7510"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(1,0,0,0.92467276,0,1.6584141)"
+       cx="182.89905"
+       cy="24.915091"
+       fx="182.89905"
+       fy="24.915091"
+       r="13.693534" />
+    <radialGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient5873-9-8"
+       id="radialGradient7510-9"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(1,0,0,0.92467276,0,1.6584141)"
+       cx="182.89905"
+       cy="24.915091"
+       fx="182.89905"
+       fy="24.915091"
+       r="13.693534" />
+    <linearGradient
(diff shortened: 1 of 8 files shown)


draffelt:
Looks good to me. However the hide all toggle was not working properly when you connect the button to the updateGL directly. If I add a `TRACE` then I can see updateGL is being called, however the window does not update. By adding `hide_all_slot()` seems to fix the issue. 

For some reason calling `updateGL()` directly from the detail Qspinbox does not have this issue.

jdtournier:
Thanks for the fix... 

Anyway, I wouldn't mind getting feedback on the UI aspects of this from the various parties who'd expressed an opinion previously. I'd like to merge sooner rather than later, but I'd like there to be some vague consensus that this is probably the better approach before I go ahead with that...

jdtournier:
OK, just had a tinker with the issues surrounding ODF wobble in the main view when inspecting ODFs close up (as raised by @thijsdhollander in issue #194), and there's no easy fix. 

For general ODF overlay, the current setup works relatively well - only issue being that the grid does jitter about when the viewing plane is not aligned with the ODF image, as the voxel used to anchor the grid changes in ways that change the grid location with respect to the image. 

When inspecting ODFs close up, this jitter becomes really annoying: the ODF displayed will wobble about and change as its values are fetched from subtly different (interpolated) locations, might suddenly jump to a different location altogether as the voxel used as the anchor switches to a different location...

For close up inspection, things work pretty well if you tick off interpolation and lock-to-grid. In this case, one ODF is guaranteed to be displayed at the centre of the viewport, and the ODF won't change provided the focus is close enough to the centre of the viewport that rotations won't induce a change in the centre voxel. 

So there are ways around all of this, but there isn't a one-size-fits-all approach that caters for all use cases - as far as I can tell... The most common use case will undoubtedly be to display the ODF atop their original data (or at least an image generated from the same data, and hence with the same grid), in lock-to-axes mode, with the expectation that ODFs will be displayed at voxel centres. This is what the default settings are geared towards. As soon as the plane tilts, all bets are off... 

One option is to disable lock-to-grid as soon as we move off-axis, but there's a few issues with that. One is that it's not trivial to detect robustly since there is no requirement that the image displayed be in any way related to the ODF image. The bigger issue though is probably that if you don't use the lock-to-grid approach, panning around the image feels strange, as the ODFs remain in the same location in the window, smoothly changing through interpolation as the ODF image pans around... Turning off interpolation also feels strange when off axis.

So not sure how best to handle this. I think we can reduce some of the random grid 'switches' that happen when rotating or moving the focus around by being smarter about how the grid in anchored. But there will always be issues when inspecting up close with the default settings.

Ideas...? I think so far Thijs' suggestion of a separate popup for inspection is looking most promising...

Lestropie:
If having the individual ODF view as a separate window rather than embedded within the toolbar solves some of the programmatic issues, as well as stopping the absurd toolbar sizing issues, then I'm all for it. Also highlights that this is an 'addon' whereas the overlay is really the key feature.

jdtournier:
OK, looks like there is a consensus... Not sure whether to merge now and add the separate window later, or get it right before merging. I think probably the latter...

Anyone want to help out...?

Lestropie:
I'm knee deep in GUI right now so I'll join in.

Fundamental question: Should it be an external window managed by `mrview`, or should it actually spawn an instance of `shview` and feed it SH data through Qt D-Bus? Suspect the latter might be more work than it's worth...

jdtournier:
> Suspect the latter might be more work than it's worth...

Yes, unfortunately I'd tend to agree... I think it's not a bad idea in principle, but you'd need to modify `shview` so extensively to get it to listen to dbus events that you might as well reimplement it as a separate window managed by MRView...

Lestropie:
OK @jdtournier give this a crack.

thijsdhollander:
Ok, I just extensively tested this: behaves just as I expected!  Everything is nicely separated/independent from the ODF overlay and its settings as well.  It's actually better than the previous ODF preview, even if only because it can now fill an entire screen if desired.  Finally a *really* good use for that second screen. :-)

The only hiccup I managed to find is that, even though "auto align" is selected, I'm able to freely rotate the preview.  It does reset to the correct alignment upon switching off and on again, or doing other types of interaction with the main viewer (e.g. just clicking to the focus is moved, or pressing A/S/C, etc...).  Not sure if this behaviour should be seen as either a bug or a feature.

Finally, just me nitpicking:
1) It'd be nice to have (the option to have) the ODF preview window "always on top" relative to the main mrview window (just like the toolbars when they are detached).  
2) Again probably just due to urge of consistency: a tiny cross in the upper right corner to close the preview.  Also because currently, when you open the ODF preview but then close the ODF display toolbar, you need to open up the toolbar again to be able to close the preview. :-)

Lestropie:
RE Auto align: That's the same behaviour as before. Could have that check box prevent any and all rotation, but I don't see any benefit in doing so:
* If you've put the cursor in the preview window and clicked and dragged, chances are it's because you want to rotate; and on the off chance that it wasn't, re-alignment is just a focus click away.
* If any and all rotation was disallowed, the user would sit there clicking and dragging, trying to figure out what's wrong, then (if they figure it out) disable the check box before they can go back to rotating.

So I think it's better the way it is.

1. 6eac4ac

2. Must be your window manager; my system gives me the usual minimize/maximize/close controls on the preview window. I'll have a look.

jdtournier:
@lestropie, haven't had a chance to check it out yet, but any reason you didn't implement this as a QDockWidget? It should solve both of Thijs issues, and allow docking alongside the other tools.

@thijsdhollander: the rotate-on-demand feature _is_ a feature. The use case is that 90% of the time, you'll be dragging your cursor along to inspect your ODFs, and you'll want them shown in the same orientation. You may want to occasionally take a closer look at one of them and spin it about, then carry on looking at the other ODFs. So allowing free rotate even when officially 'locked' saves you a couple of mouse clicks - and doesn't change functionality otherwise, as Rob says.

Lestropie:
Well, was trying to avoid anything docking because part of what spawned these modifications was the ODF renderer screwing up toolbar sizes when docked. Also with tools being docked as tabs, not convinced that it makes sense to have this particular one dock as a tab. But I'll try it and see what happens.

jdtournier:
I don't remember sizing being a particular issue with the ODF tool? I mean, it was too big, but that's just because there was too much stuff on it. What was a problem was the interaction of the tool's OpenGL draw with the main window's delayed render. I'm guessing this will also be an issue with these changes, although it's probably already handled by whatever we did to fix things the last time... 

I reckon you could use a QDockWIdget, and start it as floating - would achieve a similar effect to what you have now, with mini dock/close controls and 'stay on top of main window' behaviour...

Lestropie:
OK, so changed it to a dock; needed a little hacking to get it to work. Have it initially opening as a float, as otherwise it tries to dock itself underneath the main ODF tool. Also, the preview button in the main ODF tool is now a momentary (i.e. _not_ checkable), and tries to give focus to the preview tool wherever it is (opening it if necessary).

Only weird behaviour we've come across is that if you have multiple floating tools, and the ODF preview window is behind other tools, clicking the ODF preview button in the main ODF tool fails to promote it to the top of the stack.

jdtournier:
@Lestropie : just gave it a crack, seems to work fine for me. I don't think we should worry massively about subtle visibility issues, you call `raise()` on the widget, presumably that should to just that - possibly it has more to do with Qt and/or the your window manager than anything else. 

Only thing was that I was a bit surprised to see the preview had its own lighting parameters, I would have thought it would have made more sense to use the same ones for both. The previous implementation actually used the preview's lighting object rather than holding its own structure... But this is really a very minor niggle. Happy for you to merge when you're ready.

thijsdhollander:
Ok, so:

1) I understand the auto align now; as I was writing about it initially, I was also starting to feel it was more of a feature than a bug.  @Lestropie also explained me the name was changed some time ago from something like "*lock* orientation" to the current "auto align"; and indeed, now that I think of it, the "auto align" naming is actually not contradicting the current behaviour.  :-)

2) About...

> Only weird behaviour we've come across is that if you have multiple floating tools, and the ODF preview window is behind other tools, clicking the ODF preview button in the main ODF tool fails to promote it to the top of the stack.

... I couldn't replicate that behaviour on my computer, even performing exactly the same set of actions we did yesterday on @Lestropie 's computer: it works correctly here on mine.  It must indeed be an obscure difference between the window managers.  Sometimes it works correctly instantly, but sometimes it takes up to a good -yet single- second.  I have absolutely no clue why that is, but it's not intervening with the correct behaviour; so I wouldn't worry about it.

3) When first testing it, I actually expected the preview to have its own lighting parameters, because it has its own set of all the other settings as well (e.g. use lighting, hide neg. lobes, detail, etc...).  Probably again my reflex to expect a certain sense of consistency though... :-)

Lestropie:
I think the independent lighting settings for the preview pane was just trying to minimize the cross-talk between the tools. Particularly with the preview now being dockable, the ODF display and preview 'tools' are the first where one depends on the other, which isn't something we want to get into the habit of. But I'm happy for the lighting to be either a single or two independent controls... any others have an opinion?

jdtournier:
Regarding cross-talk between tools: the preview window clearly already depends on the ODF tool - you can't launch it without it, and the data it uses is provided by the ODF tool. I don't consider it to be a Tool as such - it doesn't derive from Tool::Base, and doesn't have an entry in the Tools menu. It just happens to be dockable...

I also notice you instantiate the preview window right in the ODF tool constructor, so you'd have access to its lighting class straight away. 

So the question is what independent settings this preview should have. I see a good case for level of detail, auto-align & interpolation, but I'm not convinced we need to have independent controls for lighting, lmax, hide negative lobes, or colour by direction. It is supposed to be a close up view of what's already on the main display, after all... What do you guys think?

Also, given the above I'm not sure 'preview' is the right name for this...

Lestropie:
As long as you're happy with the cross-talk, sure let's homogenize the appropriate settings. (Though note it does now inherit from Tool::Base.) I think your list is correct.

I'm not convinced by the name either. How about "ODF Close-up"? Or "ODF Selfie"? :-P

jdtournier:
Wow, didn't realise it inherited from Tool::Base, no wonder you were worried about cross talk... What's the rationale for not deriving directly from QDockWidget? I guess you did this to keep the spacing and font size consistent... Fair enough, I'll take a look in a minute to make sure there's no other funny interactions - I have a feeling as long as it's not listed in tool/list.h, it should be fine to do that.

But nonetheless, I think conceptually this is not a standalone tool...

jdtournier:
As to name, I was leaning towards 'ODF inspector' - use it to verify the legality of your ODFs ;)

Kissing aside, something along those lines would be OK?

Lestropie:
I kept swapping and changing last night of how I was defining it and what I was deriving from. I think deriving from Tool::Base was necessary because I needed to first define a Tool::Dock and keep the pointer to that in the main ODF tool (to be able to show/hide it), and that class stores the pointer to the Tool::Base.

I think I prefer "ODF inspect" to "ODF inspector"; the main tool isn't the "ODF displayer". Then again, maybe now the main tool should now be "ODF overlay"?

jdtournier:
I don't mind the current name for the main ODF tool - it encompasses both functions, which is correct, the 'preview' is _not_ a tool...

How about 'microscope' for the preview window...? You can look at your ODFs in more detail... :)

Or 'magnifier'? We don't actually need to call it anything, just stick a big magnifying glass icon on it...

jdtournier:
Just had a quick look through, I think deriving from Tool::Base is the right thing to do, it provides a lot of the back-end necessary for interacting with the main window, etc. and sets the font sizes and layouts appropriately...

Lestropie:
OK, that should be all apart from the renaming.

thijsdhollander:
Just tested: works great!  Having it as a dockable, being always on top, it's still handy to change those "shared" settings using the ODF display dock.  Having them out of the way on the ODF preview also turns out to be much cleaner.

On the naming: all of the above suggestions are fine by me... but indeed, since the preview is not a tool, it actually doesn't necessarily need a name.  How about: no name (in the title bar of the dock), and the button's tooltip reading "Inspect ODFs" ?

Names are always (possibly) a bit ambiguous though.  Not to bring this up (sorry, can't resist the urge), but you could also question the "ROI analysis" tool's name.  It's more of a ROI editor (and inherently viewer) one could reason... but I have to admit: I don't think any of the current tools' names would actually confuse anyone in practice.

jdtournier:
Sounds great, I'll check it out when I get to work. 

As to the name thing, it's true that it really doesn't matter, it's only shown as an icon (or at least it was the last time I checked), so we can go wild on the tooltip. I reckon something like "inspect ODF at cursor in separate window"...?

And I like @thijsdhollander's suggestion of renaming 'ROI analysis' to 'ROI editor' - much closer to what it actually does...

jdtournier:
Just had a go with this, it works beautifully - good work. I reckon this is good to merge. Just a couple of suggestions that we could do before that:

- update the tooltip as above
- make the scale AdjustButton apply also to the preview: this is most useful when the data scaling is orders of magnitude out (e.g. displaying a SH fit to the signal), and it wouldn't be possible to inspect the ODF in that case if the scaling isn't adjusted manually, using the mouse scroll to adjust the scaling just doesn't get you far enough...
- switch to detail 5 for the preview...?

thijsdhollander:
Did a bunch of little things:

1) Tooltips in of ODF display tool's button updated, including the one to open the ODF preview window

2) ODF preview window's title (in title bar) is gone now

3) Switched to default detail 5 for preview

4) Changed the ordering of a few options at the bottom of the ODF display tool, so they are grouped a little bit more logically.  The rationale is: left column is scale/lmax/griddish things, right column is more on rendering features of the individual ODFs.  In the right column, detail and hide negative lobes are on top (change what is actually shown), below that the colour/light stuff (change the surface's prettiness).  This makes it so that functionally "grouped" check boxes are closer to each other for fast 'n furious clickery.

5) Changed the main grid / overlay choice of grids things, to a simple checkbox reading "use main grid".  Defaults to unchecked/off (i.e. as before).  When I first started using MRtrix3, it actually took me an unusually long while to figure out this thing: I was confused by the "overlay" choice of grid, thinking that it was the grid... off the current selected overlay in the Overlay tool.  You know, overlay here, overlay there, and everyone gets confused.  The "main grid" is pretty unambiguous though.  So a check box to choose just that makes sense.  If not selected, it also makes sense that the grid of the ODF image itself is used without any "external" manipulation by other images.  Finally, a check box only takes a single click to switch back or forth.

I never worked with QT before, so these things were all done by pattern matching and (trying to) see the logic in the existing code.  Quite confident I didn't break anything though. :-)

Finally, 2 remaining issues:

1) If you rescale the ODF display tool, only the left column of the Display settings part will stretch (it was like that already before I touched it, I swear!).  I'm not sure how to make both columns just stretch equally.

2) Haven't figured out yet (or looked into it yet) how to make the scale box also work for the preview.  This also begs a small question by itself: does it mean that using the scroll wheel for the preview window should be disabled?  Or would the scroll wheel (in the preview) work as well to change the scale box in the ODF display (a bit like in the old days...)?

jdtournier:
Sounds all good. The resize issue sounds like it might be to do with the horizontal stretch factors in the `GridLayout` (I'm assuming the problem is in the horizontal direction...?). Can't remember where they're set, it'll be either in the layout constructor, or in the individual `addWidget()` calls.

As to the ODF scale, I agree the scroll mouse should probably be disabled - it was never particularly obvious anyway. Problem is, that means we'll probably need to remove the functionality from the DWI::Renderer, and need to find an alternative way to scale the ODF in `shview`... Simplest for now is probably just to add some flag to the DWI::Renderer to ignore mouse scroll events...

jdtournier:
@thijsdhollander: just one more thought, is there a clear tooltip for the new 'main grid' check box? We should probably have one to explain exactly what this thing does....

thijsdhollander:
Okay, slightly changed and cleaned up the layout of the ODF display tool a bit more.  Note the button for the preview tool is now below the list of loaded ODF images.  It didn't really make much sense amongst the buttons on top.  Also, it now has a bit of text (next to the icon) on the button as well.  Still has a slightly more elaborate tool tip as well, which mentions that the thing will open a separate window.
The main grid check box also comes with a tool tip now.
Finally, the mouse scroll is disabled in the preview; the scaling is now controlled by the equivalent setting in the main ODF display tool.  Right mouse button "scaling" of the whole preview (axes included) is still possible.  The preview tool now also has a reasonable minimum size of 300x300 (similar to shview).  It's not much use at smaller sizes anyway.
Probably did some other small clean-ups here and there that I forgot to mention as well.

Whoever's interested/concerned: give it a spin.  I think it's all done now, ready to go in master.

jdtournier:
Just gave it a quick once-over, everything looks good, so I went ahead and merged. There may be a few remaining niggles, but they'll probably get reported and fixed much more quickly if they're in master anyway...

PR #228, opened 2015-04-23
added median to mrmath. performance could be improved...

maxpietsch:
Regarding issue #147 

Could someone please explain to me why I need to copy the vector `values` into `v`?

The error is: assignment of read-only location '(& list)->std::vector<_Tp, _Alloc>::operator[]<float, std::allocator<float> >(num)'

### DIFF
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -116,12 +116,10 @@ class Median {
         ++count;
       }
     }
-    value_type result () const { 
-      if (!count){
+    value_type result () { 
+      if (!count)
         return NAN;
-      }
-      std::vector<value_type> v(values); // Why do I need this copy?
-      return Math::median(v);
+      return Math::median(values);
     }
     size_t count;
     std::vector<value_type> values; 
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -31,6 +31,7 @@
 #include "image/utils.h"
 #include "image/voxel.h"
 #include "math/math.h"
+#include "math/median.h"
 
 #include <limits>
 #include <vector>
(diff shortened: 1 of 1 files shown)


jdtournier:
Because your `value ()` method is declared `const`, and so can't modify its member variables, but `Median:: value ()` takes a non-const container - this is because the container will be sorted in-place during the call. Just take out the `const` at the end of the declaration of your `value ()` method...

jdtournier:
By the way, the `Math::Median` takes care of NaN's - no need to explicitly check for them or accumulate a count...

Otherwise, this looks good to me. Feel free to merge to master once you've removed the redundant bits and you're satisfied it works as you expect.

maxpietsch:
Merci. Well, that was easy...

jdtournier:
By the way, I notice you still have a check for infinities. Technically they're not a problem for the median - unless more than half the values are infinite of course. The median function already ignores NaNs. Any reason to keep this extra check in there...?

maxpietsch:
My reasoning was, that it might be faster to check for isfinite than copy the voxel values into a vector and sort it if there are many NaNs around. But that was just a guess. Happy to take it out.

jdtournier:
Yes, I see. I thought you explicitly didn't want Inf's to be taken into account... 

In general, I don't think the performance difference is going to be all that noticeable. I'd rather we kept this as close to the median function as possible, which means we do include Inf's in the computation. It's only NaN's that explicitly shouldn't be included. If you're dead keen on keeping this check in, you could use a `!std::isnan()` instead, which at least would keep the behaviour consistent...?

maxpietsch:
Ok, changed it. (4c29571)

jdtournier:
OK, 25% is appreciable... Good call.

Issue #192, opened 2015-03-25
MRView produces lots of "Attempt to set a screen on a child window." messages


jdtournier:
This is using Qt5.4 on Arch Linux. I don't get this with Qt4.8. Assuming it's a Qt bug, but need to watch this. Might be related to issue #191 - bugs in the deprecated QGLWidget class - who knows?

jdtournier:
fixed with #266

Issue #191, opened 2015-03-25
MRView: add support for newer Qt5 OpenGL API


jdtournier:
Unfortunately, the [QGLWidget class is deprecated in Qt5](http://doc.qt.io/qt-5/qglwidget.html#details)... And there are hints of various bugs in the Qt5 QGLWidget, which I'm guessing aren't going to be high priority for them to fix. There is a new way of setting up an [OpenGL viewport](http://doc.qt.io/qt-5/qtgui-index.html#opengl-and-opengl-es-integration), which we'd be well advised to support eventually. This can probably be done with minimum disruption by setting up a class to abstract away the details, by using their new [QOpenGLWidget class](http://doc.qt.io/qt-5/qopenglwidget.html#details), which seems very similar to the QGLWidget class. We'll need some care to make sure the context is set up properly for OpenGL3.3, but that shouldn't be too hard.

Note that we'll need to support Qt4 for some time, so we can't simply switch entirely to the new API.

jdtournier:
Seems we may need to look into this pretty quickly. The recent changes I made to Tool sizing (commit 8f75f6e) seems to break the ODF display panel somewhat... On Arch with Qt5.4, if Tools open as tabs (the default), the OpenGL panel in the ODF tool is not properly initialised, and nothing gets displayed in there. Works again if the tool is detached and floated, and reattached as a tab again, or if it was opened as floating directly (via the config file). Also works fine on Qt4.8, so again, I'm guessing this is a bug in the now deprecated QGLWidget (similar to issue #192), which probably won't be getting much attention from the developers any time soon...

That said, given the discussion around issue #194, simplest thing at this stage might just be to remove the panel altogether...

jdtournier:
done with #266

Issue #222, opened 2015-04-17
mrview: random "[ERROR] your OpenGL implementation ..."


thijsdhollander:
On Ubuntu, when using `mrview`, from time to time I *randomly* get the `[ERROR] your OpenGL implementation is not sufficient to run MRView - need version 3.3 or higher` at the command line.  The viewer itself just keeps running fine, and all features work correctly as well.  At first sight, I'm not able to reproduce the behaviour consistently.  Sometimes it just happens, sometimes it doesn't.  Even when it happens right at the start (by opening an image), and I exactly repeat that command, I'm still not able to reproduce it.  When it happens during use of the viewer, it can happen at different times (upon opening an image, upon loading ODFs, ...).  So just to be clear: everything is working fine at all times; it's just the message popping up on the command line at random times.

The configure script nicely reports `Checking for OpenGL: 3.3.0 NVIDIA 304.125`, so OpenGL 3.3 definitely is there, as it should be.

jdtournier:
Sounds like a bug either in Qt or in your OpenGL drivers... Either way, I reckon it will magically heal itself in time. Keep monitoring this, and if you find it's still happening in a few months' time (by which time you'll have hopefully received a few software updates), we'll worry about it... How does that sound?

I'll mark it as `wontfix` for now, but I'll leave the issue open until we _know_ whether the issue is with MRtrix...

thijsdhollander:
Yep, no worries.  It's not really affecting anything in practice after all.  If I remember it in some months and still spot it occasionally, I'll report back here.

jdtournier:
:+1:

thijsdhollander:
Just accidentally ran into some extra (strange) evidence: as I mentioned above, the configure script nicely reports `3.3.0 NVIDIA 304.125` ... but the OpenGL information dialog box, as opened from the Help menu in MRView, reports `2.1.2 NVIDIA 304.125`.

Lestropie:
From memory my work desktop shows this in debug mode, but not release mode. Just to add some more confusion.

jdtournier:
Could this be related to OpenGL Core vs Compatibility profiles? Can you guys set `NeedOpenGLCoreProfile: 1` in your config files, see if that sorts things out...?

Lestropie:
Another clue...
```
==6126== Conditional jump or move depends on uninitialised value(s)
==6126==    at 0x5C5BE9: MR::GUI::GL::init() (gl.cpp:42)
...
```

[...]

Issue #246, opened 2015-05-06
Lighting dialog box should be a dock widget


jdtournier:
Dialog for lighting is a normal window, when all other settings boxes in MRView are dock widgets. This makes it inconsistent, but also it doesn't have stay-above-main-window behaviour, and can't be docked.

This will probably impact `shview` as well, which might need to have its main window changed to a QMainWindow (or whatever it's called...).

rtabbara:
Had a go at this in  4835759. <code>GUI::Dialog::Lighting</code> has moved to <code>GUI::LightingDock</code> with corresponding .h/.cpp files also renamed/moved.

For the standalone <code>shview</code>, the lighting panel remains as a QDialog.

thijsdhollander:
@rtabbara : I quickly gave it a test before I'll head off home.  It seems to function well, there's 2 small details that might (or not) be issues:

1) It opens in a docked state.  Was a bit surprised at first because I expected it to "float" by default; possibly because I was used to the behaviour of the old dialog.  Because I was testing with the ODF display tool docked as well, it "popped up" docked underneath, which then actually *did* seem nice to me.  But then reasoning a bit further: if a user undocks the ODF display tool first, and then opens the lighting panel, it also still opens docked to the main window (and that's a bit odd).  So probably it makes more sense to have it initialise as floating...

2) It's maybe also good to bring it to the top of the docking stack whenever the button is pressed; so when one or more tools are used in an undocked state (the preview sub-tool probably being the most likely one), it comes back into view when the user might have "lost" it.

thijsdhollander:
...so on second thought, maybe having it docked by default is actually fair enough for the majority of use cases.  I reckon that this does guarantee (when the ODF display tool or tractography tool are docked in their default state) the lighting controls popping up "near the cursor", which is actually quite handy in practice.  So probably this issue is fixed as it is now...?

Side note: when testing it with the tractography tool, I noticed the "object colour" in the lighting settings is never used... is that correct? (I know the actual colour *can* be set by right clicking on the tractogram's entry in the list)

Lestropie:
> Side note: when testing it with the tractography tool, I noticed the "object colour" in the lighting settings is never used... is that correct?

It shouldn't have any effect in that context; then again, it's trivial to create the lighting settings widget with the object colour control omitted... the solution is left as an exercise for the reader :-P

thijsdhollander:
> It shouldn't have any effect in that context

[...]

PR #266, opened 2015-06-09
QOpenGLWidget and other OpenGL related issues

jdtournier:
This branch introduces a few changes to the OpenGL backend. I've tried to test most use cases on my system, but I'd prefer to have as many people try it out before I commit something that might very well screw up - especially now that we've publicised it at the ISMRM...

I'd like confirmation that this works on all platforms and for all common combinations (especially Windows, I'll test MacOSX tomorrow), for both Qt4.8, 5.2/3 & 5.4 (where applicable), on as many different hardware combinations as possible. So far, I've only tested on Arch Linux 64-bit with the nvidia drivers, for Qt 4.8.6 and 5.4.2. Feedback appreciated before a merge to master.

As to the changes themselves, in a nutshell: 

- This branch replaces the `QGLWidget` (deprecated with Qt 5.4) with its replacement `QOpenGLWidget` when compiling with Qt >= 5.4. This would close issue #191, and seems to fix issue #192.

- It also removes the glrefresh timer trick that we'd introduced to get around delayed even handling when the render was slow. This was due to the fact that calls to `updateGL()` were immediate, so further events in the event loop would get queued up and processed in turn much later. The new `QOpenGLWidget` class no longer has this function, relying instead on the generic `QWidget::update()`, which _schedules_ a repaint (exactly what we were trying to achieve with the refresh timer) - and it's available in Qt 4 as well... 

- This also now forces a core OpenGL profile by default on all platforms. Previously, this was only the case on MacOSX (can't get OpenGL 3.3 at all otherwise), but since people have been having issues with their ATI cards, and it seems to work fine on NVidia too, I guess we might as well switch over completely. This will also help sort out issues lile #222.

### DIFF
--- a/src/gui/dialog/opengl.cpp
+++ b/src/gui/dialog/opengl.cpp
@@ -32,7 +32,7 @@ namespace MR
     namespace Dialog
     {
 
-      OpenGL::OpenGL (QWidget* parent, const QGLFormat& format) : QDialog (parent)
+      OpenGL::OpenGL (QWidget* parent, const GL::Format& format) : QDialog (parent)
       {
         TreeModel* model = new TreeModel (this);
 
@@ -59,9 +59,14 @@ namespace MR
         bit_depths->appendChild (new TreeItem ("depth", str (format.depthBufferSize()), bit_depths));
         bit_depths->appendChild (new TreeItem ("stencil", str (format.stencilBufferSize()), bit_depths));
 
-        root->appendChild (new TreeItem ("Double buffering", format.doubleBuffer() ? "on" : "off", root));
+#if QT_VERSION >= 0x050400
+        root->appendChild (new TreeItem ("Buffering", format.swapBehavior() == QSurfaceFormat::SingleBuffer ? "single" :
+               ( format.swapBehavior() == QSurfaceFormat::DoubleBuffer ? "double" : "triple" ), root));
+#else
+        root->appendChild (new TreeItem ("Buffering", format.doubleBuffer() ? "double" : "single", root));
+#endif
         root->appendChild (new TreeItem ("VSync", format.swapInterval() ? "on" : "off", root));
-        root->appendChild (new TreeItem ("Multisample anti-aliasing", format.sampleBuffers() ? str(format.samples()).c_str() : "off", root));
+        root->appendChild (new TreeItem ("Multisample anti-aliasing", format.samples() ? str(format.samples()).c_str() : "off", root));
 
         gl::GetIntegerv (gl::MAX_TEXTURE_SIZE, &i);
         root->appendChild (new TreeItem ("Maximum texture size", str (i), root));
--- a/src/gui/dialog/opengl.h
+++ b/src/gui/dialog/opengl.h
@@ -35,7 +35,7 @@ namespace MR
       class OpenGL : public QDialog
       {
         public:
-          OpenGL (QWidget* parent, const QGLFormat& format);
+          OpenGL (QWidget* parent, const GL::Format& format);
       };
 
     }
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -52,7 +52,7 @@ namespace MR
     {
 
       RenderFrame::RenderFrame (QWidget* parent) :
-        QGLWidget (parent),
+        GL::Area (parent),
         view_angle (40.0), distance (0.3), line_width (1.0), scale (1.0), 
         lmax_computed (0), lod_computed (0), recompute_mesh (true), recompute_amplitudes (true), 
         show_axes (true), hide_neg_lobes (true), color_by_dir (true), use_lighting (true), 
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -88,7 +88,6 @@ namespace MR
       {
         GL::init();
         renderer.initGL();
-        gl::ClearColor (lighting->background_color[0], lighting->background_color[1], lighting->background_color[2], 0.0);
         gl::Enable (gl::DEPTH_TEST);
 
         axes_VB.gen();
@@ -147,6 +146,8 @@ namespace MR
 
       void RenderFrame::paintGL ()
       {
+        gl::ColorMask (true, true, true, true); 
+        gl::ClearColor (lighting->background_color[0], lighting->background_color[1], lighting->background_color[2], 0.0);
         gl::Clear (gl::COLOR_BUFFER_BIT | gl::DEPTH_BUFFER_BIT);
 
         float dist (1.0f / (distance * view_angle * D2R));
@@ -223,6 +224,14 @@ namespace MR
           gl::Disable (gl::BLEND);
           gl::Disable (gl::LINE_SMOOTH);
         }
+
+        // need to clear alpha channel when using QOpenGLWidget (Qt >= 5.4)
+        // otherwise we get transparent windows...
+#if QT_VERSION >= 0x050400
+        gl::ClearColor (0.0, 0.0, 0.0, 1.0);
+        gl::ColorMask (false, false, false, true); 
+        gl::Clear (GL_COLOR_BUFFER_BIT);
+#endif
 
         if (OS > 0) snapshot();
 
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -57,23 +57,15 @@ namespace MR
         lmax_computed (0), lod_computed (0), recompute_mesh (true), recompute_amplitudes (true), 
         show_axes (true), hide_neg_lobes (true), color_by_dir (true), use_lighting (true), 
         normalise (false), font (parent->font()), projection (this, font),
-        focus (0.0, 0.0, 0.0), OS (0), OS_x (0), OS_y (0),
-        glrefresh_timer (new QTimer (this))
+        focus (0.0, 0.0, 0.0), OS (0), OS_x (0), OS_y (0)
       {
         setMinimumSize (128, 128);
         lighting = new GL::Lighting (this);
         lighting->set_background = true;
-        connect (lighting, SIGNAL (changed()), this, SLOT (updateGL()));
-        glrefresh_timer->setSingleShot (true);
-        connect (glrefresh_timer, SIGNAL (timeout()), this, SLOT (base_updateGL()));
+        connect (lighting, SIGNAL (changed()), this, SLOT (update()));
       }
 
 
-
-      RenderFrame::~RenderFrame()
-      {
-        delete glrefresh_timer;
-      }
 
 
 
@@ -85,17 +77,10 @@ namespace MR
         p[6] = rotation(0,2); p[7] = rotation(1,2); p[8] = rotation(2,2);
         Math::Matrix<float> M (p, 3, 3);
         orientation.from_matrix (M);
-        updateGL();
+        update();
       }
 
 
-
-      void RenderFrame::updateGL ()
-      {
-        if (glrefresh_timer->isActive())
-          return;
-        glrefresh_timer->start();
-      }
 
 
 
@@ -255,11 +240,11 @@ namespace MR
         if (event->modifiers() == Qt::NoModifier) {
           if (event->buttons() == Qt::LeftButton) {
             orientation = Math::Versor<float>();
-            updateGL();
+            update();
           }
           else if (event->buttons() == Qt::MidButton) {
             focus.set (0.0, 0.0, 0.0);
-            updateGL();
+            update();
           }
         }
       }
@@ -288,17 +273,17 @@ namespace MR
 
             Math::Versor<float> rot (angle, v);
             orientation = rot * orientation;
-            updateGL();
+            update();
           }
           else if (event->buttons() == Qt::MidButton) {
             focus += projection.screen_to_model_direction (QPoint (dx, -dy), focus);
-            updateGL();
+            update();
           }
           else if (event->buttons() == Qt::RightButton) {
             distance *= 1.0 - DIST_INC*dy;
             if (distance < DIST_MIN) distance = DIST_MIN;
             if (distance > DIST_MAX) distance = DIST_MAX;
-            updateGL();
+            update();
           }
         }
         else if (event->modifiers() == Qt::ControlModifier) {
@@ -306,7 +291,7 @@ namespace MR
             view_angle -= ANGLE_INC*dy;
             if (view_angle < ANGLE_MIN) view_angle = ANGLE_MIN;
             if (view_angle > ANGLE_MAX) view_angle = ANGLE_MAX;
-            updateGL();
+            update();
           }
         }
       }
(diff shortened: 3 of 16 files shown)


rtabbara:
@jdtournier Tried this on my setup at work (Quadro 600/ GL 4.2.0 / Qt 5.2.1). Seems to work fine. I just went through all the different tools and played around with the various tunable parameters. Not sure if you had specific use cases in mind that you think we should check.

I've also got a Mac at home, so I'll also give that a whirl when I get the chance.

thijsdhollander:
Ok, I'm taking this for a spin and will be using it for a while; seeing if I can break something.  So far, I ran the configure script, which reports:

```
Detecting OS: linux
Checking for C++11 compliant compiler [g++]: 4.8 - tested ok
Detecting pointer size: 64 bit
Detecting byte order: little-endian
Checking for variable-length array support: yes
Checking for non-POD variable-length array support: yes
Checking for zlib compression library: 1.2.8
Checking for GNU Scientific Library: 1.16
Checking whether GSL compiles with -DHAVE_INLINE: yes
Checking shared library generation: yes
Checking for Qt moc: moc (version 5.2.1)
Checking for Qt qmake: qmake (version 5.2.1)
Checking for Qt rcc: rcc (version 5.2.1)
Checking for Qt: 5.2.1
```

Building/compiling went fine.

Running the viewer so far didn't show the `[ERROR] your OpenGL implementation is not sufficient to run MRView - need version 3.3 or higher` *yet* (whereas I just had it pop up again using the `master` branch an hour ago); but I'll try to play around with various things that seemed to result in a higher probability of it showing up in the past.  Finally, the OpenGL dialog, as opened from `mrview`, now nicely reports version `3.3.0 NVIDIA 304.125`; whereas before, it used to report 2.1.2 (as I mentioned in #222).

My computer here has a GeForce GTX 650.  The NVIDIA X Server Settings tool/dialog on Ubuntu reports `4.2.0 NVIDIA 304.125` in the OpenGL information section.

thijsdhollander:
Ok, I just threw everything I possibly could at it (combining a bunch of tools/settings until the thing slowed down to a near stop, and still drawing ROIs on top of all that).  Nothing breaks so far, no dodgy OpenGL errors as well. :smiley:

Lestropie:
Verified on Windows 64-bit with QT 5.3.2 and 5.4.1.

chunhungyeh:
Hi all. I have just given it a try on my Mac laptop (with Qt5.4.2). The configure/build stages are both fine. But I received errors when I tried running mrview. Here are the error messages:

chyeh@CHYeh-rMBP $ mrview
mrview: [ERROR] your OpenGL implementation is not sufficient to run MRView - need version 3.3 or higher
mrview: [ERROR]     operation is likely to be unstable
mrview: [ERROR] GLSL log [vertex shader]: ERROR: 0:1: '' :  version '330' is not supported
ERROR: 0:1: '' : syntax error: #version
ERROR: 0:2: 'layout' : syntax error: syntax error

mrview: [ERROR] error compiling OpenGL vertex shader ID 1
chyeh@CHYeh-rMBP $

jdtournier:
You guys rock! Thanks for the fast response. Glad I hadn't committed to master then... I'll take a look into the MacOSX issue later today.

jdtournier:
OK, that last commit seems to fix the issue on MacOSX (with Qt 5.4.2), and seems to work fine on my Arch desktop with Qt 4.8.6 & 5.4.1 - although there is this strange behaviour that I've never seen before:

screenshot from 2015-06-10 12-43-17

Pretty sure that's a bug in Gnome Shell, or the Nvidia drivers. I can't see how that can be allowed to happen. I'll update my system and reboot, maybe that'll fix it...

In any case, more testing please! :grin:

jdtournier:
OK, weird, update & reboot didn't fix the funny transparency issue... No idea why that would happen. Only happens with Qt 5.4.1 & 5.4.2 (haven't tested earlier versions). And only on this branch. I guess there must be something funny with their need `QOpenGLWidget`... I'll have to investigate.

jdtournier:
OK, having tried a few things, it looks like the simplest thing to do is to make sure the final buffer has its alpha channel set to fully opaque... So this should be fixed with commit 2d8cabd.

draffelt:
Tested fine on my laptop: Ubuntu 14.04, Nvidia 349.16, qt 4.8.6

jdtournier:
Sorry to push, but I reckon this is good to merge. Could everyone give it another shot with the latest commit, particularly on Windows & MacOSX...? Just want to make sure I'm not about to trigger an avalanche of bug reports on Google+...

chunhungyeh:
Successfully compiled/built on my laptop (MacOSX 10.10.3 with Qt 5.4.2); no funny transparency issue. Hmm...but it crashed when I clicked the 'Inspect ODF at focus' button in ODF display. (I have also tried on the master branch, where the 'Inspect ODF at focus' worked nicely.)

jdtournier:
@chunhungyeh - thanks, that's really useful feedback! I'll need to see what's wrong with the DWI renderer... I assume `shview` also doesn't work?

jdtournier:
'Inspect ODF at focus' segfault is also present on Linux. I'll investigate...

thijsdhollander:
Not sure if it's helpful, but I don't get the "Inspect ODF at focus" segfault here on Ubuntu.

`shview` works here as well, but testing it, I just noticed its menu bar is gone...?  No way of modifying any appearance settings, and if I don't load an SH coefficient file through the command line, I can't even do anything with it (apart from interacting with some empty axes...).

thijsdhollander:
Oh wait, that latter thing about 'shview'... it's also happening (for me) on the master branch.  Last time I used it (that might've been more than a month ago though), it still worked fine.

jdtournier:
You wouldn't get the segfault on Ubuntu because it uses Qt 4.8 by default - I expect this would be specific to Qt >= 5.3.

The lack of menu bar is unexpected though. I'm guessing this'll be related to the changes someone made to the DWI renderer to allow the ODF pop-up window in MRView...

jdtournier:
@thijsdhollander : can you double-check your lack of menubar issue with `shview`? The master branch works fine on my Arch system with both Qt 5.4.2 & Qt 4.8.6...

jdtournier:
@thijsdhollander : you wouldn't happen to be using Unity by any chance...? Canonical have done funny things to the menubar, IIRC - it ends up integrated into the title bar or the top bar, depending on which version you're running...

jdtournier:
OK, I think that last commit should iron out the issues raised by @chunhungyeh. No idea about @thijsdhollander's menubar issue though, it's not a problem for me. Everything seems to work fine for me on Arch Linux 64 bit, Qt 4.8.6 & Qt 5.4.2.

May I request a further round of testing...? :grin:

chunhungyeh:
With the last commit by @jdtournier, there is no segfault on my laptop. The `shview` is working now; but the funny transparency issue (plus some weird artefact) appears on the window of 'inspect odf at focus.' Here is my screenshot:

inspect_odf_at_focus

jdtournier:
Right, hopefully these next commits will have fixed those remaining issues...

chunhungyeh:
The latest commit has fixed the displaying issues of the ODF preview tool. :+1:
Yet, I have bumped into another problem when trying merging the ODF preview window into the tool panel list. Here are the error messages:
````
mrview: [ERROR] GLSL log [OpenGL shader program]: ERROR: duplicate definition of function 'main'
ERROR: Multiple declarations of 'color' have different types (vec3 and vec4)

Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt. You must
reimplement QApplication::notify() and catch all exceptions there.

mrview: [ERROR] error linking OpenGL shader program ID 3
````

thijsdhollander:
Okay, another two-fold answer:

1) I do experience a minor odd thing with the ODF preview tool now as well.  After I open it using the button in the ODF display tool, but then subsequently either dock or undock the ODF *display* tool (not the preview!), the ODF *preview* tool closes.  Not a big issue though, as I can reopen it again using the button in the ODF display tool.  Just a bit funny that the ODF preview tool depends in such a way on the action of (un)docking the ODF display tool.  I don't see any immediate issues or weirdness when (un)docking the ODF preview tool itself.

2) @jdtournier : yep, using Ubuntu and Unity.  But I don't have the actual "normal" set of menus for applications integrated in the top or title bars, everything sits nicely in their own windows.  Below is a screenshot that shows off most things.  It looks like this for both the `master` and `qopenglwidget` branches.  I opened `shview` providing an SH coefficient file at the command line.  No menus to be seen in the `shview` window, and the one in the top bar only provides a "Quit" item.  By comparison, other windows (e.g. gedit and gimp, as shown) have their menus where they should be...

screenshot from 2015-06-19 11 42 17

jdtournier:
Darn... OK, there's a lot of issues at play here.

First off, I can confirm the issues @thijsdhollander is having - this does happen on Ubuntu 14.04 when using Qt 5.2 - works fine with Qt 4.8. I'm guessing this is a Qt bug, but given that it worked before, there's probably a simple fix somewhere. By the way, since this happens on both master and this branch, it's unlikely to have anything to do with the present changes. Besides, the new QOpenGLWidget is only used with Qt >= 5.4... So I reckon this should be investigated on the master branch, see if you can identify which commit caused the problem.

By the way, @thijsdhollander: your screenshot looks to me like GNOME 3 in classic mode - not Unity. But we reproduced this on @maxpietsch's desktop (also 14.04), which runs the wacky i3 window manager, so unlikely to have anything to do with that side of things.

The other issue highlighted by @thijsdhollander is a bit weird, but I have a feeling it might have to do with the `hide_event()` handler in the ODF tool, which will also hide the preview and lighting dialogs when the main tool is hidden. There is every chance that detaching a dock widget involves a hide event in the main window, and a subsequent show event later on once the widget's resources have been reassigned to a fresh top-level window. Would be trivial to check by adding a `TRACE;` in the ODF's `hide_event()` callback, see if it gets invoked when detaching. Again, I don't expect this has anything to do with the changes in this branch, since you're not even using the new API. See whether this also happens on the master branch.

Finally, the remaining issue mentioned by @chunhungyeh is most likely genuinely due the new QOpenGLWidget API (I'll look into whether this happens on Linux too when I get the chance). The idea with Qt's new QOpenGLWidget API (as far as I can tell) is to avoid having separate X11 windows for all the different viewports (each OpenGL viewport gets its own X11 window normally, regardless of whether its embedded in another actual top-level window or not). The X11 approach keeps things much cleaner from a resource management point of view, but the new API is probably more efficient, and a closer match to how Wayland works (and probably also MacOSX, no idea about Windows). But it does mean that when docking, the preview tool goes from having its own separate window to sharing the same window as the main viewport. My guess is, this is a bug in Qt - the QOpenGLWidget really should take care to ensure that resources are managed such that it at least 'looks like' each QOpenGLWidget context is independent, but clearly this isn't the case - the issue here being most likely due to the shader program IDs getting mixed up between the two OpenGL contexts, and clashing. It's probably trying to link a shader program using shader IDs from one context mixed in with IDs from the other. We need to check whether there is any possibility that the shader programs are genuinely being compiled and linked in the wrong context, but assuming this is not the problem, I have no idea how to get around that. Some options: 
- wait for Qt to fix this (we could help things along by reporting the bug, even providing a minimal working example if needed); 
- prevent the preview from floating; 
- prevent the preview from docking. 

I'd personally favour the latter (assuming the problem is not with MRView, that is), but obviously opinions are welcome...

maxpietsch:
My opinion: wicked not wacky ;)

thijsdhollander:
Quick one before the weekend:

So first off, yes, I seem to indeed be using GNOME (not Unity).  Not sure how I came to assume I would've been using Unity... :-)  So it "just" seems to be a general occurring issue on Ubuntu 14.04 when using Qt 5.2.

My first guess as to "when" this happened, would be -as you also earlier mentioned- when we were fiddling around with the ODF preview tool.  More specifically (maybe), the activity which ensued afterwards as a part of #246 ....?

Next, the hiding behaviour of the ODF preview tool in function of the (un)docking of the ODF overlay tool indeed also happens on the master branch.  And (indeed again), it happens also to the lighting settings (again in function of the ODF overlay tool).

Finally, if preventing the preview from docking can solve the issue @chunhungyeh reported, I'm all for it as well.  Even not taking into account the actual issue at hand, I think it's a good suggestion for the preview tool as well as the lighting settings "tool": at the moment, when you drag around these things in the vicinity of the main window, it's in most cases actually quite annoying that they're trying to dock.  Most of the time, I just want them to only float in front of the main window (e.g. putting the ODF preview somewhere in a corner without it taking up a whole side bar region.  Preventing the ODF preview to dock, also prevents an awkward "empty tab" if it's accidentally being docked in the same region of the real tools.

PS: just googled/wikipedia'd that i3 window manager... I guess I'll leave my opinion on it in the middle of the wicked-to-wacky spectrum.  But it does seem to be one of the ultimate testing grounds for mrview displaying correctly in a variety of scenarios... :grinning:

jdtournier:
Issue reported by @chunhungyeh upon docking of the preview also occurs on Arch Linux (also using Qt 5.4). I'll investigate a little when I have a minute...

jdtournier:
By the way, clearly the other issues (those that also exist on master) should be dealt with as a separate issues...

jdtournier:
OK, I think I've narrowed down the issue. It looks like Qt destroys the OpenGL context for the preview tool when it gets docked, and calls `initializeGL()` again to allow the application to set things up again within the existing context of the main window. This means all the vertex array, vertex buffer and texture IDs are suddenly no longer current - they've already been deleted. But since they've already been instantiated, the classes that I've set up to deal with their management don't try to create new IDs for them - why would they, they've already been allocated... 

So this would require a rethink, I hadn't anticipated that the whole context might disappear from under our feet like this...

jdtournier:
OK, latest commits should polish this off. I've decided trying to support docking of the preview tool into the main window is probably too fragile - it seems wrong to me that the whole OpenGL context would simply be discarded without warning and `initializeGL()` called again, it breaks the whole resource allocation backend. I'm not sure Qt's approach is necessarily going to remain as-is, I'd be surprised if they themselves considered that to be the best way to handle this... 

So to keep things simple, the preview tool can't be docked any more - it's now a straight QWidget displayed using the Qt::Tool window hint. The second commit should also address the issue with the preview disappearing when floating the main tool (now #273). 

One final round of testing and hopefully this should be good to go...?

chunhungyeh:
The ODF preview tool is working well on my laptop. :+1:

thijsdhollander:
Just tested: everything seems to work well now.  Docking or undocking the ODF overlay tool does not affect the ODF preview window anymore; and docking the preview is indeed impossible, which I think is a great improvement in and of itself!

draffelt:
Works fine on my laptop and desktop.

On Tue, Jun 23, 2015 at 11:11 AM, Thijs Dhollander <notifications@github.com

Lestropie:
Windows is good.

jdtournier:
OK, thanks everyone! Just waiting on news from a WIndows install just to make sure everything is good to go...

jdtournier:
OK, spoke too soon... Let's merge.

PR #275, opened 2015-06-22
Off-grid rendering of vector plot tool

rtabbara:
@jdtournier I know we're in the midst of qopenglwidget changes, so didn't want to crowbar this into master unless it's ok with you.

* Supports off-grid rendering
* Other small cosmetic changes


jdtournier:
Agreed. Not worried about interactions with the qopenglwidget branch, they should be totally independent. More worried about the updated syntax branch, but it'll still be a while before we start merging these changes with MRView anyway...

PR #281, opened 2015-06-30
Alternative implementation for geometry shader in tractography tool

jdtournier:
Had a crack at the tractography tool - see what you think. Not as clean-looking as the current implementation (I'll work on that later), but the framerate on my system seems much better with this approach... 

Opinions...?

### DIFF
--- a/build
+++ b/build
@@ -26,13 +26,13 @@ bcolors = {
     "note" : '\033[94m'}
 
 def colorize(s):
-  out = ''
+  out = []
   for l in s.splitlines():
     for st in bcolors.keys():
       if st in l:
         l = l.replace (st, bcolors[st] + st) + '\033[0m'
         break
-    out += l + '\n'
+    out.append(l + '\n')
   return out
 
 
@@ -61,7 +61,7 @@ def pipe_errors_to_less_handler():
       [ fid, name ] = tempfile.mkstemp()
       try:
         fid = os.fdopen (fid, 'wb')
-        fid.write (colorize(error_stream).encode (errors='ignore'))
+        [ fid.write (x.encode (errors='ignore')) for x in colorize(error_stream) ]
         fid.close()
         os.system ("less -R " + name)
       except Exception as e:
@@ -505,10 +505,18 @@ def execute (message, cmd):
         disp ('ERROR: ' + message + os.linesep)
       error ('\nERROR: ' + message + '\n\n' + ' '.join(cmd) + '\n\nfailed with output\n\n' + stderr.decode (errors='ignore'))
       return 1
+
+    errstr = None
+    if len(stdout) or len(stderr):
+      errstr = '\n' + ' '.join(cmd) + ':\n\n'
+
     if len(stdout):
-      error ('STDOUT: ' + message + '\n' + stdout.decode (errors='ignore'))
+      errstr += stdout.decode (errors='ignore') + '\n\n'
     if len(stderr):
-      error ('STDERR: ' + message + '\n' + stderr.decode (errors='ignore'))
+      errstr += stderr.decode (errors='ignore') + '\n\n'
+
+    if errstr:
+      disp (errstr);
 
   except OSError:
     error (cmd[0] + ': command not found')
--- a/cmd/mrtransform.cpp
+++ b/cmd/mrtransform.cpp
@@ -79,7 +79,8 @@ void usage ()
         "specify a 4x4 linear transform to apply, in the form "
         "of a 4x4 ascii file. Note the standard 'reverse' convention "
         "is used, where the transform maps points in the template image "
-        "to the moving image.")
+        "to the moving image. Note that the reverse convention is still "
+        "assumed even if no -template image is supplied")
     +   Argument ("transform").type_file_in ()
 
     + Option ("flip",
@@ -113,8 +114,8 @@ void usage ()
 
     + OptionGroup ("Non-linear transformation options")
 
-    + Option ("warp", 
-        "apply a non-linear transform to the input image. If no template image is supplied, "
+    + Option ("warp_df",
+        "apply a non-linear deformation field to the input image. If no template image is supplied, "
         "then the input warp will define the output image dimensions.")
     + Argument ("image").type_image_in ()
 
@@ -234,6 +235,25 @@ void run ()
   }
 
 
+  opt = get_options ("noreorientation");
+  bool do_reorientation = false;
+  Math::Matrix<value_type> directions_cartesian;
+  if (!opt.size() && linear_transform.is_set() && input_header.ndim() == 4 &&
+      input_header.dim(3) >= 6 &&
+      input_header.dim(3) == (int) Math::SH::NforL (Math::SH::LforN (input_header.dim(3)))) {
+    do_reorientation = true;
+    CONSOLE ("SH series detected, performing apodised PSF reorientation");
+
+    Math::Matrix<value_type> directions_el_az;
+    opt = get_options ("directions");
+    if (opt.size())
+      directions_el_az.load(opt[0][0]);
+    else
+      DWI::Directions::electrostatic_repulsion_60 (directions_el_az);
+    Math::SH::S2C (directions_el_az, directions_cartesian);
+
+  }
+
 
   opt = get_options ("template"); // need to reslice
   if (opt.size()) {
@@ -272,27 +292,6 @@ void run ()
 
     Image::Stride::List stride = Image::Stride::get (input_header);
 
-    Options opt = get_options ("noreorientation");
-    bool do_reorientation = false;
-    Math::Matrix<value_type> directions_cartesian;
-    if (!opt.size() && linear_transform.is_set() && input_header.ndim() == 4 &&
-        input_header.dim(3) >= 6 &&
-        input_header.dim(3) == (int) Math::SH::NforL (Math::SH::LforN (input_header.dim(3)))) {
-      do_reorientation = true;
-      CONSOLE ("SH series detected, performing apodised PSF reorientation");
-
-      Math::Matrix<value_type> directions_el_az;
-      opt = get_options ("directions");
-      if (opt.size())
-        directions_el_az.load(opt[0][0]);
-      else
-        DWI::Directions::electrostatic_repulsion_60 (directions_el_az);
-      Math::SH::S2C (directions_el_az, directions_cartesian);
-
-      stride = Image::Stride::contiguous_along_axis (3, input_header);
-      Image::Stride::set (output_header, stride);
-    }
-
     InputBufferType input_buffer (input_header, stride);
     if (replace) {
       Image::Info& info_in (input_buffer);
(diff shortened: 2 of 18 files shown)


jdtournier:
OK, spoke too soon... Framerate on my laptop (admittedly only a NVIDIA GeForce G105M) is _slower_ with these changes. Go figure... I guess newer GPUs are better equipped to deal with geometry shaders (probably larger GPU cache or something). I'll keep playing around, but there's probably no point getting excited at this stage...

thijsdhollander:
Gave it a quick run.  Overall findings at this point:

1) For me, performance seems, at this point, very similar in the master versus this branch.  I'm testing with 1 million iFOD2 tracks generated from a WM FOD image: it's a 700 MB tractogram.
2) At first sight, the "black ends" I had before, are gone.  Lighting seems to work fine as well, as far as I can tell.
3) `Colour by track ends` is quite broken though: it actually introduces black segments here and there, as well as unlit ones.  The colouring is also wrong (not by the "by track ends" logic).  Switching (away) *from* this mode to another one again, opens up Pandora's box.  From segfaults to funky easter eggs; it's all there.  Unless this one is intentional:

screenshot from 2015-07-01 11 45 16

:rainbow: :sparkles: It's pretty though. :heart_eyes:

rtabbara:
@thijsdhollander I think the handling for colour by track ends has yet to be implemented which is probably why you're getting this funky behaviour.

@jdtournier If we can get away with outputting only 4 rather than 6 vertices per segment then I definitely think it's worth pursuing this approach.

If it's possible, it might further help to try to move more of the computation into the vertex shader. As suggested [here](http://gamedev.stackexchange.com/questions/48432/why-does-this-geometry-shader-slow-down-my-program-so-much), making use of other parts of the pipeline while the GS is bottlenecked is likely to help. Otherwise, it seems very strange that generating a simpler primitive would slow things down.

jdtournier:
Mmh, I thought I had the colour by track ends working the last time I checked... I'll look into it. The one thing I don't understand though is the `scalarfile_by_direction` option - I can't see what it's supposed to achieve even in principle. What am I missing here...?

As to frame rate difference, after some slightly more 'formal' tests, I have a feeling they're about the same for both master and the current branch on my laptop (would be nice to have #282 just for that :). So at least it doesn't seem to hurt performance.

@rtabbara, you'd mentioned trying to use the vertex shader as much as possible before, so I looked into it briefly yesterday, and also at the StackOver post you linked to above. I'm not sure I agree with the recommendation of moving as much work to the vertex shader though - unless we were using indexed rendering, which would allow caching of previous VS outputs, but this isn't the case here. What seems to matter most is (a) using a geometry shader at all in the first place; (b) how much output each invocation of the GS produces (and hence how much needs to be stored on the limited GPU cache); and probably (c) how many registers each invocation requires (although I haven't seen that last one mentioned yet, but it would affect how many distinct calls can be run simultaneously on the same core - more instances means reduced impact of latency in data fetches). It might very well be that (b) & (c) are actually the same thing, but from my limited reading on GPU programming, it seems they have distinct registers and shared on-die cache, so probably not.

So with this in mind, I'm thinking that anything that isn't a strictly per-vertex calculation should be left alone in the GS - none of the changes we'd make would affect the amount of data produced by the GS, and none of the changes we're talking about would avoid the use of the GS altogether. Where they might have an impact is on the number of registers needed by the GS, and that would affect performance, but I don't think any of the changes we could conceivably make would have a significant impact on that either - we could potentially shave off one or two registers, but I'd be surprised if it was any more than that, the shader compilers are pretty aggressive with their optimisations, and keeping the number of registers low would be very high on their priority list...

Which basically means that I don't think we should worry about this a huge amount. Anything that needs knowledge of adjacency can be done in the GS, everything else in the VS (apart from the colour by direction thing, that needs more adjacency than the GS can provide without resorting to lines_adjacency, and that would place severe limitations on the number of vertices in each streamline). 

That said, there is still some hope that this can be done without the GS altogether. Maybe I'll look into it at some point, but the chances are it'll require so many redundant calculations in the VS that it'll negate any advantages of avoiding the GS...

jdtournier:
OK, had a good go at getting this to work well, I think it's about as good as I'm going to get it. Handles my 1.6Gb, 500K iFOD1 tracks file with sub-second framerate on my system, but then the current master seems to perform just as well, if not better... Go figure. I have a feeling this might be due to heavier downsampling on the master - I changed the criteria for downsampling a bit, it won't downsample quite so readily. Hopefully that'll explain the difference in performance - otherwise I've kinda wasted my time here...

thijsdhollander:
Ok, seems like the `colour by track ends` bug is successfully fixed now!  Using the FPS counter as a pragmatic reference, I'm still getting similar performance on master as compared to this branch (there's no clear winner, it sort of depends on how (wildly) I interact with the tractogram).  But taking into account the downsampling kicks in less quickly... I guess that's a good thing (while maintaining performance).
Overall, I find the current balance between number/size of tracks and usability (framerate) actually quite a good one.  For denser tractograms, I'm still getting a workable framerate when zooming in up to where I need to zoom in anyway to differentiate individual tracks in such a dense environment.

rtabbara:
Seems to run fine for me as well, however strangely the FPS labels do not appear with my build (using Qt5.4).

jdtournier:
@rtabarra, not sure about the FPS not showing up, I'll investigate. So far I'd limited my testing to Qt4 to avoid any funny issues. Hopefully it'll be something simple...

On a different note, I tried this branch (well, 2 commits back, but close enough) on a 14.04 system with ATI hardware (HD7500) and the stock fglrx drivers: segfault deep in the driver library when compiling the geometry shader... Obviously a driver bug, but I'm not sure what to make of that. It wouldn't be great to ship code that we know will crash out on really common configurations... 

Given that the setup is now more focused towards the vertex shader (it's actually closer to @rtabarra's implementation than a few commits back), it might be time to look into ditching the geometry shader altogether. It will mean twice as many invocations for the vertex shader, but that might actually still be quicker than using any form of geometry shader - and would stand a better chance of not crashing out on different platforms... I might give this a shot if I find the time.

jdtournier:
OK, FPS counter in Qt5.4 issue fixed.

jdtournier:
Had a go at getting a version going without a geometry shader - it wasn't too hard, although the approach is admittedly a little unorthodox... I've also simplified a few things that didn't seem to particularly help performance in my limited testing (essentially the early discard for segments outside the FoV, since there's no easy way to do this without a geometry shader).

In terms of performance, it made essentially no difference: my system maxes out at ~130 FPS for a 100K iFOD2 track file, regardless of the approach. For the 500K, 1.6 GB iFOD1 monster, I get ~38 FPS with the previous approach, ~30 FPS with this new one. So even then, it's no big deal. I'll try it on my laptop, which I anticipate isn't so keen on geometry shaders, see if there's any major differences there. 

In general, I think we'll probably be better off not using the GS if that means it's more likely to work across different systems...

Please have a play with this, see if there's any obvious issues (there's plenty of problems introduced by Qt5.4, but I won't bore you with that here). Colour by scalar file in particular needs testing - haven't tried that at all...

jdtournier:
Right, gave various versions a shot on my laptop using the FPS counter (I've added a new `tractogram_alternate_noGS` branch for the GS-free version, and reverted the `tractogram_alternate` branch to the simplified geometry shader version). Basically, testing consisted of loading a 100K iFOD2 tractogram on the command line, getting a reading for the framerate by moving the crosshairs about, then unticking crop to slab and getting the framerate for that. I think the results are pretty clear:

1. tractogram_alternate (with GS): 
    - crop to slab: 14 fps
    - no crop to slab: 7 fps

2. tractogram_alternate_noGS (VS only):
    - crop to slab: 5 fps
    - no crop to slab: 9 fps

3. master (with GS also):
    - crop to slab: 7 fps
    - no crop to slab: 4 fps

So it looks like using a geometry shader definitely helps, if only by discarding a lot of primitives early on (when using crop to slab). Not sure why master was that much slower on my laptop, but maybe those extra quads are slowing things down more than I'd anticipated.

In any event, it really doesn't look like avoiding the GS is a good idea. It seems fine when not cropping to slab, faster than the GS version in fact, but it really loses when cropped to slab. So I think it's probably best to stick with the proper GS version, which is cleaner anyway (so probably easier to maintain), and those with second-rate drivers will just have to update their systems...

Thoughts...?

rtabbara:
:+1: Sounds good to me.

thijsdhollander:
So I gave it a quick shot as well.  Didn't test master, just compared the alternative versions with and without the GS, using a similar testing scenario (but another tractogram obviously).  Just for information's sake, I added in the numbers when using lighting too (results are always "without lighting (with lighting)"):

1. `tractogram_alternative` (with GS):
   * crop to slab: 4.28 fps (2.88 fps)
   * no crop to slab: 2.85 fps (1.67 fps)
2. `tractogram_alternative_noGS` :
   * crop to slab: 1.58 fps (1.58 fps)
   * no crop to slab: 1.62 fps (1.60 fps)

So yes, GS please. :smile: :+1:  I think it definitely matters for those scenarios where one is pushing the limits a bit: it's good to know that doing the "sensible" thing in those cases (e.g. making the slab size thinner while navigating) also results in an expected boost to performance.  I often find myself doing that to get the location/slice/... I need, and then cranking things again up to the settings I need to show off certain things.

jdtournier:
Ok, good to hear. I might try this on MacOSX and that ATI system before doing anything drastic like merging to master...

I might also try a few things to see if we can improve performance any further. The idea would be to move as much work as possible onto the fragment shader, to lighten up the vertex and geometry shaders. These two get invoked for all vertices and primitives, no matter what, and might then throw out a very significant proportion of the primitives at that point, when in crop to slab. When not cropped to slab, most of the fragments will be discarded by the depth test before the fragment shader anyway. So there might be some benefit in reducing the work required to get to the fragment shader stage, even if that means the fragment shader itself then has to work harder.

I might also remove the visibility test - remember that the whole shader function will typically need to run to completion anyway since it's operating in a SIMD context. The rasterisation engine will probably have no trouble in discarding these primitives anyway, that's what it's optimised to do. Obviously, the test based on slab position needs to stay, that's what gives the performance boost when cropped to slab...

jdtournier:
I just committed a few more minor changes, more to clean things up than anything else. I removed the visibility tests, as mentioned above, which seems to have made precisely no difference to performance on my work system - but then it is a bit higher performing than your average. I'll test it on my laptop tonight. I've also checked it on MacOSX and a 14.04 ATI-based system, it runs fine (well, at least it doesn't crash - some nasty issues with incomplete screen updates though, no idea why that would be). Please test the hell out of it.

The only change planned is to move the scalar file colouring from the vertex to the fragment shader, but that's not really necessary. Only problem is that I don't have test data for that - maybe someone can share some with me...?

I might look into the partial screen refresh issue on ATI if & when I get the chance. This isn't specific to this branch (it's on master too), or to the tractography tool for that matter. I've now seen it with Qt4 & Qt5, and on 12.04 & 14.04. I'll try running a debug version on one of those systems, see if there's anything obvious. I think I'll create a separate issue for that...

jdtournier:
Just gave this a spin on my laptop - seems the latest version is very slightly faster, but not by much (14 vs 15 fps). Interestingly though, the Qt5 version is ~2 fps faster than the Qt4 version... 

Assuming everyone is happy with this after their own testing, I guess I'll go ahead and merge?

thijsdhollander:
Everything still seems to work fine on my system!  I don't think I'm having issues with incomplete screen updates.  Performance-wise, I'm also getting almost exactly the same frame rates as before.  Funny/interesting thing, is that I'm getting better frame rates now (as compared to before) with lighting switched on.

jdtournier:
OK, everyone happy to merge to master...?

draffelt:
Yep

thijsdhollander:
:+1:

jdtournier:
Done. Thanks!

PR #283, opened 2015-07-02
Connectome tool

Lestropie:
OK, still got some things to fix up in this, but it's at the point where I'd like people to have a play with it and give some feedback before it's distributed to the masses.

The wiki page will hopefully make things more clear; coming in to it with a clean slate will likely be initially confusing, but it's intentionally designed to be very flexible, and designing it around an easier entry for new users I don't think is possible without losing that flexibility, which will be of particular use when NBS2 is released. I should also add some more tool tips...

https://github.com/MRtrix3/mrtrix3/wiki/Connectome-tool-tutorial


jdtournier:
Crap, this is not going to be fun to merge into the updated_syntax branch...

chunhungyeh:
@Lestropie : I had a quick run on the connectome tool today (and I had fun). It is indeed very nice, flexible, and handy for a variety of applications. :+1:  Here are some feedback from my very initial experience:

*    While loading the 'LUT' and 'Config,' the file format is not limited to text/ascii format, which could be confusing to new users who may be surprised that any format is actually supported. It would be beneficial to confine the selection of file format in order to avoid users selecting irrelevant files for the LUT or Config (e.g. today I accidentally clicked the obj file for the config file, but mrview did not give me warning message.) Likewise, when selecting mesh file for node visualisation and exemplar file for edge visualisation, it would be helpful to limit the option to relevant format as well.
*    The exemplar 'streamline' worked well, but the 'streamtube' did not show up on my MacOSX laptop. There was not any error or warning message when I switched to streamtube, so I could not provide any further information so far. Nevertheless, streamtube did work on my Linux desktop computer.
*    The transparency sliding bar for node worked perfectly, but the one for the edge did not work if the node had no transparency (i.e. I could not change the transparency of edge before making nodes transparent).
*    I think the 'node list and selection' panel would be used quite often, and therefore it might be good to make a separate window showing this table. Currently, the space for the panel is a bit limited, which would be inconvenient when there are many nodes. In the future, perhaps it would be nice to place just the indices instead of the full names or abbreviations (which might be different case by case) near the nodes in the main viewing window.
*    I think it would be great to have options to select the view position. Currently, the default axial view, for instance, is viewing from inferior to superior. If I want to have a S-I view, I need to do manually rotation with my smooth mouse. Similarly, the users may want to view from the right to left or form the posterior to anterior, depending on which network they are interested in.

That's it. I'll keep you informed when I meet any issues or come up with some ideas about the tool.

Lestropie:
Thanks Jimmy:

* (1). Fixed. Just one of those things I intended to do but forgot.
* (2). I'll need to borrow a Mac at some stage to hunt this one down.
* (3). Fixed. (Note to self: Never write convenience functions at 5AM)
* (4). Not a bad idea. Might try to make this its own dockable widget, that by default docks underneath the connectome tool. That way the default will not look too different to current, but the node list can be moved independently, e.g. docked on the opposite side or left floating so the list is longer. I also want to 'compress' the list vertically (i.e. have each row occupy less pixels vertically), but haven't yet figured out how to.
* (4.1). (Showing node names in the main window) This is definitely on the to-do list. But I don't want to have this functionality delay the release; people can always add their own annotations for the sake of figures. I also want to discuss ideas regarding how this could be done with people first.
* (5). Agreed; this is not specific to the connectome tool though. Particularly in OrthoView, I'd like to be able to control which direction each projection is generated from. Created new issue #286.

jdtournier:
Just to give a few hints with some of the above:

- setting the font size could be done using an approach similar to what is done in `src/gui/mrview/tool/base.cpp`, line 44. Note that this is already supposed to set the font size for tools to 2 points less than the default - I'd be wary of reducing the font size further. That said, there's every chance this approach doesn't work as advertised - I seem to get different results depending on which version of Qt I'm using, although I haven't looked into it.

- displaying text on the display should be trivial. Take a look at `src/gui/opengl/font.h`, and its actual use for annotations in for example `Projection::draw_orientation_labels ()`, in `src/gui/projection.cpp`, line 107.

- I guess this is the rationale behind #286...? As discussed on that issue, I don't think we want to introduce any ambiguity in the display's viewing convention. I appreciate most users of the connectome tool will probably want or expect neurological convention, but they're at least not about to make any clinical decisions on that basis... That said, there may be something we can do about this - I'll expand in #286.

Lestropie:
1. I don't think it's the font size so much as the addition of the pixmaps that's causing the thick rows. I'm sure it won't be too hard to fix though.

2. A 2D text display isn't too hard. But you have to choose an appropriate location to display it, manage shortened node names in the list (including manual editing since most lookup tables don't provide them), provide control for which labels are displayed and which are not, etc.. There's also a few other possibilities (e.g. colouring the labels to match the nodes, drawing labels in 3D space rather than directly on the 2D projection plane) that I'd like to discuss with people. So I'd rather leave this out for now and add it later; its absence won't make the tool unusable.

3. Yep. Kind of blurted it out without thinking about it a great deal, I'll need to think about it harder at some stage when my brain is working.

Another thing we need to consider, only slightly related to this, is the lighting. At the moment the connectome tool assumes the light position is defined in real space, and therefore does not rotate with the camera. If I change it to become consistent with other tools (i.e. light position is defined in projection space), OrthoView may look weird as the light position would be different in the three views...?

jdtournier:
Cool, just thought I'd add my 2 cents...

Interesting issue with the lighting in orthoview... Does it actually look that weird if the lighting is anchored to the camera...?

Lestropie:
> Interesting issue with the lighting in orthoview... Does it actually look that weird if the lighting is anchored to the camera...? 

Don't know, haven't tried it. I think initially I got confused by the lighting implementation in the shaders, and by process of re-doing it the light ended up in a fixed position. I also find that when the light position is fixed in real space and you rotate the camera, you get a better contrast between the diffuse and specular components. But it should be consistent between tools, and that was my main concern with going to an anchored light.

jdtournier:
So the issue with a fixed light position is that in some views the light will end up behind the object - it might make for some dramatic screenshots, but not necessarily optimal for viewing...

To get around that, we could change the lighting shaders to assume two lights with opposite positions - this probably wouldn't be too much extra work for the shader, I think the only thing that'd need recomputing would be the specular component. And the UI to control the light position would remain as brain-dead simple as it is now. Might be worth a shot?

You could add more lights, but at some point it'll start to look like ambient lighting, so you can't push that too far. And you'd need controls for all of them too...

Lestropie:
Yeah, that problem is currently highlighted as the viewer initializes to an axial slice but the default light position is superior. I think I'll just revert the lighting in this tool to the standard behaviour; it was mostly me trying to wrap my head around the shader code.

Eventually with central lighting settings I'd like to have the option of multiple light sources; shouldn't be too hard UI-wise or shader-wise. I think I'd prefer that to forcing dual lights with opposite positions. Should look OK as long as the diffuse components aren't too high and the specular exponent is large enough.

jdtournier:
Multiple lights might be good, but you might find it rapidly gets very messy... You'd have to update all the relevant shaders to match, and there may be an impact on rendering performance (although probably nothing to worry about a great deal). 

But then, maybe we could move to [spherical harmonic lighting](http://www.eecs.berkeley.edu/~sohamum/files/spharm_lighting.pdf) - it even involves a bit of spherical convolution! I think that would be most appropriate for MRtrix...

Lestropie:
OK @chunhungyeh, node list is now a separate dockable widget and the rows are much more narrow to fit more in. Break it if you can!

chunhungyeh:
@Lestropie : I had another fun time playing with the connectome tool today. And nope, I couldn't break it (should I feel sorry about it? :smile: ). Here are some thoughts from my experiment today:

*    The dockable node table looks great. It is very handy and may actually have several extensions. I have thought about highlighting the node when it is selected from the node list (e.g. higher intensity/brightness for the selected node), and apparently you've already implemented such mechanism! And even you have allowed multiple selections, that's great! Just a minor issue: when playing with it, I found no way switching off this 'highlighting' function. I finally realised I simply do an 'unselect' by pressing Ctrl+mouse button. I think this would be documented somewhere for the new users?

*    Another feature comes into my mind: I think it would be excellent to automatically highlight the relevant nodes and edges when a node is picked up from the node list. They could be highlighted by the brightness (e.g. slightly lower intensity than the selected one but higher than the rest of the nodes) on the main viewing window and also be reflected on the node table, for example, using different highlight colours. (Currently, it is blue for the select node, so maybe red (for instance) for the associated ones.) This would be great because it means that one may quickly know the importance of the selected node as well as the potential network system in which that node has participated. I think such functionality would make the connectome tool not only for visualisation purpose but also, importantly, for 'exploring brain network through visualisation.'

*    I also verify that the previous issues (file selection / transparency control) have been fixed on my laptop. :+1:

Lestropie:
> I found no way switching off this 'highlighting' function. I finally realised I simply do an 'unselect' by pressing Ctrl+mouse button. I think this would be documented somewhere for the new users?

Yep: QT provides a small list of selection 'behaviours' for such lists; the one that's most appropriate in this case requires a ctrl-click to deselect, which also means if you've currently got multiple nodes selected, you need to reduce the selection to one node before you can ctrl-click to achieve zero node selection. I think that rather than documenting, I'll add a "De-select all" button.

> I think it would be excellent to automatically highlight the relevant nodes and edges when a node is picked up from the node list. They could be highlighted by the brightness (e.g. slightly lower intensity than the selected one but higher than the rest of the nodes) on the main viewing window and also be reflected on the node table, for example, using different highlight colours. (Currently, it is blue for the select node, so maybe red (for instance) for the associated ones.) This would be great because it means that one may quickly know the importance of the selected node as well as the potential network system in which that node has participated.

I _think_ I get where you're trying to go with this. You're suggesting that once a node is selected from the list, the nodes to which it is connected should also be highlighted in some alternative fashion.

This idea stems from a presumed foundation of a binary connectome; not only do we promote and encourage the use of weighted connectomes, we don't have a 'central' connectome from which data are drawn (individual data files are loaded for different visual properties).

Currently, if a single node is selected, all edges emanating from that node will be highlighted. If a matrix file is used to threshold the visibility of edges, then only a subset of those edges will be visible. It would therefore theoretically be possible to identify those nodes connected via such edges, and apply some other form of highlighting to them. However this would _only_ work if some mechanism is used to limit edge visibility; otherwise _all_ other nodes would be highlighted by such a mechanism (as all possible edges exist until you have data that says otherwise). It would also add an extra row 'block' to the selection highlight settings. But at least the logic is sound enough, and expands to the case of multiple nodes being selected...

> ... 'exploring brain network through visualisation.'

Personally I prefer 'assessing brain network through data-driven analysis'... but that's just me :-P

chunhungyeh:
@Lestropie : I just played with your new commits that have enhanced the node table functionality, and I think that's cool and interesting...pretty sure that it will be a very useful feature. And yes, I do agree with you about the data-driven approach, but I also enjoy those 'visual effects' as they are quite amusing. :smirk:

jdtournier:
:+1:

PR #288, opened 2015-07-09
Streamline length

Lestropie:
Would like some independent eyes on this one. See commit message of 6fc76f5 for particular details.

Main point of contention is src/dwi/tractography/editing/worker.cpp, line 190ish: Because streamline lengths are now dealt with in floating-point lengths rather than point counts, it introduces a floating-point comparison problem. Ie. You don't want the user to request streamlines with a maximum length of 25mm, only to have it reject streamlines with a length calculated as 25.01mm. So I've thrown a bit of a fudge factor in there.

Also: `tckgen` code is _unchanged_. Mostly these changes are to deal with the down-sampling that occurs by default with iFOD2, so aren't applicable while the streamlines are being generated. Disadvantage of not changing the `tckgen` code is the rare use case where `-act -crop_at_gmwmi` is used: if the streamline has a short segment at either end caused by the interface cropping, it may be under the maximum length but get rejected because of the number of points. But I'm not too concerned about this.

### DIFF
--- a/cmd/tckedit.cpp
+++ b/cmd/tckedit.cpp
@@ -117,7 +117,6 @@ void run ()
   // Make sure configuration is sensible
   if (get_options("tck_weights_in").size() && num_inputs > 1)
     throw Exception ("Cannot use per-streamline weighting with multiple input files");
-  // TODO Anything else?
 
   // Get the consensus streamline properties from among the multiple input files
   Tractography::Properties properties;
--- a/cmd/tckstats.cpp
+++ b/cmd/tckstats.cpp
@@ -30,6 +30,8 @@
 
 #include "file/ofstream.h"
 
+#include "math/median.h"
+
 #include "dwi/tractography/file.h"
 #include "dwi/tractography/properties.h"
 #include "dwi/tractography/weights.h"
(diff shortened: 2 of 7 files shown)


jdtournier:
I'm not sure I understand your concern. What's wrong with rejecting streamlines whose length exceed the requested maximum...? I think you'd be better off leaving that kind of decision to the user. If they genuinely want streamlines of length 25.01 to be included, they should request a maximum of 25.1 themselves. Or am I missing the point here...?

Lestropie:
The point is, if the user has used a step size of 1.25mm during tracking, and requests streamlines of 25mm or less, they would expect streamlines with 21 points to be included in that. The fact that two particular streamlines are 24.99mm or 25.01mm is not up to the user, those two lengths are equivalent as far as the user is concerned, it's just how the floating-point calculations have come out. We can't expect the typical user to understand the nuances of floating-point calculations and compensate as such, especially given the current master branch does not exhibit such behaviour.

jdtournier:
Ok, I can see the issue, just not sure how much of a problem it might be in practice. In any case, I reckon I'd be happier if your error margin was a smaller fraction - with such simple floating point operations, you really wouldn't expect errors larger than ~1e-5...

Lestropie:
In cases where it doesn't have the streamline step size _a priori_, it gets the step size from the midpoint, and uses that as the expected step length for all but the first and last points (rather than manually calculating every inter-point distance). So any error in that quantification is going to accumulate based on the number of points.

I was going off the following logic: Say your step size is 1.25mm after down-sampling by a factor of 3 (the default). It's possible that the start & end steps may be as small as 0.417mm if tracking terminates mid-iFOD2-arc. So if you add 0.2mm to your maximum length, you should never include streamlines that are one point longer than you desire. It's effectively compensating for the fact that the operation has changed from integer-based to floating-point-based, by adding a half-integer to the threshold.

jdtournier:
OK, I see. If you're estimating the step size from the streamline, any errors in that estimate would be amplified. Seems good to me!

Issue #290, opened 2015-07-10
"intersection" of clip planes (in addition to union)


thijsdhollander:
Background: mrview doesn't have that typical "3 perpendicular slices/planes in 3D" view that you sometimes see in other viewers.  I just wanted to hack myself something like that using the clip planes.  So I started off by creating 2 axial clip planes, and moving them a bit apart.  Playing with the invert settings for both of them, with the intention of getting a "slab" in between both planes, I suddenly realised it couldn't be done.

The thing: when adding multiple clip planes, you basically get the "union" of the volumes that are left over after cutting of each of the clip planes separately.  So after the first clip plane, by adding additional ones, you can only add "volume"; not chisel off volume.  Yet another way to put it: you can only make concave cuts (like dents) in the volume.  Or also: you can not cut off the top of the brain as well as the side.  I hope it's a bit clear what I'm trying to describe here... :smile:

jdtournier:
Yes, this was a question that I faced when designing this feature. I figured the most useful scenario would almost invariably be what you call the 'union', so that the objects you're trying to view can be seen in as much context as possible. I thought about doing something like what you suggest, but with an arbitrary number of clip planes, it gets _really_ messy if you want to support all the different combinations possible... Personally I don't see much point in it, but if you have a solid use case, we may need to think about this some more.

As to the 3 orthogonal planes view, you really don't want to use the volume render for that, it's a sledgehammer to crack a poppy seed... There's a full-blown whole-volume ray-trace going on in there. All you want is to use the already existing slice render backend, and just modify the positions of the corners for each texture. But that assumes you really think this view mode is useful... I had that implemented in a precursor to MRView (way back circa ~2003...), but never bothered reimplementing it for later versions - I've never missed it much...

jdtournier:
Yes, this was a question that I faced when designing this feature. I figured the most useful scenario would almost invariably be what you call the 'union', so that the objects you're trying to view can be seen in as much context as possible. I thought about doing something like what you suggest, but with an arbitrary number of clip planes, it gets _really_ messy if you want to support all the different combinations possible... Personally I don't see much point in it, but if you have a solid use case, we may need to think about this some more.

[...]

PR #309, opened 2015-08-01
A few additions to the clip plane handling

jdtournier:
As discussed on issue #290.

### DIFF
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -68,7 +68,7 @@ void usage ()
     + "See also 'mrcalc' to compute per-voxel operations.";
 
   ARGUMENTS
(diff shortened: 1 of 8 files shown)


jdtournier:
I do think there's some useful features in there. Even just disabling clip plane highlighting is actually quite handy - it's true the highlight can obscure features otherwise. The intersection mode is interesting - not sure how often I'd use it in practice, but it's very unobtrusive and easily supported, so we should keep it. Only question is whether we use one of the newer options for the highlighting, or stick with the status quo...

jdtournier:
No idea what happened here... Ignore this one, obviously pressed the wrong button somewhere - see #310 instead.

PR #310, opened 2015-08-01
Clipplane interaction

jdtournier:
See #290 for prior discussion.

I do think there's some useful features in there. Even just disabling clip plane highlighting is actually quite handy - it's true the highlight can obscure features otherwise. The intersection mode is interesting - not sure how often I'd use it in practice, but it's very unobtrusive and easily supported, so we should keep it. Only question is whether we use one of the newer options for the highlighting, or stick with the status quo...

### DIFF
--- a/src/gui/mrview/mode/volume.cpp
+++ b/src/gui/mrview/mode/volume.cpp
@@ -20,6 +20,7 @@
 
 */
 
+#include "file/config.h"
 #include "gui/opengl/lighting.h"
 #include "math/vector.h"
 #include "gui/mrview/mode/volume.h"
@@ -72,6 +73,14 @@ namespace MR
         {
           std::vector< std::pair<GL::vec4,bool> > clip = mode.get_active_clip_planes();
           const bool AND = mode.get_clipintersectionmodestate();
+          std::string clip_color_spec = File::Config::get ("MRViewClipPlaneColour");
+          std::vector<float> clip_color = { 1.0, 0.0, 0.0, 0.1 };
+          if (clip_color_spec.size()) {
+            auto colour = parse_floats (clip_color_spec);
+            if (colour.size() < 4) 
+              WARN ("malformed config file entry for \"MRViewClipPlaneColour\" - expected 4 comma-separated values");
+            clip_color = colour;
+          }
 
           std::string source = object.declare_shader_variables() +
             "uniform sampler3D image_sampler;\n"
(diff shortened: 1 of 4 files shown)


thijsdhollander:
So 3 separate things:

1) This is definitely much better now!  I also like that the highlight-logic is independent from the other intensities.  If you tune the alpha value of the main image way down, the highlights remain the same -- as they should, because they're an interface element, much like the focus crosshairs, etc... (currently in master, the highlight still fades away with alpha loss as well).

2) It's also a robust solution for the "I don't know where my planes are any more" problem, in any mode (union/intersection).  Perfect as such.

3) One thing that I was initially also trying to cram in, was some sort of visual hint of the "polarity" of a clip plane (that would also only show as part of the highlight).  The ghosts in intersection mode were an (confusing, admittedly) attempt at that.  Having said that, I think the current state of things could potentially be extended to include a better hint as well, if wanted.  I'd think of making the highlight slab thicker (`selection_thickness`, that is?), and having it fade out with distance to the actual plane, *but only on the relevant side* (basically the side where bool `show` applies for the default union mode).  Now, I thought of quickly testing this and showing it off here, but as part of that I ran into something weird... where I'm not sure whether it is correct/intentional in the current state of the code.  What I basically started off doing --just for testing purposes-- was making the highlight slab way (too much) thicker, by changing line 477:

`gl::Uniform1f (gl::GetUniformLocation (volume_shader, "selection_thickness"), 3.0*step_size);`

to

`gl::Uniform1f (gl::GetUniformLocation (volume_shader, "selection_thickness"), 100.0*step_size);`

...but then I started noticing the highlight slab's actual thickness appears to change in the viewer, depending on the angle you're looking at it... Here's an example (after the above manual change -- didn't commit it) of what I mean.  So, volume initially at coronal viewing orientation, sagittal clip plane added.  Then I each time deselected the clip plane, so I could rotate the volume a bit (without changing the clip plane's relation to the volume!), and then selected the clip plane again to show the highlight.  As the viewing angle becomes more perpendicular to the clip plane, the highlight slab becomes thicker...











So before I would change anything based on the current state of the code: is this right?  If so; please explain. :stuck_out_tongue_winking_eye:

jdtournier:
Well, isn't it obvious...? 

OK, maybe not... This fudge of making the thickness of the selected plane dependent on the viewing angle was to get around an issue I had with the commit immediately prior to that one: the highlighted plane would appear much more transparent when viewed face on. This was because you'd end up with more samples through the plane when it was tilted, just because of the geometry of it, and more samples mean more opacity. So making the thickness depend on the viewing angle means I was able to guarantee an average of 1 sample through the plane regardless of its angle. Try comparing 3df88ed with 750fe38 to see what I mean...

thijsdhollander:
Good good, it's indeed all clear to me now.  I guessed it was something like that, but I think my intuition was always used to treating the clip plane highlight as an actual "real" translucent surface/slab.  Given that they have a certain thickness, the more opaque appearance at more shallow angles actually always made sense to me (it's just plain optics).  So I guess my mind broke a little bit when you (intentionally) "broke physics" there. :smile: 

It does make my suggestion to have the fading-out hint for the polarity a quite a bit more annoying to deal with; but then again, having such a thing would by itself cause trouble with opacity from certain angles as well (and for a range of viewing angles probably make those hints *a lot* less useful as well).  So I think at this point, it's probably best to leave things as simple and clean as possible.  Having the planes highlighted even when they don't intersect with anything, is already a huge step forward.  At least there's always some visual feedback now.

I've tested the current state of things extensively now, and I find it actually works really well in practice.  If the plane highlights do get a bit "in the way" (especially when selecting and manipulating multiple ones as a group sometimes), there's always the highlight check box now to toggle them.

So I will merge this one for now; it's a nice step (strictly) forward. :+1:

jdtournier:
All good. I did have an idea about providing visual feedback related to the 'polarity' of the clip plane: use a different colour if the plane normal is pointing towards or away from the viewer. So for example red might indicate that content between the plane and the viewer would be clipped away, and green that said content would not be affected. What do you reckon?

Issue #292, opened 2015-07-15
ACT: Enable seeding in SGM


Lestropie:
Something I keep forgetting about...

It should be possible to seed inside SGM, but currently is not.

Need to explicitly detect this case, and:
* Allow the streamline to exit from SGM into WM, but only for one of the propagations from the seed point;
* Reject the streamline if neither propagation reaches WM.

Lestropie:
Was implemented in the `act_modifications` branch, which has now been merged into `updated_syntax`.

Issue #131, opened 2014-10-25
ACT: Tweak back-tracking algorithm


Lestropie:
Was never entirely happy with how this worked; if back-tracking is required to get through one area, then another problem is encountered further along the trajectory, the back-tracking step is immediately large.

What should instead occur is that as successful steps are performed, the back-tracking step should decrease, potentially down to zero (so that if a new poor termination is encountered, only a single step is erased).

Lestropie:
Was implemented in the `act_modifications` branch, and has now been merged into the `updated_syntax` branch.

PR #225, opened 2015-04-21
Updated syntax for Image classes 

jdtournier:
This branch introduces a new cleaner API for the Image classes, as discussed in #211 

Currently, the only command working is `bogus` - and even then it only works on some inputs. Still need to implement preloading and scratch buffer support. A long way to go yet...

But it does make things look a bit cleaner. For example, from `cmd/bogus.cpp`:

```C++
void run ()
{
  value_type power = 2.0;
  auto opt = get_options ("power");
  if (opt.size())
    power = opt[0][0];

  auto in = Header::open (argument[0]).get_image<value_type>();
  auto header = in.header();
  header.datatype() = DataType::Float32;
  auto out = Header::create (argument[1], header).get_image<value_type>(); 

  ThreadedLoop (in).run ([&](decltype(in)& vin, decltype(out)& vout) {
    vout.value() = std::pow (vin.value(), power);
    }, in, out);
}
```

Anyway, not ready to merge by any stretch, but thought I'd give people a sneak preview :smile: 

Once it's stable, I wouldn't mind some help porting all the apps to the new framework...

PR #311, opened 2015-08-03
ACT modifications

Lestropie:
Need to make sure this isn't going to break something, kind of a crucial functionality...

* The original back-tracking algorithm was a bit naive: each time a bad termination was encountered, an integer was incremented, and the streamline was truncated with length according to that integer. This meant that if back-tracking was used to traverse one problematic region, but then another was encountered, the length of truncation would immediately be high. Now, the behaviour is as follows:

  * The length of truncation is 'reset' if the streamline length exceeds that at the previous back-tracking event, i.e. the first problematic area has been passed.

  * If a truncation has occurred, but the streamline has not reached the same maximum length before back-tracking is again invoked, the truncation length is with reference to the _maximal_ streamline length, not the current length.

  * Each length of truncation is repeated for three back-tracking attempts before the length of truncation is increased (trade-off between effectiveness and speed).

* It is now possible to seed streamlines within sub-cortical grey matter. The streamline must reach white matter for it to be accepted. If tracking bi-directionally, only one of the unidirectional streamline projections may enter the WM; if both attempt to do so, the second is treated as exiting the SGM as per the normal ACT priors.

* When approaching the edge of the image, iFOD2 will now terminate with EXIT_IMAGE flag if _any_ calibrator path exits the image, rather than requiring that they all exit the image.

### DIFF
--- a/cmd/tckgen.cpp
+++ b/cmd/tckgen.cpp
@@ -52,7 +52,7 @@ using namespace App;
 
 
 
-const char* algorithms[] = { "fact", "ifod1", "ifod2", "nulldist", "sd_stream", "seedtest", "tensor_det", "tensor_prob", NULL };
+const char* algorithms[] = { "fact", "ifod1", "ifod2", "nulldist1", "nulldist2", "sd_stream", "seedtest", "tensor_det", "tensor_prob", NULL };
 
 
 void usage ()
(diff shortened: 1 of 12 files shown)


Lestropie:
Closes #292.
Closes #131.

Lestropie:
These changes have instead been merged into the `updated_syntax` branch (#225). Figure we might as well push all major changes in a single go. Still needs some testing to make sure it all works as expected though, since this was all written pre-updated_syntax.

PR #313, opened 2015-08-04
Fixel tool threshold fix

Lestropie:
Wanted someone to test before merging.

... Whoops, pushed something that was supposed to go somewhere else. Oh well.

### DIFF
--- a/scripts/revpe_distcorr
+++ b/scripts/revpe_distcorr
@@ -2,14 +2,15 @@
 
 # Script for performing EPI distortion correction using FSL 5.0 tools eddy / topup / applytopup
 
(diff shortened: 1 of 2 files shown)


jdtournier:
Just to say I won't be able to review this until I get back to my desk in September. If anyone else can give it a shot, please do so. Otherwise, let's just merge and wait to see if bug reports start coming in...

thijsdhollander:
Just tested as much as I could with a .msf file, and a narrow thresholding range (i.e. lower/upper thresholds close to each other, so only a few fixels make the cut), going back and forth through the scaling options and the scaling factor.  All seems to work well; once the thresholds are set, tuning other things does not change the actual fixels shown.  Didn't notice any other bugs or otherwise unexpected behaviour.

PR #321, opened 2015-08-12
Spin box

Lestropie:
Couldn't find a better way of doing this. It works, but need to think about whether or not this approach is going to have unwanted effects.


jdtournier:
Nice simple fix - and it works...? I can't see it having undesired side effects, but I do wonder whether it'll turn out to work on all platforms... 

In any case, if it works for you, happy for it to be merged.

Lestropie:
Sure I'll hear about it if it borks something...

PR #322, opened 2015-08-13
Connectome tool list

### LINKED ISSUE #320: Connectome tool: 'Primary' matrix file list view
Hopefully this will give an appropriate compromise between users not being able to figure out the logic behind the tool, and me not wanting to lose functionality I need by simplifying the initial entry point.

Have a list view where matrix files can be loaded, but only one selected at a time. Relevant visual properties can be set through their comboboxes via the 'primary connectome' (i.e. the matrix file selected), or an external matrix file (as is done currently). If the selected 'primary connectome' (lemme know if you think of a better term) changes, all relevant visual properties update.

Importantly, set certain edge-wise visual properties such as edge visibility and edge colour should be set to this by default; so when someone opens a connectome file, they immediately get some sensible visualisation of it. But these properties can also be set via external files that are independent of the particular connectome currently selected.

Also, change the current command-line option for this tool to `-connectome.init` (where the parcellation file must be passed), and `connectome.load` can then be used to load matrix files.

Lestropie:
Enhancing the connectome tool to hopefully make the initial entry point a bit easier for users: Initialise using the parcellation image, load a connectome, and you can see the connectome (with somewhat sensible defaults). Also gives the possibility of scrolling through a set of connectomes (though I still need to enable use of the keyboard arrows in the QListView).

But throwing it up here now in case someone wants to take up the challenge of breaking it, or has any other comments about the UI design, e.g. in #320.

Closes #320.

### DIFF
--- a/src/gui/mrview/tool/connectome/connectome.cpp
+++ b/src/gui/mrview/tool/connectome/connectome.cpp
@@ -90,9 +90,9 @@ namespace MR
             node_fixed_alpha (1.0f),
             node_size_scale_factor (1.0f),
             voxel_volume (0.0f),
-            edge_visibility (edge_visibility_t::NONE),
+            edge_visibility (edge_visibility_t::CONNECTOME),
             edge_geometry (edge_geometry_t::LINE),
-            edge_colour (edge_colour_t::FIXED),
+            edge_colour (edge_colour_t::CONNECTOME),
             edge_size (edge_size_t::FIXED),
             edge_alpha (edge_alpha_t::FIXED),
             have_exemplars (false),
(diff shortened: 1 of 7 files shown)


chunhungyeh:
Verified on MacOSX. The only issue that I have on my laptop is that the current layout of the connectome tool is too long. See the attached image of my screenshot: I could not see the connectome node list unless I undocked the window. Perhaps it might be good to decrease the font size or gap between the parameters, or even make the 'node visualisation' and 'edge visualisation' sections foldable?
 
screen shot 2015-08-17 at 8 56 11 am

Lestropie:
Hmmm, looks like the vertical spacing in the grid layout is larger on your system, even though it's set explicitly. Either that or it's adding some space where there are empty rows. I'll have to play with this on a Mac.

I also need to move the LUT loading back out from the node colouring to the main setup. In cases where the atlas is already provided with sequential indices from 1, there's no need to provide a config file, you can just provide the LUT to get both node names and colours. So I might reconsider moving that and the config file load out of the 'basic setup' to somewhere else.

I can also potentially lose the 'node selection' label, and replace it with a button (with an appropriate icon) right near the top.

I have commented previously that I would like to make a foldable group box; added as Issue #325.

Lestropie:
OK, just did a little bit of fiddling of the UI ordering. Should be able to merge this independently of the `updated_syntax` work without too much drama.

@chunhungyeh We need to look at what's going on on your Mac.

Lestropie:
OK @chunhungyeh: Give that a spin, see if it further reduces the tool sizing issue on Mac.

PR #340, opened 2015-09-02
Argparse

Lestropie:
OK, I think this stuff is just about ready to merge. Would anyone like to test a couple of scripts with it beforehand? I think I've now gone through all of the scripts at least once, unfortunately FSL is playing funny-buggers on my work system (writing hexadecimal floating-point notation to output text files and not reading it) which has limited my testing.

### DIFF
--- a/scripts/act_anat_prepare_freesurfer
+++ b/scripts/act_anat_prepare_freesurfer
@@ -25,11 +25,10 @@ from lib.errorMessage import errorMessage
 from lib.printMessage import printMessage
 from lib.runCommand   import runCommand
 
-numArgs = 2
-if len(sys.argv) <= numArgs:
-  sys.stderr.write('Usage: ' + os.path.basename(sys.argv[0]) + ' <input freesurfer parcellated image> <output image> (options)\n')
-  exit(0)
-lib.app.initialise(numArgs)
+lib.app.initParser('Generate a 5TT image suitable for ACT from a FreeSurfer segmentation')
+lib.app.parser.add_argument('input',  help='The input image series; this should be one of the \'aseg\' images provided by FreeSurfer')
+lib.app.parser.add_argument('output', help='The output 5TT image')
+lib.app.initialise()
 
 config_file_name = 'FreeSurfer2ACT.txt'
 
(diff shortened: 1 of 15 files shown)


bjeurissen:
Got exactly the same problem with FSL on my arch box! What OS are you running? We should compare the logs of our package managers to track down the update that is causing this...

Lestropie:
Pretty confident it's an Arch thing; have it on all three of my systems and all three have the bug, can't reproduce with the same data on any other system. For now I'm just hijacking other systems to run the FSL-dependent scripts and patiently waiting for a fix...

jdtournier:
I have a sneaky suspicion I've seen this bug on a Ubuntu 14.04 system... I'll check later today.

jdtournier:
OK, confirmed we did have the FSL issue on a 14.04 system, but it's no longer present... Current version of FSL on that system is 5.0.8-3~nd14.04+1 from NeuroDebian. I notice my Arch system is still running 5.0.8-1, even though I did an AUR update this morning... I think the problem is that FSL updates their generic 5.0.8 package surreptitiously without listing the release number, but the SHA1 of a fresh download of the 5.0.8 package doesn't match what's in the AUR pkgbuild... And the version history makes no mention of any release numbers - it's just 5.0.8... 

Weirder still, if I download the sources via the official page (having filled in my details), the file I get doesn't match the file I'd get using the exact same URL via a straight download (using `curl`). I assume there's some strange system in place in the backend that only gives you the up to date version if you've filled in your details, which is pretty strange given that the recommended installation route is via neurodebian, which wouldn't allow them to get that information anyway. No wonder the AUR package maintainer is having trouble keeping up...

I was thinking of tweaking the PKGBUILD to feed back to the maintainer, but I'm not sure it's possible given these issues... Unless someone can figure out how to pass the relevant POST information through curl or something...?

jdtournier:
Can' figure out how to automate the FSL download so it gets the latest version... In any case, for those interested, the workaround for Arch is to download the sources via the webpage (requires filling in your details), get the FSL PKGBUILD from the AUR, put the `fsl-5.0.8-sources.tar.gz` file in that folder, get the SHA1 checksum (`sha1sum fsl-5.0.8-sources.tar.gz`), edit the PKGBUILD file with the correct SHA1 and update the `pkgrel` entry (I put a 3 in there...), save and run `makepkg`. Seems to work so far...

bjeurissen:
@Lestropie @jdtournier: hexadecimal output seems to be due to locale issue and should be fixed in the latest version according to:

https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;8e169ec.1507

http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/WhatsNew#anchor1

jdtournier:
I can also confirm that the downloads actually match if extracted and diff'd - the mismatch is probably down to using a different compression program or something...

Lestropie:
Now that the FSL bug has been localised... Anyone able to to take this for a spin? I'd like to merge it, it shouldn't conflict merging with `updated_syntax` and the enhancements come in handy from time to time.

jdtournier:
I reckon you can merge, this won't break any other developers' apps that rely on the master branch, and you'll probably get more useful feedback from users if there is a subtle bug somewhere...

PR #343, opened 2015-09-07
tensor2metric: fix assertion failure

jdtournier:
and replicate 588851e4 since this causes failure at testing due to
dimension mismatch.

### DIFF
--- a/cmd/tensor2metric.cpp
+++ b/cmd/tensor2metric.cpp
@@ -83,6 +83,9 @@ void usage ()
 }
 
 
+
+
+
 typedef float value_type;
 
 class ImagePair
--- a/cmd/tensor2metric.cpp
+++ b/cmd/tensor2metric.cpp
@@ -62,7 +62,7 @@ void usage ()
             "specify the desired eigenvalue/eigenvector(s). Note that several eigenvalues "
             "can be specified as a number sequence. For example, '1,3' specifies the "
             "major (1) and minor (3) eigenvalues/eigenvectors (default = 1).")
-  + Argument ("image")
+  + Argument ("image").type_sequence_int()
 
   + Option ("vector",
             "compute the selected eigenvector(s) of the diffusion tensor.")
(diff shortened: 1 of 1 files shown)


bjeurissen:
Ok for me!

I think in most cases where processing a 4D data set results in a 3D data set it makes sense to remove the singleton dimension.

jdtournier:
I'd assume so - it was your commit that screwed up the testing... :wink:

I'm not so convinced about this myself - it very much depends on what command will use the data next and what assumptions it's making. A lot like Matlab, which can often choke on mismatched dimensions even though all the mismatching dimensions are singletons... If we do this, we'll need to make those commands that you'd expect to use these data don't fall over because of this.

bjeurissen:
We can also change the test data to include the singleton dimension, in case only one eigenvalue is requested. It might indeed be slightly more general that way.

However, how is requesting a single eigenvalue map different from requesting any other scalar map (e.g. FA)?

jdtournier:
The issue is that the dimensionality of the command's output would depend on the option parameter, in a way that's hard for a script to deal with. If you knew the `-value` option would always return a 4D image, even if there was only a single volume, your scripts should all be able to rely on that assumption. Changing the dimensionality of the output like this may be a problem if the command you feed the resulting image into expects 4D data. I can think of some trivial instances where this might fail - all somewhat contrived, but not ridiculously so. For example, you've written a script to average the eigenvalues specified (don't ask me why, that's not the point...), and it takes the eigenvalues you're interested in averaging as an argument. So it might look something like this:
~~~ShellSession
tensor2metric $1 -num $2 -value - | mrmath - mean -axis 3 $3
~~~
This would work fine if you left the output as a 4D image, but would probably fail or crash out with the current approach. And it's difficult to handle propertly, you'd have to check whether the output was 4D, and invoke `mrmath` or not depending. So that's the kind of scenario where I see this potentially being problematic... Although I have to say in this particular instance, I can't think of examples that make a great deal of sense scientifically - but that's not to say others might not come up with legitimate uses of that type.

So that's why I think using this approach should be coupled with a design in the general MRtrix backend to handle these cases - or at least some convenience methods to make it easy for commands like e.g. `mrmath` to accept a 3D input as if it were a 4D image with a singleton dimension...

Lestropie:
> The issue is that the dimensionality of the command's output would depend on the option parameter, in a way that's hard for a script to deal with. 

I already have to do this: `dwiextract` command output being processed by the `revpe_*` scripts. But personally I think testing and handling this appropriately makes more sense than always having images floating around with a singleton 4th axis, that you then have to handle explicitly in any binary that expects a 3D image as input.

jdtournier:
@Lestropie - I'd generally agree with that. It's just that at present, there is no formal way of handling this within the MRtrix backend. So it would be good to add something that would allow `check_dimensions()` to relax its tests to allow these types of trivial mismatches, and to ensure the whole `ThreadedLoop` and `Loop` infrastructure is able to cope with that without falling over. So maybe we just need to make a separate issue for that.

jdtournier:
On a different note: it seems this testing before merge things is going to be problematic... The testing failed, so I forced a re-run by closing and reopening the issue. Now we have two tests results, one pass and one fail, and it won't allow us to merge because some of the tests have failed - even though the code being tested was identical. 

This might have something to do with me fiddling about with the settings on TravisCI... But I couldn't get it to rerun the failing test, even though pressing the button reported that it had successfully restarted the job - no updates after that. I'll push a useless commit, see if that triggers anything...

jdtournier:
Since no one's complained...

PR #346, opened 2015-09-09
mrconvert scaling issue

jdtournier:
As discussed on [this MRtrix3 forum post](https://plus.google.com/102850892380704369664/posts/JjvrqdwRCts), the header entries specifying the intensity scaling were not being preserved by mrconvert, when they were in version 0.2.x. This fixes this to preserve these values by default.

It also introduces an option to manually specify these values if needed. For example, `-scaling 0,1` would give the same answer as it does currently.

Note that this did not affect the values read - only how they were stored. And these values would be correct given the scaling parameters that were actually used (i.e. 0,1) - within the precision of the data type. Where this would be problematic is for e.g. an FA map stored using an integer type and rescaled between 0 and 1 using these header parameters: this would convert fine to floating-point, but mrconvert would previously have produced a map of zeros when converting to the same or another integer type. This should fix that problem.

### DIFF
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -72,6 +72,14 @@ void usage ()
             "-1 at the corresponding position in the list.")
   + Argument ("axes").type_sequence_int()
 
+  + Option ("scaling",
+            "specify the data scaling parameters used to rescale the intensity values. "
+            "These take the form of a comma-separated 2-vector of floating-point values, "
+            "corresponding to offset & scale, with final intensity values being given by "
+            "offset + scale * stored_value. By default, the values in the original header "
+            "are preserved.")
+  + Argument ("values").type_sequence_float()
+
   + Image::Stride::StrideOption
 
   + DataType::options()
@@ -91,10 +99,19 @@ inline std::vector<int> set_header (Image::Header& header, const InfoType& input
   header.info() = input.info();
   header.datatype() = datatype;
 
+  Options opt = get_options ("scaling");
+  if (opt.size()) {
+    std::vector<float> scaling = opt[0][0];
+    if (scaling.size() != 2) 
+      throw Exception ("-scaling option expects comma-separated 2-vector of floating-point values");
+    header.intensity_offset() = scaling[0];
+    header.intensity_scale() = scaling[1];
+  }
+
   if (get_options ("grad").size() || get_options ("fslgrad").size())
     header.DW_scheme() = DWI::get_DW_scheme<float> (header);
 
-  Options opt = get_options ("axes");
+  opt = get_options ("axes");
   std::vector<int> axes;
   if (opt.size()) {
     axes = opt[0][0];
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -76,8 +76,11 @@ void usage ()
             "specify the data scaling parameters used to rescale the intensity values. "
             "These take the form of a comma-separated 2-vector of floating-point values, "
             "corresponding to offset & scale, with final intensity values being given by "
-            "offset + scale * stored_value. By default, the values in the original header "
-            "are preserved.")
+            "offset + scale * stored_value. "
+            "By default, the values in the input image header are preserved when writing to "
+            "an integer image; when writing to a floating-point image, any scaling present "
+            "in the input image header is applied to the raw image data, such that the "
+            "output image header contains standard (non-influential) scaling only.")
   + Argument ("values").type_sequence_float()
 
   + Image::Stride::StrideOption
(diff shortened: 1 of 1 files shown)


jdtournier:
OK, this seems to expose an issue with the enhancements added by @Lestropie in e10153d (discussed in #200)... While converting with native types is less likely to introduce differences between input and output values due to conversion to a floating-point format, it also prevent proper operation of the intensity rescaling parameters, since these allow mapping integer intensity onto floating-point values... Might need to backtrack on that one...

Lestropie:
Hmmm. If the issue is _only_ in `mrconvert`, could simply take the floating-point execution branch if offset & scaling are anything other than [0, 1]?

jdtournier:
Yes, I was thinking that might be the simplest way to fix this...

jdtournier:
Other issue up for discussion: should these rescaling parameters be reset to defaults (0,1) for floating-point outputs? This would make no difference in terms of precision, but would force a preload for any subsequent command that requests it (otherwise it would probably be memory-mapped directly). Only issue, and the reason I haven't already done this, is that this means converting an image with large scale factor to floating-point and back to integer would most likely introduce nasty discretization errors. Think FA mapped over integers, rescaled from (0,1000) to (0,1). If the rescale parameter is reset, that would essentially result in an image of zeros and ones when converted back. 

So not sure which of these two should be prioritised...

jdtournier:
For the record, the command that triggers the issue with using native types was:
~~~ShellSession
$ mrconvert fa.mif -datatype int16 -scaling 0,1000 fa2.mif
~~~
Causes a floating-point exception on my laptop. Not entirely sure why, I must admit... Works fine if scaling parameters are 0,1.

jdtournier:
OK, above command executes fine on my desktop - no crashing out with floating-point exception. But the output image is clearly rubbish (pure zeros).

jdtournier:
OK, having thought about it a little, I've decided it's probably safe to reset the scaling parameters to (0,1) for floating-point outputs - see message in commit 928cdc0 for details...

jdtournier:
OK, good to go?

Lestropie:
If someone converts a scaled integer image to floating-point, then back to integer using anything other than precisely the same scaling & offset values, and complains about discretisation errors, they only have themselves to blame :-P

Just tested back and forth a couple of times, seems to do as it should. Agree that resetting those scaling parameters when going to float is probably best - worth having a `-keep_scaling` option in `mrconvert`? Also updated the help info for the `-scaling` option.

Couldn't reproduce your error on this machine.

jdtournier:
Good to hear. Thanks for testing and updating the doc.

> worth having a -keep_scaling option in mrconvert?

Good suggestion. Although I'd be tempted to use the current `-scaling` option, and interpret `-scaling 0,0` or `-scaling nan,nan` as a request to preserve scaling parameters (ahem, if properly documented, of course...). I don't expect this to be a high use case...

> Couldn't reproduce your error on this machine.

Did you try with `-scaling 0,0.001`? I updated my post, since `-scaling 0,1000` wasn't triggering the error... Besides, the image produced once `mrconvert` was fixed was useless with that setting, it's the wrong way around.

jdtournier:
OK, decided to detect `-scaling preserve` instead - it's probably cleaner - and documented too...

jdtournier:
Let me know if any further opinions - I'll merge tomorrow if all good.

Lestropie:
OK, so this just happened to pop up as an issue today. Not in `mrconvert`, but another command. The input image was stored as integer with a multiplier of ~ 1e-5, the multiplier was not preserved in the output header, and correspondingly the output image was almost all zeroes.

Arises because `Image::Header` resets the offset and scale on copy-construction. The above changes get around the issue for `mrconvert`, but it doesn't generalise to other commands.

Perhaps header copy-construction should preserve these if the data type is an integer, but reset if it's floating-point, and also reset if the datatype is manually changed to floating-point?

jdtournier:
Yes, that might be sensible. I have to admit, I've never quite figured out a neat way around these scaling issues... Problem is we have to support this since DICOM & NIfTI (and mif) all allow it. But it does make things tricky when the output image can have any datatype that may or may not be compatible with the range. 

I'd set up the default of explicitly resetting these values on Header copy since _most_ commands will produce floating-point outputs, but I agree it's not going to be appropriate in all cases. Problem is, even if the datatype doesn't change and the rescale parameters are preserved, there is no guarantee that these parameters will still be suitable after processing - it very much depends on what is actually being done to these images... There is a `Header::set_intensity_scaling (const Header& H)` function to allow commands to explicitly copy these values across where that might be sensible, but I agree, once again it would depend on the output datatype as to whether this might or might not be appropriate.

Maybe your suggestion is the most sensible default all-round though. I would also add bitwise to the list of datatypes where the scaling parameters would be reset though, since it would be extremely unlikely for someone to actually want their mask scaled funny (it certainly would likely introduce problems in other apps that rely on the mask values being in the range [0 1]). But other than that, it sounds like a good plan. I'll re-open the branch and give it a crack...

In the meantime, any other opinions on the matter are of course most welcome...

PR #347, opened 2015-09-09
Fix configure to support building with msys2 on windows.

bjeurissen:
Fix OS being reported as mingw_* instead of windows on msys2.
Fix configure to handle include paths relative to tmp.
Replace GL_ with gl:: to make OpenGL related stuff compile with Qt 4.

### DIFF
--- a/configure
+++ b/configure
@@ -366,6 +366,8 @@ exe_suffix = ''
 lib_prefix = 'lib'
 
 system = platform.system().lower()
+if system.startswith('mingw'):
+  system = 'windows'
 report ('Detecting OS: ' + system + '\n')
 if system == 'linux':
   cpp_flags += [ '-pthread', '-fPIC' ]
@@ -876,8 +878,6 @@ int main() { Foo f; }
       f.close();
 
       qmake_cmd = [ qmake ]
-      if system == 'windows':
-        qmake_cmd += [ '-spec', 'win32-g++' ]
 
       try: 
          (retcode, stdout, stderr) = execute (qmake_cmd, QMakeError, raise_on_non_zero_exit_code = False, cwd=qt_dir.name)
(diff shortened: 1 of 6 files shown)


Lestropie:
Compiled, built & run on my Windows box.

jdtournier:
@Lestropie, can I just confirm this was using the standard Qt download as recommended on the wiki?

Lestropie:
No, I'm currently on (this)[http://sourceforge.net/projects/qt64ng/files/qt/x86-64/5.4.1/mingw-4.9/sjlj/]; it's newer than what's offered by MinGW-builds, but looks like the project has crashed and burned.

jdtournier:
@Lestropie, interesting that you're on Qt 5.4 - can you confirm that everything works as expected in MRView...? @bjeurissen & I are slightly concerned about things breaking with the move to Qt's `QOpenGLWidget` - seems to cause issues on his system. But then it could be because he's running a full 5K display, which most likely requires special handling due to HiDPI rendering...

jdtournier:
OK, just gave it a shot on my Windows 7 VM - `configure` ran fine with the Qt 5.2.0 download as per the wiki. I think this is good to merge - the discussion about `QOpenGLWidget` doesn't have any impact on that.

Lestropie:
Haven't had any issues specific to Qt 5.4.

jdtournier:
Good to hear. 

One thing worth fixing though is that `./configure` falls over on my system due to a space in my user profile path... The only way around I could find was to rename my user folder - not entirely trivial. It would be good to fix that properly. The issue is I think with importing the `CFLAGS`, `LDFLAGS`, and `GSL_CFLAGS` & `GSL_LDFLAGS` environment variables - currently they're imported using the `shlex.split()` method, but that relies on spaces having been escaped properly when setting up the variables. I can't think of a trivial fix for this...

bjeurissen:
Just tested the new configure using the wiki instructions and it works fine. So I guess this branch can be pulled then? Or should we also fix the issue with the spaces?

@jdtournier : have you tried using msys2? It could be that the spaces are not an issue there...

jdtournier:
Just looking into the space in filename issue - there's really no trivial fix for this. Even if everything is escaped properly so that the arguments are properly split (I can verify they are immediately prior to executing the command), the compiler fails to find `zlib.h`, despite the correct folder being listed in the include path. 

I'll try msys2 now...

jdtournier:
Yep, msys2 works a treat - no trouble running `./configure`. It's really good to work with too - much better than the git bash approach...

Issue #299, opened 2015-07-22
MRView: handle OpenGL resource deallocation at exit


jdtournier:
There is a problem with releasing OpenGL resources, explained in the [QOpenGLWidget Qt docs](
http://doc.qt.io/qt-5/qopenglwidget.html#resource-initialization-and-cleanup), in that the context is not guaranteed to be current when members classes are destructed - in C++, destruction occurs in the opposite order to construction, so the destructor for the OpenGL context will be invoked after all the destructors for the member objects. This means there is no mechanism to ensure the right context is current when these members are destructed - and that includes all the textures, framebuffer objects, vertex array objects, etc. This is bound to cause random segfaults somewhere down the track. It would be good to figure out a way to handle this more cleanly. The `aboutToBeDestroyed()` signal is one such option.

Lestropie:
I reckon this has been dealt with in full; I certainly haven't been able to get any errors of late (which includes a number of fixes that were done on the `odf_dixel` branch and are now in `updated_syntax`. So closing this now; if errors pop up down the track, the framework for handling it is in place, and they can be dealt with on an individual basis.

Issue #21, opened 2014-05-08
MRview: segfault on close


Lestropie:
This one may be a little hard to track down... Initially I thought it was only a Windows thing (get the error popup window after closing), but now it looks like Linux is segfaulting on close as well.

There's a chance that the problem is buried deep within QT somewhere; I get a lot of "Conditional jump or move depends on uninitialised value(s)" within shared libraries. Backtrace on gdb is completely useless.

Thought I'd just list what little we have, so maybe an idea comes out of it.

Got this from valgrind on Linux:

```
==9519== Thread 2 QXcbEventReader:
==9519== Jump to the invalid address stated on the next line
==9519==    at 0xEF562B9: ???
==9519==  Address 0xef562b9 is not stack'd, malloc'd or (recently) free'd
==9519== 
==9519== 
==9519== Process terminating with default action of signal 11 (SIGSEGV)
==9519==  Access not within mapped region at address 0xEF562B9
==9519==    at 0xEF562B9: ???
```

Had noticed at one point (think it was on my home Windows machine) that I'd get the crash window popping up if I opened MRview on its own, but if I provided an image path on the command-line, I wouldn't get the error window...?

May have come from moving from QT4 to QT5; I certainly wasn't getting segfaults on my work machine previously. And I seem to recall when I did my first Windows installation that I used a pre-built QT4, and I wasn't getting the error popup on close like Donald said he had with a Windows install. Edit: Yup, no segfault on QT4. Do get a whole lot of invalid reads deep within QT code though.

bjeurissen:
Just tested on Windows 10  with QT 4.8.7. I still get the segfault message on the commandline on closing mrview (unless I provide an image path on the commandline).

bjeurissen:
Same problem with Qt 5.5

jdtournier:
> unless I provide an image path on the commandline

Well that's interesting... Any chance we can get a backtrace on a debug build...?

jdtournier:
Just thought I'd add: there is a chance this might be related to #299...

PR #348, opened 2015-09-14
Fix opengl dealloc

jdtournier:
An attempt at addressing #299 and potentially #21. This introduces more thorough checks in debug mode to verify that the original OpenGL context is current when resources are deleted. These were used to ensure that all deallocation happen in the right order, and under the right context. It seems to fix most issues, but there is still a problem with the ODF tool failing an assert at exit if used, I'll investigate further before merging.


Lestropie:
Confirm no more segfault on `mrview` close on Windows.

bjeurissen:
Works

PR #352, opened 2015-09-17
Header: preserve intensity scaling for integer types

jdtournier:
As discussed in #346.

Turned out to be a bit more complicated than anticipated due to the need to update the intensity scaling on datatype changes, which required `Header::datatype()` to return a proxy object. All seems to work OK, but I would really like someone to put this through its paces to see if I've missed something...


jdtournier:
I should also point out that these changes now _force_ the intensity scaling to defaults for non-integer types. It is not possible to set scaling parameters other than [0,1] for floating-point or binary images, even using `mrconvert`'s new `-scaling` option. I don't think that's a great loss - in many ways it is probably the most sensible thing to do, but if anyone can think of any reasons why this might be a problem, please let me know.

PR #357, opened 2015-09-22
tck2connectome: Fix length quantification

Lestropie:
This was fixed in 2cbfa942 but re-introduced in 7f70bbfc.


jdtournier:
Again, feel free to merge when ready...

Lestropie:
Ha; ab23308 was necessary to pass the testing, even though it didn't fix the issue it was originally proposed for... _shrug_

PR #358, opened 2015-09-22
tck2connectome: Fix segfault with -assignment_end_voxels option

Lestropie:


jdtournier:
Yup, seems sensible... Go for it.

Issue #255, opened 2015-05-13
tckmap hang on Windows


Lestropie:
On my Windows machine, `tckmap` is hanging at random percentage markers.
* Percentage at which it hangs changes every time, or sometimes it completes without issue.
* Doesn't occur in debug mode.
* Not fixed using `-quiet`.
* Fixed using `-nthreads 0` or `-nthreads 1`; seems to be more frequent with more threads than less.

jdtournier:
Sounds nasty - debugging multi-threading issues is really hard (and this does sound like some kind of race condition issue). I tend to find it's easier to inspect the code than trying to debug it any other way. 

What surprises my is that this wouldn't be an issue on Linux. I'd expect if this was a race condition, it would manifest on all platforms. Strange...

In any case, you're in for some fun... I typically find the only real way to debug race conditions like this is to inspect the code very carefully to spot which of the objects I'm writing to might not have copy-constructed as a fully independent copy, or something like that.... Happy hunting!

Lestropie:
Yeah, not looking forward to it. My initial thought was that the ProgressBar fix hadn't propagated, but it's not that.

thijsdhollander:
I'm also a bit stumped as to why it's not an issue on Linux... I'd be looking for exotic -rather than generic- explanations first.  It's really `tckmap` fully hanging, right? (like, not only the progressbar not updating or something, while the track mapping is still running)

If it is a multi-threading thing gone haywire though, the fact that it hangs rather than crashes at least naively points to a deadlock (rather than a race condition).  This is where I can (finally!) bring out the :sparkles: formally trained informatician :sparkles: in me.  As far as I can remember from a course in parallel systems and distributed computing, we were once taught/drilled deadlocks are the "easiest" from all parallel problems to fix (as in: find the cause, not necessarily fix for real), because you can "simply" inspect a stack trace when deadlocked, which should provide you with important clues as to the cause of the deadlock...

Apart from that, I only remember the course coming with 2 absolutely massive books that resulted in a 10 day studying marathon that resulted in a trauma that resulted in me banning most other useful information from my memory after the exam.  So far for the formal training. :neutral_face:

Lestropie:
... You really just wanted an excuse to use :sparkles: , didn't you? :-P

Trouble is, it refuses to lock when running in debug mode, making the stack untraceable. I think I just need to throw a whole lot of `TRACE` macros everywhere, and hope to place one inside the erroneous loop.

[...]

PR #371, opened 2015-10-04
Prevent tckmap hang on Windows

Lestropie:
Make use of Thread::batch() to reduce the load on multi-threading queues, which can hang in heavy processing on Windows.
Related to #255.

### DIFF
--- a/src/dwi/tractography/tracking/exec.h
+++ b/src/dwi/tractography/tracking/exec.h
@@ -44,7 +44,7 @@
 
 #define MAX_NUM_SEED_ATTEMPTS 100000
 
-#define DYNAMIC_SEEDING_BATCH_SIZE 8
+#define TRACKING_BATCH_SIZE 10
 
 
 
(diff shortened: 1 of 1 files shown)


jdtournier:
noticed the checks for `tckgen` didn't pass the first time round, I'm assuming this is due to the stochastic nature of the beast... I got TravisCI to re-run the tests, and it passed fine second time round. But maybe we want to look into modifying the tests to at least reduce the probability of failure...?

No big deal for this merge though, let's hit the button...

Lestropie:
Yeah still need to tweak them slightly. The initial thread batching made the dynamic seeding test fail because it was introducing too much of a delay in receiving tracks and updating the seed probabilities :-/  I'll up the numbers on a couple of them and see how we go.

But I reckon that if my TDI-based testing still fails on a regular basis, testing Hausdorff distances with probabilistic tracking is doomed to failure...

jdtournier:
> But I reckon that if my TDI-based testing still fails on a regular basis, testing Hausdorff distances with probabilistic tracking is doomed to failure...

May very well be true... The reason I'd be hopeful though is that the idea would be to generate a large set of test data (say 10k streamlines), and during testing only generate maybe 250 streamlines. The chances of generating a streamline that doesn't have a sufficiently close match in the testing data should be minimal. And we would allow some small number of failures to match anyway... 

But this is all speculation. I need to give this a crack - just need to find the time...

Lestropie:
But once you've allowed for enough error to not fail spuriously, your tests are probably only going to be sensitive to outright tracking failures and not small regressions; and the existing tracking tests on the SIFT phantom should catch those. The whole idea just gives me nightmarish flashbacks of trying to do streamline clustering... and look how that panned out. :-P

Alternatively if the main goal is to test tracking on real data rather than just a phantom, maybe specify a seed point, do your tracking, generate the TDI, and look for differences in that?

PR #377, opened 2015-10-08
header --> original_header

maxpietsch:
I changed quite a lot of files in a semi-automatic way. So it would be good if somebody with more experience of the affected code could eyeball the diffs as these changes affect quite a lot of code.

### DIFF
--- a/build
+++ b/build
@@ -465,7 +465,7 @@ class Entry:
       self.deps = self.deps.union (list_headers (cc_file))
     self.deps.add (config_file)
 
-    self.deps = set([ cc_file ])
+    self.deps.add( cc_file )
 
     skip = False
     if 'Q' in file_flags[cc_file]: flags += qt_cflags
--- a/cmd/5tt2gmwmi.cpp
+++ b/cmd/5tt2gmwmi.cpp
@@ -146,11 +146,11 @@ void run ()
 
   Header H;
   if (mask.valid()) {
-    H = mask.header();
+    H = mask.original_header();
     H.datatype() = DataType::Float32;
     H.datatype().set_byte_order_native();
   } else {
-    H = input.header();
+    H = input.original_header();
     H.set_ndim (3);
   }
   auto output = Image<float>::create (argument[1], H);
--- a/cmd/5tt2vis.cpp
+++ b/cmd/5tt2vis.cpp
@@ -90,7 +90,7 @@ void run ()
   auto input = Image<float>::open (argument[0]);
   DWI::Tractography::ACT::verify_5TT_image (input);
 
-  Header H (input.header());
+  Header H (input.original_header());
   H.set_ndim (3);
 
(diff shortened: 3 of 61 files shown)


thijsdhollander:
Okay, I did a first check by

1) Switching to this branch and pulling everything.
2) Deleting the configuration file and entire build that I still had sitting there locally.
3) Nicely running ./configure and ./build again.

...so as to safely force everything to recompile.  And it does, so I reason there's no more occurrences of references to `.header()` in the image class.  I'll quickly go over the actual changes to see if I spot anything that shouldn't have been changed.

thijsdhollander:
Went over each change in the mean time; at first sight, this looks all good.  A few ones I wanted to highlight though, so people are aware or can check if needed:

* In fixel_map.h, [line 88](https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/src/dwi/fixel_map.h#L88) : so it propagates to the Fixel_map class (but it makes sense)
* In src/gui/mrview/volume.h, [lines 97-98](https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/src/gui/mrview/volume.h#L97-L98) : @Lestropie , you're working on this one I guess; does it make sense over there?
* [It also applies to adapters now.](https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/lib/adapter/base.h#L52)
* [And to filters.](https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/lib/filter/base.h#L64-L66)
* And I'm not really sure myself if it's correct to use the original header in lines (102)[https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/lib/header.h#L102] and (132)[https://github.com/MRtrix3/mrtrix3/blob/8620226c023db1f380e852ef62791d9c3ec56f9e/lib/header.h#L132] of header.h .  So it's maybe best if someone else checked those 2 lines closely as well (the functionality hasn't changed as to before though).

Apart from that, after this stuff is merged into the actual updated_syntax branch: it's quite handy to explicitly see original_header everywhere where ... well ... it is.  So it's maybe a good thing to check whether that is definitely the wanted behaviour in all of those places, especially if the actual IO might use another type or other strides.  That's what tripped me up with the previous header() anyway. :-)

jdtournier:
All seems sensible to me - including those last two question marks you had about the `lib/header.h` lines: they'll be necessary to allow constructing a `Header` from an `Image` or `Adapter`.

thijsdhollander:
Aha, yes, that makes sense. :-)  Good, I'll merge things again in the updated_syntax branch before any other divergences happen.

PR #382, opened 2015-10-22
Changes to header handling in updated syntax

Lestropie:
Related to #377 and #225.

Kind of two changes bundled in to the one branch since I was doing all kinds of gymnastics at the time:

* Reverted `original_header()` back to `header()`. Although I appreciate that the behaviour when copy-constructing a header from an image with manual strides may not have been as expected (#225), the potential marginal gain from this rename is more than offset by the confusion arising from having `original_header()` called everywhere, yet there being no corresponding 'not-original' `header()`. As I think Donald said, in such an instance it shouldn't be too hard to copy-construct the header, then set the strides just as was done for the `.with_direct_IO()` call. I'll probably also look into potential ways of doing this within the `Image` class. But the rename solution is an absolute no-go IMO.

* Something that I thought would happen as part of the new syntax, but perhaps wasn't entirely complete. Prior to the current `master` branch, we standardized the '`DataSet`' interface, where anything that could be looped over had to define a particular set of member functions. To me, the new `Header` / `Image` construct is different to this; but there was a lot of remaining code using e.g. templated constructors for a '`HeaderType`' / '`ImageType`', that was in fact only using members of the `Header` class; moreover, I found a number of places were constructing a `Header` from an `Image` using this templating, even though there's a `header()` member right there in the `Image` class.

So, what I've done is: _Remove the `Header` template copy-constructors_. These to me could only cause confusion as to what information was actually being used to construct the header (particularly with the `.original_header()` rename, where the templated copy-constructor initialized the new `Header` using `= in.original_header()`, but then manually set _some_ members using direct '`DataSet`' function calls). So now these are gone. Take a look and see what you think.

This did however highlight an alternative location where discriminating between `.original_header()` and `.header()` makes sense: Adapters and filters (well, arguably anything that requires this shouldn't be a _filter_, but let's leave that for now). I've set this up so that such classes retain a copy of the template image's header on construction, but construction of a specific e.g. Adapter class may modify e.g. image dimensions in the _base_ `Header` class. That way, `.header()` returns a fully-formed const header to whatever the output image will look like, whereas `.original_header()` shows the template image header (though nothing's actually using it as yet).

Note that this means that Adapter classes should produce and provide access to a _fully-formed_ `Header` class, not just individual member functions (and I've done this for a couple of examples where it was required). This to me makes a helluva lot of sense, given you probably want to construct your output image based on whatever your adapter is going to provide; but relying on overloading individual '`DataSet`' member functions just introduces ambiguity as to whether the combination of those data and the rest of the header makes sense, or indeed whether those functions are even being called in the relevant construction. Making an image using `Adapter.header()` is better.

### DIFF
--- a/cmd/5tt2gmwmi.cpp
+++ b/cmd/5tt2gmwmi.cpp
@@ -131,7 +131,7 @@ void run ()
 {
 
   auto input = Image<float>::open (argument[0]);
-  DWI::Tractography::ACT::verify_5TT_image (input);
+  DWI::Tractography::ACT::verify_5TT_image (input.header());
 
   // TODO It would be nice to have the capability to define this mask based on another image
   // This will however require the use of interpolators
@@ -146,11 +146,11 @@ void run ()
 
   Header H;
   if (mask.valid()) {
-    H = mask.original_header();
+    H = mask.header();
     H.datatype() = DataType::Float32;
     H.datatype().set_byte_order_native();
   } else {
-    H = input.original_header();
+    H = input.header();
     H.set_ndim (3);
   }
   auto output = Image<float>::create (argument[1], H);
--- a/cmd/5tt2vis.cpp
+++ b/cmd/5tt2vis.cpp
@@ -86,11 +86,10 @@ void usage ()
 
 void run ()
 {
-
   auto input = Image<float>::open (argument[0]);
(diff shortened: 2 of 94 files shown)


Lestropie:
Rejected in favour of #491.

PR #387, opened 2015-10-30
tcksample: add option to ouptut resampled tracks

jdtournier:


jdtournier:
I assume this isn't going to cause any upsets... I'll merge now.

Issue #385, opened 2015-10-30
MRView tractography tool: last vertex is not displayed.


jdtournier:
The last vertex of each streamline is not being rendered. It's really obvious to see when using the new `tcksample -locations` output. The tracks display fine in the old viewer (0.2.x)...

PR #388, opened 2015-10-30
MRView: fix missing vertex from streamlines

jdtournier:
Should fix issue #385.


jdtournier:
OK, there's a faint chance this might cause a vertex buffer overrun, but it seems to work just fine on my system. Wouldn't mind people trying to break it before merging...

jdtournier:
Since no one seems to object...

Issue #389, opened 2015-11-03
advanced debugging: movie does not want to open a .mih file


kimbraeckman:
Hi,

I am trying to open a .mih file with the mrview command. It works fine but the file I want to open is converted from raw bruker data:

gast096e:~ medisip$ mrview /Users/medisip/NODDI_shell2.mih 
mrview: [ERROR] cannot stat file "/Users/medisip//Users/medisip/Dropbox": No such file or directory
Segmentation fault: 11

also a window pops up that states:
mrview quit unexpectedly

I tried the advanced debugging:
gast096e:mrtrix3 medisip$ ./configure -debug debug; ./build debug

MRtrix build type requested: debug

Detecting OS: darwin
Checking for C++11 compliant compiler [clang++]: 4.2.1 - tested ok
Detecting pointer size: 64 bit
Detecting byte order: little-endian
Checking for variable-length array support: yes
Checking for non-POD variable-length array support: no
Checking for zlib compression library: 1.2.5
Checking for GNU Scientific Library: 1.16
Checking whether GSL compiles with -DHAVE_INLINE: yes
Checking shared library generation: yes
Checking for Qt moc: moc (version 5.5.1)
Checking for Qt qmake: qmake (version 5.5.1)
Checking for Qt rcc: rcc (version 5.5.1)
Checking for Qt: 5.5.1

writing configuration to file './config.debug': ok

[CC] lib/thread__debug.o
[CC] src/dwi/tractography/ACT/gmwmi__debug.o
[CC] cmd/tckstats__debug.o
[CC] lib/image/handler/default__debug.o
[CC] cmd/peaks2amp__debug.o
[CC] src/dwi/tractography/SIFT2/tckfactor__debug.o
[CC] cmd/fixelcfestats__debug.o
[CC] cmd/tckfixcount__debug.o
[CC] cmd/fixel2tsf__debug.o
[CC] src/dwi/tractography/SIFT2/fixel_updater__debug.o
[MOC] src/gui/mrview/tool/vector__debug_moc.cpp
[CC] cmd/roigen__debug.o
[CC] lib/image/adapter/reslice__debug.o
[CC] lib/image/format/mrtrix_gz__debug.o
[CC] lib/file/dicom/tree__debug.o
[CC] lib/image/format/mri__debug.o
[CC] src/gui/mrview/mode/slice__debug.o
[CC] src/gui/mrview/adjust_button__debug.o
[...]

PR #391, opened 2015-11-03
mih format: handle cases where the file is specified as an absolute path

jdtournier:
Addresses issue #389


jdtournier:
merging since this one is unlikely to cause any trouble...

Issue #397, opened 2015-11-16
Zooming and ODF rendering broken on Mac


thijsdhollander:
Both on (latest) master branch, as well as on updated syntax branch.  As originally reported by @dchristiaens in https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-153653893 .

jdtournier:
Any idea which commit caused the problem? Probably simplest to track down on the master branch. My money is on 231ab0cc - that changed a fair bit with the OpenGL handling, most likely to have caused problems...

dchristiaens:
OK, finally got some time to look into this. I have some trouble zooming in or out with apple's magic mouse from time to time; not with an old-fashioned scroll wheel. I have the feeling it used to be better, but I've not yet traced this back to its origins. By any chance, is there a keyboard shortcut for zooming in or out? That might come in as a handy alternative at times...

A more pressing issue is the ODF rendering. Here's a screenshot in the current master branch:
![screen shot 2015-11-16 at 13 21 51](https://cloud.githubusercontent.com/assets/7466983/11184456/fa60db2e-8c77-11e5-864e-4c9853aaf737.png)
I'm not sure what's going on with the ODFs. Notice that the intensity bar in the lower right corner looks weird too. I was able to track this one down to commit e9a0dac and pull request #266. That means it's been around since June and I really need to update more often :-)

dchristiaens:
Specifically, the Mac OS X rendering issue was introduced in b2f4ca9, the very first commit in this pull request. Subsequent commits did not fix it.

jdtournier:
OK, for a minute I thought it sounded similar to [the funny transparency issue](https://github.com/MRtrix3/mrtrix3/pull/266#issuecomment-110714924) we had with this branch, but if you say the problem was introduced earlier, then I'm not so sure. All this commit did was change the code responsible for requesting an OpenGL core profile from requesting it per-context, to setting it once application-wide. Is there any difference in the OpenGL attributes reported in the Help->OpenGL dialog...?

jdtournier:
As to scrolling going funny, 3ed77660 is the commit most likely to have introduced problems - although it was supposed to help...

Regarding zooming: I'd added support for pinch to zoom a while back in commit fc6013af, although I haven't really had a chance to test it out thoroughly (won't work on my Linux laptop. unfortunately). Does this not work...?

jdtournier:
[...]

PR #399, opened 2015-11-16
MRView: a few fixes mostly for MacOSX

jdtournier:
As discussed on #397, this fixes issues with rendering on MacOSX, fixes mouse scroll support for the Apple magic mouse, and adds keyboard shortcuts for zooming.

As always, any attempts to try to break this are appreciated...


jdtournier:
Since there are no objections so far...

PR #400, opened 2015-11-17
ACT: minor changes to documentation

jdtournier:
Clarify the expectations for a 5TT image, and how to add the pathology
mask to the 5TT image.

### DIFF
--- a/cmd/5ttgen.cpp
+++ b/cmd/5ttgen.cpp
@@ -45,7 +45,8 @@ void usage ()
 	AUTHOR = "Robert E. Smith (r.smith@brain.org.au)";
 
   DESCRIPTION
-  + "concatenate segmentation results from FSL FAST and FIRST into the 5TT format required for ACT";
+  + "concatenate segmentation results from FSL FAST and FIRST into the 5TT format required for ACT"
+  + "Note: if you need to add a pathological tissue mask, use the 5ttedit command.";
 
   ARGUMENTS
   + Argument ("in_fast_one",   "first output PVE image from FAST (should be CSF)").type_image_in()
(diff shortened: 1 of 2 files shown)


jdtournier:
@Lestropie - just thought this might help prevent some of the confusion some people have been encountering over here... Basically, you'd expect `5ttgen` to allow adding in the pathology mask, but it doesn't - you need to use `5ttedit`. This is fine, but there was no documentation to that effect. On that front, should we consider modifying `5ttgen` to accept an optional pathology image...?

So to get around this (before going through the code and coming across the comment that suggested using `5ttedit`...), we tried a manual `mrcat` approach to generate the 5TT image, but that was flatly refused by `tckgen` as invalid. Turns out this was because the datatype was not floating-point. So I just added in some additional information to the error message to at least allow the user to figure this out. But on that front, given that you can achieve floating-point with an appropriately scaled integer image, is this requirement not overly stringent? Could we allow a integer image as long as the rescale factor was (much) less that 1...?

Lestropie:
Fine with me. Note however that `5ttgen` disappears in `updated_syntax`, I just do the necessary calculations using `mrcalc` in the `act_anat_prepare_fsl` script, which should avoid some confusion.

Not really inclined to support scaled integer images here. You'd have to well and truly go out of your way to generate such an image. Can understand supporting scaled integer images in apps where they come straight from the scanner (e.g. `mrview`), but not really here.

jdtournier:
Cool, using `mrcat` seems like a simpler strategy. 

As to the integer image issue, what surprised me was that you would explicitly _reject_ integer images. I can see the rationale, but really if that's what people want to feed in, why prevent it? The code will handle it just fine, there is nothing technically wrong with doing this. If it turns out that the integer image is unscaled, meaning their mask is binary, that's not an actual error. Besides, the most obvious solution to the error message about needing the datatype to be floating-point is to convert their (binary) mask to floating-point as-is, which allows the program to run but clearly isn't what we'd be hoping they'd do. Personally, I think checking the datatype is probably going to prove unnecessary and/or not helpful in practice...

PR #402, opened 2015-11-23
mrview overlay interpolation was swapped for the overlay value

maxpietsch:
The trilinear and nearest neighbour interpolation were swapped for the overlay value display in mrview.


jdtournier:
Confirmed... Fix seems fine, I'll merge.

Issue #31, opened 2014-05-08
Consider switch to non-GPL license


Lestropie:
(direct copy from previous discussion)

I've been thinking for a while now that it would be good to use a different license than the GPL to allow integration of MRtrix into other projects, whether proprietary or that simply use different licenses. Something like the Mozilla Public License or even MIT/BSD. This would require stripping out all GSL dependencies - it's the only GPL component used in MRtrix.

I've had a quick look through to find which bits would need to be modified. Obviously all the linear algebra stuff would need to be modified. For this, I've just come across the [FLENS project](http://apfel.mathematik.uni-ulm.de/~lehn/FLENS/index.html), which is basically a templated C++ header-only implementation of all of BLAS and LAPACK, which is perfect. That would basically replace a lot of the `lib/math` folder in one go...

Other things that would need re-implementation are actually not too bad. I'd already re-implemented the SH & Legendre functions, so that's no issue. The FFT stuff is currently not used in the main branch, but it would be trivial to use something like [kissFFT](http://sourceforge.net/projects/kissfft/) if needed. The only other sticking point is a decent multidimensional optimiser, which is currently only used in gendir, so no big deal. I've come across [one decent-looking implementation of the BFGS conjugate gradient optimiser](https://www.lri.fr/~ilya/pbfgs.html), which looks relatively simple to adapt to our needs.

So all in all, nothing too difficult to do, but would nonetheless require a fair bit of work. Question is, is this something that we'd consider important, and if so, is this something that we want done before release, or are we happy to wait till later...?

-------------------------

It might also be worth looking into the boost library. In particular the Math, Random and uBLAS parts might be useful. Math has special functions implemented, but apparently you already coded those. uBlas provides templated c++ BLAS. As fas as I know LAPACK is not included though... Also [Eigen](http://eigen.tuxfamily.org/) might be interesting for linear algebra. Pretty sure it will cover most of your linear algebra needs... Not sure if it's license is liberal enough.

-------------------------

Eigen looks good - it's C++98, while FLENS uses C++11 features. I'd like to move to C++11, there's a lot of features that we could use straight away, for example the [Mersenne Twister random number generator](http://www.cplusplus.com/reference/random/mt19937/), which is what we use with the GSL currently. Only issue with moving to C++11 is that it might prevent deployment on older clusters and HPC setups - but then they can always compile static binaries if required. Maybe it's time we tried it...

[...]

PR #415, opened 2015-12-09
Global tractography

dchristiaens:
This branch adds a command `tckglobal`, which implements my multi-shell multi-tissue global tractography method.

The code is ported to the new syntax (#225), and builds without warnings on my Mac OS X and Ubuntu systems. I've added docs and I'll create a detailed wiki page for it too. This latest version was tested in many datasets, but anyone is invited to try and break it :-)

Finally, I will also add tests for Travis CI at some point, but currently there's no multi-shell data available in the testing repo. I guess we should add this for testing msdwi2fod too.


jdtournier:
Happy to merge, but just thought I'd check whether you're happy for this to happen before we update the license (#31)? We also need to have a frank discussion about the copyright notice (#9). I'll get the discussion going on both of these issues in their respective threads...

dchristiaens:
Regarding the global tractography code, ideally I adopt the same licence as the rest of MRtrix. Currently it's very simple: mrtrix is GPL thus tckglobal is GPL. I'm happy with a more permissive license along the lines of MPL or LGPL. However, I do want modifications to the algorithm implementation be made open source, and therefore think MIT or Apache licences are too permissive. Hence, should you decide to switch to MIT/Apache, I'd need to insert a separate license for this part of the library, or we should revisit the original plan to distribute it as a separate module after all. But I'd much rather have it go in the main library under MPL/LGPL terms.

thijsdhollander:
I'd say, let's have it in the main one under MPL terms. :-)

thijsdhollander:
Just checked some more: it all looks nice and solid!  I'd be happy to have a reference to the TODI paper included in the reference list though (possibly accompanied by a "(when using the -todi option)"), since you directly allow it as an optional output. :-)

I do like your choice to have certain more advanced parameter options clearly separated from the rest.  It's an approach of categorising options that might actually also work well for a few other MRtrix commands that are heavy on optional parameters...

jdtournier:
OK, guess everyone's happy to merge this in... We can sort out any remaining issues later.

Issue #141, opened 2014-11-18
rotate bvecs in revpe_distcorr


bjeurissen:
Currently, the revpe_distcorr script does not rotate the gradient directions.

This can be implemented fairly straightforwardly as explained in http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/EDDY/Faq#Will_eddy_rotate_my_bevcs_for_me.3F

It doesn't seem to be a lot of work (apart from getting the coordinate system conventions right) and might be a useful feature.

What do you guys think?

tjroine:
Hi,

Here's a script which can be used to rotate bvecs based on the results from FSL's eddy. Unfortunately I don't have a c++ version yet, but maybe this will be of help in implementing this step into the revpe_distcorr script. In any case this script can be used individually (in Matlab) before performing CSD. I personally use it in a bash script with an extra exit command included in the end of the function, to be able to continue after the rotation to the next step in the script.

Cheers,
Timo

https://www.dropbox.com/s/fu233yuyrxnp5n3/rotate_gradients.m?dl=0

bjeurissen:
The posted script does not correctly rotate around the x- and y-axis. Also, as Rob suggested, it should not rotate the mrtrix grad file, but the fsl bvec file and then re-import into the header to avoid issues with axis conventions.

@Lestropie, I could easily include this in the revpe_distrcorr python script (and/or the libraries it depends on), the only issue is that the matrix operations would result in a numpy dependency, which might be undesirable. What do you think?

tjroine:
Hi, 

I wasn't aware of this difference between mrtrix grads and fsl (or other software) bvecs coordinate system. If I understood correctly, although a same gradient table in the scanner is used, the resulting bvecs will depend on the possible rotation of the FoV? Does this mean that a separate bvecs file should always be saved for each subject? Can this be done if the scanner doesn't write gradient information into the DICOM header?

I think this difference is an important distinction, which will affect many researchers and could be emphasized more in the wiki.

Anyway, the new version of the function will require the intermediation through bvecs and should work correctly irrespective to the possible differences in the coordinate systems.

https://www.dropbox.com/s/co1drajyuv97ona/rotate_bvecs.m?dl=0

Cheers,
Timo

jdtournier:
[...]

PR #416, opened 2015-12-09
Rotate gradients in eddy and provide wrapper for use without reverse phase encoding

dchristiaens:
As discussed in #141.


Lestropie:
This has been made redundant by the changes in `updated_syntax` based on #419: `dwipreproc` can operate in the absence of reversed phase-encoding, and will use the rotated bvecs provided by `eddy` from 5.0.8 onwards.

Issue #422, opened 2015-12-15
(ms)dwi2response & sh2response: Resolving naming before release


Lestropie:
We need to decide what to do about these before we release. Supersedes #105.

Here's my current thinking, feel free to modify or suggest complete alternatives:

* Keep `dwi2response`, but rename it to `dwi2sf`, and output the single-fibre voxel mask by default; only give the final response function if explicitly requested by the user.

* Rename `sh2response` to `sf2response`, and receive as primary input a DWI rather than SH; this will additionally allow for selecting the shell(s) for which response(s) should be estimated (which should simplify `msdwi2response`).

Hopefully that makes it clear that there are in fact two steps: single-fibre voxel selection, and response function estimation based on those voxels. Just using the (optional) response function output from `dwi2sf` is a shortcut that will probably satisfy most users, but hopefully the fact that the SF voxel mask is the default output will make people question that and look into alternative ways to get the SF mask rather than just reporting that they can't get a response function.

Hopefully it also separates `msdwi2response` from what is currently `dwi2response`, as their operations are actually quite different currently.

dchristiaens:
I agree with your argument for `dwi2sf` (I might call it `dwi2sfmask`, to avoid confusion with a sinlge fibre _kernel_) as a more appropriate name for implementing Tax et al. (2014). However, I don't like the `sf2response` rename. That command will always require _both_ a mask and the input data, and it just makes more sense to me to call it after the type of input data, rather than the type of mask. Besides, `sh2response` is used for fitting GM and CSF response functions just as well.

As to whether it should receive DWI input rather than SH, I would still prefer SH. As long as `dwi2fod` and `msdwi2fod` live happily next to each other, we need to distinguish between the singe-shell and multi-shell response functions at some point. Otherwise how do you handle the b=0 in good old CSD? In my opinion, `sh2response` has the benefit of being 100% clear on being single-shell only.

Nevertheless, I fully agree that `dwi2response` and `msdwi2response` are currently very different operations, and that this may cause some confusion. What I would propose we do instead is to introduce `dwi2response` as a script, identical to `msdwi2response` but operating on the shell of highest b-value only. That way, we would effectively have the two-step process you root for: single-fibre masking and response function estimation.

thijsdhollander:
[...]

PR #425, opened 2015-12-15
Estimate response

jdtournier:
As discussed in #422. This is currently a very simple proof of concept script, just to get feedback on whether the approach is worth implementing properly. Feedback welcome...


jdtournier:
Sorry, wasn't supposed to be against `master`...

PR #446, opened 2016-02-19
A few suggestions following installation in a CentOS Docker container

glatard:
- Added list of CentOS packages to README.md
- Added a Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM centos:latest
+RUN yum update -y
+RUN yum install git gcc-c++ zlib-devel gsl-devel qt-devel -y
+ENV PATH=$PATH:/usr/lib64/qt4/bin
+RUN git clone https://github.com/MRtrix3/mrtrix3.git
+RUN cd mrtrix3 && ./configure && ./build
+ENV PATH=$PATH:/mrtrix3/bin


jdtournier:
Hi Tristan,

Thanks for your suggestions, they're sure to be useful for others. However, there's a few reasons why I think merging these changes might not be quite the right thing to do. I think both of these would be best added to the wiki, with the added bonus that this would also make them more readily discoverable. 

The main issue I have is that we've kept the README deliberately short and to the point - there are no distro-specific instructions in there. If we were to add instructions for CentOS, we'd have to include instructions for Ubuntu, Fedora, RHEL, SuSE, Arch, etc. - and probably also specific instruction for each _version_ of these distros (Ubuntu 12.04 certainly requires a bit a extra handling). This is information that belongs in the wiki, where there is freedom to go into much more detail. 

The docker file is also no doubt going to be very valuable, but only for a small subset of users. It is also specific to the particular distro that you intend to run within your container. I've no direct experience with Docker myself, but I understand it would be perfectly possible to run different distros within them (in fact, I think our community site runs within a Ubuntu 14.04 Docker container). Including a single Dockerfile would not allow all these various combinations to co-exist. So again, I think this would best be documented in a dedicated section on the wiki. The file is short enough that its contents can be included as-is. 

The other issue is that we are about to change things around a fair bit when the `updated_syntax` branch gets merged into `master`, and this would invalidate your changes... The list of packages would need to drop GSL and include Eigen, and the PATH would need to point to `mrtrix/release/bin` rather than `mrtrix/bin`. It would be best to delay until we do this to upload these suggestions. 

On top of that, we're in the process of moving the wiki over to readthedocs, so again, it's best to wait until this is complete before trying to include these modifications... Not sure what the process would be to modify the documentation at this point, hopefully we can do it via a pull request too. The others might be better placed to comment on this...

glatard:
Hi J-Donald,

Thanks for the answer. 

About the CentOS packages, feel free to add them wherever you think is more appropriate in the documentation, I just thought it could be useful for others since it took me some time to figure out the exact package names (though this is no big deal indeed). 

About the Dockerfile: keeping a Dockerfile in the actual software project is useful because it allows to automatically build Docker containers any time the git repo is updated. That could provide an "official" container for MRtrix, preventing people from creating their own (there are 3 at the moment on DockerHub, including mine: https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mrtrix&starCount=0). If you find it useful, I could easily provide a Ubuntu-based Dockerfile, if this is your preferred distribution. Otherwise, no problem, I'll stick to my own custom container image.

jdtournier:
> About the CentOS packages, feel free to add them wherever you think is more appropriate in the documentation, I just thought it could be useful for others since it took me some time to figure out the exact package names (though this is no big deal indeed).

I appreciate it's not always easy to figure these out... We currently list commands to install the dependencies for Ubuntu, Fedora/CentOS and Arch, but as this clearly demonstrates, the package list for CentOS isn't quite right... So we should definitely update that section as per your suggestion. 

> About the Dockerfile: keeping a Dockerfile in the actual software project is useful because it allows to automatically build Docker containers any time the git repo is updated. That could provide an "official" container for MRtrix, preventing people from creating their own (there are 3 at the moment on DockerHub, including mine: https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mrtrix&starCount=0). If you find it useful, I could easily provide a Ubuntu-based Dockerfile, if this is your preferred distribution. Otherwise, no problem, I'll stick to my own custom container image.

Again, I can see why this would be convenient. The problem is that I don't have a preferred distribution for this - if I did, it would be Arch Linux, which would probably not be a popular decision given that it's a moving target... But really, the point is that I don't think we should prevent people from creating containers with whatever flavour of Linux they wish. I hadn't heard about DockerHub before, but that does actually strike me as the right place to list such a file. The issue with providing an 'official' Dockerfile is that this immediately implies we take on some level of support for such an installation, which I'd be uneasy with given that I've no experience with this. I would be happy to include a folder containing different Dockerfiles for popular options, but at this point I think we might as well direct people over to DockerHub from the wiki...  

The ability to have the container auto-update when the repo changes is attractive though. Is there no way to trigger such a rebuild otherwise? Something like an [IFTTT recipe](https://ifttt.com/github)...? On the other hand, this might also not be such an issue in the long run. At the moment, we're still in beta testing, so haven't released an official 3.0 version yet. There's still a few features missing before we actually start making proper releases as such. But at that point, you'll probably want your container to build the latest tagged release, rather than track the `master` branch...?

glatard:
On 02/19/2016 04:19 PM, J-Donald Tournier wrote:
>
>     About the CentOS packages, feel free to add them wherever you
>     think is more appropriate in the documentation, I just thought it
>     could be useful for others since it took me some time to figure
>     out the exact package names (though this is no big deal indeed).
>
> I appreciate it's not always easy to figure these out... We currently 
> list commands to install the dependencies for Ubuntu, Fedora/CentOS 
> and Arch, but as this clearly demonstrates, the package list for 
> CentOS isn't quite right... So we should definitely update that 
> section as per your suggestion.
>
>     About the Dockerfile: keeping a Dockerfile in the actual software
>     project is useful because it allows to automatically build Docker
>     containers any time the git repo is updated. That could provide an
>     "official" container for MRtrix, preventing people from creating
>     their own (there are 3 at the moment on DockerHub, including mine:
>     https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mrtrix&starCount=0).
>     If you find it useful, I could easily provide a Ubuntu-based
>     Dockerfile, if this is your preferred distribution. Otherwise, no
>     problem, I'll stick to my own custom container image.
>
> Again, I can see why this would be convenient. The problem is that I 
> don't have a preferred distribution for this - if I did, it would be 
> Arch Linux, which would probably not be a popular decision given that 
> it's a moving target... But really, the point is that I don't think we 
> should prevent people from creating containers with whatever flavour 
> of Linux they wish.
>
Having a container released with one particular distribution wouldn't 
prevent people from building their own from another distribution. It 
would just be convenient for people who want to use MRtrix3 without 
having to install it.
>
> I hadn't heard about DockerHub before, but that does actually strike 
> me as the right place to list such a file.
>
In DockerHub, automated builds need a Dockerfile in a git repository. 
Container images are built automatically every time there is a push to 
the git repo (they can also be triggered manually). So Dockerfiles are 
stored in git, not on DockerHub.
>
> The issue with providing an 'official' Dockerfile is that this 
> immediately implies we take on some level of support for such an 
> installation, which I'd be uneasy with given that I've no experience 
> with this. I would be happy to include a folder containing different 
> Dockerfiles for popular options, but at this point I think we might as 
> well direct people over to DockerHub from the wiki...
>
Yes, it may take some support indeed, in particular when installation 
instructions change.
>
> The ability to have the container auto-update when the repo changes is 
> attractive though. Is there no way to trigger such a rebuild 
> otherwise? Something like an IFTTT recipe <https://ifttt.com/github>...?
>
For Docker containers, the way to do it definitely seems to be through 
Dockerfiles in git repositories.
>
> On the other hand, this might also not be such an issue in the long 
> run. At the moment, we're still in beta testing, so haven't released 
> an official 3.0 version yet. There's still a few features missing 
> before we actually start making proper releases as such. But at that 
> point, you'll probably want your container to build the latest tagged 
> release, rather than track the |master| branch...?
>
Yes, I will do it and host my Dockerfile and containers in my own 
repositories. Feel free to ping me in the future if you want to work on 
this.

jdtournier:
OK, I see why you'd need it to be on the repo proper. Do you have any freedom into what its actual name? Would you be able to use a `Dockerfile.CentOS` file, for example? I would be a bit less uneasy about this, it wouldn't give the impression that this is the only Docker that would work or that we might support (which I don't think we want to anyway, unless there are enough users using it to help other out?). 

Otherwise, if you're happy to stick to point releases anyway, it might be easier to stick with the status quo. 

Not sure what the other devs think about this...?

draffelt:
I think it'd be a good move to have containers for a few common
distributions in the future. It might be handy for people who have access
to cloud computing resources (e.g Amazon research grants
<https://aws.amazon.com/grants/application/>). That said, I don't know much
about Docker so I'll need to do some more reading. I had a brief look today
at http://kubernetes.io/ as a possible way to utilise everyone's desktops
at night.

dchristiaens:
I gave Docker a try a while ago, though I'm by no means an expert. It's a great platform, and it would be good if we could offer support in the future -- especially as it's quickly becoming a default in AWS and other cloud services. It could be a very accessible and user-friendly way for people to get their hands on MRtrix, both on their own system and in cluster computing environments. But because of this, I agree with Donald that it's best to build Docker containers for stable releases only, similar to how we would otherwise provide binaries for the most common platforms. I would expect people who want the latest bleeding-edge developments to be a bit more technically skilled and compile mrtrix themselves.

Nevertheless, I can see how this is a bit of an issue today: it's been over a year since release 0.3.12 and new developments have gone a long way since. This is about to change though, and hopefully we can move to a system of regular releases soon, including binaries and Docker containers.

jdtournier:
@glatard: I'm trying to clean up the backlog of pull requests at the moment. This one is pretty old and outdated now, and I think the general conclusion was that we'd put this on hold for now. I'm going to close this, if you don't mind? I think we'd like to have Docker support in the long run, but it would need someone to drive it forward, and I don't think I'm the right person for it... 

One option might be to have a separate repo with the Docker file hosted on it, and basically have it track the main repo's releases. But to be useful, I think it would need to also have FSL and ANTs available within it - many of our scripts rely on them. So it would be a bit of extra work to get this done...

Issue #398, opened 2015-11-16
ROI "layer" is currently visualised "below" overlay


thijsdhollander:
As I originally brought up in https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-147906230 .  Not a bug, just wondering what makes the most sense in use cases where an image, overlay(s) and ROI editing come together.

maxpietsch:
I am in favour of ROI on top of overlay. Comparing two images, one as main image the other as overlay, hides the ROIs. I reckon ROIs usually cover less area than overlays do. So I don't think the current behaviour makes much sense.

maxpietsch:
The the order of display layers depends on the order of tools in the tool menu. Are there any objections to moving the ROI tool to the end?

thijsdhollander:
Yeah, so I'm definitely also in favour of "ROIs on top": I actually often use combinations of a main image and one or even more overlays to get a good view on certain anatomical features; which I then intend to use for drawing ROIs.  As you mention, it's quite annoying that the ROIs then get covered up (or still difficult to see even when using partial opacity of the/some overlay(s)).

I wasn't aware it simply depended on the order in the tool menu though.  If that's the case, I'd be very happy for the ROI tool to be moved (more) towards the end.  So what do you reckon?  Just swapping the ROI and overlay tools is probably enough?  The other tools don't seem to be able to cause problems, since they're not showing scalar images.

maxpietsch:
Done in c442cc3.

PR #457, opened 2016-03-03
mrview: changed order of tools 

maxpietsch:
so that ROIs are displayed on top of the other tools such as overlays as discussed in #398

### DIFF
--- a/src/gui/mrview/tool/list.h
+++ b/src/gui/mrview/tool/list.h
@@ -32,12 +32,12 @@ The second argument is the name of the mode as displayed in the menu.
 The third argument is the text to be shown in the menu tooltip. */
 
 TOOL(View, View options, Adjust view settings)
-TOOL(ROI, ROI editor, View & edit regions of interest)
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Ah, hadn't spotted yet you'd made a pull request already.  As I mentioned in https://github.com/MRtrix3/mrtrix3/issues/398#issuecomment-192013657 , maybe moving the ROI tool just beyond the overlay tool might have been enough.  But this is definitely fine for me as well. :+1:

Lestropie:
I'd prefer it to be shifted to just below the overlay tool. I don't think you want your ROI voxels being drawn over the top of streamlines, ODFs or fixels.

jdtournier:
> I'd prefer it to be shifted to just below the overlay tool. I don't think you want your ROI voxels being drawn over the top of streamlines, ODFs or fixels.

I agree. I also think the ROI and overlay tools are both somehow more 'fundamental', so probably ought to come earlier in the list anyway...

PR #465, opened 2016-03-14
Improve error handling in matlab functions

kloewe:
Using read_mrtrix_tracks.m, I have noticed a very minor problem with the error handling.
For example, when fopen() fails because the filename was not correctly specified, this is currently handled by printing an error message and then returning. This leads to the wrong type of error.
```
>> t = read_mrtrix_tracks('wrong_filename.tck');
error opening wrong_filename.tck
Error in read_mrtrix_tracks (line 10)
image.comments = {};

Output argument "tracks" (and maybe others) not assigned during call to
"read_mrtrix_tracks.m>read_mrtrix_tracks".
```
I changed the error handling in read_mrtrix_tracks.m and the other functions so that the functions assert() and error() are used instead of disp()/return().

### DIFF
--- a/matlab/read_mrtrix.m
+++ b/matlab/read_mrtrix.m
@@ -8,15 +8,11 @@ function image = read_mrtrix (filename)
 image.comments = {};
 
 f = fopen (filename, 'r');
-if (f<1) 
-  disp (['error opening ' filename ]);
-  return
-end
+assert(f ~= -1, 'error opening %s', filename);
 L = fgetl(f);
 if ~strncmp(L, 'mrtrix image', 12)
   fclose(f);
-  disp ([filename ' is not in MRtrix format']);
-  return
+  error('%s is not in MRtrix format', filename);
 end
 
 transform = [];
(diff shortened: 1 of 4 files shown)


jdtournier:
This seems fine to me - I didn't know Matlab had `assert()` and `error()` calls, good to know.

Only snag right now is that our testing framework doesn't seem to have picked up your pull request... And technically I shouldn't merge until the build passes. Maybe the others can help figure out what's going on...?

I've just checked on TravisCI, and the automatic building of pull requests was disabled for some reason - although it was enabled for pushes, which is probably why it's never been a problem before. I think this problem might be due to the fact that this pull request lives on a forked repo (at least I assume it does?), so never got pushed to our repo as such. Maybe we should assign to only building pull requests, rather than pushes...?

In the meantime, I can't figure out how to get TravisCI to test this pull request... Any ideas? I can force the merge for now, but I'd like to figure out what's going on here and fix it...

kloewe:
I guess you're right. The build is probably not working because the pull request comes from my fork of your repo. Maybe the workaround outlined [here](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests) could be implemented.

jdtournier:
Looking at the link you sent through, I think the problem was that at the time you created your pull request, building pull requests wasn't enabled in TravisCI. It is now enabled, so maybe the simplest is to close this request and make a new one for the same commit, see if that triggers TravisCI into action.

The rest of the post was about dealing with sensitive data within the build environment - these data just won't be available for external pull requests. We don't have any problems on that front, I don't think there's any issues there.

jdtournier:
:+1:

PR #470, opened 2016-03-16
Update mac_install.rst

jdtournier:


jdtournier:
Crap, @maxpietsch beat me to it with #469...

PR #471, opened 2016-03-17
update to mac install instuctions

maxpietsch:
Separated homebrew, port and from-source installation instructions. port installation is untested (don't use it)


jdtournier:
@dchristiaens, @bjeurissen: feel free to merge this pull request - I don't have much of a clue about installation on MacOSX, so if you think it's good to go, press the button...

dchristiaens:
Oops, thought I spotted a typo but then it wasn't... Happy with the doc as it is now; I will merge when TravisCI is done.

dchristiaens:
@jdtournier: I just read your latest comment on the [forum](http://community.mrtrix.org/t/build-fails-on-mac/121/18), in which you suggest to recommend Homebrew by default after all. That is also my preference. Shall I move it to the top of the list then? @bjeurissen, @maxpietsch, do you agree?

jdtournier:
Like I said, that depends on whether or not Qt5 can also be installed easily via macports...

jdtournier:
So it seems Qt5 is also easily installed via macports. In this case we shouldn't give too much preference for one over the other. Why don't you all have a vote as to which you prefer, and we'll list that one first (since it will be better supported by you guys that way), but also make sure we clearly spell out that both work equally well. What I'd like to do is verify that both work well without needing to use the direct Qt5 download, which will simplify installation quite a bit...

dchristiaens:
Agreed. I vote Homebrew, but happy either way.

maxpietsch:
I use Homebrew as well so I am more familiar with it and can try to find out what might be wrong if something is not working. The internet seems to be undecided on which is better. Homebrew has less packages but seems to integrate better with native OS X libraries which makes it more robust on system upgrades (that's why I switched).

bjeurissen:
As far as I understand, the fact that it integrates with the native OS X
libraries makes it _more_ likely to break on system upgrades.

dchristiaens:
It seems the majority of us prefer Homebrew, so I moved it back to the top of the list. Instructions for Macports and for installing from source are obviously still there, so people that prefer one or the other can easily get set up too.

However, for novice users who may never have compiled software before, I believe Homebrew is a good choice. It is installed with a single command line and since it symlinks packages from `/usr/local/` they don't need to set their PATH or anything. As @maxpietsch already mentioned, I like how it installs packages against native system libraries whenever possible (yet still gives a warning when it can't for some reason). Macports approach of installing all dependencies in an isolated environment often leads to a huge dependency tree, which not only increases build time but may also increase the likelihood of it crashing somewhere down the line. As @bjeurissen mentioned, the Macports community claims that their approach is more robust to system upgrades. However, I have personally experienced Macports still crashing when upgrading Mac OS X. Like Max, I gave up at some point and switched to Homebrew, and never had any issues since. (Disclaimer: this was just me...)

In any case, it would be wise to test these install instructions on a clean system. Did anyone buy a new Mac lately? :-) Or do you reckon we just merge this pull request and wait for user feedback?

jdtournier:
So just to reiterate one thing: I don't really care which is _better_ - clearly there are already lots of opinions on the matter, we're not going to resolve this here. Both seem good and well supported. What I care about is having enough experience within our team to support users, so I'd rather have the most commonly used within our dev team be listed first. But we should in no way emphasise one over the other - this is purely about which happens to be listed first, which unfortunately will give it some precedence.

Otherwise, I just had a look at the proposed changes, and I've a few comments before merging:

- there needs to be a little section heading before that 'multiple choice' section. Currently it's part of the larger list of instructions, which are sequential steps to install - this bit is clearly not sequential, you have to choose one or the other, this needs to be clear from the layout as much as the text.

- there needs to be a little intro paragraph to that section, explaining what the aim is here (installing Eigen and Qt - probably should be the title, actually), that there are several options, that some of these might already be a good fit for some users (in which case they are strongly encouraged to use the relevant option), that option 3 should be regarded as a last resort if nothing else works (and even then, you'd be well advised to consult the forum first, chances are switching tactics is only going to make things worse, as @bjeurissen pointed out), and that users without existing homebrew or macports installs are free to pick whichever they like the look of - as currently stated, it seems that if you don't have either homebrew or macports installed, you have to go the direct download route...

- for the direct download option, it might be worth emphasising that the `Eigen` folder should not itself be included in the path, only the folder enclosing it (the most likely reason this issue started in the first place...).

Everyone agree...? If so, it'd be lovely if someone could implement these changes... :grin:

dchristiaens:
Thanks for your comments Donald. I modified the file accordingly.

jdtournier:
Thanks @dchristiaens - I made a few changes on top of yours. I'll merge now, I think it's good to go - unless anyone has any further input...?

dchristiaens:
Looks good.

Issue #481, opened 2016-03-21
New command: tensor2dwi


Lestropie:
Generate a predicted DWI based on a tensor image, and a template image. Optionally import gradient table from file(s). As requested [here](http://community.mrtrix.org/t/tensor2dwi-to-compute-sse-map/109/2).

Would be a cleaner interface if the _b=0_ term was included in the tensor image output (#17)...

thijsdhollander:
With respect to the request on the forum specifically: wouldn't it be easier to have dwi2tensor provide an option to return the SSE?

Lestropie:
For that application yes. There's theoretically more that could be done with a pseudo-DWI generated from a tensor image than just an SSE output, so it's more general that way; I guess it depends on whether or not that potential warrants having an extra command. I'm content either way.

draffelt:
I don't think its really an issue having one more command. We already have
over 100 commands (including scripts), and therefore it's already hard for
a user to look through and discover what they want. One more at this stage
is not a deal breaker. The number of commands is only going to grow bigger.

What we probably need is the ability to assign keywords/tags to each
command, then have the ability to list commands based on tags on the user
docs. We then *might* want to consider splitting mrfilter and maskfilter
back into separate commands (with 'filter' tags), more in line with the
unix philosophy.

thijsdhollander:
Definitely not against another command.  The name would be quite clear anyway, so it's not confusing in any way.  But for this purpose, it's indeed a very slight bit annoying that the b=0 has to come in from somewhere again.  Maybe `dwi2tensor` could have an option to return the actual predicted signal (i.e., not only the b=0, but a full evaluation using the gradient table of the input DWI)?  It could even use the kurtosis tensor part of the fit automatically as well then (so the output is really reflecting the predicted signal).
So not against an extra command, but this might help to avoid juggling around that b=0.

thijsdhollander:
(I'll take a quick look into it)

Lestropie:
I suppose if you were to output the prediced signal from `dwi2tensor`, people might expect the same capability to be present in `dwi2fod`; but then, that can already be achieved after the fact using `shconv | sh2amp`. So it'd still be kind of inconsistent between different models.

There's no reason why using the kurtosis part would be exclusive to generating the predicted signal image straight from `dwi2tensor`; I'd expect `tensor2dwi` to be able to do that as well?

thijsdhollander:
[...]

PR #484, opened 2016-03-22
dwi2tensor: added option to output predicted dwi signal

thijsdhollander:
I've added an option to `dwi2tensor` to output the predicted dwi signal; as per the user request and accompanying issue at #481 .
Another command `tensor2dwi` can always be added later on as well.  This is just a simple solution that will help the user; and in most use cases avoid juggling around a b=0 image, tensor image, and a gradient table (all required otherwise as input); and sometimes a kurtosis tensor image as well.


thijsdhollander:
I'll merge this for now, so as to help out the user on the forum.  It's not breaking or even touching any existing functionality; just added an independent option.

PR #485, opened 2016-03-22
documentation update

thijsdhollander:


jdtournier:
What's all these `.rste` files that got added with this commit?!? They seem to duplicate the existing `.rst` files. I assume this wasn't what was intended?

rtabbara:
I believe they are intermediate files that are created when locally building the docs into html using the provided Makefile. I think Thijs just wanted to check how the modifications turned out before a commit.

But I do not think we need to add them to the repo in order for readthedocs to work - sorry, I should have added .rste to gitignore.

thijsdhollander:
Hmmm... ok, not intended at all.  I just enabled the -dev option and rebuild with that, so the documentation of dwi2tensor would rebuild.  Should've checked the actual commit more thoroughly.

Issue #479, opened 2016-03-21
Volume render + anisotropic voxels


thijsdhollander:
Not sure how to exactly describe this, but there's something seriously wrong with the volume rendering in conjunction with an image with anisotropic voxel size.  You'll immediately notice it when you freely rotate such an image.  Here's a few screenies; I've set interpolation to off, so it's clear in what dimension the voxel size is anisotropic, but this problem happens in the same manner when interpolation is on as well.  Prepare for funny business:

![screenshot0001](https://cloud.githubusercontent.com/assets/11952694/13908658/a48e1d50-ef5b-11e5-8740-9a07a02747cc.png)
![screenshot0002](https://cloud.githubusercontent.com/assets/11952694/13908657/a467f27e-ef5b-11e5-97fd-1b93c2a5903b.png)
![screenshot0003](https://cloud.githubusercontent.com/assets/11952694/13908653/a46780b4-ef5b-11e5-8540-ff64c1ce79bf.png)
![screenshot0004](https://cloud.githubusercontent.com/assets/11952694/13908654/a467a54e-ef5b-11e5-9e60-f7ab8d9959e1.png)
![screenshot0005](https://cloud.githubusercontent.com/assets/11952694/13908656/a467dc8a-ef5b-11e5-989c-a58efdb09d71.png)
![screenshot0006](https://cloud.githubusercontent.com/assets/11952694/13908655/a467c65a-ef5b-11e5-8fd5-80c851c51e99.png)
![screenshot0007](https://cloud.githubusercontent.com/assets/11952694/13908652/a4672254-ef5b-11e5-9e6c-fb9dce1d78f9.png)

draffelt:
Freaky...

jdtournier:
Whoa, it's like wobbly mirrors! 

No idea how this could happen, the distortion doesn't seem linear. I'll have a look into it when I get the chance.

thijsdhollander:
You can easily "investigate" what happens by, e.g., overlaying a tractogram and making the volume transparent or something.  As you rotate, the warping happens.  It's very nonlinear...

Note that, while maybe not *that* obvious at first, this also happens for even slightly anisotropic voxel sizes.  So if you've been combining an mprage with subtle anisotropic voxels with a tractogram, clip planing it a bit, and rotating to investigate, your tracks might have slightly mismatched spatially with the anatomy; depending on viewing angle and location in the volume...

jdtournier:
[...]

PR #486, opened 2016-03-22
Fix volume rendering when image voxels are anisotropic

jdtournier:
as discussed in #479


thijsdhollander:
Triple checked for highly anisotropic voxels along each major axis separately.  All is good. :+1:

Issue #489, opened 2016-03-23
mrtransform with reorientation segfaults depending on datatype


maxpietsch:
When using the -datatype option with FOD images mrtransform segfaults. It works with -noreorient.

~~~~
mrtransform template/csd4/13_csd4.mif input_transformed/13.mif -linear linear_transforms/13.txt -debug -datatype float 64

mrtransform: [INFO] opening image "csd4/13_csd4.mif"...
mrtransform: [DEBUG] reading key/value file "csd4/13_csd4.mif"...
mrtransform: [DEBUG] sanitising image information...
mrtransform: SH series detected, performing apodised PSF reorientation
mrtransform: [DEBUG] mapping image "csd4/13_csd4.mif"...
mrtransform: [DEBUG] memory-mapping file "csd4/13_csd4.mif"...
mrtransform: [DEBUG] file "csd4/13_csd4.mif" mapped at 0x7f40d9813000, size 33177600 (read-only)
mrtransform: [DEBUG] image "csd4/13_csd4.mif" loaded
mrtransform: [DEBUG] image "csd4/13_csd4.mif" initialised with strides = [ 15 1440 138240 1 ], start = 0, using direct IO
mrtransform: [INFO] image will not be regridded
mrtransform: [INFO] creating image "input_transformed/13.mif"...
mrtransform: [DEBUG] sanitising image information...
mrtransform: [DEBUG] resizing file "input_transformed/13.mif" to 66358688...
mrtransform: [DEBUG] sanitising image information...
mrtransform: [DEBUG] mapping image "input_transformed/13.mif"...
mrtransform: [DEBUG] creating RAM buffer for file "input_transformed/13.mif"...
mrtransform: [DEBUG] file "input_transformed/13.mif" held in RAM at 0x7f40d58ca010, size 66355200
mrtransform: [DEBUG] image "input_transformed/13.mif" loaded
mrtransform: [DEBUG] image "input_transformed/13.mif" initialised with strides = [ 15 1440 138240 1 ], start = 0, using indirect IO
mrtransform: [100%] copying from "csd4/13_csd4.mif" to "input_transformed/13.mif"...
mrtransform: [DEBUG] initialising threads...
mrtransform: [DEBUG] launching 8 threads "loop threads"...
mrtransform: [  0%] reorienting... Segmentation fault
~~~~

jdtournier:
Guess that's fixed then...

maxpietsch:
[...]

PR #491, opened 2016-03-23
Remove original header

jdtournier:
As discussed as part of the `updated_syntax` branch (#225), this removes the `Image::original_header()` method and modifies the `Header` copy-constructors and assignment operators for the reasons set out in [this post](https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-151808441). See commit log for 	2e78df5 for details.

These changes were actually remarkably easy to implement - too easy in fact. The changes required to update the apps and associated code were trivial (and often simplified the code further), and everything checks out in the testing. However, these changes are very likely to have introduced subtle issues specifically with datatype handling. I'd really like everyone to check out this branch and put the code through its paces, and verify that (a) things work, and (b) the images produced have the expected datatype (or at least a sensible datatype...).


Lestropie:
Personally I prefer #382. You've had to add pass-through functions to both the `Image` and `Adapter::Base` classes to make them 'look like' Headers, just so that you can pass them to the `Header` template constructor, which requires that all the member functions in `Header` be available in the template class, yet it's _not_ the `Header` copy-constructor.

`Image` already has a `.header()` member, and as shown in #382 it's trivial to have Adapters provide a fully-formed `Header` rather than just overloading a select few 'HeaderType' functions. So it's possible to construct `Header` exclusively from `Header`; and by doing so, it's explicit in the code which function calls are only accessing header information and are not reading any image data.

Toy example: Let's say you have `Image<float> image; Header H (image);`, and somebody changes it to `auto H (image);`: H is now an `Image<float>`, rather than a `Header`. Obviously this would be a dumb thing to do, but it highlights the unnecessary ambiguity; I don't see any need for that template `Header` constructor, `Header H (image.header())` is so much clearer.

It's been a while since I fully wrapped my head around the nuances of this particular problem, so I may well be missing some benefit of this approach that can't be achieved with mine. But I feel like a big benefit of `updated_syntax` was dropping the `Info` / `ConstInfo` / `Header` / `ConstHeader` / `Buffer*` complexity that necessitated a whole lot of templating, yet we're still using `HeaderType` templates despite there being no benefit from it, and in fact certain ambiguities are lingering around by doing so.

If there's still a problem with data scaling / axis strides in my solution, we should look at that directly, rather than obscuring the solution to the problem in the nuances between `Header (const Header&)` and `Header (const HeaderType&)`.

jdtournier:
Yes, I did spend quite a bit of time going over #225 again to make sure I remembered all the nuances around this issue...

However, the ambiguity this pull request is trying to address isn't quite what you're talking about. Unless I'm getting the wrong end of the stick, the main point of your proposal is to make it clear what _type_ of object is being used where. So your example using the `auto` keyword would be somewhat more explicit, and another advantage is that all the template functions that currently take a `HeaderType` would no longer need to be templated, which would simplify the code somewhat. 

I'm not going to reiterate the [original post](https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-151808441) detailing the rationale, but the main problem this pull request is trying to resolve is more fundamental. Basically, the question is what _information_ should the `Header` obtained from an `Image` contain? When you instantiate a `Header` (whether via `Header::open()` or `Header::create()` or similar), there is no ambiguity as to what information you might expect this structure to contain. When you get a `Header` from an `Image` or `Adapter`, there is scope for ambiguity, _regardless of the mechanism by which you get the `Header`_. This is because the image file on disk will typically be stored using different parameters from those used while the image is manipulated in the process itself. So which version should be returned to the user? The problem is that the most appropriate version is going to be to some extent dependent on what the user wants to do with the `Header`.

In terms of the actual differences, the most common one is the datatype: you can't create an `Image` without giving it a datatype as the template parameter, and this is the type that will be visible in your process. But more often than not, that datatype will not match what is stored on file. The scaling parameters are also problematic: it's obvious that the `Header` (not obtained from an `Image`) needs to have the scaling parameters _as stored on file_, but what should the `Header` reported by the `Image` provide, bearing in mind this will often be used to instantiate another (output) `Image`, possibly with a different datatype? There is also the `.with_direct_IO()`call, which introduces a mismatch between the strides of the `Header` (as stored on file) and the `Image` (as loaded in RAM). 

For the reasons outlined in the  [original post](https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-151808441), currently the backend tries to preserve the input image's attributes as much as possible, to prevent unexpected changes in the strides of the output image, and intensity scaling issues when processing integer-type images with non-default scaling parameters. So the `.original_header()` as it is now returns the `Header` as stored on file, while a `Header` constructed from an `Image` or `Adapter` will return a hybrid between the on-disk and on-RAM versions (in-RAM dimensions and voxel sizes, on-disk version for most other attributes - rationale being that this is likely to be the version used as a template for the output image). This is what led to unexpected differences depending on exactly which version of the `Header` was used to create a new image file, and kicked off this whole discussion. 

So the change here is to make the behaviour consistent with expectation, by ensuring a `Header` obtained from an `Image` or `Adapter` contains strictly the same information as that `Image` or `Adapter` provides via its methods. This means if the user needs information relating to how the data are stored on file, they need to use the `Header` obtained from the initial `Header::open()/create()`. If they want to use the information about the image as it is stored in RAM, then the `Header` obtained from the corresponding `Image` or `Adapter` is what they should use. It is no longer possible to request the original on-disk `Header` from an existing `Image`, it just led to too much confusion. This works because 99% of the time, the in-RAM version is the better one to use, and only in selected cases does the application need to pay special attention to the on-disk `Header`: this includes `mrconvert` in particular, but also `mrinfo`, `mrview`, `mrpad`, `mrcrop`, etc - basically any app that expects to preserve the input datatype without modifying the intensities as such, or simply query these attributes. The rest of the time, typical usage will preserve the strides anyway where it makes sense, and the in-RAM datatype is more appropriate to store the processed output file anyway. 

So coming back to your proposal: the main difference is one of implementation: my changes retain the current behaviour, with `Image` and `Adapter` providing very thin and lightweight views into the relevant data, and providing _methods_ to access the header information; whereas your changes _also_ add a full `Header` class into each of these classes - or at least it should, I've noticed you use the parent `Header` most of the time, which can't work: the attributes of an `Image` will in general differ from those of its parent `Header`, and similarly for an `Adapter` . As I've mentioned previously, I think this add bulk to classes that are designed to be lightweight, makes usage more verbose (you have to add `.header()` everywhere), and doesn't clarify things particularly. Where I think we differ is that I don't see any ambiguity in using an `Image` where a `HeaderType` is expected - it's very much integral to how the software was designed in the first place. Besides, you can use any of these types directly even when the function does expect a proper `Header`, thanks to C++'s implicit type-casting mechanism (this is actually used in quite few places in this changeset, when `DWI::get_DW_scheme(const Header&)` is invoked with an `Image<value_type>` as the argument). I also find it a bit weird that you can use `Image::stride(n);` or `Image::header().stride(n);` to access the same information - it feels needlessly redundant, and people are bound to wonder why, and whether one version is better and/or different from the other. Besides, [as currently implemented](https://github.com/MRtrix3/mrtrix3/blob/header_changes/lib/image.h#L62), your `Image::header()` call returns the original on-disk `Header`, so this last example would in general give different answers - this would need to be resolved to prevent the issues that led to these changes in the first place. This wouldn't be difficult, you'd need to have each `Image` own its own `Header`, I'm just saying that further changes would be needed.

I know that the specific example you used was just to illustrate the kinds of issues that might arise, but I do think that the only time this would be a problem is when using the `auto` keyword. It really should come as no surprise that `auto H = image;` means `H` is of whatever type `image` was... But this can lead to surprises particularly when using Eigen - I've certainly been guilty of assuming that things like:
~~~C++
auto vec = mymatrix.row(0);
~~~
would return an object of type `Eigen::VectorXd`. It doesn't: you get some proxy `Eigen::Block` class that doesn't necessarily have all the methods you'd expect a full-blown vector to have. Eigen calls will return proxy objects most of the time, and using `auto` too liberally will cause trouble - it 's probably a good idea to use it sparingly.

In all other cases, I don't see the point of forcing the use of a `.header()` call where the original `Image` or `Adapter` can currently be used with no ambiguity. What's wrong with this:
~~~C++
// auto in = Image::open (argument[0]);
Header header (in);
auto out = Image<float>::create (argument[1], in);
auto grad = DWI::get_valid_DW_scheme (in);
~~~
that we need to replace it with this:
~~~C++
// auto in = Image::open (argument[0]);
auto header = in.header();
auto out = Image<float>::create (argument[1], in.header());
auto grad = DWI::get_valid_DW_scheme (in.header());
~~~
It's pretty clear in all cases what's going on, I don't see any scope for ambiguity. 

As I see it, the only tangible benefit to such an approach might be to reduce the amount of templated calls in the code, which would help to reduce compile times somewhat. But bear in mind that most of these calls are not templated only on the `HeaderType` - look at the `Image<float>::create()` call for instance: it's templated on both the `ValueType` _and_ the `HeaderType`, so forcing a proper `Header` here still won't allow us to avoid this call being a template. 

Anyone else have opinions on this?

jdtournier:
Having thought about these changes a bit more, it's occurred to me that the following `Header` methods are completely redundant for `Images` and `Adapters`: 

- `datatype()` 
- `intensity_offset()` 
- `intensity_scale()` 
- `format()`

Removing them from these classes makes it very clear that this kind of information can _only_ be obtained from a proper `Header`, and that creating a `Header` from an `Image` or `Adapter` can only set these attributes to defaults (with the exception of the datatype, which is instead obtained from the `value_type`). 

So I've just had a go at removing them altogether, and relying on the compiler to figure out which `Header` constructor to use. Although it did require a little bit of extra SFINAE magic to make sure the right constructor was invoked for classes derived from `Header` (notably `Filters`), this actually compiles and tests just fine with no further changes elsewhere. Clearly, these methods genuinely weren't being used anywhere else, so removing them seems like the right thing to do. 

As always, thoughts welcome.

maxpietsch:
> Removing them from these classes makes it very clear that this kind of information can only be obtained from a proper Header, and that creating a Header from an Image or Adapter can only set these attributes to defaults (with the exception of the datatype, which is instead obtained from the value_type).

I haven't looked into the inner workings of all of this Header / Image business so I hope I can provide a perspective from a layman's point of view...

I absolutely agree that it is confusing if a derived Image or Header does not return its properties but that of the original file as stored on disk as the new Header / Image classes make it not explicit on where the data resides. I think that is more valuable than not needing to save this information in an app if it needs it for write-out.

I also like the idea of separating Header and Image as much as possible to avoid confusion about whether the information it contains actually correspond to use it. But why can't I create a Header from an Image? Essentially, why is there a separation between value_type and datatype?

~~~~
// auto in = Image::open (argument[0]);
Header header (in);
auto out = Image<float>::create (argument[1], in);
~~~~

Wouldn't it make sense to specify once - on load - what datatype, stride, ... the image (`in`) should have and all derived images/ headers (`out`) use the same as default? If I want to load an image with a certain type / stride and then create intermediate images with different type / stride it is pretty clear that one has to create a new Header with those set manually. For instance in mrtransform one can specify the datatype of the input header but the in-RAM images are hard coded with template argument `float`. But what if I don't want to change the type? Or what if I want to use the same precision of the input image for all intermediate and output images?

draffelt:
I've had a "quick" read through the discussion here, and the original discussion in #225. 

I think this pull request solves all the issues from my point of view.

> This means if the user needs information relating to how the data are stored on file, they need to use the Header obtained from the initial Header::open()/create(). If they want to use the information about the image as it is stored in RAM, then the Header obtained from the corresponding Image or Adapter is what they should use. It is no longer possible to request the original on-disk Header from an existing Image, it just led to too much confusion. This works because 99% of the time, the in-RAM version is the better one to use, and only in selected cases does the application need to pay special attention to the on-disk Header

Completely separating the original header on disk from an image makes good sense to me. :+1:  New developers are less likely to get caught out or confused. And I've said before that I think the output should be based on the working images and not the original input header.

> Having thought about these changes a bit more, it's occurred to me that the following Header methods are completely redundant for Images and Adapters:

Even better :+1: 

I know I was a fence sitter in the original #225 discussion, but I think it's obvious enough what is happening when passing in an Image/Adapter as an argument to a constructor that expects a Header. I also prefer the brevity. That said, passing image.header() would be more explicit for non-C++ coders who may not be as aware of implicit type casting, however we can't be expected to not use basic C++ features on the basis of this.

Now that we have a clearer separation between an Image and the original header on disk, it would be nice to include the Transform matrices in `Image`! I propose we rename `header.transform()` to `header.image2scanner_transform()`. And add `image.voxel2scanner_transform()`, `image.scanner2voxel_transform()`, and `image.image2scanner_transform()`. 

While I'm at it. Can I also put forward the idea again to have a convenience function in `Image` to return the current scanner position as an `Eigen::Vector`. And possibly one to return the current index as an `Eigen::Vector` too?

thijsdhollander:
Hey guys, fun times in the Easter weekend all over the place I see... Well, just to add to this one: I'm very happy with the proposed solution in this pull request as well.  In fact, my opinion from https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-152416012 (which was a reply to Donald's original https://github.com/MRtrix3/mrtrix3/pull/225#issuecomment-151808441 ) remains unchanged (spoiler: I was very happy with the solution back then as well).  I fully support the clear separation between `Header` and `Image`.  And indeed, in the vast majority of use cases, the output depends on the working images.  Stripping out those redundant methods will help new developers seeing this logic even more.

@draffelt , about those Transform matrices: I remember this was brought up and discussed a while ago indeed, but I lost/forgot the main arguments in that context.  I feel like I'm the fence sitter in that context now, but I might tip to your side of the fence if I'd only recall why... :smile: I suppose it's mostly to make things more explicit?  That would make sense indeed.

Completely unrelated and useless: whenever I read the words "fence sitter" now, an image of a cow on a fence comes to mind.

draffelt:
It's purely for convenience, not having to create a Transform object every time I want a scanner space position. 

For example in `warpinit`:

~~~C++
  Transform transform (header);
  auto func = [&transform](Image<float>& image) {
    Eigen::Vector3 voxel_pos ((float)image.index(0), (float)image.index(1), (float)image.index(2));
    Eigen::Vector3f scanner_pos = (transform.voxel2scanner * voxel_pos).cast<float>();
    for (auto l = Loop (3) (image); l; ++l) 
      image.value() = scanner_pos[image.index(3)];
  };
~~~

Would be replaced with:

~~~C++
  auto func = [](Image<float>& image) {
    Eigen::Vector3 voxel_pos ((float)image.index(0), (float)image.index(1), (float)image.index(2));
    Eigen::Vector3f scanner_pos = (image.voxel2scanner_transform() * voxel_pos).cast<float>();
    for (auto l = Loop (3) (image); l; ++l) 
      image.value() = scanner_pos[image.index(3)];
  };
~~~

If we had an image method to get the current spatial index as a vector it could shortened to this:
~~~C++
  auto func = [](Image<float>& image) {
    Eigen::Vector3f scanner_pos = (image.voxel2scanner_transform() * image.voxel_pos()).cast<float>();
    for (auto l = Loop (3) (image); l; ++l) 
      image.value() = scanner_pos[image.index(3)];
  };
~~~

And if we had a method to also get the current image position we could write this:
~~~C++
  auto func = [](Image<float>& image) {
    Eigen::Vector3f scanner_pos = image.scanner_pos().cast<float>();
    for (auto l = Loop (3) (image); l; ++l) 
      image.value() = scanner_pos[image.index(3)];
  };
~~~

Now that Rob has generated a base class for the interpolators, transform.h is very small. The `get_default` method is only used in header.sanitise_transform(), so it could be moved there. And the rest could be just members of the Image class. 

I guess the only downside is the (possibly) unnecessary computation of 3 (small) matrices every time an image is created. @jdtournier, I can see the appeal of you wanting `Image` being a lightweight representation that matches the image on disc as close as possible, and I know the extra CPU cycles "jar with your general minimum-effort approach to life". However, I think a minimal effort API for future developers is also important. It would be also easier for new developers to discover this functionality in the `Image` documentation, instead of having to stumble across the documentation in a separate Transform class (though perhaps I would think to look first in the Image API as this is where it is found in ITK). The image class already goes beyond a simple disk representation by tracking the state/position of the current index. Given the strong emphasis on doing everything in scanner space in MRtrix, I think it makes sense to also provide direct convenient access to the scanner space position of the current index as well.

I admit that it benefits code related to registration and so I probably care more than most. However it would also tidy up some tractography code, and simplify the base interpolator class since it would only derive from Image.

jdtournier:
@maxpietsch:

> But why can't I create a Header from an Image? Essentially, why is there a separation between value_type and datatype?

Looks like there's a misunderstanding here... You definitely _can_ create a `Header` from an `Image` - it's just that the attributes that go into that `Header` will be the same as the `Image` would report. You get the in-RAM version, not the on-disk version. The latest changes simply remove information that is not necessary for the in-RAM representation.

The separation between `datatype` and `value_type` is that the former is (a method returning) a _run-time variable_, whereas the latter is a _compile-time type_. The former has to be resolved at runtime from the input, or potentially from user parameters for the output. But the application has to pick a type _at compile time_ with which to handle the data in the app, there's no way around that. What happens here is that the `DataType:: from<value_type>()` is used to return the correct _variable_ corresponding to the _type_ `value_type`. You can always change that if you want by assigning some other variable to the `Header `'s `datatype()` method before creating your output image. But this reflects the type that will be written back to disk, not what you might use to manipulate the data in RAM (that's the `ValueType` template parameter to the `Image`). 

Does that make sense...?

jdtournier:
@draffelt: 

> Now that we have a clearer separation between an Image and the original header on disk, it would be nice to include the Transform matrices in Image! 

Glad to see you haven't fallen off your hobby horse yet... :wink:  I was wondering how long it would be before this idea resurfaced again. I have to say, this idea still doesn't agree with me - difference being that I rarely need to access this information with the type of processing I do most of the time. But really, the main reason I haven't been keen up to now is because I couldn't see a clean way to implement this that didn't involve all manner of hacks to get ensure consistency.

But I think we might now have a satisfactory solution for this. What I'm thinking is that the `Image::Buffer` class (the internal version of the `Header` maintained by the `Image` and shared across instances via a shared pointer) would override the `Header::transform()` to return a fully-formed `Transform`, and the `Image::transform ()` would return a reference to that (now shared) `Transform`, providing direct access to all the matrices you're after. The `Adapter` classes would also be modified to match the `Image`.

This would require the interface between `Header` and `Image` (and `Adapters`) to differ. Alternatively, we could make this an _additional_ method called something other than `transform()` - but that might make things confusing, since we'd now have two methods with similar purposes. Might be better to have a clean separation between what `transform()` means for the `Header`, and what it means for the `Image`...?

But the advantage here is that the internal `Image::Buffer` class is instantiated at the same time as the `Image` - i.e. at a stage where these attributes are locked down, no further change possible. This avoids all the issues related to keeping the various transform matrices consistent. And since it's shared across `Image` instances, there's no wasted memory and needless allocations, which keeps me happy...

Doing this is a relatively big deal though, implying changes throughout the codebase. We'd also need to change the relevant `Header` constructor and assignment operator to match, and _all_ the bits of code handling transformations... Not that it's stopped me before, but it might be worth making sure we're all happy with this idea before proceeding... 

In any case, I won't be able to do anything about this until next week, so there's plenty of time to discuss.

jdtournier:
> To be clear, I realise that #382 doesn't currently solve the data type / scaling / stride ambiguity problem, I should have been more explicit there. What I was referring to was making all Image / Adapter / etc. classes provide a fully-formed Header in a .header() member function, that would accurately depict what is provided by that class, as opposed to being required to provide all individual HeaderType template functions. 

Yes, I figured that's what it was about... What I'm still unclear about is whether your intent is to clarify the code _at the point of use_ (i.e. it's more obvious on inspection that this call will use header information whereas this one might use voxel data, without referring to the function declaration), or to simplify the code that provides that functionality (i.e. by not having to prefix every one of these calls with `template <class HeaderType>`). Or maybe both...?

> So e.g. on Image formation the Image::Buffer header contents would be updated to reflect these parameters; whereas the original result obtained from Header::open() would still provide the image-on-disk representation.

So that would require a few more modifications - the original on-disk `Header` needs to be preserved so that the image write-back can be done correctly for `Images` opened `.with_direct_io()` - an issue that just needed fixing recently in #489. Not impossible to get around, we'd just need to store the original strides, datatype and scaling parameters - but the current implementation makes it easy to use the template `threaded_copy()` function easily as-is in the write-back, which is neat. 

> In functions expecting a Header, passing Image<value_type> as input will require implicit construction of a temporary Header using the template constructor; whereas if the Image::Buffer members were updated appropriately as described above and .header() were used, it'd be implicit downcasting. 

Yes, quite right. I wouldn't necessarily advocate using this feature in performance-critical code for this reason. But it makes sense for something like the `DWI::get_valid_DW_scheme()` call, which is heavy anyway. Using a `Header` argument here allows the function to be regular (non-templated), and compiled in the `.cpp` file, which should help reduce compile times... But in general, anything where performance is important should be a template call. 

Note also that if the function is genuinely performance-critical, it should be declared `inline` in the header file, in which case it doesn't matter a great deal whether it's templated or not - it will be compiled in every compilation unit that includes that header. The compile time for a template function is probably a little bit longer than the equivalent regular function, but probably not by a great deal...

> And since Image::Buffer is shared, it wouldn't make Image any larger or more expensive to copy-construct.

So this is the crux of the problem: this may be true for the `Image` class, but not for a general `Adapter`. For instance, the `Adapter::Subset` or `Adapter::Extract` classes need to modify the origin of the transformation matrix, and hence can't use their parent's `Header` as-is. So either their `.header()` would construct a modified version and return that every time (a bit intense...), or _each instance_ of the `Adapter` needs to store a full-blown `Header`. It may be possible to store that `Header` using a shared pointer as is done in the `Image` class, which would alleviate some of the duplication, but really the main point is that `Adapters` would need a fair bit of modification to work properly, and would become heavier objects.

> Personally I still prefer the explicit .header() calls when reading code: it makes it perfectly clear what methods are only using header information as opposed to image data. 

I get your point. There are calls where this would make things more explicit, in that you could see immediately that the function is only using the `Header` information, not accessing voxel data. But if for instance `Image` derived from `Header`, there would be nothing weird about passing an `Image` to a function expecting a `Header` - it's perfectly sensible C++. The syntax in this case would be identical to what it is now, as far as the user code is concerned. So I don't see this causing a great deal of confusion to C++ programmers.

As an aside, you can already make that distinction from the function _declaration_ - this is why we use the `HeaderType` and `ImageType` template arguments. I appreciate this isn't as clear a distinction _at the point of use_ though - you still have to look up the function declaration to figure this out. But I don't think this is something developers would find odd in any way. And the purpose of the function call should already give a solid hint as to what information it might rely on. 

But more to the point:

> But hopefully appropriate const Header& usage makes it clear enough. And I agree that the image.stride() / image.header().stride() duplicity isn't ideal.

To keep things really consistent, you'd probably need to provide access to the header attributes purely via the `.header()` call, which would only add bulk to the user code, typically without any increase in clarity of intent. So we'd want to preserve access to these attributes via the existing `Image` and `Adapter` methods, which introduces this duplicity. It's really not a bit deal, but I personally prefer things as they are...

jdtournier:
Coming back to [my earlier suggestion regarding the transformation matrices](https://github.com/MRtrix3/mrtrix3/pull/491#issuecomment-201208579): having thought about this some more, I think this is still not as simple as it sounds.

First, my suggestion of having the `.transform()` call return different things for the `Header` and the `Image` is not a good idea. It would entail modifications throughout to ensure that whatever function expects a `HeaderType` and looks at the transform can actually accept an `Image` as well - not impossible, but messy. It would be easier to _add_ additional methods of the type that @draffelt suggested, i.e. `voxel2scanner_transform()`, etc. I do find these call names a bit awkward, but I can't think of anything better... 

On the other hand, adding a `voxel_pos()` convenience call is a no-brainer, requiring essentially no modification. But it would need to be reimplemented for any `Adapter` whose index might differ from its parent (i.e. `Subset`, `Extract`, `Replicate`, etc.). 

However, things get much messier for the other convenience calls to access the transformation matrices for `Adapters`. As I'd mentioned in [a previous post](https://github.com/MRtrix3/mrtrix3/pull/491#issuecomment-203354819), some of these `Adapters` need to modify the transform. This means they would need to recompute all the derived matrices as well, _and_ store them separately from their parent `Image`. As things stand, they already store their own `transform_type`, but they would need to store a fair bit more again if these changes were implemented, which would make them heavier objects. Again, this isn't necessarily a deal-breaker, but it's starting to sound like a lot of extra work both in terms of code update and maintenance, and in terms of runtime overhead. I'm really not convinced it's worth all the hassle...

From another point of view: maybe the problem is that `Adapters` exist at all...? Realistically, there's rarely used more than once in the code (most of them are only used in `Filters`), and while they provide a bit of convenience, it's not obvious that they're worth preserving if they cause all these issues... Most of the time, they're only used for convenience in Loops, etc, so having to maintain additional unused information such as transforms, etc. on the off-chance that they might be used elsewhere where it might matter just feels like a waste of time. 

So I can think of a few options here:

1. maintain things as they are: maintain `Adapters` as _bona fide_ image access classes, with equivalent status to the `Image` class; and don't provide convenience calls to access the transformation matrices.

2. maintain `Adapters` as _bona fide_ image access classes, but also provide convenience calls to access the transformation matrices, with all the mess that entails.

3. downgrade `Adapters` to a lower status, essentially providing _only_ the functionality they were designed to provide. So for instance, the `Adapter::Subset` could be simplified to _not_ maintain its own transform(s), and wouldn't provide a `.transform()` method at all. It would _only_ be used to access data from an ROI. The modified transform would be obtained via a dedicated separate call that would return a fully-formed `transform_type` by copy (i.e. it wouldn't be stored in the `Adapter` class itself), which could be used to initialise the `Header` for a destination `Image`. The idea being to keep Adapters lightweight and agile, and focused on specific use-cases. The documentation would need to be updated to match.

4. remove `Adapters` as a user-visible means of accessing image data, instead declare them at the point of use when required. So for instance, the `Adapter::Extract` would move into `cmd/mrconvert.cpp`, the `Adapter::Median` would move into `lib/filter/median.h`, etc. (i.e. pretty much the only places they're being used at the moment). It would then be up to the individual developers to ensure their `Adapters` work as they expect for their particular application. So this would preserve the flexibility that `Adapters` provide, without giving developers any guarantees that they can be used everywhere an `ImageType` is expected. 

5. remove the ability to use `Adapters` altogether. All calls currently accepting an `ImageType` would instead accept an `Image<value_type>` (note they'd still need to be template functions). All applications that currently use `Adapters` would need to be modified to match. Personally, I'd find that a real shame, there's a few cases where the ability to operate on other types than an `Image` is very helpful (for instance, the `Image` preload & write-back is performed using a [custom TmpImage class](https://github.com/MRtrix3/mrtrix3/blob/master/lib/image.h#L277L300) that can be fed directly into a `threaded_copy()` call - very convenient). 

I have to admit, I'm not convinced either way. It seems a little drastic to remove what is after all a fairly neat feature, just to make it easier to add some convenience functions that in practice only remove a few lines of code here and there... On the other hand, I have to admit that `Adapters` are rarely used in practice, certainly very rarely outside of their initial intended application. So maybe a compromise like options 3 or 4 might be sufficient? It seems to strike the right balance, allowing lightweight `Adapter` classes without having to add all the bulk that would be required to make them self-consistent to the same extent as a full-blown `Image`... 

Thoughts...?

Lestropie:
> What I'm still unclear about is whether your intent is to clarify the code at the point of use (i.e. it's more obvious on inspection that this call will use header information whereas this one might use voxel data, without referring to the function declaration), or to simplify the code that provides that functionality (i.e. by not having to prefix every one of these calls with template <class HeaderType>). Or maybe both...?

Primarily the former, with a sprinkling of the latter (with added bonus of moving non-performance-critical helper functions to .cpp files).

> The original on-disk Header needs to be preserved so that the image write-back can be done correctly for Images opened .with_direct_io()

Yes, I was thinking some kind of 'direct-IO' wrapper that would store this relevant information to be recovered at write-back time. Hard to know how clean / dirty it'd be without actually doing it.

> This may be true for the Image class, but not for a general Adapter. For instance, the Adapter::Subset or Adapter::Extract classes need to modify the origin of the transformation matrix, and hence can't use their parent's Header as-is. So either their .header() would construct a modified version and return that every time (a bit intense...), or each instance of the Adapter needs to store a full-blown Header

This is somewhat done in #382 already. In `master`, these classes initialize relevant member data on construction, and provide access to them via the relevant `HeaderType` functions. In my changes, a helper class builds the fully-formed `Header` class on `Adapter` construction, which modifies the `Header` data members as necessary, and provides const reference access to it via the `.header()` member function. All it would need to reduce copy-construction overhead would be to plonk this in a `std::shared_ptr`.

> But if for instance Image derived from Header, there would be nothing weird about passing an Image to a function expecting a Header - it's perfectly sensible C++.

I guess fundamentally the ambiguity making me squirm is the fact that `Image` quite deliberately does _not_ derive from `Header`, yet has to provide identical member functions in order to be compatible with `HeaderType` functions. Feels contradictory to me. :neutral_face:

jdtournier:
OK, I've pushed a few more things to do with the developer doc, and a couple of minor cleanups. Assuming this checks out OK on TravisCI, I think we should merge what we have now. We can carry on discussing the transform and adapter issues in a separate issue, but I think everyone is happy enough with the current changes...?

thijsdhollander:
Yep, happy to merge.

I'll create a separate issue for the transform/adapter stuff; let's not make any overly drastic changes on that front until we've given ourselves the time to let that one sink in a bit and discuss it properly. :+1:

thijsdhollander:
No one, including TravisCI, seems to complain... I'll go ahead and follow @jdtournier suggestion to merge this in for now.  Further Transform and Adapter discussions have been carried on in #503 .

PR #493, opened 2016-03-23
New testing framework

jdtournier:
As discussed previously on multiple channels, it would be good to have the testing framework within the main repo, to keep the tests in sync with the code. All the machinery now resides in the `testing` folder, and tests can be run using the enclosed `run_tests` script. To avoid having to include all the testing data within the main repo, the testing script will now clone/pull the data from a separate dedicated repo: [test_data](https://github.com/MRtrix3/test_data), which I've forked from the current testing repo to preserve the history and to allow the current `master` branch to perform its tests using the existing framework. Eventually, we will delete the current [testing](https://github.com/MRtrix3/testing) repo, and rely on the new system. Note that since the `test_data` repo is a fork of the current `testing` repo, we won't lose any of the history. 

Comments/feedback welcome. Now lets see how TravisCI handles this one...


jdtournier:
Well I guess that worked OK then... I'll wait to hear back from everyone before merging, just to make sure everyone is aware of and comfortable with this change...

draffelt:
Sounds ok to me...except I guess it's still a bit of a pain that the
testing data is not synced with the code. So it doesn't completely solve
it.

jdtournier:
Sure, but it's a better compromise, right? Would you rather have the testing data in the main repo also? I appreciate it would make things cleaner from a developer's point of view, but I just think it's a bit unnecessary for users...

draffelt:
I'm ambivalent at this stage. It does annoy me having to download GB worth
of stuff for FSL and Freesurfer. However the current testing framework is
only 54MB.

Probably the best solution is to have the test data repo included as a git
submodule <https://git-scm.com/book/en/v2/Git-Tools-Submodules>. Here is a good
blog <http://blogs.atlassian.com/2013/03/git-submodules-workflows-tips/>about
submodules. The run_tests script would just need to initialise and fetch
the data when first run. As long as devs remember to commit changes to the
test data submodule before committing the changes to the tests everything
should be peachy and rosie.



On Thu, Mar 24, 2016 at 7:02 AM, J-Donald Tournier <notifications@github.com

jdtournier:
Yes, I agree the amount of data really doesn't add all that much. And by the time we create proper releases, we might be able to strip out these bits anyway to keep the downloads small - although I have a feeling that might not integrate well with GitHub releases, not sure.

But having had a look into submodules, I'm struggling to get my head around how well that would fit our needs. I think the newer git subtree approach would be an even better fit, since we'd be able to commit to it directly while working in the main repo. As you say, all we'd need to do is have the `run_tests` script set up the `test_data` repo as a subtree when first run, and it should just work...

draffelt:
Yes, I also considered that we might be able to remove the test data temporarily while tagging and building binaries for a release. It's probably not too hard to manage. The only issue is that if for some reason we wanted to go back and test a particular release, we would need to go to the commit before deleting the files. I think a submodule/tree is cleaner. 

You can [commit to a submodule](http://stackoverflow.com/questions/5542910/how-do-i-commit-changes-in-a-git-submodule) too. A Git submodule just keeps track of a single commit that should correspond to (work with) the state of the parent repo (sounds like what we are after?).  I thought that trees were only needed when [more flexibility is required](https://developer.atlassian.com/blog/2015/05/the-power-of-git-subtree/).  [See here](http://stackoverflow.com/questions/31769820/differences-between-git-submodule-and-subtree) for differences too. Also, the benefit of a submodule vs a tree is that the test data is not automatically pulled when cloning the main repo.

jdtournier:
Just had a look through the links you posted. I think I have a clearer idea of how this works, you're probably right that a submodule is probably the way to go. I'm a bit concerned about the git assault course you have to go through to add commits to the submodule though, it seems pretty heavy and error-prone. I can't see an easy way around it though, the minute you request a particular commit, the submodule will be in detached-HEAD state, you can't commit until you checkout `master` again... And then you have to switch back and make a commit on the main repo to tell it to use the new commit. I'll need to think about this a bit more...

jdtournier:
OK, had a go at using git submodules as a backend for the testing data. Let me know what you think...

draffelt:
Had a play today and added some new data and a test. Everything seems easy enough. 

For anyone who normally uses SSH to clone and access the Github remote (like me), note that you need to create a personal access token to push new data to the repo since the testing_data submodule URL is HTTPS (took me a minute to figure out why I could not push with my normal credentials). I think HTTPS is the right choice.

jdtournier:
> note that you need to create a personal access token to push new data to the repo since the testing_data submodule URL is HTTPS

Alternatively, look at the [README](https://github.com/MRtrix3/mrtrix3/blob/master/testing/README.md) to see how to change the remote URL on your local installation...

PR #506, opened 2016-04-01
Changing strides in scripts for input to FSL

thijsdhollander:
I kicked off with the first changes, but we might want to consistently do this for some other scripts as well (even when it might not matter per se).


thijsdhollander:
Ok, so 2 main commits so far; essentially: in `dwipreproc` every occurrence of `-stride +1,+2,+3,+4` has been replaced with `-stride -1,+2,+3,+4`.  There's strong evidence that this change is correct, by testing on a real dataset with macroscopically known motion.  The flipping that this caused before to the output of `topup`, is hypothesised to no longer be present; so the flip that was manually done on the field map, has been removed.  This remains to be tested using data where the `rpe_all` option can be applied (and this latter particular change also only applies for that option).  Finally, in the other commit, I fixed some comments to no longer mention this flipping, and changed a file name to also not explicitly suggest that some flipping had to be or was applied.

thijsdhollander:
So, investigated the potential impact on `5ttgen fsl`: changing `+1,+2,+3,+4` to `-1,+2,+3,+4` does not affect the actual segmentation, and the results still match the anatomy in exactly the same manner as before.  What does appear to be affected is the outline of the (`bet` generated) mask.  Note `5ttgen fsl` by default uses a pipeline consisting of calls to `bet`, `fast` and `first`; so worst case there might be a mix of conventions causing all sorts of unexpected interactions for the pipeline as a whole.  Going by the *assumption* that the need for the stride change seems to be more likely than the previous version, I made this change in commit c02ab52bbc7c32db96a5cf0bca1018dce7cb0ab0 .  But consider this still under investigation... (any further evidence welcome).

Here's the output before and after the change.  Just showing the CSF tissue type, since that one shows the subtle differences along the brain mask:

Before:
before

After:
after


Note again that all segmentations *inside* the brain mask were unaffected by the change... the only difference really seems to be just the brain mask outline.  You might want to open the above images in separate tabs or something, and flick back and forth to observe the small differences.

Lestropie:
If those strides were going to affect anything in this script, I would expect it to be FIRST, and/or the `standard_space_roi` step (provides initial masking prior to BET). Might be worth looking more closely at those two - the latter is probably having a concomitant effect on the BET outline.

Having said that, it's entirely possible that FSL isn't handling the expected axis directions equivalently between all commands, in which case whether we should provide `-stride +1,+2,+3` or `-stride -1,+2,+3` will be command-dependent. I'd rather hear back from FMRIB as to what they can discover rather than doing a blanket change (even if it more-or-less doesn't affect the result).

Lestropie:
Have run `dwiprep -rpe_all` on this branch, and am confident that the field map image aligns with the input images with the `mrtransform` axis flip _omitted_, as was implemented in 45defc1c.

thijsdhollander:
Agreed that we still need to check with "the source" on that front.  I'd be surprised if it were a mix of standards/spaces (but, admittedly, it's only an assumption); that would've (potentially) led to issues for FSL users as well, if they used some of those commands in conjunction.  Even if the change is justified as-is, I'd still explicitly add the stride option for every separate input.  Those changes should then probably also be restricted to the `fsl.py` script (at the moment, the most important one sits in the `5ttgen` script itself, which is a bit weird, since at that point it's not yet specified that the fsl implementation/algorithm will be used).
Just based on output, it's a bit too hard to tell which one (or none, or a mix) would be the correct one.  Some details seem to be slightly in favour of the change, but it's too subtle to be true evidence I'd say.  Here's one such detail that seems to suggest it's super-slightly better after the change:

before_detail after_detail

The hot scale is the T1 image; the grey scale overlay is the CSF segmentation.  Note the very small change in the middle of the image... It's something, but definitely not enough to be convincing.

Good to hear the flip-issue makes sense as well.  This also seemed to make more sense from the point of view that "native" FSL users never reported issues with chaining those 2 commands together...

Lestropie:
> I'd be surprised if it were a mix of standards/spaces (but, admittedly, it's only an assumption); that would've (potentially) led to issues for FSL users as well, if they used some of those commands in conjunction.

I think I'll be surprised no matter what the answer is :-/  It would be weird if their own documentation says they conform to RAS, but the entire code based has been built around LAS for over a decade.

> Those changes should then probably also be restricted to the fsl.py script (at the moment, the most important one sits in the 5ttgen script itself, which is a bit weird, since at that point it's not yet specified that the fsl implementation/algorithm will be used).

This can be fixed: just write the input image to e.g. `input.mif` without modifying any strides, then each individual algorithm is responsible for modification from there. Was probably just to save an `mrconvert` call; and the `freesurfer` algorithm doesn't care particularly.

> Just based on output, it's a bit too hard to tell which one (or none, or a mix) would be the correct one. 

One I'm more interested to see is whether the good-ol' `run_first_all` silent failure is more or less frequent with this change. Not sure if I have data to test that. Though FIRST even has its own internal coordinate system... but let's not go there.

Lestropie:
> Have run dwiprep -rpe_all on this branch, and am confident that the field map image aligns with the input images with the mrtransform axis flip omitted, as was implemented in 45defc1.

Just recalled that in the version history, FSL 5.0.8 made a change to the header output of `topup`. I suspect they fixed that image having an identity transform, but it may also be worth testing this specific part against different FSL versions.

jdtournier:
Right, seems the correct fix is actually to modify the bvecs handling - x-component needs to be flipped when transform determinant is positive. Regardless, I think it's probably still a good idea to make this change, the images will then be pretty much exactly as FSL has historically expected - less likely to introduce issues. But this would be entirely optional - as long as we fix the bvecs handling...

Note that this will _definitely_ mess up people's existing data... If anyone tries to import RAS NIfTI data with bvecs/bvals that worked fine in MRtrix before, they won't work once we fix the bvecs handling. We'll have to make some serious announcements about this.

bjeurissen:
@thijs, those subtle unexpected differences you see when switching between RAS and LAS, could they be due to the fact that some of these algorithms are not entirely deterministic?

jdtournier:
I'm thinking it might be worth merging this branch now, since the changes are beneficial and don't interfere with anything else. We can start a fresh pull request for the bvec handling - and then worry about the implications of that change and how best to handle the fallout...

thijs:
Ok, people, I think you might have messed up. This is not the @thijs you're
looking for... ;)

thijsdhollander:
This is so weird; just got an email from myself from GitHub... well not really. @thijs : I'll fix this in @bjeurissen 's post. ;-)

thijsdhollander:
@bjeurissen : yes, that's what I expected at some point as well.  But I double checked it by rerunning (after the change) a second time, the outcome being exactly the same.  So at least for (the combination of) all steps and FSL programs involved in `5ttgen`, the full process appears to be entirely deterministic.

In the case of `dwipreproc`, I noticed the change didn't only affect the corrected gradient table, but also the specific (image) outcome for all diffusion weighted images (but not the b=0 images).  I reckon this is because `eddy` uses the gradient directions to know which DWIs can be expected to appear more similar.  If these gradients would be reoriented throughout the algorithm for this purpose, then it becomes really crucial to provide them correctly at the input.

I agree with @jdtournier about merging these changes already: they now do result in the *correct* output (images as well as gradients).  The bvec/bval issue remains crucial (and should of course be followed up in a separate pull request), but these changes at least already *avoid* the modified scripts being afffected by that specific issue.

Apart from `dwipreproc` and `5ttgen`, there's some other scripts that interact with FSL as well; we might just as well change the strides in those as well, so as to conform with FSL's historical expectations.

bjeurissen:
I am also in favour of merging

jdtournier:
Right, let's merge. Before we do though, there is the slightly thorny issue of how we handle this publicly. We'll have to come clean about the issue and the fact that it was caused by our incorrect interpretation of the bvecs. We'll need to provide some recommendations for users to avoid further problems, and under what circumstances they might need to rerun their processing. 

I'm actually not too worried about `dwipreproc`, or any of the scripts for that matter. Those users who care about getting the reorientation right can simply rerun their scripts and move on. I'm far more concerned about what'll happen when we fix the bvecs handling. That'll mean a decent proportion of people's data that worked fine in MRtrix before won't work properly afterwards. Thankfully it should be trivial to detect when they're wrong, so it's not as bad as the SH basis change for instance. But we'll probably need to provide instructions for how to verify the bvecs, and how to correct them if needed...

Also, given that the issues are so closely related, it might be worth fixing the bvec handling now as part of the same merge, and make a single announcement about the problem. I'll try to work on that tonight - although I'll probably need some help with testing (I guess it would be a simple matter of running `dtifit` and `dwi2tensor` and verifying the PPDs in FSLView and MRView respectively). Also, any thoughts on how to handle the announcement would be welcome...

jdtournier:
OK, just gave this a crack. Seems to work for RAS/LAS images (i.e. strides 1,2,3,4 / -1,2,3,4). However, this will need to be tripled-checked for all kinds of possibilities using both MRtrix and FSL.

Lestropie:
Personally I'd prefer to revert the scripts to RAS NIfTI export, and fix the fundamental underlying issue. That'll make more sense to people in terms of us discovering what caused the issue and fixing it, rather than pushing a patch that works but is slightly indirect.

> worry about the implications of that change and how best to handle the fallout...

I wore the last one; it's somebody else's turn :-/

jdtournier:
> Personally I'd prefer to revert the scripts to RAS NIfTI export, and fix the fundamental underlying issue. That'll make more sense to people in terms of us discovering what caused the issue and fixing it, rather than pushing a patch that works but is slightly indirect.

I'd be happy with that - we just need to verify that everything does work as expected...

>> worry about the implications of that change and how best to handle the fallout...

> I wore the last one; it's somebody else's turn :-/

:confused: 

I'll be carrying the can on that one, if that's what you're asking...? I'd just like some feedback as to what we tell everyone and what advice we give them as to how to deal with their data being inconsistent...

jdtournier:
> I wore the last one

Or should that have been 'wrote'...?

Lestropie:
No no, I definitely wore it. Like a torn suit that's never been washed. With a "kick me" sign on the back.

jdtournier:
> No no, I definitely wore it. Like a torn suit that's never been washed. With a "kick me" sign on the back.

Couldn't have been as bad as the bug I had in my FA calculation for years: took Kerstin to point out that (3/2) isn't the same as (3.0/2.0) - seems copy/pasting from Matlab isn't always a good idea...

jdtournier:
@lestropie: if we don't make the strides changes in the scripts, are we sure _everything_ will work as expected...? I'm thinking about the flip thing that needed to be done for the `-rpe_all` option, for example. Thinking about this, I do think the _safest_ option is to use the `-1,2,3,4` strides, given that this is likely to be the format presented to FSL the vast majority of the time, and hence the combination that would have received the most community testing...

thijsdhollander:
Yeah, I still agree on that one as well... ever since I noticed (the requirement of) that flip and subsequently learned about FSL's history, I became quite uncomfortable feeding `+1,+2,+3,+4` into any FSL command...  And now noticing that it's causing (deterministically) different outputs as well on other fronts...  I'd rather be safe as well and assume as little as possible.  The proper fix for the bvec/bval is there as well, and obviously our communication should focus (only) on (the consequences of) that when making an announcement; but behind the scenes, we might just as well keep everything as safe as possible.  Going `-1,+2,+3(,+4)` for every FSL input seems to be the most conventional after all.

jdtournier:
Anyone had a chance to check out the latest changes to the bvecs handling? 

Given that we screwed this up the first time round, I'd really appreciate it if someone could feed data to & from MRtrix, test everything works as expected in FSL for all relevant combinations. For the purposes of this test, I think it would be sufficient to try simple tensor tracking (assuming that can be done in FSL...?) to avoid long bedpostx runs. I'd love to say it's enough to try just displaying the tensor PPD orientations to test, but I'm wary of that bug in FSLView causing yet more confusion...

Lestropie:
> I do think the safest option is to use the -1,2,3,4 strides, given that this is likely to be the format presented to FSL the vast majority of the time, and hence the combination that would have received the most community testing...

It's not clear to me from Mark's message whether neurological or radiological is the 'standard' / 'conventional' image format for FSL; only that bvecs assumes radiological irrespective of how the image is stored, and therefore this must be explicitly checked. And radiological / LAS appears to be what FSL handles 'internally'. I can't remember who made the comment about another conversion software providing 'NIfTI' and 'FSL NIfTI' export; I'm curious to know if this is the distinction.

Having said that, as you say we're looking for a 'safe' solution, so maybe we should keep the -1,+2,+3. I'm more thinking ahead that whatever approach we use here will likely propagate to become our 'recommended' handling of FSL compatibility in the documentation outside of these scripts, so we want to be confident that this is what we're sticking to.

thijsdhollander:
> I can't remember who made the comment about another conversion software providing 'NIfTI' and 'FSL NIfTI' export;

...that would've been me. :smile:  I picked that up here:

https://lcni.uoregon.edu/downloads/mriconvert/mriconvert-and-mcverter

I agree we have to be quite confident on different things before we merge and communicate.  It's just from subtle things like that `topup` output that needs the flip when feeding in `+1,...` data, I'd think we might be set for an issue-filled future if we wouldn't make the change to `-1,...` as soon as possible.  Let's say that specific one might be a minor FSL bug that they fix in a next version, then requiring the flip in a `+1,...` data feeding scenario will become version dependent; whereas in a `-1,...` data feeding scenario the script (without the flip) would remain resilient to such changes.

But again, I agree the main communication after all of this gets merged should be focussing on the bvec/bval changes; since that's bound to influence people who have data sitting that's been "converted" for FSL usage (but that they might, in that form, input in MRtrix again as well).

Lestropie:
That sounds like a decision then.

Though this bit irked me:

> Data is flipped top bottom, as in the Analyze 7.5 output, to avoid problems with FSL's coordinate system.

Do they mean left-right? :worried: 

Is it possible to export bvecs / bvals from DICOM even if the image output is NIfTI rather than 'FSL NIfTI'? This should be handy to compare against.

thijsdhollander:
Yep, it can export bvecs/bvals as well for "regular" NIfTI.  The bvecs were the same in both cases (for this dataset), but it appears that the first stride (`-1`) is as well; so that's not providing any new info.  However, just to mess with our collective minds: for "regular" NIfTI, the output had strides `-1,-2,+3,+4`.  Yes, that's a `-2` in there.  This could either relate to that "flipped top bottom" reference, or (and I think this might be it) those strides might be the order in which the data was initially acquired and written (in the DICOM).  If anyone has, e.g., sagittal or coronal acquisitions lying a round, it'd be interesting to see the difference.

thijsdhollander:
So the latter I mentioned above appears to be correct.  Here's the relevant `mrinfo` portions of...

`mrinfo` applied directly to the DICOM folder:
```
  Dimensions:        96 x 96 x 60 x 67
  Voxel size:        2.5 x 2.5 x 2.5 x ?
  Data strides:      [ -1 -2 3 4 ]
  Format:            DICOM
  Data type:         unsigned 16 bit integer (little endian)
  Transform:               0.9994  -4.773e-08     0.03451        -120
                         0.009393      0.9622      -0.272      -119.5
                          -0.0332      0.2722      0.9617      -86.89
```

`mrinfo` applied to the "regular" NIfTI output from `MRIConvert`:
```
  Dimensions:        96 x 96 x 60 x 67
  Voxel size:        2.5 x 2.5 x 2.49982 x 8.8
  Data strides:      [ -1 -2 3 4 ]
  Format:            NIfTI-1.1
  Data type:         unsigned 16 bit integer (little endian)
  Transform:               0.9994  -2.782e-05     0.03451        -120
                         0.009417      0.9623      -0.272      -119.5
                          -0.0332      0.2721      0.9617      -86.89
```

`mrinfo` applied to the "FSL" NIfTI output from `MRIConvert`:
```
  Dimensions:        96 x 96 x 60 x 67
  Voxel size:        2.5 x 2.5 x 2.49982 x 8.8
  Data strides:      [ -1 2 3 4 ]
  Format:            NIfTI-1.1
  Data type:         signed 16 bit integer (little endian)
  Transform:               0.9994  -5.049e-06     0.03451        -120
                         0.009395      0.9623      -0.272      -119.4
                          -0.0332      0.2721      0.9617      -87.56
```

Not to have people worry even more, but still: also notice the changes/differences in the right-most column of the transformation matrices...

thijsdhollander:
Last update for now: those latter changes in the transformation matrices are actually along the vertical / z axis in scanner space.  Note that there was quite some slice angulation applied in this particular scan around the x-axis (in the sagittal plane).

jdtournier:
Glad you guys are having fun with NIfTI... :wink:

So in no particular order:

- > It's not clear to me from Mark's message whether neurological or radiological is the 'standard' / 'conventional' image format for FSL; only that bvecs assumes radiological irrespective of how the image is stored, and therefore this must be explicitly checked. And radiological / LAS appears to be what FSL handles 'internally'.

  Well, the relevant bit is that a lot of FSL is pre-NIfTI, and hence built around the old Analyze format, which is LAS / radiological. This is why I'd argue to ensure our interactions with FSL are via LAS images. 

- >  However, just to mess with our collective minds: for "regular" NIfTI, the output had strides -1,-2,+3,+4. Yes, that's a -2 in there. This could either relate to that "flipped top bottom" reference, or (and I think this might be it) those strides might be the order in which the data was initially acquired and written (in the DICOM).
  
  Spot on. The conversion direct from DICOM (whether using MRIconvert or `mrconvert`) will preserve the strides in the original, which will use the DICOM patient-centered system (PCS): essentially the same as NIfTI but with the x and y reversed. If you just read the voxels in the order that they appear in the DICOM files, they come in with strides -1,-2,3,4 - i.e. a LPS system. Anything else requires reordering the voxels on file. Note this is true for axial images - wouldn't be the case for sagittal or coronal acquisitions, the stride would come out different again. 

- >> Data is flipped top bottom, as in the Analyze 7.5 output, to avoid problems with FSL's coordinate system.
  > Do they mean left-right? 

  Nope, that's what you need to do to produce a LHS image... The issue is that the NIfTI qform, being a quaternion, can only represent rigid body rotations. So it can't represent a LHS coordinate system by itself, since the NIfTI standard coordinate system is RHS. So there is a field called `qfac` whose purpose is to flip one of the axes so that the system can be made LHS - and you guessed it, the axis that gets flipped is the z axis... If you're really interested, have a look at `lib/file/nifti1.h` (basically downloaded direct from the relevant website), it's very well documented in the comments. So you can imagine the fun required to figure the right qform to go along a z-flip to represent the coordinate system you're actually after...

- > Yep, it can export bvecs/bvals as well for "regular" NIfTI. The bvecs were the same in both cases (for this dataset), but it appears that the first stride (-1) is as well; so that's not providing any new info.

  Actually, it does... MRIconvert might be correct here. The difference is that -1,-2,3,4 corresponds to a RHS system, whereas -1,2,3,4 corresponds to a LHS. According to Mark, we'd need to flip the x component of the bvecs if the determinant is positive (RHS). So the fact that the bvecs come out identical in both cases tells you that MRIconvert is taking this into account. 

  I am surprised that the y component wouldn't be affected though..?!? Can you check whether our import of these bvecs (after the fix applied on this branch) is correct, i.e. `mrinfo dwi.nii -fslgrad bvecs bvals -dwgrad` returns the same for both cases...? I have a strong suspicion we might still be wrong... This needs double-checking within FSL, I think, maybe by computing the PPD of the tensor and displaying in FSLView (as long as you have a non-buggy version...). 

- > Not to have people worry even more, but still: also notice the changes/differences in the right-most column of the transformation matrices...

  Well that backfired: I'm very worried about this! This shouldn't be happening, I'm pretty sure I've interpreted all the standards correctly, and that the origin refers to the _centre_ of the voxel at (0,0,0) - I'm pretty sure that's true for DICOM and for NIfTI - it is for MRtrix. So how did this discrepancy come in...? Presumably these images don't align properly in MRView...? Can we check for alignment using any other packages?

jdtournier:
OK, touchscreens are bad. Sorry about that...

thijsdhollander:
> Glad you guys are having fun with NIfTI... :wink:

grumpy-cat-circle-of-no

> Well, the relevant bit is that a lot of FSL is pre-NIfTI, and hence built around the old Analyze format, which is LAS / radiological. This is why I'd argue to ensure our interactions with FSL are via LAS images. 

Agreed.

> Nope, that's what you need to do to produce a LHS image... The issue is that the NIfTI qform, being a quaternion, can only represent rigid body rotations. So it can't represent a LHS coordinate system by itself, since the NIfTI standard coordinate system is RHS. So there is a field called qfac whose purpose is to flip one of the axes so that the system can be made LHS - and you guessed it, the axis that gets flipped is the z axis...

Very enlightening... it's amazing how we all learn new things every day. :astonished: 

> MRIConvert might be correct here.

...ok, so everything that follows ***assumes*** this is the case.  But I wouldn't take it for granted...

So, ***assuming*** MRIConvert is correct in both its functionality to export "regular" NIfTI, including bvecs and bvals for that; and also in its functionality to export "FSL" NIfTI, again including bvecs and bvals for that..... well, then I'm now 100% sure we are still not getting it right.  In the case of the dataset I described before, the case that we are clearly getting wrong is the "regular" NIfTI one, i.e., the `-1,-2,3,4` one (RHS).

I double checked: both bvec files as output by MRIConvert are the same.  However, importing into MRtrix clearly yields different results.  Running `mrinfo dwi.nii -fslgrad bvecs bvals -dwgrad` gives very different results.  Not just a flipped sign of a given column, different values overall.  Running an MRtrix tensorfit and visualising the principal eigenvectors in MRview shows the effects of a sign flip along the y-axis in *image* space (the aforementioned "regular" NIfTI, `-1,-2,3,4`, RHS one being wrong).  So it appears as if the non-trivial difference that I saw from the `-dwgrad` output was caused by a (wrong) flip of the y component of the bvecs (in image space); but of course the `-dwgrad` output is in scanner space and recall that there is significant slice angulation in this dataset...

> Presumably these images don't align properly in MRView...?

Nope.  There's a shift of about 1/4th of a voxel (voxel size was 2.5mm isotropic) along the y axis in image space between both images...

Again, we're going down the path of assuming MRIConvert is correct of course... I'd still be careful about that part at this point.  If anyone else wants to independently test MRIConvert outputs, definitely go ahead and give it a check.

About testing import in FSL, or running FSLview: since I have the NeuroDebian version here, I'm not too keen on trusting FSLview... would be good if someone could do a bit of testing with an FSL compiled from source (or any other set of binaries from their website).

jdtournier:
OK, we've had a bit of fun over here looking into this too. Short version: looks like everything checks out OK... 

- Seems the bvecs for the MRIconvert _standard_ NIfTI output don't match - computing tensors in FSL and displaying in FSLView (version 3.2, not affected by the display bug...) shows that the orientations are wrong using that combination. OK using their FSL NIfTI output.

- The transform produced by MRIconvert seems to suffer from some inaccuracies, although I can't put my finger on it... Using `mrinfo` on the raw DICOM, then a NIfTI generated using `mrconvert` as-is (producing strides -1,-2,3,4), then a NIfTI generated using `mrconvert -stride -1,2,3,4`, and finally a NIfTI generated using `dcm2nii` (as recommended on the FSL wiki) shows all of these are entirely consistent. I've no idea why MRIconvert produces transforms that don't match perfectly...

- Tested `mrconvert -export_grad_fsl` as-is, and with `-stride 1:4` (NIfTI-standard RAS) `-stride -1,2,3,4` (FSL NIfTI LAS), and `-stride -2,-1,3,4` (PLI, still a RHS coordinate system, but with a 90° rotation when displayed in FSLView). All data fed through FSL's `dtifit`, and all check out OK - the PPD's line up with the anatomy as expected (although the data are oriented funny in the PLI case...). 

So I _think_ I'm satisfied that we are now handling the bvecs correctly. Assuming you're all happy with this, I'll draft up a blog post and circulate, just to make sure we're all happy with how this will be handled before we merge and announce...

thijsdhollander:
> Seems the bvecs for the MRIconvert standard NIfTI output don't match

Good to hear. :-)  I was also getting worried about this; as you mentioned before, I'd had been expecting a sign flip in the y component.

> The transform produced by MRIconvert seems to suffer from some inaccuracies

Even better; that should be all worries out of the way then!

> So I think I'm satisfied that we are now handling the bvecs correctly. Assuming you're all happy with this, I'll draft up a blog post and circulate, just to make sure we're all happy with how this will be handled before we merge and announce...

As far as I can tell, there's no more evidence of the updated bvec handling being incorrect now; so happy to proceed as suggested. :+1:

thijsdhollander:
So while I'm on the move, I just remembered there's still other scripts that interface with FSL as well.  If we're going to consistently recommend the `-1,...` approach for interfacing with FSL, it'd probably be a good idea to update those as well, while we're at it?
`dwipreproc` is covered, as well as `5ttgen` (but I always welcome double checking); one other one I can think of off the top of my head is `dwibiascorrect`?
Also, at the moment, `5ttgen` is changed overall to `-1,...`; even though there's separate FSL and Freesurfer options for that one.  In this case, the change doesn't / shouldn't affect the Freesurfer version.  But for, e.g., `dwibiascorrect`, there's also the ANTS version; where I'm (just here and now, without testing) not fully sure whether that one can live with a `-1,...` change, without introducing a flip in the output bias field (I'm actually thinking this is what will specifically happen for the ANTS version; that one should probably still be fed `+1,...` NIfTI's).
Any other scripts on the TODO list?

Lestropie:
Just went through and changed what seemed appropriate. Also changed `5ttgen` and `dwibiascorrect` to only map to LAS if FSL is being used. Yet to test everything through though.

thijsdhollander:
Great :+1: , thanks for that!  As to testing: I *think* we should be safe enough if, for any script that was changed, a simple/common use case is run before and after the script change.  In principle, there shouldn't be any gross changes to the output (but minor ones, on the other hand, might be expected).  Sudden obvious flips are of course a major warning; in which case we should panic and run around in disarray before we proceed looking into the matter. :-)

Issue #448, opened 2016-02-25
mrmath product was doing a sum


thijsdhollander:
What the title says.  I quickly patched it up in f731f6711a88b4bc0d8b551f06d2e7c0f22c49ca .  Just highlighting it over here; is everyone happy with the actual behaviour of it?  One slightly weird case I could think of, is if every value fails `isfinite`, the result will be 1.  Feel free to change that if it would be considered not the correct/wanted behaviour for that scenario.

jdtournier:
Well that's not great... How long has this been the case? I'm assuming this isn't being tested properly on TravisCI then? I hope it wasn't one of my sleep-deprived screwups... 

Otherwise, I'm thinking it might be worth having a `-finite_only` option or something to allow the user to decide how to treat inf and nan. In which case getting a 1 might not be so bad? Otherwise it is indeed quite tricky to catch that case efficiently...

jdtournier:
Hehe... This wasn't my screwup - I feel better now... :+1:

jdtournier:
Just had a thought about this issue when all volumes fail `isfinite()`. How about this approach: 
~~~
out = NaN;
for (auto in : incoming)
  if (isfinite (in))
    out = isfinite (out) ? out * in : in;
~~~

That should leave the value set to NaN if there genuinely weren't any finite values for all volumes for that voxel, and do the expected product otherwise. What do you reckon?

thijsdhollander:
That probably makes the most sense indeed; better to highlight the meaninglessness of that case by a NaN, rather than a 1.

thijsdhollander:
I'll quickly give it a fix.

thijsdhollander:
Fixed it for now (as suggested above).  As mentioned in #509 , it might be worthwhile to check some of the other operators as well, as to what they yield for these special cases.

PR #509, opened 2016-04-04
mrmath product sensible outcome for all isfinite fails

thijsdhollander:
...to deal with #448 in a reasonably efficient manner.


thijsdhollander:
Ok, I'll merge this in for now.  @Lestropie just mentioned me it might be worthwhile at some stage to look at whether the behaviour of each operation in `mrmath` makes (similar) sense for these special cases.  The code suggests that different operators have been written by different people over time: it's clearly a bit of a "mix of styles", if that makes sense.

PR #511, opened 2016-04-04
Connectome no config

Lestropie:
Would appreciate a second set of eyes on this one, since it's a decent change and has a widespread effect in how we handle connectomes.

More details in commit messages, but the fundamental change is:

There's no more 'connectome config file' concept whatsoever. Instead, everything is based entirely on 'lookup tables': `labelconvert` (formerly `labelconfig`) permutes indices from one lookup table to another. Hopefully this makes things easier to manage, since you only have to track which lookup table is associated with your parcellation image, without worrying about whether or not the indices have been permuted relative to that table.


Lestropie:
Adding a new label 'test wanted'. Will reserve 'help wanted' for cases where someone needs assistance in finding a bug, implementation, contributions simply due to the size of the issue and number of hours required, etc.. 'Test wanted' can be used particularly for pull requests, to state that the primary developer feels that the code requires further review & testing by an independent person before it can be merged to master.

Lestropie:
Superseded by #653.

PR #512, opened 2016-04-05
Fixes issue in GLM scale contrasts function. 

draffelt:
For certain design matrices, the Math::pinv (which uses a LDLT decomposition) was falling over. It now uses a FullPivLU decomposition instead.


draffelt:
I'm going to merge this now. Tests for stats commands are on my todo list, but I won't have time to get it done soon and this fix is urgent.

PR #516, opened 2016-04-06
A few tweaks to the build script and developer documentation

jdtournier:


jdtournier:
Just thought I'd flag up that this would require a fresh `./configure -dev` to preserve the effect of that `-dev` option...

Otherwise, this is relatively benign. Everyone happy for me to merge...?

thijsdhollander:
:+1:

PR #520, opened 2016-04-07
mrview ODF tool: Fix preview in dixel mode

### LINKED ISSUE #500: ODF preview in "amps" mode
screenshot from 2016-03-31 14 04 21

...there's something wrong with it.  In the above case, I loaded up a DWI dataset (in `.mif` format) straight into the tool.  It automatically detected that, correctly assumed to use the included gradient table for the directional information, and the overlay itself seems to be as expected as far as I can tell.  So it's just specifically the preview that is broken in some way...

Related, but not really crucial: I noticed the "amps" mode removes the "detail" setting from the interface (which makes sense), but there's still a detail setting in the preview.

Lestropie:
Amplitude data sent to the preview renderer was not being filtered based on the selected shell, resulting in noisy ODF shapes as the wrong amplitude would be read for each direction.
Also enable / disable the level of detail control in the preview window according to whether the data is dixel-based (and hence whether this control has any meaning).
Closes #500.
Also modified so that b=0 shells no longer appear in the list of available shells to display.

@thijsdhollander Have a quick go at breaking it?

### DIFF
--- a/src/gui/mrview/tool/odf/odf.cpp
+++ b/src/gui/mrview/tool/odf/odf.cpp
@@ -333,7 +333,12 @@ namespace MR
             for (int y = -ny; y <= ny; ++y) {
               for (int x = -nx; x <= nx; ++x) {
                 Eigen::Vector3f p = pos + float(x)*x_dir + float(y)*y_dir;
-                get_values (values, settings->image, p, interpolation_box->isChecked());
+
+                // values gets shrunk by the previous get_values() call
+                if (settings->mode == GUI::DWI::Renderer::mode_t::DIXEL && settings->dixel.dir_type == ODF_Item::DixelPlugin::dir_t::DW_SCHEME)
+                  values.resize (settings->image.header().size (3));
+
+                get_values (values, *settings, p, interpolation_box->isChecked());
                 if (!std::isfinite (values[0])) continue;
 
                 switch (settings->mode) {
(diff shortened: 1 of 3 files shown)


thijsdhollander:
As usual, I can break it (that's what I do). :wink: 

No, so, overall it seems all good now; and switching back and forth between something where the detail setting applies or not, also correctly enables and disables it.

The only case where I can break stuff a bit, is with multi-shell data: upon switching shells, the preview is broken in a similar manner as the original issue.  Upon changing the focus (or basically doing anything that enforces re-evaluating the preview), it's fixed again.

Lestropie:
How'd I miss that?  ><  Fixed.

thijsdhollander:
Ok, so the dixel preview is unbreakable now; but the axes break upon switching shells.

screenshot from 2016-04-07 16 23 08

 :confounded:

jdtournier:
Pretty cool though. 
I suspect there's a buffer overrun on the OpenGL side somewhere, corrupting the data for the axes?

Lestropie:
Unfortunately this is taking a little time to get to; I can't reproduce the fault on any of my own systems. But I want to fix and merge before we push #506 and potentially tag.

It has something to do with the DWI renderer getting an OpenGL "Invalid operation given current state" error when the dixel mesh vertex buffer is re-written, and therefore that data is instead going to the previously bound buffer, which happens to be the axes vertices.

jdtournier:
A few more assert failures:

- when opening the ODF preview tool: 

  ~~~
  QObject::connect: No such slot MR::GUI::MRView::Tool::ODF_Preview::lighting_update_slot() in src/gui/mrview/tool/odf/odf.cpp:536
  ~~~

- when switching between SH and dixel ODF images on the GUI a bit too quick with the ODF preview open:

  ~~~
  mrview: src/gui/dwi/render_frame.h:87: void MR::GUI::DWI::RenderFrame::set_lmax(int): Assertion `mode == mode_t::SH' failed.
  ~~~

Issue #513, opened 2016-04-05
don't use delayed write-back for temporary images


jdtournier:
Currently, the `Image` backend will always allocate a RAM buffer for images accessed read-write, and commit this buffer to storage in the destructor. This was done to avoid nasty network saturation issues when writing data out-of-order on a memory-mapped image backed by a network-based filesystem (i.e. pretty much every cluster setup...). 

Ideally, we'd like to avoid this delayed write-back for images backed on local filesystems, since the system copes just fine (typically) with scattered writes to a memory-mapped file (at least when the filesystem isn't mounted with the `sync` option, which isn't the default anyway). The problem is that it's almost impossible to detect whether the filesystem is local in a portable fashion... (if anyone has some bright ideas on that front, I'm all ears). But one situation where it might be safe to assume that the filesystem is local is for piped temporary images - on all platforms, the default location for these images is local, and in most OSs, it's even RAM-backed, the data won't even get written to disk. 

So maybe we can forego the delayed write-back for piped images, which would speed up all pipelined operations with relatively little effort. Thoughts?

jdtournier:
I should also add that this would halve the RAM usage for pipelined data: since the temporary file will be in `/tmp`, which on a modern distribution would be mounted onto a RAM-backed `tmpfs` filesystem, writing directly to a memory-mapped file avoids the need to store both the buffer and the file in RAM simultaneously while performing the write-back... Seems pretty worthwhile to me!

Lestropie:
Makes sense to me, I don't see any issue with it.

With regards to non-temporary images: is there anything stopping us from using a config file entry to toggle between enforced delayed write-back and memory-mapping?

draffelt:
I agree. It's unlikely that the tmp directory will be a networked drive.
That said, you never know with a cluster environment so it might be good to
have a config option as Rob suggested

jdtournier:
OK, bit confused here. Seems @lestropie was suggesting a config file option to avoid delayed write-back _every time_, whereas @draffelt suggests the option would apply only to temp files. I think for temp files, it should always be safe to use direct memory-mapping. There is already an option to specify a different temp file prefix if/when `/tmp` isn't adequate.

[...]

PR #522, opened 2016-04-07
Use memory-mapping for output images if held on local FS

jdtournier:
As discussed recently in #513.

This detects the filesystem that the file is held on using [statfs()](http://man7.org/linux/man-pages/man2/statfs.2.html), and uses memory-mapping for RW files if the FS isn't reported as NFS, CIFS or FUSE (which should also catch SSHFS & GlusterFS). Otherwise the current delayed write-back approach is used. This also checks whether the FS is mounted with the synchronous flag, and switches to delayed write-back in this case too, since that would also be expected to kill performance (and hammer the drive...).

This also applies to temporary (piped) images, so this will use direct memory-mapping for these data too if the temp folder is local (which it _really_ ought to be). So this should avoid double RAM allocations when using pipelines on `tmpfs`, for instance.

This works fine on Linux, and performance is slightly better - but nothing to get all that excited about. I guess the real driver is the reduced RAM usage this implies, together with allowing more flexibility to the system to manage swapping if necessary on RAM-limited systems.

However, I've only tested this on Linux... It looks like [it should work fine on MacOSX](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/statfs.2.html), given that the same call is available. Not so sure about Windows though... haven't had a chance to try it yet. Would appreciate some thorough testing on all platforms. :smiley:

Also, if anyone knows of other network filesystems we should be checking for, please speak up!

### DIFF
--- a/lib/file/mmap.cpp
+++ b/lib/file/mmap.cpp
@@ -69,7 +69,7 @@ namespace MR
           delayed_writeback = true;
         }
 
-        if (fsbuf.f_type == 0xff534d42 /* CIFS */|| fsbuf.f_type == 0x6969 /* NFS */ || fsbuf.f_type == 0x65735546 /* FUSE */) {
+        if (fsbuf.f_type == 0xff534d42 /* CIFS */|| fsbuf.f_type == 0x6969 /* NFS */ || fsbuf.f_type == 0x65735546 /* FUSE */ || fsbuf.f_type == 0x517b /* SMB */) {
           DEBUG ("\"" + Entry::name + "\" appears to reside on a networked filesystem - using delayed write-back");
           delayed_writeback = true;
         }
--- a/lib/file/mmap.cpp
+++ b/lib/file/mmap.cpp
@@ -119,11 +119,6 @@ namespace MR
           DEBUG ("\"" + Entry::name + "\" appears to reside on a networked filesystem - using delayed write-back");
           delayed_writeback = true;
         }
-
-        if (fsbuf.f_flags & 0x0010 /* ST_SYNCHRONOUS */) {
-          DEBUG ("\"" + Entry::name + "\" resides on a synchronous filesystem - using delayed write-back");
-          delayed_writeback = true;
-        }
 #endif
 
         if (delayed_writeback) {
--- a/lib/file/mmap.cpp
+++ b/lib/file/mmap.cpp
@@ -17,6 +17,13 @@
 #include <unistd.h>
 #include <zlib.h>
 
+#ifdef MRTRIX_MACOSX
+# include <sys/param.h>
+# include <sys/mount.h>
+#else
+# include <sys/vfs.h> 
+#endif
+
 #ifdef MRTRIX_WINDOWS
 #include <windows.h>
 #else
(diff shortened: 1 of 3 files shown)


jdtournier:
OK, just added SMB to the list of network filesystems - hadn't spotted it till now, I thought CIFS would cover that...

Lestropie:
`sys/fs.h: No such file or directory` on my Windows box. I'll have a go at this tomorrow night.

jdtournier:
OK, had a brief look into the Windows side of things - we can't rely on the `statfs()` call, it's not provided within mingw. Not too surprising really, it's a POSIX call...

However, I reckon we could still get this to work using the Windows native API [GetDriveType](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364939(v=vs.85).aspx) function: one of the possible return values is `DRIVE_REMOTE` - which seems just the ticket... Just need to figure out how to get the `lpRootPathName` from the filename, and we should be in business...

Won't have time to look into this further any time soon, so if someone else wants to have a go... :grin:

maxpietsch:
Had a quick look at it: Windows has a POSIX layer but statfs is not implemented: http://research.microsoft.com/pubs/183461/posix-emulation-submitted.pdf

And [gnuwin32](http://gnuwin32.sourceforge.net/packages/libgw32c.htm) has some [code](https://sourceware.org/git/?p=glibc.git;a=tree;f=sysdeps/unix/sysv/linux;hb=0c5b8b5941e036dcaac69cecee9f01fdf9218e6e) about getting file system information but I don't know what exaclty is implemented. fstatfs is [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/fstatfs64.c;h=bba99186b5ab5352f4e796dcde5007e1069dd481;hb=0c5b8b5941e036dcaac69cecee9f01fdf9218e6e).

jdtournier:
@maxpietsch: I don't think it's worth trying to find a Windows-friendly version of `statfs()` - in many ways, it's not ideal. What we want is something that tells us whether the file is local or remote. The Windows GetDriveType() function seems to give us precisely this information, and we don't then need to rely on third-party hacks. In many ways, that's the function I would have liked to have access to on the other platforms...

Lestropie:
That correctly detected drive type on both local and network drives on my laptop. :+1:

jdtournier:
> That correctly detected drive type on both local and network drives on my laptop. 

I assume this was _not_ a Windows laptop...?

Anybody tried on a Mac yet?

Lestropie:
> I assume this was not a Windows laptop...?

Prior commit:
> File::MMap: Detect drive type on Windows

jdtournier:
You beaut! Glad you took the hint :wink:

Just need to hear back from Mac users now...

maxpietsch:
Also works on OS X for sshfs mounted folder and USB stick.

jdtournier:
Assuming this checks out OK, everyone happy to merge...? This does have the potential to create issues for users on e.g. clusters... But I _think_ it does what it's supposed to...

jdtournier:
I think I'm going to merge this, since no one has complained so far... Let's see whether any issues arise as we go.

PR #525, opened 2016-04-08
fixed small typo from #524

maxpietsch:


jdtournier:
Darn, merged too quickly... :wink:

PR #526, opened 2016-04-08
Update Windows install instructions, DLL location, and config file location

bjeurissen:


jdtournier:
Just added a few more changes as discussed elsewhere. Main one is that the DLL will now be created in the `release/bin` folder, removing the need to set the `PATH` to include the `release/lib` folder (and allowing running commands directly without setting the PATH at all, i.e. by invoking `~/mrtrix3/release/bin/mrconvert`). The location for the config file is now set to the same as for Unix: `/etc/mrtrix.conf` for the system-wide version, `~/.mrtrix.conf` for the user version. Unfortunately, `~` doesn't resolve to the MSYS2 home folder, but to the actual Windows user home, i.e. `/c/Users/Joe Bloggs` - not a big deal, but not unlikely to trip people up either...

bjeurissen:
Great!

There must be a way to fix the home folder issue, no? How would all the other apps function properly if they were reading and writing config files to the windows home dir?

jdtournier:
> There must be a way to fix the home folder issue, no?

Actually, there is... MRtrix3 was set to [look in the USERPROFILE environment variable](https://github.com/MRtrix3/mrtrix3/blob/master/lib/file/path.h#L33) on Windows. I'll change it back to `HOME` now, check whether it works. Note this probably means that the config file will be ignored if the commands aren't run within the MinGW shell. Personally I think that's fine, it's _so_ much better than the native shell... but others might disagree.

jdtournier:
Using `HIOME` works fine. Let's roll with that.

jdtournier:
@bjeurissen, @lestropie: feel free to hit merge if you're happy with these changes...

Lestropie:
Couple of things from playing around:

1. `Path::is_file ("/etc/mrtrix.conf")` returns 0 on my system, even though the file is there and it resolves just fine in the terminal using that exact path. Not quite sure what to do about that one.

2. There was a circular dependency between `load_standard_options()` and `File::Config::init()`. Hopefully the changes I've pushed sort it out.

3. Looks like syntax colouring & line clears actually work in the MSYS2 terminal. Would people be happy for it to default to on for all systems (when not piping to file)? Perhaps with instructions somewhere in the FAQ of what the VT100 codes look like when a terminal doesn't support them & how to disable them?

jdtournier:
1. Not sure if be too worried about this at this stage. Dunno about you guys, but I've never actually used the system-wide config file... Probably simplest just to document it and create an issue about it, just so we don't forget.

2. Looks good. Not sure what the problem was, but this seems innocuous enough.

3. Great! Happy to default to that if it works.

Lestropie:
> > There was a circular dependency between load_standard_options() and File::Config::init(). Hopefully the changes I've pushed sort it out.

> 2. Looks good. Not sure what the problem was, but this seems innocuous enough.

* `File::Config::init()` contains `INFO()` and `DEBUG` calls, which require `App::log_level` to be set - this is done primarily in `load_standard_options()`, where `-info` / `-debug` / `-quiet` are parsed.

* `load_standard_options()` called `File::Config::get_bool()`, which relies on `File::Config::init()` having completed.

Change is just to do `load_standard_options()` first, and delay the `File::Config::get_bool()` call.

> > Looks like syntax colouring & line clears actually work in the MSYS2 terminal. 

> Great! Happy to default to that if it works.

I'll do this when I get home; I can get a text grab from my old Git Bash terminal with the terminal colouring on so I can give an example of what it looks like when the terminal doesn't support it.

PR #528, opened 2016-04-11
mrview ODF tool: Change ODF type handling

### LINKED ISSUE #521: mrview ODF tool: Select ODF type at image open
Can't find where the original discussion occurred.

Instead of having a combobox in the ODF tool to select the type of orientation overlay (SH / tensor / dixel), have separate open buttons at the top of the tool. Lose the ability to open multiple overlays of different types with a single file selection, but reduces the amount of sanity checking required & potential for obscure bugs due to flicking between types. Will also stop tensor images from initially being opened as SH images & displaying with wrong shape / scaling.

Lestropie:
Setting the type of ODF information present in an image is now done at image load (using one of three buttons / command-line options), rather than being part of the interface.
Closes #521.


jdtournier:
This breaks MRView... If I open a CSD file using the ODF display button, I get a frank segfault. This is on a 12.04 system with ATI graphics, and an Arch system with NVidia graphics. Unfortunately it's merged now and several further merges build on it, so it's not simple to revert cleanly. This needs fixing ASAP...

PR #529, opened 2016-04-11
MRview: Windowing reset fixes

### LINKED ISSUE #405: Resetting windowing (Esc) "lags behind"
Cryptic title to basically describe this: when looking at a slice, then changing the slice direction (e.g. from axial to sagittal by pressing 's'), then hitting escape to min-max window... it uses the min and max of the slice before the slice direction was changed (e.g. the axial slice of my previous example, not the sagittal one).
To easily see this, scroll to, let's say, an axial slice near the top of a brain-masked T1 or something (i.e. where the intensity range shown is quite small), then switch to a coronal or sagittal slice by just hitting 'c' or 's' once, then pressing escape.  Given you set the "experiment" up as such, it will be quite obvious that the range is way to small for the coronal/sagittal slice shown.

This is using `mrview` in the updated syntax branch (haven't checked master).

Lestropie:
Anybody want to test this?

When the user requests an intensity windowing reset, this will now be based on the min/max of the current slice if locked to axes, or the min/max of the volume if not.
In addition, if the current intensity windowing is invalid, the first non-empty slice encountered will force a windowing reset, as discussed in: http://community.mrtrix.org/t/trouble-with-nibabel-generated-nii-file/102/12.
Closes #405.

### DIFF
--- a/src/gui/mrview/gui_image.cpp
+++ b/src/gui/mrview/gui_image.cpp
@@ -177,7 +177,7 @@ namespace MR
 
         std::string cmap_name = ColourMap::maps[colourmap].name;
 
-        const bool windowing_reset_required = (display_range < 0.0f);
+        const bool windowing_reset_required = (!std::isfinite (display_range) || (display_range < 0.0f));
 
         if (cmap_name == "RGB") {
 
--- a/src/gui/mrview/gui_image.cpp
+++ b/src/gui/mrview/gui_image.cpp
@@ -123,7 +123,9 @@ namespace MR
           ImageBase (std::move (image_header)),
           image (header().get_image<cfloat>()),
           linear_interp (image),
-          nearest_interp (image)
+          nearest_interp (image),
+          slice_min { NaN, NaN, NaN },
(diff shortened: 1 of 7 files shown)


jdtournier:
Just gave it a spin - everything seems to work fine, only issue is that that colourbar and actual display range aren't actually updated when the min/max is finally set. I just tried with an off-centre ROI image, the display is black at first (better than full white), and the ROI is correctly displayed when it comes into view. But the colourbar still report zero as the min _and_ max, and hitting Esc to force a reset clearly does affect the image too. It seems the `reset_windowing()` slot isn't being invoked?

Lestropie:
OK, so there was differential behaviour between if the first slice displayed when `mrview` is run is empty (colourbar from 0 to 0, displayed colourbar limits failed to update on windowing reset), and if the windowing is reset on an empty slice (colourbar from 0 to -1, displayed colourbar limits update fine). I was testing on the latter. f2b8579 should get it.

jdtournier:
Yep, that works fine now. I'll merge that too...

Issue #424, opened 2015-12-15
ODF tool's sagittal preview is wrong


thijsdhollander:
Ok, this suddenly came to mind; saw it just before I left for Belgium and didn't have the time to write up an issue or investigate further.  This was on updated_syntax before @Lestropie merged his branch that introduced dixel and tensor overlays in that tool.  Basically, when using the ODF preview "sub-tool", and looking at the image in sagittal orientation (locked to axes), the preview was in another orientation (which was also not the axial or coronal one, but something like upside-down or rotated 90 degrees in another orientation).  Don't know if it's still an issue, and if it's consistently reproducible in the way I just described though...

jdtournier:
I can't reproduce... Anyone having similar issues?

thijsdhollander:
Just pulled the current/latest version of the updated_syntax branch, and I still get the aforementioned issue.  Here's a few screenies (region shown is the splenium).  Axial and coronal view are fine, but the preview's orientation is wrong for the sagittal one.

Axial:
![screenshot from 2016-01-06 12 45 28](https://cloud.githubusercontent.com/assets/11952694/12132872/219f5ffc-b474-11e5-8a6d-85d3ca0261a7.png)

Coronal:
![screenshot from 2016-01-06 12 45 36](https://cloud.githubusercontent.com/assets/11952694/12132876/256399f0-b474-11e5-9b48-e6579cc1de44.png)

Sagittal:
![screenshot from 2016-01-06 12 45 44](https://cloud.githubusercontent.com/assets/11952694/12132879/281236f2-b474-11e5-8321-be10e19e0a72.png)

jdtournier:
OK, confirmed: sagittal is definitely not right. The other issue is that the ODF preview doesn't update the first time on a change of projection (i.e. pressing A, C or S) - only on the second key press...

Lestropie:
That should be the orientation sorted; with the delayed update, it seems to be calling the update function as it should, but the projection that the ODF preview window is accessing has itself not yet been updated to reflect the new orientation. But that's the same as the master branch. So gonna have to dig around in the guts a bit.

jdtournier:
@Lestropie, can we close this?

Lestropie:
The delayed update issue is still there (i.e. the preview window doesn't immediately rotate on a change of projection).

Lestropie:
[...]

PR #531, opened 2016-04-11
Fix lag in ODF preview

jdtournier:
See discussion in #424.


Lestropie:
@jdtournier Moved some of the updates I made as part of the `odf_preview_update` branch into this branch, that I thought we should keep given that branch won't be merged. They're split into separate commits so it's clear what belongs to what. Make sure you're happy with all of these.

Issue #530, opened 2016-04-11
sh2power is wrong (I think)


dchristiaens:
Say I compute the RMS amplitude of some SH function as
```
sh2amp sh.mif dirs500.txt - | mrcalc - - -mult - | mrmath - mean - -axis 3 | mrcalc - -sqrt rms0.mif
```
Normally, one should be able to compute the same thing with `sh2power`, i.e.,
```
sh2power sh.mif - | mrmath - sum - -axis 3 | mrcalc - -sqrt rms1.mif
```
However, I noticed that these have different contrasts:
![rms0.mif](https://cloud.githubusercontent.com/assets/7466983/14423407/0022aef2-ffdc-11e5-8927-5fde0fedd627.png)
![rms1.mif](https://cloud.githubusercontent.com/assets/7466983/14423435/1b246c9a-ffdc-11e5-9cc7-31d18dcdfe78.png)

However, if I compute the root-sum-squared of all SH coefficients, divided by sqrt(4pi) as such 
```
mrcalc sh.mif sh.mif -mult - | mrmath - sum - -axis 3 | mrcalc - 12.5664 -divide -sqrt rms2.mif
```
then I do get the correct result:
![rms2.mif](https://cloud.githubusercontent.com/assets/7466983/14423494/666463f4-ffdc-11e5-9fcf-4f0ae740001b.png)

The difference is of course the division with (2*l+1) in sh2power on [line 63](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/sh2power.cpp#L63). Therefore, I think this division should be deleted, unless I'm missing something?

What I think happened here is that this originates from the legacy SH basis implementation, which used different normalization. The new orthonormal basis is then supposedly normalized per SH order l, rather than for each individual (l,m) basis function.

I take at least part of the blame on this one, since I'm the one who ported `sh2power` to the new syntax and wrote the test for it. At the time, I only tested whether its output was consistent between the old and new syntax...

jdtournier:
Well, it's both right and wrong, depends on how you view it... The correction of the SH basis is done for the m!=0 terms, and you can see it being applied a few lines earlier. The idea behind the division by (2*l+1) is to produce the power _per coefficient_, rather than as a total. I can't remember what the rationale was for that decision... But it would mean that the sum of the powers wouldn't match the expected value for the total power. 

[...]

PR #532, opened 2016-04-11
Compute total SH power by default

dchristiaens:
As discussed in #530.

### DIFF
--- a/cmd/sh2power.cpp
+++ b/cmd/sh2power.cpp
@@ -30,6 +30,10 @@ void usage () {
   ARGUMENTS
     + Argument ("SH", "the input spherical harmonics coefficients image.").type_image_in ()
     + Argument ("power", "the output power image.").type_image_out ();
+  
+  OPTIONS
+    + Option ("full", "output the power of each SH order in a 4-D image.");
+  
 }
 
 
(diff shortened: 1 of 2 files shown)


jdtournier:
Not sure I like the option name `-full` - doesn't really convey what it means, I'd almost expect this to mean the overall power, which is now the default. How about `-per_order`, `-per_band`, `-per_l` or something like that...?

thijsdhollander:
As I suggested in https://github.com/MRtrix3/mrtrix3/issues/530#issuecomment-208652023 , I'd keep the default behaviour as before, and add a `-total` option.  That'll also guarantee that people relying on it, are not affected.

...and then mostly clearly document the definitions being used of course.  I reckon that's what caused the most confusion in the first place.

thijsdhollander:
There could even be an option to output the powers without the `(2*l+1)` normalisation as well (for completeness sake).

jdtournier:
> There could even be an option to output the powers without the (2*l+1) normalisation as well (for completeness sake).

So this is where I'm not sure. Can you think of any reason to normalise in the first place...?

thijsdhollander:
Doesn't need to be there for me (I typically only eyeball the power spectrum's contrasts, so scaling is not relevant for me).  I did notice that particular scaling as well in Flavio's anisotropic power maps ISMRM abstract. (there he applies the scaling per band, sums all bands except the 0 one)

dchristiaens:
OK, renamed the option and provided some more docs. One last thing I'm wondering is if we should divide the output with 4*pi. Then it should exactly equal the mean squared amplitude (the docs now say "linearly proportional").

jdtournier:
> One last thing I'm wondering is if we should divide the output with 4*pi.

Fine by me. Again, it just needs to be documented...

jdtournier:
We good to merge...?

jdtournier:
Just looking at the test: is there any way we can have a more exact test here? A tolerance of 200 seems rather extreme... Maybe you can add an additional test that checks for regressions specifically against  the current output of the command (assuming you've checked it behaves as expected)?

dchristiaens:
Just added the rescaling with 4*pi to have an exact implementation of Parseval's theorem. Now the command is good to go as far as I'm concerned, if you agree.

Regarding the tests, I thought it would be nice to test correspondence to the mean-squared-amplitude directly. But the test image dwi.mif contains 60 directions, so the SH fit is obviously smoothing a few things out. This yields a max-abs-error of about 150, hence I put the tolerance to 200. Relative to amplitudes up to 500^2 that's a relative error below 0.001. So I think it's OK, but I also agree it looks bad on first sight... A better way may be doing it the other way around: start from an SH image and sample this to evaluate the mean-squared amplitude, akin to my first post on this thread.

jdtournier:
OK, so that tolerance of 200 isn't actually as bad as it sounds. I'll merge.

PR #533, opened 2016-04-12
Args no default values

Lestropie:
Remove the default parameters for command-line arguments, since they were not being used in any constructive way. This includes default integer / floating-point values, default selection for choice-based arguments, and default text for string arguments.


jdtournier:
You gotta love the testing framework... :wink:

Lestropie:
:+1:  Yep, came in handy there. There's still plenty missing from it though...

PR #540, opened 2016-04-14
File::GZ: fix getline() when line is exactly 63 bytes

jdtournier:
This fixes a bug that caused errors when loading an compressed `mif.gz` image when one of the lines in the header was exactly 63 bytes long...


bjeurissen:
Works! Thanks!

PR #542, opened 2016-04-14
MRView: fix segfault when loading ODF image in ODF tool

jdtournier:
Fix a few issues with #528

### DIFF
--- a/src/gui/dwi/renderer.cpp
+++ b/src/gui/dwi/renderer.cpp
@@ -158,18 +158,18 @@ namespace MR
         }
 
         source +=
-          "out float amplitude;\n"
+          "out float amplitude"+VSout+";\n"
           "void main () {\n";
 
         if (mode_ == mode_t::SH) {
           source +=
-          "  amplitude = r_del_daz[0];\n";
+          "  amplitude"+VSout+" = r_del_daz[0];\n";
         } else if (mode_ == mode_t::TENSOR) {
           source +=
-          "  amplitude = 1.0 / dot (vertex, inv_tensor * vertex);\n";
+          "  amplitude"+VSout+" = 1.0 / dot (vertex, inv_tensor * vertex);\n";
         } else if (mode_ == mode_t::DIXEL) {
           source +=
-          "  amplitude = value;\n";
+          "  amplitude"+VSout+" = value;\n";
         }
 
         if (use_lighting_ && (mode_ == mode_t::SH || mode_ == mode_t::TENSOR)) {
@@ -212,7 +212,7 @@ namespace MR
 
         if (mode_ == mode_t::SH || mode_ == mode_t::TENSOR) {
           source +=
-          "  vec3 pos = vertex * amplitude * scale;\n"
+          "  vec3 pos = vertex * amplitude"+VSout+" * scale;\n"
           "  if (reverse != 0)\n"
           "    pos = -pos;\n";
           if (orthographic_) {
@@ -227,7 +227,7 @@ namespace MR
         } else if (mode_ == mode_t::DIXEL) {
           source +=
           "  vert_dir = vertex;\n"
-          "  vert_pos = vertex * amplitude;\n"
+          "  vert_pos = vertex * amplitude"+VSout+";\n"
           "  if (reverse != 0) {\n"
           "     vert_dir = -vert_dir;\n"
           "     vert_pos = -vert_pos;\n"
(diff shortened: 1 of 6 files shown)


Lestropie:
... How that didn't cause a fault on my work machine, I will never know. Must not have run it in debug mode though :-/

jdtournier:
:grin: 

And there's more coming...

Lestropie:
Trying to get stuff merged so we can tag after the FSL fix, but stuff keeps ending up sitting there stale...  ><  Need some kind of a buddy system for pull requests.

I'm slightly handicapped here, I can't run `gdb` on my Windows machine (antivirus segfaults before it gets anywhere, even when the antivirus is turned off...). I'll keep playing until you're confident you're on top of it.

Lestropie:
@jdtournier There's one. If someone could please explain to me how that hasn't caused an issue previously, that'd be delightful.

jdtournier:
Probably because it's all in the shader. If it doesn't cause an error when compiling the shader, the fact that the variable names don't match between the CPU and GPU code isn't an error - it just gets ignored by the GL when you try to set these variables, and the shader will have undefined variables...

jdtournier:
Here's another one with a `-assert` config:

    mrview: src/gui/dwi/renderer.cpp:614: virtual void MR::GUI::DWI::Renderer::Dixel::set_data(const vector_t&, int) const: Assertion `data.size() == vertex_count' failed.

happens when displaying an dixel overlay, and hitting the ODF preview button...

jdtournier:
These are the relevant values at that point in the code:
~~~
mrview [src/gui/dwi/renderer.cpp: 612]: data.size() = 167
mrview [src/gui/dwi/renderer.cpp: 613]: vertex_count = 145
~~~

Looks like the data size is set to the `DWI.ndim(3)`, the other to the actual directions count in the relevant shell. 

Might also explain #520...

Lestropie:
> `mrview: src/gui/dwi/renderer.cpp:614: virtual void MR::GUI::DWI::Renderer::Dixel::set_data(const vector_t&, int) const: Assertion 'data.size() == vertex_count' failed`

That one should be fixed in #520. The remaining issue of the wonky axes is still there, that's a deeper GL error; but I caught that one.

I'm in the process of merging #520 into here, because I'm bug hunting and I keep re-finding bugs that are fixed in #520.

jdtournier:
fair enough...

Lestropie:
There we go; see if any additional problems you came across are still there.

jdtournier:
A few more assert failures:

- when opening the ODF preview tool: 

  ~~~
  QObject::connect: No such slot MR::GUI::MRView::Tool::ODF_Preview::lighting_update_slot() in src/gui/mrview/tool/odf/odf.cpp:536
  ~~~

- when switching between SH and dixel ODF images on the GUI a bit too quick with the ODF preview open:

  ~~~
  mrview: src/gui/dwi/render_frame.h:87: void MR::GUI::DWI::RenderFrame::set_lmax(int): Assertion `mode == mode_t::SH' failed.
  ~~~

Lestropie:
> when opening the ODF preview tool: 

I removed that function because it looked like it shouldn't be required given the new way of re-drawing the preview. But I forgot to remove the `connect()` call, and my system didn't give an error even in debug mode. But doing it that way, if you hide the ODF overlay, changing lighting settings won't update in the preview window. So put it back.

Lestropie:
> when switching between SH and dixel ODF images on the GUI a bit too quick with the ODF preview open:

Can't reproduce no matter how hard I mash my keyboard. And can't see a logical way to have pulled that off. Hopefully no users will be madly switching between ODFs trying to break my hurried patching until I get to work tomorrow.

jdtournier:
> Can't reproduce no matter how hard I mash my keyboard.

That's because it's a mouse interaction... :wink:

Try running a full debug build with asserts (i.e. `./configure debug -debug -assert && ./build debug/bin/mrview`), that slows the display down quite a bit... I think what might be happening is that the different slots get called in some order that doesn't agree with the ODF preview, probably because `update_preview()`is called within the main ODF tool's `draw()` function, which itself is invoked as needed asynchronously (via `updateGL()`, which schedules a redraw, but doesn't actually do it there and then). So because of the delayed update, there may be scope for some other signal to be processed in between, introducing a race condition between the actual render and any other GUI interaction that might have happened in the meantime...

Lestropie:
> That's because it's a mouse interaction...

Was mashing shift + up/down to switch ODF images as quickly as possible. Tend to agree it'll be a race condition, but hard to figure out at 3AM with no `gdb`.

Lestropie:
> when switching between SH and dixel ODF images on the GUI a bit too quick with the ODF preview open:

Still unable to reproduce, regardless of mouse vs. keyboard / speed of switching / speed of rendering.

PR #549, opened 2016-04-15
ODF tool: fix assert failure when loading SH ODF with non-SH ODF in current display

jdtournier:


jdtournier:
Carrying on from #542... I hadn't described to issue correctly, the problem was an assert failure when opening an SH image (with a different lmax than currently set) when a non-SH ODF was currently being displayed with the preview open...

PR #550, opened 2016-04-15
updates to user docs

jdtournier:
Note that I've also created a branch called `doctest` that now gets built automatically on [readthedocs](http://mrtrix.readthedocs.org/en/doctest/) - this makes it easy to verify updates to the documentation without having to install readthedocs locally (also avoids unintentional commits of lots of `*.rste` files... :wink:). Feel free to use that branch if you need to push documentation updates.


Lestropie:
Good idea :+1: Any idea of the delay between pushing to the repo and the changes being visible?

jdtournier:
Seems to take a couple of minutes or so...

jdtournier:
The rebuild is triggered immediately via a webhook, so it's basically only the time taken for readthedocs to pull the changes and run the build....

PR #572, opened 2016-04-22
Fix qform output when qfac is -1

jdtournier:
As discussed in [this community topic](http://community.mrtrix.org/t/dwipreproc-error-inconsistent-left-right-order-stored-in-sform-and-qform/179).


jdtournier:
Guys, feel free to merge as soon as we have confirmation that this fixes the problem - I'm off to bed...

thijsdhollander:
I didn't seem to be affected by the issue mentioned in the community topic (i.e., in the context of using `dwipreproc`).  Using the fix in this branch, I'm still not affected; so at least it's not breaking anything from my point of view.  I'm using the NeuroDebian version of FSL 5.0.9.

The fix for the qform (and the explanation of the bug in the first place) seems to make sense though.

jdtournier:
I think I'll merge this branch, even though the [issue seems to remain for the user affected](http://community.mrtrix.org/t/dwipreproc-error-inconsistent-left-right-order-stored-in-sform-and-qform/179/25) - I can't figure out why there would be a problem, it looks more likely to be a bug in the compiler or Eigen or something... Everyone happy with this...?

thijsdhollander:
👍

Issue #81, opened 2014-07-18
mrview: AdjustButton significant figures


Lestropie:
By default, the value shown as text in the AdjustButton is just a float-to-string conversion, which results in 6 significant figures. But often when you first load a file and the buttons get their default values, 6 significant figures doesn't fit within the width of the button because the window / toolbar is too small; and because it's right-justified, you can't actually see the magnitude of the parameter, just a whole lot of decimal places.

It might be nice to determine an appropriate number of significant figures to display initially, and perhaps also reduce the number shown when the mouse is dragged (e.g. use the rate parameter to determine how much the value changes per pixel, and use that to set the significant figures to display on a drag event). The user can still set the value at any precision they like and it will be displayed as such.

Thoughts?

PR #573, opened 2016-04-26
str() function: Fix for unsupported types

Lestropie:
Awaiting testing on Ubuntu 16.04, which fails compilation on current master: see [here](http://community.mrtrix.org/t/compile-error-on-ubuntu-16-04/183).

### DIFF
--- a/cmd/fixel2tsf.cpp
+++ b/cmd/fixel2tsf.cpp
@@ -83,7 +83,7 @@ void run ()
   DWI::Tractography::ScalarWriter<float> tsf_writer (argument[2], properties);
 
   float angular_threshold = get_option_value ("angle", DEFAULT_ANGULAR_THRESHOLD);
-  const float angular_threshold_dp = cos (angular_threshold * (M_PI/180.0));
+  const float angular_threshold_dp = cos (angular_threshold * (Math::pi/180.0));
 
   const size_t num_tracks = properties["count"].empty() ? 0 : to<int> (properties["count"]);
 
--- a/cmd/fixelcfestats.cpp
+++ b/cmd/fixelcfestats.cpp
@@ -152,7 +152,7 @@ void run() {
   value_type cfe_c = get_option_value ("cfe_c", DEFAULT_CFE_C);
   int num_perms = get_option_value ("nperms", DEFAULT_PERMUTATIONS);
   value_type angular_threshold = get_option_value ("angle", DEFAULT_ANGLE_THRESHOLD);
-  const float angular_threshold_dp = cos (angular_threshold * (M_PI/180.0));
+  const float angular_threshold_dp = cos (angular_threshold * (Math::pi/180.0));
 
   value_type connectivity_threshold = get_option_value ("connectivity", DEFAULT_CONNECTIVITY_THRESHOLD);
   value_type smooth_std_dev = get_option_value ("smooth", DEFAULT_SMOOTHING_STD) / 2.3548;
@@ -259,7 +259,7 @@ void run() {
   value_type gaussian_const1 = 1.0;
   if (smooth_std_dev > 0.0) {
     do_smoothing = true;
-    gaussian_const1 = 1.0 / (smooth_std_dev *  std::sqrt (2.0 * M_PI));
+    gaussian_const1 = 1.0 / (smooth_std_dev *  std::sqrt (2.0 * Math::pi));
   }
   {
     ProgressBar progress ("normalising and thresholding fixel-fixel connectivity matrix", num_fixels);
--- a/lib/mrtrix.h
+++ b/lib/mrtrix.h
@@ -72,7 +72,7 @@ namespace MR
     };
 
   template <typename X>
-    struct max_digits<X, typename std::enable_if<std::is_fundamental<typename X::value_type>::value, int>::type> {
+    struct max_digits<X, typename std::enable_if<std::is_fundamental<typename X::value_type>::value && !std::is_fundamental<typename X::Scalar>::value, int>::type> {
       static constexpr int value () { return std::numeric_limits<typename X::value_type>::max_digits10; }
     };
 
--- a/lib/mrtrix.h
+++ b/lib/mrtrix.h
@@ -58,6 +58,11 @@ namespace MR
 
   template <typename X, typename ReturnType = int>
     struct max_digits {
+      static constexpr int value () { return 0; }
+    };
+
+  template <typename X>
+    struct max_digits<X, typename std::enable_if<std::is_fundamental<X>::value, int>::type> {
       static constexpr int value () { return std::numeric_limits<X>::max_digits10; }
     };
 
@@ -70,6 +75,8 @@ namespace MR
     struct max_digits<X, typename std::enable_if<std::is_fundamental<typename X::value_type>::value, int>::type> {
       static constexpr int value () { return std::numeric_limits<typename X::value_type>::max_digits10; }
     };
+
+
 
   template <class T> inline std::string str (const T& value, int precision = 0)
   {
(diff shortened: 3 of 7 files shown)


thijsdhollander:
Slightly related, but maybe not worth a full blown issue: with all the Eigen shenanigans lately, should the installation instructions mention, or stress more clearly, that it's in peoples' best interests to use the latest non-beta version of Eigen?  It sounds trivial, but it seems some (linux) distributions may come with a pre-installed older version, and if people aren't confronted with an error at the `./configure` stage (because Eigen is in fact present), they may run into issues later on when compiling.  And that's even still a reasonably good scenario (it may be worse if things compile but there's significant bugs in an older or beta version of Eigen).

jdtournier:
I'm not convinced the issue is that `Eigen::Transpose` lacks the `Scalar` typedef, as you mentioned in the community post - that shouldn't cause any ambiguity, the compiler should default to the standard implementation (return 0). I have a suspicion the problem might be that they've added a `value_type ` typedef to that class... I'd need to check, but if that's the case, it's going to be difficult to fix. In this case, we might need to check whether this is a change the Eigen devs intended to make long-term, or just a temporary glitch...

In the meantime, the changes here look sensible, although they might change behaviour a bit, forcing full precision in places where they weren't before? I wonder whether this might affect things a little... Stuff like the output of `mrinfo`, etc. - I remember there was an issue where subtle rounding errors caused the voxel size to show up as 2.4999999999 instead of 2.5 or something...

thijsdhollander:
> I remember there was an issue where subtle rounding errors caused the voxel size to show up as 2.4999999999 instead of 2.5 or something.

About that: I still keep running into these kinds of things all the time.  The most notable place where this happens is actually in the `mrview` GUI, for many different numerical text boxes (i.e., things as simple as the window range).  Entering a number like "0.1", and then seeing it immediately changed to "0.0999999997" or "0.1000000003" or something upon hitting Enter or clicking in another field.  Same thing with certain parameters being either visualised in `mrinfo`, or actually being saved in headers (i.e., tracking parameters, etc...).

jdtournier:
@thijsdhollander: I think the MRView issue is due to the way these changes are actually applied. When you change some of these parameters, this triggers a redraw, where the parameters are re-calculated based on everything else, and the UI is update to match. This is to allow changing these parameters using the mouse interactions or keyboard shortcuts as well, which are handled completely separately from the view tool (by necessity, the view too won't necessarily even exist by then).

For `mrinfo`, `tckinfo`, and the like, the issue is likely to be either double->float conversions happening somewhere, and/or other operations happening somewhere along the line - in the case of NIfTI for example, the transform is either scaled by the voxel size (if the sform) or stored as a quaternion (if the qfrom), as floats in both cases; so it requires some operations to compute the MRtrix3 internal representation...

Lestropie:
> that shouldn't cause any ambiguity, the compiler should default to the standard implementation (return 0)

I'd have thought that on `master`, if neither `Scalar` nor `value_type` are defined, it would run the `std::numeric_limits<X>::max_digits10` version? I had assumed that would cause a compiler error. The change in 4a4d53a should force precision to zero for non-arithmetic types.

>  I have a suspicion the problem might be that they've added a value_type typedef to that class...

Having both `Scalar` and `value_type` defined would make sense in terms of the compiler error. But I don't see the latter anywhere in their beta code. I'll install the beta code and see if I can reproduce.

> The most notable place where this happens is actually in the mrview GUI, for many different numerical text boxes (i.e., things as simple as the window range).

#81 is related. For instance, if the user enters a value manually, the precision displayed should be based on the user entry; whereas if it's a sliding value, the precision should be based on the rate of change of the value per pixel.

Lestropie:
Eigen 3.2.8:
```
namespace internal {
template<typename MatrixType>
struct traits<Transpose<MatrixType> > : traits<MatrixType>
{
  typedef typename MatrixType::Scalar Scalar;
  typedef typename nested<MatrixType>::type MatrixTypeNested;
  typedef typename remove_reference<MatrixTypeNested>::type MatrixTypeNestedPlain;
  typedef typename traits<MatrixType>::StorageKind StorageKind;
  typedef typename traits<MatrixType>::XprKind XprKind;
  enum {
    RowsAtCompileTime = MatrixType::ColsAtCompileTime,
    ColsAtCompileTime = MatrixType::RowsAtCompileTime,
    MaxRowsAtCompileTime = MatrixType::MaxColsAtCompileTime,
    MaxColsAtCompileTime = MatrixType::MaxRowsAtCompileTime,
    FlagsLvalueBit = is_lvalue<MatrixType>::value ? LvalueBit : 0,
    Flags0 = MatrixTypeNestedPlain::Flags & ~(LvalueBit | NestByRefBit),
    Flags1 = Flags0 | FlagsLvalueBit,
    Flags = Flags1 ^ RowMajorBit,
    CoeffReadCost = MatrixTypeNestedPlain::CoeffReadCost,
    InnerStrideAtCompileTime = inner_stride_at_compile_time<MatrixType>::ret,
    OuterStrideAtCompileTime = outer_stride_at_compile_time<MatrixType>::ret
  };
};
}
```

Eigen 3.3 beta 1:
```
namespace internal {
template<typename MatrixType>
struct traits<Transpose<MatrixType> > : public traits<MatrixType>
{
  typedef typename ref_selector<MatrixType>::type MatrixTypeNested;
  typedef typename remove_reference<MatrixTypeNested>::type MatrixTypeNestedPlain;
  enum {
    RowsAtCompileTime = MatrixType::ColsAtCompileTime,
    ColsAtCompileTime = MatrixType::RowsAtCompileTime,
    MaxRowsAtCompileTime = MatrixType::MaxColsAtCompileTime,
    MaxColsAtCompileTime = MatrixType::MaxRowsAtCompileTime,
    FlagsLvalueBit = is_lvalue<MatrixType>::value ? LvalueBit : 0,
    Flags0 = traits<MatrixTypeNestedPlain>::Flags & ~(LvalueBit | NestByRefBit),
    Flags1 = Flags0 | FlagsLvalueBit,
    Flags = Flags1 ^ RowMajorBit,
    InnerStrideAtCompileTime = inner_stride_at_compile_time<MatrixType>::ret,
    OuterStrideAtCompileTime = outer_stride_at_compile_time<MatrixType>::ret
  };
};
}
```

\o/

jdtournier:
> I'd have thought that on master, if neither Scalar nor value_type are defined, it would run the std::numeric_limits<X>::max_digits10 version? I had assumed that would cause a compiler error. 

It does run `std::numeric_limits<X>::max_digits10()`, but that shouldn't cause a compiler error, since [the non-specialised version of that function returns zero](http://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10). So thinking about it, it seems the change in this pull request shouldn't make any difference...(?)

> Having both Scalar and value_type defined would make sense in terms of the compiler error. But I don't see the latter anywhere in their beta code.

Maybe from an inherited class...?

Lestropie:
> It does run std::numeric_limits<X>::max_digits10(), but that shouldn't cause a compiler error, since the non-specialised version of that function returns zero. So thinking about it, it seems the change in this pull request shouldn't make any difference...(?)

Suspect so. It's more explicit, but nothing a code comment couldn't have solved.

> Maybe from an inherited class...?

It's in there somewhere. This compiles:
```
auto temp = response.transpose();
typename decltype(temp)::value_type temp2;
```
See 7f055ce for proposed fix.

jdtournier:
Yes, I was thinking something like this latest commit might be the simplest fix... Happy with that.

Lestropie:
Merged; @thijsdhollander Feel free to open another issue RE: setting precisions if you want. But we'd need to decide whether to reduce the default precision for `default_type`, or go through and explicitly provide a precision in those cases where 15 digits is not desirable.

(From memory running `str()` with a precision of 6 will change 2.4999999999 to 2.5 rather than 2.50000)

Lestropie:
Just noting this for later: At least on my Windows box, `str()` seems to be setting 6 decimal places rather than 6 significant figures for floats, which is then highlighting rounding errors compared to older code.

PR #578, opened 2016-04-27
Fix Eigen-related alignment issues 

jdtournier:
As originally reported here:
http://community.mrtrix.org/t/mrview-segmentationfault-core-dumb-ubuntu-16-04/185

The issue is that Eigen expect fixed-sized types to be aligned for vectorised instructions - or rather, SSE/SSE2 expects data to be aligned for these operations. While Eigen does what it can, the standard `operator new()` doesn't provide any alignment guarantees, being type-agnostic. So Eigen provides a macro that is used to add an appropriate `operator new()` member to classes that contain fixed-size Eigen classes such as `Vector3f`, etc. Failing to do this means that dynamic allocation of classes containing fixed-sized Eigen types may fail to align the data properly, and lead to crashes when the classes are subsequently used in vectorised operations. This may not manifest on all hardware, since it depends on the alignment that the default `operator new()` call will use (no problem if it's actually aligning on 16-byte boundaries anyway), and whether the CPU supports SSE4 instructions, which don't require alignment anyway... 

So this adds the requisite `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` to all classes that I could see that contain a `transform_type` (which is what caused the crash in the original post). I expect there'll be lots more cases cropping up as time goes on, since this will probably affect _all_ `Vector3f`, `Vector4d`, `Matrix3d`, etc. types, which are used extensively in MRtrix3. Thankfully, this is only a problem with _dynamic_ allocation of a class that have one of these as a member...

### DIFF
--- a/lib/adapter/extract.h
+++ b/lib/adapter/extract.h
@@ -31,6 +31,8 @@ namespace MR
         using Base<ImageType>::parent;
         typedef typename ImageType::value_type value_type;
 
+        EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
         Extract1D (const ImageType& original, const size_t axis, const std::vector<int>& indices) :
           Base<ImageType> (original),
           extract_axis (axis),
--- a/lib/adapter/reslice.h
+++ b/lib/adapter/reslice.h
@@ -76,6 +76,8 @@ namespace MR
       public:
         typedef typename ImageType::value_type value_type;
 
+        EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
         template <class HeaderType>
           Reslice (const ImageType& original,
                    const HeaderType& reference,
--- a/lib/adapter/subset.h
+++ b/lib/adapter/subset.h
@@ -28,6 +28,7 @@ namespace MR
     {
       public:
         typedef typename ImageType::value_type value_type;
+        EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
         using Base<ImageType>::name;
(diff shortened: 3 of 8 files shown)


Lestropie:
Might be worth seeing if the fault can be reproduced using e.g. `-mno-sse4` option for `gcc`. If it can, we should `./run_tests`.

jdtournier:
> Might be worth seeing if the fault can be reproduced using e.g. -mno-sse4 option for gcc. 

Good call. Trying that now...

jdtournier:
Nope, can't reproduce with `-mno-sse4`...

jdtournier:
Can't even reproduce with an `ARCH=x86-64` build, which should only enable SSE2. I have a feeling things will also depend on whether the compiler will align dynamic allocations on 8 or 16 byte boundaries...

Lestropie:
Not having any luck either. But we could ask Ralf to run the test suite in debug mode?

jdtournier:
Potentially. Let's see whether these changes actually make any difference first. I'd also be interested to know what CPU he's using - this has never come up as an issue before, and I run MRView on my old 2008 Core2 Duo laptop - pretty sure that doesn't have SSE4 either... But maybe he's running in a 32-bit environment. I'll inquire...

jdtournier:
OK, issues seems related to using the beta version of Eigen 3.3. I'm going to can this pull request. We'll need to figure out what the problem if/when the issue still occurs when Eigen 3.3 hits release...

PR #606, opened 2016-05-16
Documentation modifications

### LINKED ISSUE #539: Docs: Re-arrange
I think some elements of the wiki layout were a bit unusual, and they really show now with the move to readthedocs. They could do with a bit of moving around.

Feel free to comment / concatenate.

* Separate "FAQ" and "Troubleshooting". "Troubleshooting" could be in either "Install" or "Getting Started" really; FAQs could potentially move to "Tutorials" since I think most entries are going to be "how do I do this?" type questions.

* Move the command and script listings out of "Getting Started". "Reference" would seem appropriate in a non-academic context... Would probably also need a smaller page somewhere describing the command naming convention, that then links to the full list.

* "Scripts for external libraries": Needs renaming. Not all depend on external libraries, and I wouldn't want anyone to shy away from development using the script backend because they think it's only for interfacing with other softwares.

* Standardize inclusion of references in user documentation.

* Kind of related to this, kind of not: `generate_user_docs.sh` queries the `bin/` directory, which may still have a compiled binary left over from another branch, which then tries to get its `.rst` included in commits even though the corresponding command isn't part of the branch you're working on. This needs to instead query the `cmd/` directory and then run the corresponding binaries to generate the docs.

* Should add to "Troubleshooting" a comment about the new memory-mapping solution; titled e.g. "MRtrix hangs when writing an output file to a network drive" so if someone encounters a network file system type that's not currently flagged, they might find this comment.

Lestropie:
Opening this up for inspection / comment / modification.

Primary changes are:

- Moving command and script listing to 'References' rather than 'Getting Started'
- Splitting 'Troubleshooting' from 'FAQ'

, but there's been other changes along the way as well.

Closes #539.


thijsdhollander:
I just pulled this branch and tried to `./build` after a `./configure -dev` , but getting this error at the end of it:
```

Error: please specify a directory
```

jdtournier:
That'll probably have something to do with changes I made a while back to this process. I'll have a quick look and report back. In the meantime, the simplest thing when making changes to the user doc is to push to the `doctest` branch - I've set this up to build immediately on readthedocs. This is exactly what Rob did, so you can have a look at the formatted docs directly at this address:

jdtournier:
Ok, the changes I'd made were to add the special targets `doc` and `bash` to the `build` script to allow building the docs and/or bash completion file as needed without having to re-configure, and hence re-build the whole thing. The `-dev` option is still in `configure` and should work. I also can't find where in the `generate_user_docs.sh` that error might be raised... but my first guess is that the script relies on the existence of the `docs/getting_started/commands` folder. That or it tried to build the user docs before building any executables...?

Lestropie:
> my first guess is that the script relies on the existence of the docs/getting_started/commands folder.

Given I've moved the location of that directory (and hence it won't exist), and it's the first time that error's been seen, I reckon that'd be a pretty good guess.

jdtournier:
Why did you remove that folder, BTW? I thought that was supposed to remain consistent with the commands themselves, and only updated as needed to match changes to the executables...?

Lestropie:
> Why did you remove that folder, BTW? I thought that was supposed to remain consistent with the commands themselves, and only updated as needed to match changes to the executables...?

Not sure what you mean here. The files generated by `generate_user_docs.sh` now go into `docs/reference/` rather than `docs/getting_started/`, since that's where these contents now appear in the actual docs.

But thinking about it again, if this branch is checked out, the new locations of the generated user doc files should be checked out at the same time, since I've already run the script prior to pushing. So I'll just have to stop supposing and have a look.

jdtournier:
Ah OK, sorry.  Everything does seem to match, not sure what the problem is then...

jdtournier:
BTW, this might be a good opportunity to mention using `shview` to inspect the responses estimated by `dwi2response` in the basic DWI processing section - as [suggested in this post](http://community.mrtrix.org/t/csd-on-dti-data-with-30-directions/81/23?u=jdtournier)...

Lestropie:
@thijsdhollander: Can you check to see if that error was related to generating the documentation for `dirsplit` that was erased in f4c1b83? It just regenerated on my system.

`dwiintensitynorm` and `population_template` require NumPy, can somebody install this on Mac and add the instructions? Alternatively, it looks like `dwiintensitynorm` doesn't even use it, and `population_template` only uses it to import/export transform text files and do a dot product; do we really need it (yet)?

Lestropie:
> BTW, this might be a good opportunity to mention using shview to inspect the responses estimated by dwi2response in the basic DWI processing section - as suggested in this post...

Done.

Lestropie:
Added `dwi2tensor` encoding description as mentioned [here](http://community.mrtrix.org/t/order-of-coef-in-dwi2tensor/220/3).

And all those non-float fractions in `src/dwi/tensor.h` were making me nervous.

jdtournier:
> And all those non-float fractions in src/dwi/tensor.h were making me nervous.

Eeek...! Does this change the output...?

Lestropie:
> Eeek...! Does this change the output...?

Well currently it looks to me like adding the `-dkt` output option totally corrupts the fit; even the tensor coefficients (which come out as a separate image to the DKI coefficients) are just noise. Omitting `-dkt` works fine. And that's the case with and without the changes in 6273cb5.

jdtournier:
OK, at least the regular tensor fit isn't affected... @bjeurissen, any thoughts as to what might be going on...?

bjeurissen:
@Lestropie: Thanks for making the integer fractions in the bmatrix calculation less ambiguous in terms of type promotion. The resulting bmatrix is the same as before, but your fixes make things a lot more foolproof. Regarding the noise outcome when adding -dkt, I cannot reproduce this, are you testing this on single shell data by any chance?

thijsdhollander:
> Regarding the noise outcome when adding -dkt, I cannot reproduce this, are you testing this on single shell data by any chance?

I was thinking this as well.  @bjeurissen : would it make sense to error, or at least give a warning, if people request the `-dkt` when feeding only single-shell data to `dwi2tensor`?

Lestropie:
> Regarding the noise outcome when adding -dkt, I cannot reproduce this, are you testing this on single shell data by any chance?

Hmmm... yes, it's possible that I ran it on a pre-single-shell-extracted file rather than the full dataset. Reassuring to know the problem's me and not the code :-P

Perhaps slightly better than checking the shell count would be to check the b-matrix condition number, since technically DSI data could be used here. But I'll post that elsewhere.

thijsdhollander:
> Perhaps slightly better than checking the shell count would be to check the b-matrix condition number, since technically DSI data could be used here. But I'll post that elsewhere.

...which makes me wonder: what actually comes from the shell count when applied to a "typical" DSI dataset (let's say one on a 3D rectangular grid in q-space)?  Heaps of shells?  Or all just in one shell?  Undefined?  Or other apocalyptic effects? :-)

Lestropie:
>  what actually comes from the shell count when applied to a "typical" DSI dataset (let's say one on a 3D rectangular grid in q-space)?

Haven't actually had the data to test it on. Volumes will remain in separate shells if there's a gap in the b-value distribution that exceeds 100. So it depends on whether the radii of the cartesian distribution of vectors exceeds this. If it doesn't, it may well all come out in 1 shell (and we should probably try to detect that based on e.g. the standard deviation within the shell). If it does, then hopefully the number of 'shells' exceeds the square root of the total number of volumes, which throws an error.

Also: Does anyone know why the Travis logs are getting truncated? Tests here are failing and I can't see why, they run fine on my system.

jdtournier:
Yes, I was thinking we'd eventually need to do something to support non multi-shell data... 

As to the TravisCI logs: you can always download the raw logs. Looking at them now: build completes without error, testing apps build ok, but every single test fails...

jdtournier:
Ah sorry. Yes, it seems the raw logs are truncated too. Can't get to the `testing.log` output...

jdtournier:
Had a quick look. The only thing that can have such s widespread effect is your change in `lib/app.cpp`, forcing apps to declare their `AUTHOR`. You've probably forgotten to update the testing commands to match...

Lestropie:
Superseded by #653.

Issue #599, opened 2016-05-10
Registration code: Ambiguous `push_back()` calls


Lestropie:
```
cmd/mrmetric.cpp:310:15: error: call to member function 'push_back' is ambiguous
      headers.push_back (input1);
cmd/mrmetric.cpp:311:15: error: call to member function 'push_back' is ambiguous
      headers.push_back (input2);
./lib/registration/nonlinear.h:88:23: error: call to member function 'push_back' is ambiguous
              headers.push_back (im2_image);
./lib/registration/nonlinear.h:89:23: error: call to member function 'push_back' is ambiguous
              headers.push_back (im1_image);
./lib/registration/transform/search.h:234:23: error: call to member function 'push_back' is ambiguous
              headers.push_back (im1);
./lib/registration/transform/search.h:235:23: error: call to member function 'push_back' is ambiguous
              headers.push_back (im2);
```

Need to explicitly tell compiler whether these should call the template constructor or the move constructor.

jdtournier:
Closed via #612

PR #612, opened 2016-05-16
Fix push back headers in registration

jdtournier:
As discussed here: http://community.mrtrix.org/t/build-errors/210

There's also a couple of fixes to silence some of the warnings on clang.


jdtournier:
relates to #599

PR #617, opened 2016-05-17
dwi2response: fix handling of white space in filename

jdtournier:


Lestropie:
If this is an issue, it needs to be handled in a helluva lot more places than just this line...?

jdtournier:
More than likely... Do you want to look into it? :grin:

jdtournier:
@Lestropie: merge now, or do you want to double-check whether similar changes need to be made elsewhere...?

Lestropie:
Well, I was about to say again that we might as well do the same in all locations where the user provides a path. But then again, the whole reason you can't use paths with whitespace in MRtrix commands is that the command line parsing becomes ambiguous; surely the same applies here? It might work if the option / argument in question is the last one to appear on the command line, but I'm worried this is going to introduce more problems than it fixes...?

jdtournier:
> the whole reason you can't use paths with whitespace in MRtrix commands is that the command line parsing becomes ambiguous

Since when has that been the case? MRtrix has always happily supported whitespace in paths - it's scripts that needs special handling. You can easily supply a filename with spaces or other funny characters to MRtrix _executables_ using quotations marks or the escape character - this tells the shell that this is to be interpreted a single argument, which is then provided as a single entry in the `argv` array. The problem is when these arguments are then subsequently used in other commands, either directly (in a bash script) or via `os.system()` calls (in python), if care is not taken to ensure these arguments won't then be split by the shell (the system call implies the command will run within its own shell). For `bash` for instance, I try to make sure any substitution is written in double quotes (e.g. `"$1"`, `"$filename"`, etc. ), which normally handles this appropriately. For the build script, I avoid the `os.system()` call and use the [`subprocess.Popen()`](https://docs.python.org/2/library/subprocess.html) call, which takes a _list_ of arguments rather than a single string. This avoids any ambiguity, and would sidestep the need for the fix I've committed here...

Lestropie:
Obviously with quotations / escape characters they work. I suppose I don't see where in the chain the support for those mechanisms breaks down in the scripts. If they're substituted out even before `argparse` is invoked, then I'd expect even that initial parsing to break down. If the outputs from `argparse` remove any quotation marks / escape characters, or indeed even if they don't, then it might be the implementation of my `runCommand()` function that's causing the heartache (which does use `subprocess`, but performs the string splitting manually). I'd rather see how the problem manifests before I propagate a solution throughout the whole script directory.

jdtournier:
Scripts and executables will be provided with the arguments as parsed by the shell via the `argv` array/list (or the `$1`, `$2`, ... special variables in bash). It's perfectly legal for any of those entries to contain spaces, since at that point they're maintained as distinct entries in that list. The issue comes when you subsequently form a command to be invoked _as a single string_ - which incidentally is what both your `runCommand()` and the `os.system()` call expect (your [`runCommand()` invokes `os.system()`](https://github.com/MRtrix3/mrtrix3/blob/master/scripts/lib/runCommand.py#L49) by the way - no sign of `subprocess` anywhere as far as I can see?). As soon as these arguments are concatenated into a single string, spaces revert to having special meaning as argument delimiters. Furthermore, the [`os.system()` call](https://docs.python.org/2/library/os.html#os.system) will invoke a new shell, passing your command to it for parsing and execution. This is why you can add quotations marks to your command and have them interpreted correctly (in that the command invoked doesn't see the quotation marks, and an argument with spaces remains a single argument). 

The `subprocess.Popen()` call on the other hand expects a _list_ of arguments, the same format as provided via the `argv` array. There is no scope for ambiguity there at all - an argument with spaces in it remains a single argument nonetheless. The command is also executed _directly_, not via another shell - so it's a touch faster to launch, but the downside is you lose any interpretation that you might have wanted the shell to perform (rarely a problem though, generally you're trying to _avoid_ bash getting in the way).

Lestropie:
Ah, I must have been thinking of other functions in the library that are using `subprocess` (mostly to capture the standard output).

From memory I wanted the base `runCommand()` function to take a single string to make the actual scripts easier to write and cleaner to read. But given I'm already breaking up those strings by whitespace within `runCommand()` in order to perform other cleverness, maybe I should be vectorising the string at that point, looking for quotation marks / escape characters to preserve as I do so, and then getting rid of the `os.system()` call. I've tried to avoid using any `bash` anywhere, though there might be a wildcard or two somewhere that I could Pythonise.

Currently, if quotation marks are inserted into the scripts anywhere that a user-defined path is fed to `runCommand()`, it would *typically* work, but might not always depending on user input.

I'd rather not push an incomplete fix. I'll go digging.

jdtournier:
> Currently, if quotation marks are inserted into the scripts anywhere that a user-defined path is fed to runCommand(), it would typically work, but might not always depending on user input.

Not sure what you meant here, but if you're saying you're expecting arguments provided in quotation marks to remain within quotation marks by the time the script received them, that's not how it works. The shell will do a bagload of interpretation and pass the arguments _as it figures they should be_ to the command/script. To allow these quotes to pass through, you'd have to escape the quotation marks - and then your script would probably fail to find the files, etc. if it ever tried to open them... Not sure that's what you meant, though...?

jdtournier:
Sorry, ignore my previous post - read your post again and figured out what you were taking about. I'm just being dense..

jdtournier:
Sorry, ignore my previous post - read your post again and figured out what you were taking about. I'm just being dense..

jdtournier:
Just to add my 2 cents to what you suggested: I think all the bash parsing and interpretation should be done upfront by bash itself when the command/script is invoked. Any wildcard substitutions should happen at that point. The script itself shouldn't attempt to second-guess what the user has in mind, I think it would just lead to confusion if you have two layers of interpretations. I'm not sure whether you had any particular set of circumstances in mind that would require another level of interpretation...?

Lestropie:
> Any wildcard substitutions should happen at that point.

What I meant there is that I've tried to avoid using wildcards _within the scripts_ in order to select a set of files that have been generated _by the script_. Nothing to do with user input. But if I were to change `runCommand()` to take a vector of arguments, wildcards wouldn't work.

Lestropie:
Am working on a more general solution to the whitespace issue in the `script_whitespace_handling` branch. Also moved from `os.system()` to `subprocess`. Will discuss further in a different pull request.

PR #618, opened 2016-05-17
dwi2response manual: Fix manual lmax setting

### LINKED ISSUE #616: dwi2response: fix handling of -lmax option
As reported in [this community post](http://community.mrtrix.org/t/msdwi2response-returns-only-nan/209/3), using `-lmax` with a vector of values per shell doesn't work so well. It seems only the first value is used or something?

Lestropie:
Closes #616.

### DIFF
--- a/scripts/src/dwi2response/manual.py
+++ b/scripts/src/dwi2response/manual.py
@@ -39,12 +39,20 @@ def execute():
   from lib.getHeaderInfo import getHeaderInfo
   from lib.runCommand    import runCommand
   
-  lmax_option = ''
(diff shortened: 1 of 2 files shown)


Lestropie:
On a side note: `amp2sh` gives warnings about A2SH condition number when you feed it a massive number of _b=0_ images and it tries to fit _lmax_=4 even though all of the vectors are `[0 0 1]`. `DWI::compute_SH2amp_mapping()` doesn't currently receive the b-value so can't branch there, but ideally the code should probably bypass this point entirely if the _b_-value is below `bzero_threshold`.

PR #622, opened 2016-05-19
fix compilation issues with Intel compiler

boegel:
I'm trying to compile MRtrix 0.3.14 with the Intel compilers (I tried with both versions 2016.1.150 and 2016.3.210).

719e416 fixes this problem:

```
In file included from ./lib/adapter/subset.h(20),
                from src/gui/mrview/tool/connectome/connectome.cpp(21):
./lib/adapter/base.h(28): error: a template template parameter cannot have the same name as one of its template parameters
     template <template <class AdapterType> class AdapterType, class ImageType, typename... Args>
                                                        ^
```

I'm still running into another problem I haven't figured out though (I'm not exactly a C++ expert), suggestions welcome:

```
src/dwi/tractography/seeding/base.h(65): error: expression must have class type
          ThreadedLoop (data).run ([&] (decltype(data)& v) { if (v.value()) count.fetch_add (1, std::memory_order_relaxed); }, data);
                                                                   ^
```

### DIFF
--- a/cmd/label2mesh.cpp
+++ b/cmd/label2mesh.cpp
@@ -103,7 +103,7 @@ void run ()
         from.push_back (lower_corners[in][axis]);
         dimensions.push_back (upper_corners[in][axis] - lower_corners[in][axis] + 1);
       }
-      Adapter::Subset<decltype(labels)> subset (labels, from, dimensions);
+      Adapter::Subset<Image<uint32_t>> subset (labels, from, dimensions);
 
       auto scratch = Image<bool>::scratch (subset, "Node " + str(in) + " mask");
       for (auto i = Loop (subset) (subset, scratch); i; ++i)
--- a/lib/adapter/base.h
+++ b/lib/adapter/base.h
@@ -25,7 +25,7 @@ namespace MR
   namespace Adapter
   {
 
-    template <template <class> class AdapterType, class ImageType, typename... Args>
+    template <template <class ImageType> class AdapterType, class ImageType, typename... Args>
       inline AdapterType<ImageType> make (const ImageType& parent, Args&&... args) {
         return AdapterType<ImageType> (parent, std::forward<Args> (args)...);
       }
--- a/lib/adapter/base.h
+++ b/lib/adapter/base.h
@@ -25,7 +25,7 @@ namespace MR
   namespace Adapter
   {
 
-    template <template <class AdapterType> class AdapterType, class ImageType, typename... Args>
+    template <template <class> class AdapterType, class ImageType, typename... Args>
       inline AdapterType<ImageType> make (const ImageType& parent, Args&&... args) {
         return AdapterType<ImageType> (parent, std::forward<Args> (args)...);
       }
--- a/src/dwi/tractography/seeding/base.h
+++ b/src/dwi/tractography/seeding/base.h
@@ -62,7 +62,7 @@ namespace MR
       uint32_t get_count (ImageType& data)
       {
         std::atomic<uint32_t> count (0);
-        ThreadedLoop (data).run ([&] (decltype(data)& v) { if (v.value()) count.fetch_add (1, std::memory_order_relaxed); }, data);
+        ThreadedLoop (data).run ([&] (ImageType& v) { if (v.value()) count.fetch_add (1, std::memory_order_relaxed); }, data);
         return count;
       }
 


jdtournier:
Wow, I've never tried to compile with anything other than g++ or clang++... I'm surprised these are the only issues that cropped up. Any particular reason you need to compile with the Intel compiler...?

In any case, could I ask you modify your first commit, so that the line modified reads:

    template <template <class ImageType> class AdapterType, class ImageType, typename... Args>

This is actually what it was meant to be, and compiles fine on my system. Hopefully the intel compiler will be happy with that too. I'm surprised neither g++ or clang++ complained about this, it was indeed not quite right...

I think the intel compiler is also correct about the second issue. Technically, `decltype(data)` would resolve to `ImageType&`, so `decltype(data)& v` would imply `v` is a reference to a reference, and hence not of class type. The other compilers seemed happy enough with this, but it's a simple fix in any case. Change the offending line to:

    ThreadedLoop (data).run ([&] (ImageType& v) { if (v.value()) count.fetch_add (1, std::memory_order_relaxed); }, data);

Both of the above changes compile fine on both g++ and clang++, and the tests complete without errors on my system. I'd commit this directly to your branch, but I have a feeling I'm going to have issues with permissions, etc. Better you make the changes on your end, test that they actually do work out, and we'll merge your updated pull request when everyone's happy with it.

boegel:
@jdtournier Thank you for the suggestions and detailed explanation!

I've applied the suggested fixes, and verified that they fix the compilation issues.

I'm now running into another issue later on during the compilation though (actually, two distinct issues, it seems). Any idea about how to fix these?

```
ERROR: (189/412) [CC] release/cmd/label2mesh.o

icpc -c -std=c++11 -pthread -fPIC -march=native -std=c++11 -O2 -xHost -ftz -fp-speculation=safe -fp-model source -DMRTRIX_WORD64 -Wall -O2 -DNDEBUG -Isrc -Icmd -I./lib -Icmd cmd/lab
el2mesh.cpp -o release/cmd/label2mesh.o

failed with output

In file included from ./lib/command.h(22),
                 from cmd/label2mesh.cpp(20):
./lib/app.h(224): warning #1098: the qualifier on this friend declaration is ignored
          friend void  MR::App::init (int argc, char** argv);
                       ^

In file included from ./lib/command.h(22),
                 from cmd/label2mesh.cpp(20):
./lib/app.h(225): warning #1098: the qualifier on this friend declaration is ignored
          friend void  MR::App::parse ();
                       ^

In file included from ./lib/command.h(22),
                 from cmd/label2mesh.cpp(20):
./lib/app.h(226): warning #1098: the qualifier on this friend declaration is ignored
          friend void  MR::App::sort_arguments (int argc, const char* const* argv);
                       ^

./lib/adapter/base.h(40): error: name followed by "::" must be a class or namespace name
            typedef typename ImageType::value_type value_type;
                             ^
          detected during:
            instantiation of class "MR::Adapter::Base<ImageType> [with ImageType=MR::Image<uint32_t={unsigned int}> &]" at line 27 of "./lib/adapter/subset.h"
            instantiation of class "MR::Adapter::Subset<ImageType> [with ImageType=MR::Image<uint32_t={unsigned int}> &]" at line 106 of "cmd/label2mesh.cpp"

./lib/adapter/subset.h(30): error: name followed by "::" must be a class or namespace name
          typedef typename ImageType::value_type value_type;
                           ^
          detected during instantiation of class "MR::Adapter::Subset<ImageType> [with ImageType=MR::Image<uint32_t={unsigned int}> &]" at line 106 of "cmd/label2mesh.cpp"

cmd/label2mesh.cpp(110): error: no operator "==" matches these operands
            operand types are: MR::Helper::Value<MR::Adapter::Base<MR::Image<uint32_t> &>> == const size_t
          scratch.value() = (subset.value() == in);
                                            ^

compilation aborted for cmd/label2mesh.cpp (code 2)
```

boegel:
@jdtournier As for your question on why we're using the Intel compilers, it's quite simple: we try to build and install all of the software for the users of our HPC infrastructure at Ghent University (http://www.ugent.be/hpc/en) using the Intel compilers, because they yield significantly better performance (how much depends on the software/use case).

We handle all of this via EasyBuild, cfr. http://hpcugent.github.io/easybuild

Building MRtrix 0.3.12 with the Intel compilers worked smoothly, so we want to get 0.3.14 working as well with them.

Lestropie:
Hi,

For the errors you can try:

`cmd/label2mesh.cpp`:
Line 106:
`Adapter::Subset<decltype(labels)> subset (labels, from, dimensions);`
to:
`Adapter::Subset<Image<uint32_t>> subset (labels, from, dimensions);`

If that change fixes compilation, that's the first time I've seen something like that. Looks like it's inferring a class reference rather than a class type for the template, because the code lies within a lambda function that's grabbing items by reference; it's then looking for a typedef within that template, and failing because it's not of class type.

Line 110:
`scratch.value() = (subset.value() == in);`
to:
`scratch.value() = (uint32_t(subset.value()) == uint32_t(in));`

Not the cleanest solution, let's just see if it works: my suspicion is that a large number of such implicit casts are going to cause errors with your compiler, and there's no choice but to go through and correct them one at a time.

jdtournier:
Actually, I think that second error is a red herring:

~~~

cmd/label2mesh.cpp(110): error: no operator "==" matches these operands
            operand types are: MR::Helper::Value<MR::Adapter::Base<MR::Image<uint32_t> &>> == const size_t
          scratch.value() = (subset.value() == in);
~~~

Note that the template parameter was inferred to be `MR::Adapter::Base<MR::Image<uint32_t> &` - a reference. That should be fixed by making sure the compiler infers the class itself, not a reference to it. Which might already be fixed with your first modification.

boegel:
@Lestropie Thanks for the suggestions! @jdtournier was right w.r.t. the 2nd error being a red herring, just the first change was sufficient (see 3125966)

With these changes, I'm able to get MRtrix built with the Intel 2016 update 1 compilers!

I've included the full output of the build in https://gist.github.com/boegel/a9233aec37bb240c4ad7f6eb4627ff92.

jdtournier:
:+1: 

I'll take a look at the warnings in the log when I have a minute...

boegel:
@jdtournier Do you think this is good to be merged in? It certainly works for me, cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/3084

If there are any future changes that you feel are required based on the warnings, those can be tackled later?

jdtournier:
Done!

boegel:
:+1:

PR #623, opened 2016-05-20
dwi2tensor: Check condition number of matrix

Lestropie:
This will hopefully provide the user with a warning if they attempt a tensor / kurtosis tensor estimation with inadequate variation in their b-values.

Arose initially from discussion in #606. However I'm not completely sold: It's the same approach as is used for determining _lmax_ based on the SH transform, but the numbers are completely different: I had a stable fit at condition number 2e6 and unstable at 1e9. So would appreciate people's thoughts on this, or consider reverting to a _b_value shell count method instead.

### DIFF
--- a/src/dwi/tensor.h
+++ b/src/dwi/tensor.h
@@ -16,6 +16,8 @@
 #ifndef __dwi_tensor_h__
 #define __dwi_tensor_h__
 
+#include <Eigen/SVD>
+
 #include "types.h"
 
 namespace MR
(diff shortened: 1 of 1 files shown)


jdtournier:
So thinking about this a bit, I think I'd rather revert to a shell count strategy, especially given how different the numbers are. With the SH fit, we are at least expected a condition number close to 1 in the ideal case. Not so here, clearly the problem is a bit different. Although I am puzzled as to what datasets you were using that would give you condition numbers of 2e6 and 1e9 - presumably these come from real DW encodings? And if so, I'm assuming one of them is multi-shell, the other single-shell? 

I think part of the problem with the condition number here is that one of the parameters is scaled entirely differently: the log(S0) term will typically be several orders of magnitude larger than the coefficients of the tensor. So that probably explains the really large condition number even for a well-conditioned fit...

Lestropie:
> So thinking about this a bit, I think I'd rather revert to a shell count strategy

I tend to agree. Though it would be preferable to catch a potential failure of shell clustering failure in the case of DSI data, and proceed with tensor / kurtosis fitting with a warning rather than aborting. I'll take it on.

> Although I am puzzled as to what datasets you were using that would give you condition numbers of 2e6 and 1e9

From memory I tested single-b-value / single-shell with b=0 / multi-shell, DTI / DKI, and those were the two extremes of what worked v.s. what didn't work.

Lestropie:
Closed in favour of #635.

PR #625, opened 2016-05-23
Dwi2response cleanup

Lestropie:
Following discussion [here](http://community.mrtrix.org/t/dwi2response-bus-error/236/7).

The `msmt_5tt` fix needs to be tested, I don't have data for it at home right now.


Lestropie:
Note: In the process of doing this I also found a slight glitch in `dwi2response tournier` [here](https://github.com/MRtrix3/mrtrix3/pull/625/commits/83624b9c3745eca2beaa9163518d51058ed08c57#diff-fe3ff567bfebd2f50280f2266d96a2d8R102). The intention was to select the top 3,000 voxels, dilate once, and use that as the mask for the next iteration. However what was happening was that the top 300 voxels that had been used for RF calculation were instead being dilated and used as the next mask. Whoops.

thijsdhollander:
Since merging this will change the behaviour for most/all people anyway (i.e., the change to the `tournier` algorithm's behaviour), this is probably the perfect opportunity to switch the ordering of the outputs of the MSMT algorithm to WM-GM-CSF (as opposed to GM-WM-CSF)...?

thijsdhollander:
(It's sort of implied in the above, but I'd definitely post an announcement or something on the forum after the merge, due to the changed behaviour.)

Lestropie:
> this is probably the perfect opportunity to switch the ordering of the outputs of the MSMT algorithm to WM-GM-CSF (as opposed to GM-WM-CSF)...?

I won't protest against this if it's preferable.

> (It's sort of implied in the above, but I'd definitely post an announcement or something on the forum after the merge, due to the changed behaviour.)

It'll be in the big changelist since the `updated_syntax` merge I want to post once I hit merge on all these pull requests. Do we think this needs a little extra attention?

thijsdhollander:
> Do we think this needs a little extra attention?

I think a good way to deal with this (now and in future cases), would be to specifically enlist and highlight changes that specifically impact the behaviour (and outcome) of methods or algorithms; especially if they come *without* a change in the user interface of the kind that makes them obvious or simply unavoidable to spot.  This is definitely such a case, and a change to WM-GM-CSF ordering of inputs in an of itself actually (potentially) also is.  I.e., changes of the kind that could easily go unnoticed, but will change outcomes.

It's all still tricky business though; I often wonder how much we can (safely) rely upon the fact that users will actually visit the website and read the blog and/or forum upon each `git pull` and `./build` combo that they can (all too?) easily use to update... but I must admit I wouldn't know how to highlight these kinds of changes *even more*.

thijsdhollander:
Just fixed a small bug left in after the previous commits...

Lestropie:
@thijsdhollander What was the justification for WM-GM-CSF? The current MSMT documentation uses CSF-GM-WM in the `msdwi2fod` call, and that's also the ordering necessary to reproduce the RGB tissue maps as they were presented in the manuscript. I don't particularly care what the ordering is, I had just assumed in my head that that's why you were changing it.

thijsdhollander:
WM-GM-CSF is the way people typically refer to the listing of tissue types.
**More importantly**: it's also the ordering used throughout the MSMT-CSD paper (mentioned as "WM/GM/CSF").

Don't know if there's a good "logic" to it, but I reckon putting the WM up front makes sense; it's still the main focus of what we do (and is the thing that was previously the only output of CSD, the other tissues got added "later on").  In one of the above commits, I've effectively changed that ordering in the `msdwi2fod` call example as well.
The CSF-GM-WM ordering is indeed the one that ended up matching the RGB "convention" (as far as that's already a convention).  The RGB maps are, in the end just a simple visualisation trick, I wouldn't base any standard ordering on them.  They may be a handy overview of the results, but they're also biassing the impression of the prevalence of tissue signal in mixtures quite a lot... the safer visualisation here would actually just be the individual tissue prevalence maps.

The fact that GM/CSF are "added" with respect to the single-tissue WM approach, also makes for more logical adjustments of commands where single-tissue CSD was done before: the new additions (GM/CSF) just get added at the end of the command.

Lestropie:
Superseded by #653.

PR #627, opened 2016-05-23
Moved Sn_scale_estimator from src/dwi/ to lib/math/

Lestropie:
@jdtournier This file placement struck me as odd, even though it's only used in DWI-related code. Merge if you're happy with it.


jdtournier:
I guess you're right, probably better there... I'll merge.

PR #632, opened 2016-05-25
Fix dicom import for Siemens mosaics for non-square LR phase-encode EPI 

jdtournier:
A pretty specific set of circumstances, but hopefully that will handle this case appropriately... 

Tested these changes for regression on a whole load of DICOM data that I have lying around for testing, everything checks out - this makes no difference for the data we do have.


bjeurissen:
Checks out on my data as well!

Thanks for fixing this!

jdtournier:
OK, merging. Hope it doesn't break anything...

Issue #630, opened 2016-05-25
More verbose message when configure script fails


Lestropie:
As suggested [here](http://community.mrtrix.org/t/mac-i-came-across-this-problem-in-the-process-of-install/240/3?u=rsmith), though I'm sure there's been user questions in the past that might have been non-issues. Need to assume a lower level of understanding of bash / terminal usage.

Lestropie:
Closed via #633, which was pushed via #653.

PR #633, opened 2016-05-26
Configure better qt debugging and clearer output

jdtournier:
This ensures that the full output of all steps in the Qt configure stage is properly logged, so we can get a clearer idea why things might be failing. Also changes a few of the error message with hopefully clearer feedback and instructions, as discussed in #630 .

### DIFF
--- a/build
+++ b/build
@@ -236,6 +236,14 @@ The special target 'doc' is used to update the command documentation, so that
 any modifications to the inline documentation in commands can be propagate
 through to the user documentation site.
 
+By default, 'build' will use all available cores to run a parallel build. In some
+instances, this can cause problems (notably out of RAM errors). You can control 
+how many jobs 'build' will run concurrently using the NUMBER_OF_PROCESSORS 
+environment variable. For example:
+
+  $ NUMBER_OF_PROCESSORS=1 ./build
+
+
 OPTIONS:
 
   -verbose               print each command as it is being invoked
--- a/configure
+++ b/configure
@@ -100,9 +100,14 @@ ENVIRONMENT VARIABLES:
 For non-standard setups, you may need to supply additional
 information using environment variables. For example, to set
 the compiler, use:
-   $ CXX=/usr/local/bin/g++-4.1 ./configure
+   $ CXX=/usr/local/bin/g++-5.5 ./configure
 
-The following environment variables can be set:
+Alternatively:
+   $ export CXX=/usr/local/bin/g++-5.5
+   $ ./configure
+
+Multiple environment variables can be set this way as needed. 
+The following environment variables are available:
 
 CXX             The compiler command to use. The default is "g++" ("clang" on MacOSX)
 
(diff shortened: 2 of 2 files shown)


Lestropie:
@jdtournier I was thinking even more verbose like 6f6d8e0. From my perspective the people posting configuration issues have been less technically savvy, and this may become even more so on release, so I was going for super-friendly, 'this is the issue, this is what you can try' type messages so that hopefully people are more inclined to try to sort it out rather than panic.

jdtournier:
Yes, I thought you might want to add to this... feel free to do so, but bear in mind that on release, we'll be providing pre-compiled downloads too - at least that's the plan... Also, while your changes are clearer, the reality is they're still going to be difficult for non-technical users. The problem is that the lack the fundamental concepts of a path, how the system looks for executables, what a compiler even is, etc. There's only so much you can achieve here - ultimately they'll have to get help anyway...

Lestropie:
Yeah I know. I just figure we might as well have it as non-cryptic as is reasonable, so that if someone does need to come looking for help, it's unambiguous that it's because they don't understand those fundamentals, not because our explanation wasn't good enough. I don't like being accused of being either too condescending or not beginner-friendly enough when someone comes looking for help with this stuff and their level of aptitude isn't clear.

jdtournier:
Agreed. Feel free to merge if you're happy with your additions. Although it might be worth double-checking that the first set of changes I'd made go work as intended... The rationale was originally to help debugging a troublesome Qt configure step (turned out to be a missing GL dev package), but it was impossible to figure out because for some reason the output of the Qt tests wasn't included in the logs.

jdtournier:
Actually, while we're at it: might be worth deciding on a single best recommendation in these cases. Sometimes the advice is to either set one set of flags (e.g. QMAKE) _or_ set the PATH to include its location. I reckon in cases like this, the best course of action is actually to set the PATH - the current PATH gets included in the config, and used during subsequent builds - idea being to ensure the config remains valid when subsequently invoked in a different session where the PATH might have been reverted to what it was. In general, that would be a better and often simpler thing to do, since it will generally also pull in the right `rcc` and `moc` executables too...

Anyway, just saying it's worth revising what advice is provided to make sure it's actually our best advice - and simplify things by only suggesting one course of action...

Lestropie:
@jdtournier On my Windows system, I was getting 'system could not find the path specified' without b43d4f1. Doubt it'd break any other system, but best test it just to be sure.

jdtournier:
Your change works fine here. I've taken out the 'cwd' argument just in case, it's basically as it was before. 

I've also have a good go at revising all the error messages and advice provided. See what you think...

Lestropie:
> I've also have a good go at revising all the error messages and advice provided. See what you think...

Nice. Error message looks right for the one I tested. Gotta love those commits that remove more lines than they add.

PS. The troubleshooting page on doctest branch also mentions the [running out of RAM error](http://mrtrix.readthedocs.io/en/doctest/getting_started/troubleshooting.html#compiler-error-during-build), so hopefully there'll be less of those reports coming in.

Lestropie:
Superseded by #653.

PR #636, opened 2016-05-30
Thread::Queue: Hold mutex lock for CV notifications

Lestropie:
This appears to prevent thread deadlocking on Windows, even though C++ documentation claims that it isn't required.

As reported at various times. Requires testing on Mac OSX.


Lestropie:
Superseded by #653.

PR #639, opened 2016-05-31
SIFT model: Disable heuristic GM scaling by default

### LINKED ISSUE #608: SIFT/SIFT2: Change default behaviour RE: FODs near GM
With single-shell single-tissue CSD, the FODs in voxels that are partly WM and partly GM tend to be of a comparable size as those in deep WM. However, only the WM fraction of the voxel may contain streamlines. As a consequence, these FODs would typically be under-represented by streamlines. As a heuristic correction, I multiply the integrals of all FOD lobes in such voxels by (1.0 - GM). However with methods such as MSMT, this is no longer appropriate. I think that using this heuristic should now require explicit activation at the command line, and be disabled by default.

Lestropie:
In the SIFT model, it is common for the FODs in voxels containing part WM and part GM to be under-represented by streamlines, as the streamlines are only permitted to exist within some fraction of the voxel volume.
A heuristic was provided to compensate for this by using the ACT 5TT image to down-scale the FOD size in such voxels.
However with the emergence of more tissue-specific spherical deconvolution approaches, this heuristic is no longer required in all circumstances.
Therefore, this heuristic scaling can now be activated via a command-line option, rather than the previous behaviour where it could be de-activated via an option.
Closes #608.

Don't merge until tag is happening, since default behaviour is changing.

### DIFF
--- a/docs/getting_started/commands/tcksift.rst
+++ b/docs/getting_started/commands/tcksift.rst
@@ -36,7 +36,7 @@ Options for setting the processing mask for the SIFT fixel-streamlines compariso
 Options affecting the SIFT model
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--  **-no_fd_scaling** by default, the fibre densities in voxels with grey matter partial volume contamination are scaled appropriately to compensate. Provide this option to override this behaviour and not perform any scaling.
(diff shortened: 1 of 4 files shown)


Lestropie:
Superseded by #653.

PR #640, opened 2016-05-31
mrview: Fix 3D textures for images with non-standard header offset/scaling

### LINKED ISSUE #368: MRView: scaled integer images not displaying correctly when 3D texture is used
That means when _not_ locked to axes, in volume render mode, or in lightbox mode. Issue relates to scale factor in header much smaller than 1, presumably these scalings are not properly handled when loaded into the 3D texture, whose type matches that of the original image, so we end up with small numbers rounded to zero in the texture.

Originally reported by @maxpietsch

Lestropie:
Closes #368.

@jdtournier Happy with this solution?


Lestropie:
Superseded by #653.

PR #641, opened 2016-06-01
Dwi2fod algorithms

Lestropie:
Details in individual commits.

Couple of comments from having gone through the process:

* Bug in `msdwi2fod` that was padding response functions with uninitialized memory if the requested *lmax* was greater than that of the provided response function. (Use `conservativeResizeLike(Zero())` rather than `conservativeResize()` if expanding a matrix and expecting it to be zero-padded)

* Removed some templating. Some parts were totally weird regarding which variables the templates applied to and which were hard-coded. So now it's just floats for the input / output images, doubles for processing. Shouldn't be hard to re-introduce if you *really* want to keep it; at least then you'd have to make the changes thorough in order to avoid mixing types.

* **Edit**: Bug in MSMT: The constraint matrix wasn't being zeroed, and then only blocks of it were being written to. :-(

Figure we merge this along with all the other things that are going in before the tag.


Lestropie:
Ah. One thing I commented in the code but forgot to address: Would it be more 'discoverable' to have linear spherical deconvolution as its own explicit algorithm, rather than having to use CSD with 0 iterations?

thijsdhollander:
> Bug in MSMT: The constraint matrix wasn't being zeroed, and then only blocks of it were being written to. :-(

This was in the original `msdwi2fod` ?  @jdtournier @bjeurissen : it's probably quite important these things (and their impact) are double(-triple)-checked before merge.

> Would it be more 'discoverable' to have linear spherical deconvolution as its own explicit algorithm, rather than having to use CSD with 0 iterations?

It would, but I personally don't feel linear spherical deconvolution *should* be very discoverable; so I'd leave it out as an explicit algorithm. :-)

thijsdhollander:
Just another reminder: the documentation (at least one workflow) should probably also be checked to comply with the interface changes.

jdtournier:
> Bug in MSMT: The constraint matrix wasn't being zeroed, and then only blocks of it were being written to. :-(

Good catch. It's frankly surprising this hasn't caused issues so far - more than likely because it's a sufficiently large allocation for it to get a fresh set of memory pages, which the kernel will zero by default. But this will be OS-dependent... Did it cause trouble on your system?

>This was in the original msdwi2fod ? 

Yes, it's in the original too... 

> @jdtournier @bjeurissen : it's probably quite important these things (and their impact) are double(-triple)-checked before merge.

Ideally, we'd all write perfect code the first time around. We don't, so unless the code actually bugs out, it's difficult to check these things based purely on code inspection. In this instance, the matrix comes out just fine on my system without the fix, probably for the reason mentioned above... 

Which is why we have a testing framework - but there are no tests for `msdwi2fod`? We should definitely sort that out... @bjeurissen, you're probably in the best position to generate a few tests along with data, and verify all of that against your original MatLab implementation - do you have time to do this? Bearing in mind this would have to be included as part of this branch given that the commands have changed.

> Removed some templating. Some parts were totally weird regarding which variables the templates applied to and which were hard-coded.

What was weird about that, exactly? This just allows the class to operate on any image, regardless of its `value_type` -  it's a trick I use all the time to provide a more generic interface. I agree there's less need for it now that we have a single `Image` class, but I still don't see anything weird about this that required hard-coding the input/output Images to `float` (it was already hard-coded to `double` for the computation) - the only tangible benefit as far as I can tell is that you remove the `template <class ImageType>`, etc. from the declaration of the relevant methods, everything else is pretty much as it was...? Unless you're talking about something else...?

bjeurissen:
@jdtournier: I can write some tests

@Lestropie: Sorry about the bug: I did not notice it, because I was getting straight zeros in the matrix during development. Just checked it again on a few machines here and I still consistently get zeros...

Nevertheless, I should have read the eigen docs instead of just checking the contents of the matrix after allocation... Thanks for spotting that mistake!

Lestropie:
> Did it cause trouble on your system?

No, was just valgrinding as I was porting and getting ridonculous amounts of errors.

> > Removed some templating. Some parts were totally weird regarding which variables the templates applied to and which were hard-coded.

> What was weird about that, exactly?

The algorithms themselves are hard-coded to work in double precision, so `value_type` only applied to the input / output `Image` classes, which is slightly deceiving: `value_type` is usually a processing typedef. But the header of the output image is forced to `Float32`, so there'd be no point in outputting to anything else. And for the input image, the typedef only exists in the `cmd/` file where the image values are read into a double vector by the `Processor` class and passed to the algorithm, so it's only 'providing a generic interface' to itself. And there was a `cost_value_type` presumably left over from 0.3.13 that wasn't doing anything. So I don't see the templates serving any purpose other than confusion in this instance.

Providing the algorithms themselves as templates would be a more 'generic interface'; but that's not where the templates are. If you want to provide the output images in either single or double precision, that would need to be built into the command, the templating as it was didn't achieve that; and being purely within the `cmd/` file, the `Processor` class wasn't available to other commands.

Or if you *really* want templates, could rename to `image_value_type` and have the `Processor` functions operate on `Image<image_value_type>` rather than e.g. `template <class DWIType>`, which is overkill for something outside of a header file.

> Sorry about the bug: I did not notice it, because I was getting straight zeros in the matrix during development. Just checked it again on a few machines here and I still consistently get zeros...

All good; much less severe than some of the bugs I've left lying around...

*dodges incoming rocks*

Fingers crossed it hasn't actually affected people too much. I found that when uninitialised elements are non-zero, they tend to be tiny (e.g. 1e-36), probably because even used memory is still more zero than non-zero and so the float exponent still comes out zeroed (or small if non-zero). Plus nobody's reported it having had massive effects. Nevertheless it could have either subtle or severe effects depending on what's in there, so we'll just report it as part of the tag changelist.

jdtournier:
OK, valgrind would indeed have flagged this as an error. Not sure it would have affected too many people though, given how consistently the matrix comes out zeroed in practice. But there's definitely plenty of scope for this to screw up on different systems, so good thing you spotted it...

About the template issue, the reason I tend to code templated classes directly is that more often than not, these classes end up getting pulled into their own headers and made available to the rest of the codebase. In my experience, the classes that I write as part of an app often end up being useful elsewhere, so I consciously write them that way to make it easy to copy/paste directly into a separate header when/if required. I completely agree that in the context of a signore m an m single

jdtournier:
_[sorry, goofed again...]_

OK, valgrind would indeed have flagged this as an error. Not sure it would have affected too many people though, given how consistently the matrix comes out zeroed in practice. But there's definitely plenty of scope for this to screw up on different systems, so good thing you spotted it...

About the template issue, the reason I tend to code templated classes directly is that more often than not, these classes end up getting pulled into their own headers and made available to the rest of the codebase. In my experience, the classes that I write as part of an app often end up being useful elsewhere, so I consciously write them that way to make it easy to copy/paste directly into a separate header when/if required. I completely agree that in the context of a single dedicated app, using templates is a bit overkill. But it is easy to do, and I don't find it confusing - no more so than using an explicit type. It did admittedly make a bit more sense back in the day when we had different buffers, since it made switching buffers trivial (although that could equally have been achieved with a typedef). 

However, once it makes its way out into a separate header, I'd argue that using templates makes a lot of sense: other developers might want to use it, and when they do there's no guarantee they won't need to pass an Image with a different type, or even an Adapter. The point is, it's trivial to support, so why explicitly remove the flexibility? What surprises me here is that you'd take a class that was already generic and ready to go, and remove its genericity when you pull it out into its own header - exactly the opposite of what I'd expect...

The other thing I don't get is why this would confusing. In this particular instance, I can see that the _rationale_ for using template is not obvious, and hence might have been confusing. But now that it's in its own header, it's pretty clear why we'd want it templated. I can also see that the obsolete `typedef compute_type` might have been confusing - that's obviously a relic from the past, and not a lot of use now because it's very difficult to use it with Eigen (if you want to use the much more convenient `MatrixXd`, etc. typedefs). But that's just one line to remove. The rest of the changes only affect the declaration of the method, as far as I can tell, from:

    template <class ImageType> void foo (ImageType& in)

to:

    void foo (Image<float>& in)

Everything else is more or less as-is. So it's a little bit less verbose, and explicit about the type requested (which is what I'm arguing _against_ here). The code itself is generally unaffected. So I don't think the _code_ is any harder to follow (although the compiler debugging output might...), and once it's in a header, the rationale is clear. Why go out of your way to make it less flexible?

Also, if we're going to template this, why restrict it to instances of the Image class? It's no harder for the compiler, no additional effort for the developer, no extra ambiguity for the user - and allows Adapters to be used. I know you've argued in the past that Adapters should be Images too, but they're not, so for now they should be supported unless there's a good reason against that.

As far as I tell from your post, the bulk of your aversion to the use of templates here was down to the fact that it makes no sense in a dedicated command. But you've just moved it out to its own header file, so none of that applies... I'd be strongly in favour of reverting back to a template structure here - unless you are absolutely convinced no-one will ever want to use that class anywhere other than its intended app, i.e. in `dwi2fod`.

Lestropie:
OK, let's summarize first. [Diff page](https://github.com/MRtrix3/mrtrix3/pull/641/files), line numbers are in reference to the *old* files:

* `dwi2fod.cpp`, [line 77](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-acce80360144129eca1f6bd0f62804cfL77]: '`value_type`' only applies to the loading of input & output images; it doesn't influence the datatype of the output image, or the type used for processing. Therefore name is misleading.

* `dwi2fod.cpp`, [line 78](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-acce80360144129eca1f6bd0f62804cfL78): Unused; relic as you say.

* `dwi2fod.cpp`, lines [92](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-acce80360144129eca1f6bd0f62804cfL92), [122](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-acce80360144129eca1f6bd0f62804cfL122) and [144](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-acce80360144129eca1f6bd0f62804cfL144): Only potentially reasonable use of templates as far as I can see. Completely trivial to re-implement if it did one day find another use. *However*:

  * Is inside a `cmd/` file, remains in a `cmd/` file, and makes most sense being in a `cmd/` file given its functionality.

  * If it *were* to be hypothetically moved into a header file, the sole purpose of that class is: load data from an image into a vector, run the algorithm, and write the vector result to an image. If some other command wanted to make use of CSD, this class would almost certainly *not* be used: you would either be performing some other processing prior to CSD (in which case you'd daisy-chain the functor operations rather than using a midway buffer), or you'd be doing something with the results of CSD (in which case you don't want to write to file). Not inconceivable that a use might be found, but very unlikely, and probably requiring considerable gymnastics to re-use what is in reality a very small and simple class.

> What surprises me here is that you'd take a class that was already generic and ready to go, and remove its genericity when you pull it out into its own header - exactly the opposite of what I'd expect...

> ... But now that it's in its own header, it's pretty clear why we'd want it templated. ...

> ... and once it's in a header, the rationale is clear.

> As far as I tell from your post, the bulk of your aversion to the use of templates here was down to the fact that it makes no sense in a dedicated command. But you've just moved it out to its own header file, so none of that applies...

The MSMT *initialisation* is now in its own header. If you look at [the file](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-11161ea66f0ba4509689bb000dde233fL78), that was *never templated in the first place*. The only thing templated in `msdwi2fod` was the [class](https://github.com/MRtrix3/mrtrix3/pull/641/files#diff-11161ea66f0ba4509689bb000dde233fL193) responsible for loading the data into a vector, running the algorithm, and writing the result to output files. For my thoughts on that, see above.

jdtournier:
OK, clearly I need to look at what you've _actually_ done more closely, rather than make incorrect interpretations about what I think you've done. Point taken... 

Obviously I had just assumed that you'd moved _all_ of the MSMT-CSD stuff out to its own header, but if it's going to stay in the command, I can see your rationale. 

Apologies for the confusion, I'll shut up now...

thijsdhollander:
Quick double-check of the command's user interface; I know it's trivial, but just checking I'm not overlooking anything at all:

Am I **100%** correct in that, for *any* current use case, I can simply instruct any user that this change, for them, simply means that, they need to replace occurrences of:

1. `dwi2fod` with `dwi2fod csd`
2. `msdwi2fod` with `dwi2fod msmt_csd`

...and additionally *maybe* be on the lookout for, hopefully/probably *very minor*, chances of a slight change in the MSMT-CSD output (without scaring them too much about this specifically)?

Also, just out of curiosity (definitely not arguing for/against anything here): any reasons for "msmt_csd" instead of "msmt-csd" (or even both)?

Lestropie:
> Am I 100% correct ...

Yep. You can see in 2f49412 that that's all I've done.

> Also, just out of curiosity (definitely not arguing for/against anything here): any reasons for "msmt_csd" instead of "msmt-csd" (or even both)?

Hmmm... Partly historical (thinking `SD_STREAM` and `SD_PROB` from 0.2), partly I always use underscores in directory / file paths since that seems to be 'convention' in Unix (thinking `.bash_history` etc.). Also dashes are used to prefix command-line options, so maybe the underscore helps to separate multi-word names from those. Could potentially have `.type_choice()` do a character replacement before it looks for a match so it would accept both?

thijsdhollander:
> Yep. You can see in 2f49412 that that's all I've done.

That's what I figured; thanks for confirming. :-)

> Also dashes are used to prefix command-line options, so maybe the underscore helps to separate multi-word names from those.

Good point indeed, there's quite a lot of dashes around the place already in your average command call.

> Could potentially have `.type_choice()` do a character replacement before it looks for a match so it would accept both?

Hadn't thought about that yet, but that would be perfect actually.  I'd still just keep all instructions as they are (and never really mention that it's also dash-robust), but it's nice to be robust against people accidentally/automatically typing that (I know I'd accidentally do it at times).  As far as I can tell, it should never interfere with the correct usage.

Lestropie:
Superseded by #653.

PR #653, opened 2016-06-06
Tag 0.3.15

Lestropie:
An accumulation of changes that will all be pushed at once and the 0.3.15 tag will be formed. In part these are all being done together because they either change the user interface, or the default behaviour. Awaiting addition of `dwidenoise` command.


Lestropie:
@maxpietsch `transformconvert` is failing assertions. About to have a look, but you're probably going to want to look at it even if I can find the solution myself.

maxpietsch:
OK, I'll check what is failing.

thijsdhollander:
Just did a completely fresh clone (deleted the mrtrix folder entirely before it for optimal cleanness) after the merge, followed by a `./configure release -dev` and a `./build`, but the latter one ended with:

```
...
(442/442) [LB] release/bin/tsfinfo
[SH] ./scripts/mrtrix_bash_completion

Error: please specify a directory
```

(the "`...`" part went fine)

Lestropie:
@thijsdhollander I think this issue has come up once before, when the documentation page for `dirsplit` was erroneously deleted. There appears to be another linux binary called `dirsplit`, which, if executed with the `__print_usage_rst__` argument, produces this error. Running a `git status` should confirm (the `dirsplit` doc page should be empty).

jdtournier:
[So there is](http://linux.die.net/man/1/dirsplit)! The question then is, why is the MRtrix3 version not being invoked preferentially? Does the `generate_user_docs.sh` script not set the PATH, or invoke commands using their relative paths?

thijsdhollander:
That seems to be it indeed. `git status` returns

```
	modified:   docs/reference/commands/dirsplit.rst
	modified:   docs/reference/commands_list.rst
```

and `dirsplit.rst` is empty!

Lestropie:
OK, must have flared up because of 8f1e685. I changed it to build the list of commands from the `cmd/` directory rather than `bin/`, so that when switching between git branches you don't accidentally add documentation for binaries that are lying around having been built on other branches.

maxpietsch:
@Lestropie generate_user_docs still picks up scripts that may lie in the script folder

Lestropie:
If it's lying in the scripts folder and hasn't been committed to a branch, then yes. Don't think there's anything that can be done about that, short of querying git. But you [no longer have to place a script in that directory in order to use the scripts library](http://mrtrix.readthedocs.io/en/latest/tutorials/FAQ.html#making-use-of-python-scripts-library), so just keeping such files out of there until they're committed to a branch is a better option.

maxpietsch:
Oh, didn't know that. Neat :)

PR #656, opened 2016-06-07
Docs update: installation on MacOSX

chunhungyeh:
Use `pkg-config` to fix the issue of eigen cflag


jdtournier:
Ah. I'd already pushed the same commit on Rob's pending mega-merge (#653)...

chunhungyeh:
Oh! I didn't notice, sorry about that. I will close this one.

PR #657, opened 2016-06-07
Configure script: set flags for correct OS X version

maxpietsch:
configure now uses `QMAKE_MACOSX_DEPLOYMENT_TARGET` or `MACOSX_DEPLOYMENT_TARGET` or the actual OS X version to set `mmacosx-version-min` in `cpp_flags`, `ld_flags` and `ld_lib_flags`.

@jdtournier Can you double check whether that is what you had in mind?

### DIFF
--- a/configure
+++ b/configure
@@ -406,9 +406,11 @@ elif system == 'windows':
   ld_lib_flags += [ '-pthread', '-shared' ]
   runpath = ''
 elif system == 'darwin':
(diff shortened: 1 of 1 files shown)


jdtournier:
Seems fine to me... Failure in testing was one job in `tckgen` for python2 - I'm assuming this is one of those random glitches, I've restarted that test, hopefully it'll just magically fix itself...

PR #658, opened 2016-06-08
population_template: removed intersection template mask for affine re…

draffelt:
…gistration. Improved the template mask for non-linear registration


scripts/population_template
```diff
@@ -105,6 +105,8 @@ for i in inFiles:
 noreorientation = lib.app.args.noreorientation
 dorigid = lib.app.args.rigid
 
+print dorigid
```
Lestropie: Loose debugging print?

draffelt:
Thanks Rob.

PR #667, opened 2016-06-14
Registration linear tweaks

maxpietsch:
- mrregister: added options rigid_log and affine_log to write gradient descent iterations to file
- mrregister: don't smooth input images of affine_niter or rigid_niter is zero
- population_template: crop average space to input masks + 10 voxels padding to avoid large templates if initial alignment is poor.
- population_template: write linear registration gradient descent iterations to log folder if run with -verbose option
- population template: new option linear_niter: specify number of iterations for linear registration


draffelt:
I was thinking we should also save the input_transformed and warp files for each level separately, but delete the data as soon as it's no needed. What do you think?

thijsdhollander:
Yep, sounds much cleaner; and would be very useful to "debug" (or just evaluate what's going on to get an insight in the quality of the final result).

maxpietsch:
Tested a full run of population_template with default arguments with numpy and without it. I would merge this branch soon if there are no objections, keeping it in sync with the changes to the documentation in #679.

draffelt:
Happy for you to merge. Though I was thinking maybe we could cut down on the number of linear template update iterations? What do you think?

maxpietsch:
Yes, thought the same with the data you shared with me. There the template is pretty much stable after 3 iterations. I am not sure how data quality affects the required number of steps though.

PR #669, opened 2016-06-16
generate_user_docs_path sets mrtrix path as discussed in #653

maxpietsch:


maxpietsch:
I'll merge as it doesn't seem to break anything.

PR #675, opened 2016-06-20
Macosx touchpad fixes

jdtournier:
This fixes up the touchpad scrolling issue on MacOSX (although there is residual weirdness due to the way MacOSX emulates scroll acceleration with a regular mouse). It also fixes up pinch-to-zoom (this had been implemented previously, but never tested), and adds handling of panning using the touchpad - as far as I can tell, this is triggered by a long motionless single touch (not press) and drag. 

Calling all Macbook users to test these changes!

### DIFF
--- a/src/gui/mrview/mode/base.cpp
+++ b/src/gui/mrview/mode/base.cpp
@@ -151,7 +151,7 @@ done_painting:
         void Base::mouse_press_event () { }
         void Base::mouse_release_event () { }
 
-        void Base::slice_move_event (int x) 
+        void Base::slice_move_event (float x) 
         {
           const Projection* proj = get_current_projection();
           if (!proj) return;
--- a/src/gui/mrview/mode/base.h
+++ b/src/gui/mrview/mode/base.h
@@ -84,7 +84,7 @@ namespace MR
             virtual void mouse_press_event ();
             virtual void mouse_release_event ();
             virtual void reset_event ();
-            virtual void slice_move_event (int x);
+            virtual void slice_move_event (float x);
             virtual void set_focus_event ();
             virtual void contrast_event ();
             virtual void pan_event ();
--- a/src/gui/mrview/mode/ortho.cpp
+++ b/src/gui/mrview/mode/ortho.cpp
@@ -140,7 +140,7 @@ namespace MR
 
 
 
-        void Ortho::slice_move_event (int x) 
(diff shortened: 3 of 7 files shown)


chunhungyeh:
The functions for panning (with one finger), scrolling (with two fingers), and moving crosshair (with three fingers) are working pretty well. Very interesting! :)

The pinch-to-zoom one is quite hard to control as it goes very fast.

jdtournier:
OK, thanks for the feedback. This is going to have to wait until someone with a Macbook shows up here for testing - everything worked fine on the Macbook Pro that I was testing this on last week, The `e->scaleFactor()` bit is supposed to return the fractional change in the zoom factor, which should really track what's on the touchpad exactly. If this is not reported correctly by Qt, I'm not too sure what the best way to fix it would be. @maxpietsch has  a Macbook Air - @chunhungyeh, is this what you have too...? Maybe it's specific to these models...

jdtournier:
Just tested this on my ridiculously underpowered Windows 10 netbook: pinch to zoom works just as expected, both for the touchpad and the touchscreen. Can't figure out what the gesture is for panning on the touchpad, but  with the touchscreen, I can pan around with 2 fingers on the screen. :+1:

chunhungyeh:
Mine is Macbook Pro (built in early 2013).

thijsdhollander:
> Just tested this on my ridiculously underpowered Windows 10 netbook: pinch to zoom works just as expected, both for the touchpad and the touchscreen. Can't figure out what the gesture is for panning on the touchpad, but with the touchscreen, I can pan around with 2 fingers on the screen.

Ok, now I have a seriously good reason to install MRtrix on my good old first generation Surface. :smile:  Is there a gesture for scrolling (through slices)?

jdtournier:
> Is there a gesture for scrolling (through slices)?

Couldn't find one. Problem being that scrolling is actually panning in most 2D applications...

maxpietsch:
Turns out the zoom problem is due to a bug in qt's pinch function (since qt 5.4.1) which was fixed in qt 5.6.1: https://bugreports.qt.io/browse/QTBUG-48138

Happy to merge!

jdtournier:
Cool! @chunhungyeh: any chance you could verify this? And merge if that fixes it for you too?

chunhungyeh:
Tested: updating to Qt5.6.1 fixes the problem.

Issue #594, opened 2016-05-07
mrview: No QT icons on Windows


Lestropie:
Still haven't resolved this yet, need to dig further.

jdtournier:
Check whether the SVG module is installed and available...

Lestropie:
Not so simple unfortunately. All seems to be there. I even deleted and rebuilt mrview's icons.cpp file on the same machine.

Lestropie:
OK, this one's officially got me stumped. Fresh installs using the up-to-date instructions, all the same packages: Icons on my Win7 desktop, no icons on my Win8.1 laptop :confused:. Config files are even identical between the two, excluding irrelevant stuff in `PATH`.

Lestropie:
... and guess who uninstalled `msys2` forgetting about the uncommitted branch overhauling the mesh code with ~5k line changes :angry:

jdtournier:
Ouch. That's not going to help the migraines...

jdtournier:
Coming back to the original issue: have you tried copying the executables across systems, to check whether the problems are specific to the MRtrix3 executables? Could also try copying the whole msys2 folder over, just to check whether it's something with Qt or something...

Lestropie:
> ... and guess who uninstalled msys2 forgetting about the uncommitted branch overhauling the mesh code with ~5k line changes :angry:

Got as far as running a deep filesystem file recovery and dumping git tree file contents... no avail :cry: 

> Coming back to the original issue: have you tried copying the executables across systems, to check whether the problems are specific to the MRtrix3 executables? Could also try copying the whole msys2 folder over, just to check whether it's something with Qt or something...

Either executable / dll running within the Win7 msys2 installation has icons. Either executable / dll running within the Win8.1 installation has no icons. There's clearly differences in the msys installation between the two: One has a `/bin/` directory, the other doesn't. So there must be an error in how Qt5 installs on Win8.1 (no separate svg package, it all comes bundled). Ideally I want to figure out how to fix it, users might not have a second Windows box with a different version in order to swap files.

Lestropie:
In case anybody else encounters this: The only solution I've been able to find thus far is to downgrade from Qt5 to Qt4.

jdtournier:
[...]

PR #679, opened 2016-06-21
Updates to installation instructions

jdtournier:
A couple of changes based on recent experience:

- update the `PATH` update command to deal with spaces in folder names - often a problem especially on Windows

- add a warning about issuing the `PATH` update command repeatedly, and suggest the use of an editor to modify - or at least delete the offending line before issuing the command again...

Thoughts / amendments welcome...


Lestropie:
Isn't QtCreator a bit overkill for editing a Bash file? I'd recommend Notepad++ myself. Shouldn't cause any issues with line endings.

jdtournier:
Yes, I wasn't too pleased about that either... What I'm shooting for is an easy to use graphical editor that can be installed via the `pacman` command, and launched on the MSYS2 command line. It turns out that `gedit` brings with it a ton of other dependencies, making it quite an involved download. I'm reluctant to suggest other editors since I can't guarantee they'll be in the PATH for MSYS2, and navigating to the `~/.bashrc` file is not trivial, given that the home folder is in not in an obvious location. This is why I want simple instructions that can simply be copy/pasted into the terminal and will just work without fuss. Not married to `qtcreator` in any way, I just didn't find anything else on the package repo - plus I was kinda hoping it had been included in the Qt install anyway... 

Another option is `nano`, which is available and relatively simple to use. I'd be happy to refer to that instead, but I have a feeling users might find a graphical editor more comfortable...

If you find another simpler graphical text editor,  feel free to update the instructions. Just as long as it's a foolproof process...

Lestropie:
> I'm reluctant to suggest other editors since I can't guarantee they'll be in the PATH for MSYS2, and navigating to the ~/.bashrc file is not trivial, given that the home folder is in not in an obvious location.

True; but you could instead instruct to take note of the location where Notepad++ gets installed, and execute it from MSYS2 via an absolute path. Though I suppose you've then got to deal with whitespaces due to `Program Files (x86)`... `nano` is probably the way to go I guess, not likely to be much else that's graphical and provided by MSYS2.

> This is why I want simple instructions that can simply be copy/pasted into the terminal and will just work without fuss.

Isn't that what got us into this situation in the first place? :-P

jdtournier:
> nano is probably the way to go I guess, not likely to be much else that's graphical and provided by MSYS2.

Happy to go with nano if you really don't think `qtcreator` is the way to go. Not too fussed...

>> This is why I want simple instructions that can simply be copy/pasted into the terminal and will just work without fuss.
> Isn't that what got us into this situation in the first place? :-P

Well, not really... Incorrect instructions is what got us into this situation. The alternative is to get people to edit their `./bashrc` file directly, and I can't see that being any easier. One thing we could do is include script to amend the PATH - maybe as an additional option to `build`...?

Lestropie:
> Incorrect instructions is what got us into this situation.

Me derping the installation instructions was only recent (went with the tag); people have been getting stumped on that step for ages. Been needing that extra text for a while. In fact I wonder if it needs even more elaboration.

jdtournier:
> In fact I wonder if it needs even more elaboration.

I agree and I don't. This is such a fundamental aspect of installing any software, it's really not specific to MRtrix3. Really, we're providing these instructions as a convenience. But we have to get it right since most users genuinely don't have a clue about these things. Half of me thinks we can leave things as-is and provide a link to a more detailed explanation of what the PATH is and how to set it, but that's actually asking a lot from people who have better things to do. The more I think about this, the more I'm leaning towards a script to handle this...

jdtournier:
Just had a quick go at cobbling a script together for this. This is what I came up with:

```bash
setpath="export PATH=\"$(pwd)/release/bin:$(pwd)/scripts:\$PATH\""
grep -Fxq "$setpath" "$1"

if [ $? -ne 0 ]; then
  cat << EOD 
The PATH for MRtrix3 does not appear to be set. The following line has been
added to the file "$1":

    $setpath

EOD

  echo "$setpath" >> "$1"
  eval $setpath

else
  echo "The PATH for MRtrix3 is already set."
fi
```

Save it to a file called `add_path_to` in the MRtrix3 toplevel folder, and invoke with:

    $ source add_path_to ~/.bashrc

That will only add the requisite line if it's not already present in the file. The minor disadvantage is that the line has to match _exactly_, so it won't catch equivalent settings for the PATH that might work just fine already. But it's really difficult to identify lines that might set the MRtrix3 PATH unambiguously. For example, maybe there's a line setting the PATH for an obsolete version of MRtrix3, in a folder that's long gone - in which case we might detect that as already set when it does in fact need to be updated. Maybe the user is setting the PATH to MRtrix3 along with a bunch of other software, in which case replacing that line would mean the other software would no longer work. So the approach here is simply to check whether the line exists already as an _exact_ match, and add it to the file if it's not present. 

As a bonus, using the syntax above allows us to also set the PATH in the current terminal - no need to start a fresh terminal, it should work straightaway. 

Other minor limitation is that this is `bash`-only - no `tcsh` support. But then the current instructions have always been `bash`-specific anyway...

I haven't added this to the repo yet since I think this is really an idea to be discussed. If everyone agrees, we can add the file and modify the docs accordingly.

Lestropie:
1. Your script would claim that everything was OK if it were run from the same wrong directory that the snippet was first run from. It also wouldn't clear the offending line if that script were first run from the incorrect location; not that that would matter greatly, but it's still not 'clean'.

2. The instructions still didn't actually say that the snippet needed to be run from the *MRtrix3* root directory. Probably the source of a lot of problems... 

3. Once a user has to go to the effort of installing a text editor and removing an offending line, I think they might as well manually add the necessary line while they're there; they've just installed the software, they should know where it is. Telling them to exit and re-run the snippet means they'll probably just repeat whatever mistake got them into that situation in the first place.

Lestropie:
One other thing that could potentially be added: Pick a good bash / unix tutorial page / website, and include a link to it at the top of the installation pages with a friendly, non-condescending help message.

i.e. somebody else needs to write it; I can't do non-condescending ;-p

jdtournier:
> One other thing that could potentially be added: Pick a good bash / unix tutorial page / website, and include a link to it at the top of the installation pages with a friendly, non-condescending help message.

Yes, I thought about this too, mentioned it earlier. Two problems: I couldn't find a simple clear tutorial on this. There's lots of information in different locations, so you can figure everything out if you know where to look, but it's not as simple as I'd hoped. Second problem is for most non-technical users, this is enough to completely turn them away. Telling them to figure out how to update their `PATH` is asking them to take a crash course in UNIX, and most of them will simply give up - at least in my experience of dealing with neuropsych students, etc. 

Which is why I think we do need a one-shot copy/paste instruction that will just work _most_ of the time. If it doesn't work, sure they'll have to get their editor out and fix it the hard way - just like they would have had to if we hadn't provided that one-liner. But most users will hopefully not need to think about it, and will be able to just get on with it.

So like I said, the reason I'd posted the script here rather than commit is because this is a very first stab at doing this - lots of scope for improvements, clearly. So how about this:

- the script checks for the existence of `release/bin/mrinfo` before proceeding, issues a hopefully helpful error message if it's not found.

- the script _also_ writes a comment line such as:
  ```bash
  # MRtrix3 PATH automatically generated by add_to_path script:
  export PATH=...
  ```

- the script checks for the presence of that line instead of an exact match to the line it would have been written, and if found, replaces the next line in the script, provided it starts with `export PATH=`.

I reckon that would fix most of the issues you've raised...?

jdtournier:
OK, take a look at the python script I've just pushed in #685 - does more or less what I mentioned in my previous post. It's in a separate pull request to be merged onto this one, since if we're all happy with it, we'll need to update the installation instructions to match.

jdtournier:
Just found out the package names for `numpy` should be `python-numpy` on both Ubuntu and Arch - updated the instructions to match. Can anyone confirm that this is actually correct...? Haven't tested...

jdtournier:
@Lestropie - assuming the above change is OK, are we good to merge?

maxpietsch:
it is `python-numpy` on ubuntu

Lestropie:
Hmmm. I removed NumPy from the list of "installation requirements" in 8f87e55 (it's a runtime requirement, not a prerequisite for installation), but it's still listed as a package to install, since there's one script that requires it. Do you want to leave it like that, or should it to be listed in both, or neither?

jdtournier:
Not sure how best to handle this. Is there any way to detect at runtime within the python script that needs it whether `numpy` is installed? If we can catch it, we could at least provide instructions at that point to install it. If that can be done, it would be better to remove it from the package list.

That said, if we think `numpy` might come in handy for other scripts in the near future, it's probably best left as-is. It's certainly no deal-breaker on any of the supported platforms, they all provide packages for it via the official channels. So my gut feeling is probably to leave things as they are for now - unless there's any evidence that it causes trouble in certain cases?

Lestropie:
Mostly the Windows case: MSYS version which is installed by default doesn't have numpy available, which means stuffing around with getting that uninstalled and installing the mingw version (and hopefully not breaking anything else in the process), for the sake of a script that's only required by a fraction of users. Having said that, looks like it may well remove cleanly...

maxpietsch:
The only thing numpy is used for in `population_template` is to load and linear transformations to text file and for one matrix dot product. I could replace that with non-numpy code if numpy is not going to be used anywhere else.

dchristiaens:
@jdtournier 
> Is there any way to detect at runtime within the python script that needs it whether `numpy` is installed?

Yes, `import numpy` will throw an `ImportError`, which you can catch and print instructions for.

As to the question of whether or not to leave it in, I think numpy would be useful to have at our users disposal. For example, it's used in my code for reading/writing .mif and .tck files, akin to the matlab file handlers, which I could release at some point. However, I'm not accustomed with the burden of getting it installed on Windows and if it's too much trouble I'm OK with removing it from the list of required dependencies.

maxpietsch:
Windows currently has [no official source](https://github.com/numpy/numpy/issues/5479) for a numpy wheel but there is an alternative [here](http://www.lfd.uci.edu/~gohlke/pythonlibs/#numpy). I don't have a windows installation handy but after downloading that wheel 

    pip install numpy-1.10.2+mkl-cp35-none-win_amd64.whl

should work. There is no license information on http://www.lfd.uci.edu/~gohlke/pythonlibs other than "Refer to the documentation of the individual packages for license restrictions and dependencies." I guess that means we could host the numpy wheel ourselves which should make the install completely scriptable.

Lestropie:
My MSYS2 is broken again: `mrview` stopped working entirely for a while there; now after an MSYS update I've once again [got no icons](https://github.com/MRtrix3/mrtrix3/issues/594) in `mrview`. So I might try a fresh install with shifting to the MinGW Python, update installation instructions if need be, then this can be merged.

I think I'd prefer to avoid wrapping any kind of Python support: especially given my own system has been unstable under MSYS2 updates, I don't feel we should be supporting any kind of wrapping in order to get around their temporary bugs.

jdtournier:
Agree things are a bit of a mess in MSYS2 right now. On the python issue: the version shipped by MinGW doesn't work. First off, it doesn't provide `python` as such, just `python3`, so invoking `./configure` doesn't work out of the box. That's really resolved by invoking `python3 configure`, but that's not the real issue. The bigger problem is that for some mysterious reason, trying to get the Eigen3 flags by invoking `pkg-config --cflags eigen3` (via subprocess) fails with Python 3.5 as shipped by MinGW, while it works fine with version 3.4 as shipped by the MSYS2 `python` package. I can't figure out why, the relevant environment variables are set correctly in the subprocess, as far as I can tell from limited testing, and the command works just fine when copy/pasted into the terminal. Go figure...

Lestropie:
In that case, why don't we change the suggestion back to the MSYS Python, merge this, perhaps make a new issue describing just this issue, and we can come back to it if someone actually tries to run `population_template` on a Windows machine and jumps up and down. Who knows, by then MSYS2 / MinGW might have sorted it out.

jdtournier:
OK, reverted to MSYS2 Python (didn't realise it had been changed), and removed MinGW numpy from the list of dependencies (it won't work with the standard MSYS2 version of Python anyway...). Maybe someone can open an issue about the lack of numpy on Windows, but I do suspect this will eventually get sorted out by the MSYS2 team - hopefully they do intend to bring MinGW and MSYS2 into alignment at some point... 

Other than that, I think we're ready to merge? The only thing outstanding from this issue is the reliance of `population_template` on numpy, but I don't think this is a big deal. Windows is already a second-rate option given that FSL isn't available on that platform... That said, it would be nice if all _MRtrix3_-specific scripts and commands did work out of the box on all supported platforms, but this can be sorted in due course (or we just wait until numpy does become available on MSYS2).

maxpietsch:
I removed all numpy dependencies from pupulation_template in 8074df2eb65dce4b. Haven't had time to test the branch (#667) but a quick run worked so I guess I can merge if everyone is happy with the changes.

jdtournier:
Cool, happy with that!

jdtournier:
@Lestropie, I guess you've had the most to do with this pull request, I'm happy for it to be merged when you are. Hit the button if you agree!

Issue #560, opened 2016-04-20
Add drag/drop files to tools


chrisk1:
It would speed up usage considerably if it were possible to drag/drop files into the tools in mrtrix - i.e. dragging a file into the window to use as an overlay (vs clicking "Open overlay image" and navigating to the file's folder). Thanks!

maxpietsch:
If not too much hassle, it would also be nice if the order of the images can be changed by moving them in the image list. This would be useful when ROIs overlap for instance. Definitely in the "nice to have" category, though.

jdtournier:
OK, both of these require drag & drop support, which already exists for the main image. If anyone is interested in taking this on, [take a look here](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/window.cpp#L128-L153) for inspiration...

rtabbara:
@chrisk1 @maxpietsch  This has now been added. Let me know if you encounter any problems.

PR #682, opened 2016-06-22
Adding drag-drop to tools

rtabbara:
Feature request #560

* Support to load files via drag and drop (Overlay, ROI, Tractography and Vector for now)
* Support to reorder loaded files within the list view (Overlay and ROI)

I avoided supporting reordering of files for the Tractography and Vector tools as both use depth-testing when rendering.


rtabbara:
Anyone want to look this over before merge?

Issue #672, opened 2016-06-20
Scripts: Logging & error messages


Lestropie:
* Generate a log file in the temporary directory with the commands executed.

  * This would be beneficial for the `-continue` option: Skip all but the final entry in the log (re-run that since that's the command that failed). If the error occurred earlier, users can manually erase log entries until the last entry is the first thing they want run.

  * Actually, if this works, could have a separate script responsible for script recommencement. When you feed it the temporary directory location, it checks the calling command / original working directory / command log, and executes the relevant script with (potentially hidden) flags to resume the script.

* When a command fails, capture the stderr / stdout contents of that command and write to the terminal. Much easier now that `runCommand()` is using `subprocess`.

Lestropie:
Closed by #683; might move the `-continue` enhancement to a new issue.

PR #683, opened 2016-06-22
Script changes

Lestropie:
Attacked the scripts due to various user discussions and personal frustrations. More details in commit messages.

Probably still to do:

* Don't use mutually exclusive options in `argparse`. They're missing a lot of flags, and so can bugger up the documentation production. -> Done.

* Ideally I'd like to allow to have different authorship, citation list and copyright messages for different algorithms (for those scripts that run different algorithms from the `scripts/src/` directory). But every time I try to do it I break something else. So might not be worth the hassle until there's a present need for it.

* More testing on a non-Windows machine, just to be sure...

* Other aspects of #672. -> Done: Using the command log for the `-continue` functionality can be moved to its own new issue and tackled later, since I'm not sure how many people are actually using it.

* Proper paragraph spacing, just like in the binaries. -> Done.


Lestropie:
Hope everybody enjoys the new script help pages :grin:

PR #685, opened 2016-06-23
set_path: script to automatically set the PATH

jdtournier:
As discussed in #679. This script seems pretty solid to me, but feedback on whether it works as advertised would be much appreciated...


Lestropie:
Happy to go with this option; just need to make sure it's robust against the kind of mistakes that people were making in the first place with the `PATH` update snippet and doesn't throw any new curveballs. Have pushed some suggestions.

jdtournier:
Are you happy to merge into #679 then? And update the doc accordingly...?

Lestropie:
On my Windows laptop, `.bashrc` is not sourced, but `.bash_profile` is. So I changed the default target in that case. I also noticed that the old Mac install instructions had a mixture of `.bash_profile`, `.bashrc` and `.profile` in the text; best confirm that the correct target has been selected.

jdtournier:
Well, unfortunately there's no such thing as 'the right target', as far as I can tell. Technically, `.profile` is sourced for a login shell (whatever that means), which happens to be the default on MacOSX, but not on Linux. And these defaults can be changed easily, so that the instructions may work most of the time, but not necessarily _all_ of the time. I _think_ `.bashrc` is sourced explicitly from both `.profile` and `.bash_profile` on most Linux systems, but again there's no guarantee, people can change these files if they want. If we want compete coverage, we'd need to detail all of these issues in some detail.

Also, I've been doing a few clean reinstalls of msys2 recently, and haven't had any issues using `.bashrc`. Not saying that `.bash_profile` wouldn't work also, but maybe there's something non-standard about your setup...?

jdtournier:
OK, just looked some more into which file is appropriate, and as I was saying it's not straightforward. [Here is the reference](https://www.gnu.org/software/bash/manual/bashref.html#Bash-Startup-Files) straight from the GNU doc. 

As you can see, `bash` will _only_ read `~/.bashrc` if started as a non-login shell. When started as a login shell, it reads _one_ of  `~/.bash_profile`, `~/.bash_login`, and `~/.profile`, whichever one it comes across first. The recommendation is to have `~/.bash_profile` source `~/.bashrc` explicitly so that the login shell includes all the non-login settings too. On my Arch system, this is exactly what `~/.profile` does (there's no `~/.bash_profile` file at all). On my (fresh and unmodified) MSYS2 install, I have both `~/.profile` and `~/.bash_profile` present, and both source `~/.bashrc`. From the GNU doc, it sounds like `~/.profile` is redundant and would be ignored (?), but `~/.bashrc` should be sourced in any case.

All this to say, a properly set up system _should_ source `~/.bashrc` in every case. This falls over if the user modifies `~/.profile` and/or `~/.bash_profile` and removes the lines that explicitly source `~/.bashrc` (or deletes the files altogether). In this case, all bets are off... But this should _not_ be the default for any reasonable distribution (including MSYS2).

On MacOSX, the situation is also pretty messy, but it looks like the default setting is for the terminal to start a login shell by default (the opposite to Linux), and furthermore that the default `~/.profile` or `~/.bash_profile` does _not_ source `~/.bashrc` (maybe these files don't even exist?). So on MacOSX the safest option is to default to `~/.bash_profile`. As I mentioned before, you can change the default behaviour to start a non-login shell on MacOSX, in which case `~/.bash_profile` _won't_ be read (conversely, you can change the default to login on Linux too, at least using `gnome-terminal`). 

So it's all pretty messy and ultimately relies on users not screwing up their defaults. But I think the instructions as we had them were correct for unmodified user accounts, which I think is the only reasonable situation to target...

Lestropie:
OK. Given the mucking about I've done in the past with my MSYS install, it's likely that whatever is supposed to be sourcing `.bashrc` is currently missing for me. In which case, let's default to `.bashrc` for MSYS, and hope nobody else winds up in my predicament. Besides, the instructions for changing the target if `PATH` isn't being set are right there in the docs.

PR #686, opened 2016-06-23
Fix output redirection

jdtournier:
As discussed on #161, and recently raised again on the [community forum](http://community.mrtrix.org/t/mrtrix-redirect-output-and-error-to-a-file/314). We already had a way of detecting whether `stderr` was redirected to file, but there were still issues when the redirection was in append mode (i.e. `command 2>> log`), since this explicitly prevents seeking back in the file. This meant the progressbar wasn't able to update its line correctly, and the output file was being filled with every single update. 

This change now also check the file access mode, and if the `O_APPEND` flag is set, forces the progressbar to write each update as a new line, without trying to overwrite the previous update. It also throttles back the update rate to 1/30th of the normal rate, which should keep the log file manageable without completely killing all indication of progress. 

Obviously this will need thorough testing, particularly on Windows and MacOSX. For reference, this is what I get with these changes when running a simple `mrconvert` job in various circumstances:

**No redirection:**
```ShellSession
$ mrconvert in.nii out.nii -force 
mrconvert: [WARNING] existing output files will be overwritten
mrconvert: [100%] copying from "in.nii" to "out.nii"
```

**Straight redirection:**
```ShellSession
$ mrconvert in.nii out.nii -force > stdout.log 2> stderr.log
```
results in `stderr.log`:
```
mrconvert: [WARNING] existing output files will be overwritten
mrconvert: [100%] copying from "in.nii" to "out.nii"
```

**Append-mode redirection:**
```ShellSession
$ mrconvert in.nii out.nii -force >> stdout.log 2>> stderr.log
```
results in `stderr.log` (assuming it was empty or non-existent to begin with...):
```
mrconvert: [WARNING] existing output files will be overwritten
mrconvert: [  0%] copying from "in.nii" to "out.nii"... 
mrconvert: [ 30%] copying from "in.nii" to "out.nii"... 
mrconvert: [ 60%] copying from "in.nii" to "out.nii"... 
mrconvert: [ 90%] copying from "in.nii" to "out.nii"... 
mrconvert: [100%] copying from "in.nii" to "out.nii"
```


Lestropie:
Looks pretty cool. Surprised nothing else was using `Path::is_file()`.

Just some thoughts:

* Could do with testing with SLURM too, just to be sure.

* Was already there, but will raise nonetheless: With using `fseeko()` to jump back in the file and overwrite existing contents, is there a risk of overwriting debug / error information that's being written while the progressbar is active?

* If `stderr_seekable` is false, is a factor of 30 reduction in output frequency appropriate / adequate? Or are we better off just writing e.g. `tckgen: --stderr progress updates suppressed--\n` as the second progressbar update, until the final update call when `done` is true? That way the log would look something like:
```
tckgen:        0 generated,        0 selected
tckgen: --stderr progress updates suppressed--
tckgen:   131027 generated,   100000 selected
```
I reckon people getting confused / annoyed with a long list of progress updates in a log file (even after a factor of 30 reduction) would still be more frequent than someone encountering an error mid-progressbar, using append-mode redirection, and needing to know exactly far the command progressed before it crashed; in which case you just tell them not to use append-mode redirection.

jdtournier:
> Surprised nothing else was using Path::is_file().

There's another version of that function that takes a `std::string` - the overload that was unused took a file descriptor, which is much less common.

> Could do with testing with SLURM too, just to be sure

Are you in a position to do this?

> With using fseeko() to jump back in the file and overwrite existing contents, is there a risk of overwriting debug / error information that's being written while the progressbar is active?

Just checked by running a `mrconvert -debug` job - output is strictly identical between terminal and non-append redirection.

> is a factor of 30 reduction in output frequency appropriate / adequate? 

Yes, I don't know how to really deal with this. It possible to change the way the percentage progressbar updates to something like:

    mrconvert: copying from "in.nii" to "out.nii"... 0% .. 20% .. 40% .. 60% .. 80% .. 100%

which would clean up the output a bit in that case, but it doesn't change the issue with the generic update used in e.g. `tckgen` or `dirgen`, etc. I had a go with `dirgen`, and it was (I think) useful to have regular updates as the command progressed, but for longer jobs, an update per second (which I think is roughly what this works out as) would still produce thousands of lines. One strategy that I was also thinking about was to double the sampling interval following every sample. So you'd get updates 1,2,4,8,16,32,... which would allow short jobs to provide frequent reports, and throttle right back for longer jobs. So even a hour-long job would only produce ~16 updates (assuming 30 progressbar calls per second). What do you think?

Other thought is whether we should use the same strategy for both types of redirection. It would simplify things a bit, and also makes the output more consistent - users might be wondering why the formatting changes with what might look like identical circumstances. It's a bit of a shame from my point of view given that the straight redirect actually does produces exactly the same as the terminal would have, but maybe consistency is more important here...

jdtournier:
By the way, if we do decide to make both redirects consistent, testing becomes a lot easier, since we won't need to detect whether it's running in append mode or not - just need to check whether the output is a file, as was done already...

jdtournier:
@lestropie, any thoughts on this? I think I'll go for the simpler option of treating any redirects to file the same way, avoiding the whole seek issue. What I'm not sure about is what to do when the `ProgressBar::update()` method is called to change the text _and_ the percentage is shown - this applies to `tckgen` in particular. In this case, we'd want regular updates, preferably with decreasing frequency. I think this is important so we can detect problems where the crash happens during processing, but early on (say at  3%) rather than as soon as the loop is entered - this is the kind of information that can come in handy when debugging. But when the text never gets updated it's a bit pointless to do this, when we can show progress with a one-liner. The problem is it's difficult to detect this in the backend with the current API....

Lestropie:
>> Could do with testing with SLURM too, just to be sure

> Are you in a position to do this?

I was going to get VLSCI access to test it, but they'd probably expect an acknowledgment on the MRtrix3 paper just for that. I might email a collaborator at Monash and ask them if they can test it.

> One strategy that I was also thinking about was to double the sampling interval following every sample. So you'd get updates 1,2,4,8,16,32,... which would allow short jobs to provide frequent reports, and throttle right back for longer jobs. So even a hour-long job would only produce ~16 updates (assuming 30 progressbar calls per second). What do you think?

I guess I worry that with anything too elaborate, it's only going to contribute to user confusion... but maybe I'm paranoid.

Another thing to look into (given we're trying to preserve useful progress info in such circumstances) would be to latch the most recent text update (whether it's displayed or not), and display that on destruction (if it wasn't displayed already), so that if an exception is thrown, the most up-to-date progress should be shown.

> Other thought is whether we should use the same strategy for both types of redirection. It would simplify things a bit, and also makes the output more consistent - users might be wondering why the formatting changes with what might look like identical circumstances. It's a bit of a shame from my point of view given that the straight redirect actually does produces exactly the same as the terminal would have, but maybe consistency is more important here...

Bear in mind that you identified and raised the append-mode redirection issue; the reported issues have always been to do with custom progressbar text. Wouldn't surprise me if nobody else has or will ever come across this discrepancy. Which is why I personally wouldn't mind if updates were just disabled in the append case. But we'd still need to deal with the custom progressbar text for non-append redirection.

jdtournier:
> I guess I worry that with anything too elaborate, it's only going to contribute to user confusion... but maybe I'm paranoid.

Agreed that we should keep things simple and predictable... which is why I'm leaning towards using the same approach for both types of redirection: most of the complexity currently relates to handling the line overwrite properly. Getting rid of that would actually simplify things quite a bit.

> Another thing to look into (given we're trying to preserve useful progress info in such circumstances) would be to latch the most recent text update (whether it's displayed or not), and display that on destruction (if it wasn't displayed already), so that if an exception is thrown, the most up-to-date progress should be shown.

So I _think_ that's already pretty much what happens (I'd need to check). But I don't think that's actually going to help: if an exception is thrown during processing, the chances are it'll be within a multi-threaded context, and it'll probably cause an immediate abort - no chance for the exception to be handled. And it also won't help when the issue causes a segfault. I think we really need to have a relatively up to date message committed to the log _before_ the problem occurs if we're to guarantee having a good idea of _when_ it happened.

> Bear in mind that you identified and raised the append-mode redirection issue; the reported issues have always been to do with custom progressbar text. Wouldn't surprise me if nobody else has or will ever come across this discrepancy. 

No, this affects _all_ ProgressBar updates, custom or otherwise - just checked with a straight `mrconvert`. It's only been _reported_ for jobs like `tckgen` since it'll be by far the biggest culprit in terms of filling the logs, just because it runs for so long.

> But we'd still need to deal with the custom progressbar text for non-append redirection.

Actually, that case works just fine currently - you can run e.g. `dirgen` to check, it works just fine in normal redirect mode, the output is identical to the normal terminal output. The issue with append redirection currently is that the seek back to the start of line is ignored. This affects all forms of the ProgressBar, since they all rewrite the entire line for every update: to increase the percentage, modify the busy indicator, and/or change the text displayed. 

We have to forget about updating an already written line if we want to handle append redirection - and we do, since this is what most queue systems seem to use. So that only really leaves two options: add to the current line (which we could do for a regular progressbar where the text doesn't change), or add more lines with reduced frequency (which suits a custom update scenario). I'd like to see whether we can do both of these. The decreasing frequency of updates is one way to ensure the reasonable frequency of updates for short jobs while avoiding ridiculously long logs for long jobs. I'll see if I can nut out something that handles all of this cleanly...

Lestropie:
OK hard reset, I lost the plot there somewhere; of course it'd have to be in append mode redirection to be raised :-/

I reckon the average user wouldn't be thinking about text logs in append mode v.s. non-append mode, they'd be thinking about the logs they get on a HPC vs. what they'd normally see on the terminal. Given that's a specific use case where things are typically 'a bit unusual' to begin with, I don't think having a SLURM log differing from a manual stderr redirect would throw a flag for most people. And the handling of such is fairly well confined to the `ProgressBar` class, so it's not huge amounts of code - you could maybe set update / completion function pointers to handle progress text for terminal / append-mode / non-append-mode (like what's currently done to handle GUI apps) to reduce wacky code branching. But if you'd rather simplify and homogenise that's your call.

> add to the current line (which we could do for a regular progressbar where the text doesn't change)

I'd probably go for reprinting the progressbar text with the new percentage, since debug / info calls can occur mid-progressbar and therefore 'interrupt' the appearance if just appending a new percentage.

jdtournier:
OK, I've had a bash at implementing what we've been talking about. Seems to work fine on my end. It also uses function pointers to keep things a bit cleaner, as you suggested. Both types of redirection handled identically, just makes things easier, really. Here's a couple of example outputs:

command:

    tckgen csd8.mif -seed_image ~/test_data/mask.mif poo.tck -force -num 100k 2>log.txt

produces `log.txt`:
```
tckgen: preloading data for "csd8.mif"...  [==================================================]
tckgen: [  0%]        1 generated,        0 selected
tckgen: [  0%]      141 generated,       58 selected
tckgen: [  0%]      361 generated,      150 selected
tckgen: [  0%]      781 generated,      330 selected
tckgen: [  0%]     1591 generated,      658 selected
tckgen: [  1%]     3041 generated,     1250 selected
tckgen: [  2%]     6161 generated,     2518 selected
tckgen: [  5%]    12051 generated,     5000 selected
tckgen: [ 10%]    23743 generated,    10000 selected
tckgen: [ 19%]    47451 generated,    19822 selected
tckgen: [ 39%]    95051 generated,    39628 selected
tckgen: [ 79%]   190371 generated,    79172 selected
tckgen: [100%]   240780 generated,   100000 selected
```

command:

    dirgen 200 dir200.txt -pow 8 2> log.txt

produces `log.txt`:
```
dirgen: [ .  ] Optimising directions (power 2, energy: 56544.113, gradient: 137.36828, iteration 125)
dirgen: [  . ] Optimising directions (power 2, energy: 56535.492, gradient: 26.73801, iteration 280)
dirgen: [   .] Optimising directions (power 2, energy: 56534.742, gradient: 3.8874109, iteration 481)
dirgen: [ .  ] Optimising directions (power 2, energy: 56534.563, gradient: 1.7344997, iteration 919)
dirgen: [   .] Optimising directions (power 2, energy: 56534.539, gradient: 0.030542887, iteration 1798)
dirgen: [   .] Optimising directions (power 8, energy: 3.7796542e+08, gradient: 28209153, iteration 0)
dirgen: [done] Optimising directions (power 8, energy: 3.7742391e+08, gradient: 104.19969, iteration 1122)
```

Would be good to double-check on actual HPC clusters at this stage...

Lestropie:
No go on getting my collaborator to test it on HPC. I've applied for VLSCI access, turnover is much faster than it used to be but I don't know how long until I'll be able to test it. Might be worth asking one of the MRtrix users who have reported the issue?

jdtournier:
Sure, I'll make some noise about this on the community forum.

jdtournier:
> Got MRtrix compiled natively on VLSCI. Output from those two commands look the same as your quotes.

Good to hear.

> Should I add to the docs the steps I took to compile natively on HPC?

Was that different from the standard instructions...? Or was the issue getting all the dependencies sorted? If you've got some pearls of wisdom to share, I'm sure there'll be plenty of people interested in hearing about it!

> I also need to up the Python requirement to 2.7: 2.6 no workey.

What doesn't work exactly? I'd put in a bit of effort to make sure both `build` and `configure` work with 2.6 - would be crap to hear that's no longer the case...

Lestropie:
> Was that different from the standard instructions...? 

Mostly just little gotchas in compiling `g++` from source - nothing major. But it might not be that much harder than the current suggestions, and would get the right optimisations and make updating easier.

> What doesn't work exactly?

Primarily it doesn't like the `.encode(errors='ignore')` calls from 55ee9eb. But `build` also seems to hang and I was having a bit of trouble figuring out why. I can have another go if you want it fixed.

jdtournier:
Re python 2.6: no big deal, I think it's probably best to require 2.7, otherwise it makes it quite difficult to write scripts that can run under both python 2 & 3...

Issue #451, opened 2016-02-26
Viewing each volume of a 4D image in Lightbox mode


draffelt:
I was computing an image today storing the elements of a Jacobian matrix in the 4th dimension, and thought it would be nice to see all of the volumes at the same time in a 3x3 grid.  The same would apply to tensor images.

It may also be useful to see all diffusion volumes at once for checking image quality and results of pre-processing (eg failed motion correction). 

When in lightbox mode, we could possibly have a check box in the view tool that enables the user to switch to 4D, where the slice increment becomes the volume increment.

rtabbara:
Added in #694

PR #694, opened 2016-06-28
MRView: Lightbox: View volumes of 4d image

rtabbara:
As requested in #451 

A few things:
- View options tool has been updated to include a `Show 4d volumes` checkbox (when LightBox mode is active). Toggling this will change the slice increment adjust button to a corresponding volume increment spin box.
- Similar to when viewing slices, the volume indices are computed relative to the current focused "main" image. (i.e. the image that would appear in single slice mode).
- Computed volume indices are clamped to lie between 0 and n_vols - 1. In particular, this allows for the possibility of certain volumes being rendered repeatedly across lightbox views.
- For performance reasons, I had to extend `MRView::Image` (gui_image) to allow for caching of `GL::Texture` (3d textures). Otherwise, we would be in a situation where we're reloading our textures every draw call as we render each volume per view.
- Implicitly, we need a copy constructor for `GL::Texture` in order to perform a lookup from our cache. Rather than explicitly defining one, I've instead added the function `cache_copy(const Texture&t)`. I thought this was the least invasive approach, as it means the default copy constructor remains deleted and more generally won't affect other uses of `GL::Texture`.


jdtournier:
@rtabbara, just had a chance to play with this. Works pretty well, Although there are a couple of glitches:

- things go funny when at the end of the series: the latter half of the display is filled with duplicates of the last volume.

- hitting F4 (i.e. selecting lightbox mode) again causes the currently selected volume to move to the central position - even if you're already in lightbox mode. When that happens, the volume index in the View tool does not update. 

So a few suggestions: 

- the volume index in the view tool defines which volume is shown in the top left corner of the display - not the currently selected volume.

- the volume index is not allowed to increment further than num_volumes - num_panes - although we'd still need to allow for the possibility that num_volumes < num_panes, in which case the last few panes should be empty (or at least not display an image). This obviously needs careful handling when changing number of rows or columns.

- switching to lightbox mode when 'show 4D volumes' is ticked should honour the volume index entry shown in the View tool, and set the crosshairs over the currently selected volume wherever it may land on the display, without shifting that volume to the central location. The exception is when the currently selected volume is not within the range displayed, in which case you'd need to shift the volume index to ensure it was visible. The main issue here is ensuring that switching back to lightbox mode should present the same view as was presented when it was last used (i.e. hitting F1 then F4 directly, or just hitting F4 again) - assuming the user doesn't change things that would obviously require changes (i.e. currently selected volume out of currently displayed range, etc). 

What do you reckon?

rtabbara:
@jdtournier Thanks for looking at it. All suggestions seem reasonable. I was originally clamping volumes to lie between valid indices which is why you were seeing duplicates, but I think your suggestion of rendering blank panes is clearer.

rtabbara:
@jdtournier Sorry, have neglected this - can't believe a year has passed since this PR! Where does the time go? Anyway, I'll incorporate the suggested changes soon and people can take another look.

rtabbara:
OK, I've updated this (in record breaking time too). I think it does what @jdtournier had in mind, but please have a look.

The one small nuance: If a user is viewing volumes in lightbox mode and then toggles between volumes (i.e. hitting left or right key), then when switching back to single slice the 'original' volume will have also changed. In other words, the original volume is always considered to be whatever is in the top-left pane of the lightbox. I *think* this is the right behaviour, but let me know otherwise.

jdtournier:
I like... :wink: 

screenshot from 2017-07-20 11-49-18

Minor comments:

- maybe we should rename 'show 4D volumes' to 'cycle through volumes' or something like that - I find the current wording a little cryptic...

- currently, switching back to single slice view switches back to the volume shown in the top-left corner. I think it would be more intuitive to switch back to the volume where the crosshairs are currently located, which is also the volume whose position is shown in the bottom corner. This would probably require storing the index of the first volume shown for each image, so that switching back to light box mode can resume the same state as before. I'm not sure it's absolutely critical though, more of a wishlist item. But if you have a minute... :wink: Doing it right will probably require storing the index of the first volume in the Image class itself, I reckon.

Otherwise, it looks good, and seems to work OK. :+1:

jdtournier:
One more little glitch I can't quite figure out: the auto-rescale (Esc key) sometimes doesn't behave as expected. I'd expect it to reset based on the volume currently under focus (in the crosshairs), but it looks like it's using the first volume regardless. And I mean the first volume of the whole series, not the one shown in the top left. Even if I have only b≠0 volumes shown, pressing Esc reverts to the original scaling (presumably based on the first b=0 volume). 

This also fails when switching between a DWI series (with 'cycle through volumes' ticked), then switching to a single-volume anatomical: if I then press Esc, the windowing becomes undefined (the entries in the View tool become empty). It sorts itself out by switching over to single slice view and back to lightbox, but it would be good to fix this...

rtabbara:
@jdtournier ok, think this is finally working as intended. Wasn't a fan of caching volume state info in the Image class solely for the purpose of the light box, so instead I'm simply caching the index of the selected pane in the LightBox class itself so it persists as you toggle between different view modes. Two special cases:

* for light box panes at indices < selected pane index, ensure calculated volume index is non-negative. i.e. guarantee first pane shows volume index >= 0

* if selected pane index corresponds to a volume index > n_vols -1 (e.g. after a change in the light box layout), clamp the current selection to the max pane index that contains a valid volume. From the first point, we know in the worse case, the first pane will always show a valid volume and the current selection can always be >= 0.

So those are the only two cases where abrupt changes in the light box view can occur, but I *think* users will be able to follow the the logic behind it. Anyways, please check it out.

thijsdhollander:
This looks great! :+1:   I've had requests for this sort of feature from a few different people last week; many who had different use cases that may be greatly helped by this feature.  Let's get it in and release RC2.

thijsdhollander:
Maybe one potential (future) addition to this feature: it would provide additional value to have an option (e.g. check box) to see the focus cross hairs replicated for every volume.  To distinguish visually between the "real" focus (whose voxel information is shown by default at the bottom left) and the replicated ones, the replicated ones could have a different colour (could by default just be a less bright version of the real focus, e.g. RGB values divided by a certain implicit factor).  A common use case for this feature is looking at multiple contrasts at once (well, it is essentially *the* use case of this feature): it does help to be able to pinpoint a particular location across these contrasts.

Lestropie:
+1 for multi-focus. My immediate thought was to use a dashed line for the focus in the other volumes. Alternatively, use the same RGB value as the primary focus lines but a reduced alpha.

rtabbara:
@jdtournier So is this ok to merge?

Lestropie:
Changing between slice-wise and volume-wise changes behaviour when setting the focus point in an empty window port. Not clear what the behaviour *should* be though.

With cycle through volumes off, selecting a focus point in one of those ports outside the FoV gives the correct position, and the "`?`" image value. The "A/P/I/S/L/R" labels also move between ports with the focus point. With cycle through volumes on, the focus remains with the most recent "valid" (i.e. contains image data) viewport, and thus the focus lines may become invisible and the "A/P/I/S/L/R" labels remain where they were.

This is *probably* the best bet, since scrolling to non-existent volumes is not possible in any other mode and allowing such could result in the user becoming stuck in no-mans-land if they change view modes. It just struck me as "different" at first, and so thought I'd bring it up. But I suspect any alternative behaviour would be more effort than it's worth.

---

If you have cycle through volumes on, move the focus point to a different volume, then switch cycle through volumes off, the focus point doesn't quite stay exactly where it was. The magnitude of the movement appears to be related to the distance between the old and new focus points.

Example: 4x5 grid, 5mm slice spacing (2 voxels). Start in 3D mode. Switch to 4D mode, move focus to the window to the right; when you switch back to 3D mode, the voxel position of the focus point (as reported in bottom left corner) jumps up 2 voxels. Repeat again, but instead move the focus to the window above; now when you switch back to 3D mode, the voxel position of the focus point jumps down by 10 voxels (2 voxels slice difference per window * 5 windows between old and new focus points). So I presume there's some projection of the focus point into the selected plane during the 3D/4D mode change, but it's using the incorrect slice index.

rtabbara:
> If you have cycle through volumes on, move the focus point to a different volume, then switch cycle through volumes off, the focus point doesn't quite stay exactly where it was. 

@Lestropie good pick up! Should be fixed now.

jdtournier:
Just had a quick play with this - seems to work fine. I propose we merge this, it's definitely good enough for release. We can always fix minor issues as they surface. What do you all reckon?

Issue #201, opened 2015-03-31
dwi2fod & sh2amp


steso:
Not sure if it's really a bug or if the error is located somewhere else, I'm also happy to provide some example files if necessary. I generate two synthetic crossing fiber bundles based on a tensor in x and y directions (the identical tensor, just rotated 90°). The two branches have the same signal contribution (volume fraction) and there is no partial volume/up-sampling/etc involved. The diffusion signal is calculated on gradient scheme of 252 directions. If I perform an ODF reconstruction using the FRACT approach and compare the two peaks in the crossing region, I get more or less the same ODF-amplitude (<< 0.1% difference). In mrtrix I save the dwi.nii and the corresponding encoding.b file, run dwi2response & dwi2fod. If I extract the gradient amplitudes using sh2amp for 3 directions (x,y and z direction) I get about 5% difference btw. x and y.
If I take a closer look at the branches (in the non-crossing areas) I also get some negative amplitudes, probably from negative side-lobes, which are also slightly different in the horizontal vs vertical bundle.
If I use the SH-coefficients after dwi2fod and extract the amplitudes myself, I get the same values as the output from sh2amp. I'm not sure where this "directional" bias could have been introduced. There might be a small bias introduced by the initial sampling scheme, but 4-5% seems rather large to me.
By the way I'm using version 0.3.10-455.

jdtournier:
That does indeed sound fishy. Any chance you could post the exact commands you used?

jdtournier:
It would also be great if you could post or send me the details of how the input was generated...

steso:
Sure:

dwi2response "%path%/simCross_dwi.nii" "%path%/simCross_response.txt" -grad "%path%/encoding.b" -mask "%path%/simCross_mask.nii" -force -nthreads 16

dwi2fod "%path%/simCross_dwi.nii" "%path%/simCross_response.txt" -grad "%path%/encoding.b" "%path%/simCross_csd.nii" -mask "%path%/simCross_mask.nii" -nthreads 16

gendir 3 dirs.txt
sh2amp simCross_csd.nii dirs.txt amps.nii

steso:
Sure:

dwi2response "%path%/simCross_dwi.nii" "%path%/simCross_response.txt" -grad "%path%/encoding.b" -mask "%path%/simCross_mask.nii" -force -nthreads 16

[...]

PR #699, opened 2016-07-04
Improved fod integration

### LINKED ISSUE #312: FOD segmentation: Improve integral calculation
As discussed in #201; moving here so that the issue title actually reflects what needs to be done, and the label is more appropriate.

I think the most likely approach is going to be to perform something akin to the trapezoidal rule on the sphere, rather than just summing amplitudes. This should also take into account zero-crossings.

Lestropie:
This essentially fixes the orientation bias observed in the calculation of FOD integrals using the FMLS segmenter, by deriving a set of appropriate weights to apply to the sample points when summing. The sampling density is therefore returned to the 1281-direction set, avoiding the ~ 4x speed penalty observed with the temporary fix of moving to a 5000-direction set.

Closes #312.


jdtournier:
Just for my own interest: does this use the approach I'd [outlined in this post](https://github.com/MRtrix3/mrtrix3/issues/201#issuecomment-89053428)...?

Lestropie:
Close to. Rather than random SH functions, I generate the SH2A transform up to _lmax_ `LforN(num_directions)+2`, generate SH basis functions using vectors that are zero-filled except for one coefficient, and enforce that the integrals of all of those functions (when combined with a vector of sample point weighting coefficients) should be zero (except for the vector where it's the _l=0_ term that's non-zero).

Lestropie:
Chalk up another win for automated testing.

draffelt:
Thanks Trav :+1:

PR #701, opened 2016-07-05
Float type changes

Lestropie:
Promoting this to a pull request since I hate having different bits of my own code dependent on different repo branches.

Requires inspection to make sure people are happy with the changes, and testing of stats code to make sure that the results aren't drastically changed and time / RAM requirements aren't severely affected.


draffelt:
I really need to add some stats tests anyway, so this looks like a good opportunity.  However, realistically I won't have time to look into this until next week.

Lestropie:
Superseded by #785.

Issue #411, opened 2015-12-07
Track files: Implement file format handlers


Lestropie:
Much like is done for image files, it should be possible to identify different types of track data storage based on the file extension, and have the `Tractography::Reader` and `Writer` classes call relevant handler functions to extract / write data from/to the `Tractography::Streamline` class.

This would allow any streamlines-handling command to handle streamlines data stored in any format, and may also allow the `tckconvert` command functionality to be encapsulated within the `tckedit` command.

Ideally this should also allow the `Writer` class to use buffered writeback for any format.

As first mentioned in #216.

dchristiaens:
I'm all for! As a matter of fact, I had already implemented the `tckconvert` command as a bunch of Reader/Writer classes, which should be more or less ready to go in the library. The reason I didn't do so at the time, was mainly because I did not yet implement a VTKReader class. I've written code for that in python a few years ago, to import tracks from MITK-Diffusion, but found it tricky to make it compatible with different software packages.

More importantly, reading and writing to a VTK file requires to keep its contents in RAM, making them very unsuited for streamline tracking or filtering of large datasets. Therefore, I would still advise users to stick to `.tck`, even though it would be nice to support other formats transparently.

Lestropie:
Also implement support for .trk format.

Lestropie:
As discussed in #480: Also implement piping of streamlines data.

Lestropie:
Bumping this since there's a [bit of interest](http://community.mrtrix.org/t/conversion-of-streamlines-from-trk-to-tck/806). I'll hopefully get to it as part of my fixel TWI development stuff.

Lestropie:
This is going to require quite a lot of refactoring. I'd rather remove the RC4 milestone from this, and instead address it as part of my fixel TWI updates (second major update trailing #1543).

Lestropie:
[Additional feature request](https://community.mrtrix.org/t/anonymise-nifti/773/4): `tckconvert` to receive same header modification command-line options as `mrconvert`, in particular to enable anonymisation following #1603.

Lestropie:
Note that any modifications in this respect should be made with the foresight of the most general case of streamlines data storage; current candidate for a new community-driven format is frheault/tractography_file_format#1.

PR #712, opened 2016-07-08
Port to eigen 3.3

jdtournier:
Changes required to make things work with Eigen 3.3 beta (testing with Eigen 3.2.92). This basically boils down to sorting out memory alignment issues so that whatever CPU instructions are used to vectorise operations don't crash. Eigen tries to handle these issues internally as much as it can, but there is one specific set of circumstances that causes trouble: [fixed-sized vectorisable objects](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html), in combination with dynamic memory allocation. 

The main culprit for us in _MRtrix3_ seems to be `Eigen::Transform`, a.k.a. `transform_type`. Basically, the problem is that `new` allocates memory without regard for the alignment required for the object - whereas static allocation does honour the alignment specified (via constructs like [`alignas`](http://en.cppreference.com/w/cpp/language/alignas)). This isn't a problem for variable-sized matrices, since the memory to hold the contents is dynamically allocated by the Matrix constructor, which takes care of alignment. But that's not the case with fixed-sized objects like `Eigen::Vector4f`, which basically boils down to a `float[4]` static array. 

The problem is that this propagates through to any class that has such as type as a member. This includes `Header`, and anything that derives from it or contains a `Header` member, but also other classes like the interpolator classes, since they all contain `transform_type` objects. To sort this out, the simplest thing is to define a custom `operator new` that does guarantee alignment. Eigen provides a convenient macro `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` that does precisely this. But this need to be done for every class affected - even if it derives from a class that has already been 'fixed', since the base class's `operator new` will only be invoked for that exact type. 

There is also an issue with using these types (or classes that contain them) in `std::vector` arrays and other STL containers, but that's a relatively rare occurence.

So I've gone through and added this macro to just about every class that was causing problems in testing. Everything checks out as far as I can see, apart from:

- `tckconvert`: due to minor differences in precision, causing the diff on VTK output to fail - [as discussed recently on the forum](http://community.mrtrix.org/t/errors-from-run-tests-on-mrtrix-3-1-5-on-ubuntu-16-04/340?u=jdtournier);

- `dwidenoise`: should be fixed by #711, I'll merge that in later;

- `tckmap`: it's off by more than the precision, but not by a huge margin either - might be down to tiny differences in the rounding of the position of streamline vertices - worth looking into at any rate. 

While these changes sort out all issues picked up in testing, there's bound to be cases that I've missed and aren't covered by the testing. It may be that the simplest thing would simply be to mandate that all classes should include the `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` macro just to avoid future problems - I'm pretty sure all it would do is ensure dynamic allocation respects the class's stated alignment requirements, whatever they may be.

Clearly, this is going to need some thorough testing and quite a few more pairs of eyes to make sure everything does work as expected. Although this shouldn't affect operation with Eigen 3.2.8, so it might be safe to merge, and we can pick up other issues downstream as 3.3 beta users report them - not ideal, but there hopefully won't be too many users at this stage...?


Lestropie:
RE the `tckconvert` test error: If a VTK reader were implemented, the test could re-convert back to `.tck` and then use `testing_diff_tck` rather than `diff`. But personally I'd rather just do #411.

RE testing: Those who have tried the Eigen beta have encountered issues and reverted, and the install instructions for all OS's now state explicitly to avoid the beta. So we could just merge, and agree that all devs will install and use the Eigen beta, fixing and creating tests for any problems encountered.

jdtournier:
> RE the tckconvert test error: If a VTK reader were implemented, the test could re-convert back to .tck and then use testing_diff_tck rather than diff. But personally I'd rather just do #411.

I think I prefer the `tckconvert` option. Last I checked, the VTK format needs a list of vertices, then a list of primitives using the vertices. That means we can't stream the output, it all had to remain resident in RAM until processing is done, at which point it can write the data out. So to me it sounds like a bad idea to support it to the same extent as our current format, it simply won't work for large tractography runs. If it's supported via `tckconvert`, at least it's clear that it's not the recommended format, and the actual conversion is much easier to handle anyway since we can use a multi pass approach over the existing tracks file.

But all this is predicated on the notion that this is the only format VTK can handle - it may be that we don't need to write the vertices as we write the primitives, in which case this wouldn't be an issue.

> RE testing: Those who have tried the Eigen beta have encountered issues and reverted, and the install instructions for all OS's now state explicitly to avoid the beta. So we could just merge, and agree that all devs will install and use the Eigen beta, fixing and creating tests for any problems encountered.

Yes, that's more or less what leaning towards. Let's do that... but we need to sort out the tests that are currently failing first.

On a related note: the testing failure with `dwidenoise` is as far as I can tell entirely down to the output being written as 32 bit floats. If there is a rounding difference in that final double to float conversion, that translates into a relatively large error. I checked by forcing the processing to use double precision in the app itself: it didn't seem to affect the size of the differences. So basically any value that's borderline in the double to float conversion can fall either way, and the difference between these two will be ~epsilon × amplitude. I've reduced the tolerance drastically on that basis...

Lestropie:
> So basically any value that's borderline in the double to float conversion can fall either way, and the difference between these two will be ~epsilon × amplitude. I've reduced the tolerance drastically on that basis...

Would it be better for the `testing_diff_*` commands to test fractional rather than absolute differences (or offer the option to toggle between the two), given this is likely to be the tolerance limit in a number of situations?

jdtournier:
>Would it be better for the testing_diff_* commands to test fractional rather than absolute differences (or offer the option to toggle between the two), given this is likely to be the tolerance limit in a number of situations?

Yes, that's kinda what I was thinking as I was writing this. But it's not necessarily always the case. I redirect l expect difference in `tckmap` for example to be dominated by whether or not a streamline makes it into a given voxel...

jdtournier:
Sorry, just noticed you mentioned adding an option for relative tolerance. Sounds like a good idea, all for it.

thijsdhollander:
So, with respect to http://community.mrtrix.org/t/mrview-segmentationfault-core-dump-ubuntu-16-04/185/32 , I'm updating the branch now.  Anyone has any concerns with merging after the tests have finished?

Lestropie:
The plan was to merge #762 into `port_to_eigen_3.3` before merging this; @jdtournier wasn't happy with some of the testing changes, but I'd hoped that simply switching back to an absolute difference test where deemed appropriate (and maybe regenerating some pre-0.3.13 test target data) would be satisfactory. I do think we should merge these changes and all run the beta ourselves but leave the documentation advising against using the beta for a while. If you want to do that and we'll change #762 to merge to master that's fine by me.

thijsdhollander:
Good, we'll do that.  Apart from advising against using the beta (which I agree with), let's hope these changes already resolve some of the issues with the beta.  There's bound to be users out there who may be using the beta without knowing (specifically those on Ubuntu 6.04)... better to have (some) issues resolved for them than not, I guess.

thijsdhollander:
(so I didn't delete the branch yet, just in case further useful/required changes for Eigen 3.3 compatibility pop up)

PR #726, opened 2016-07-19
Minor changes to allow the website publication list generation script to work

draffelt:


Lestropie:
@draffelt Click the button if we can push this one out of the way.

Issue #642, opened 2016-06-01
Bugs in mrfilter gradient


thijsdhollander:
Some weird things going on in the gradient filter:

- Not supplying the `-stdev` option, the command complains
- Supplying `-magnitude`, the command doesn't complain about not supplying `-stdev`
- ...but supplying `-magnitude` as well as `-stdev`, the `-stdev` value gets ignored

...and maybe other stuff.  But there's clearly something wrong with the ordering of certain actions in the filter: the magnitude option/code gets considered first, and quits before the smoothing (using stdev) can even take place...

maxpietsch:
Just came across this bug and had a go at fixing it. 

Turns out there was also a [bug](https://github.com/MRtrix3/mrtrix3/commit/87663f5ac4432f56473d97e442414225e2472299#diff-c12b7db079cffcebb210a071b705b9feR115) in the smooth filter not setting the stdev values if a vector of length 3 was supplied.

maxpietsch:
Added tests and fixed another issue use of `-scanner ` not being propagated to gradient filter if `-magnitude` is set. (fc0b6a5b4489a)

thijsdhollander:
I may be completely wrong (it's been a quite while since I was looking into this stuff), but: does the latter actually matter?  As in: isn't the magnitude just the same irrespective of it being computed with respect to the scanner space versus the image space?

maxpietsch:
If you have a location in image space `x,y` then the gradient of image `u` in this space wrt x is `du(x,y) / dx` of as central differences:

    du/dx:=(u(x+h, y)-u(x-h, y))/2h

If we want to have the gradient in a scaled space x'=a*x, then the gradient in this space is

    du/dx' = du/dx * dx/dx' = 1/a * du/dx

So if the voxel spacing is not 1mm isotropic then the magnitude of the  gradients wrt image and scanner space should differ.

maxpietsch:
Besides, the main issue was that the `-scanner ` setting is off by default for `mrfilter` called without the `-magnitude` option but was on for `mrfilter -magnitude` as gradient.h is initialised with `wrt_scanner (true)`.

thijsdhollander:
Ah, yep, the units are indeed different (but the contrast shouldn't be).  Agreed it makes sense then to just make things work consistently. :+1:

PR #746, opened 2016-08-03
mrfilter: fixed gradient stdev bug #642.

maxpietsch:


maxpietsch:
Any objections against merging this?

thijsdhollander:
Didn't have time to test it, but it seems the logic for the change is sound...?  So happy to merge if you're confident that it works. ;-)

maxpietsch:
Wasn't confident so I did some more testing. I found another bug: If -magnitude is set then the gradient filter defaults to gradients wrt scanner space, without -magnitude it defaults to the opposite. (The default in `mrfilter` is not wrt scanner space).

There is a warning in gradient.h 

   if (do_wrt_scanner && magnitude)
       WARN(For a gradient magnitude image, setting gradient to scanner axes has no effect)

which I think is not necessary. I don't see why a gradient magnitude shouldn't be computed wrt to scanner space. 

So I removed the above warning and changed the behaviour of -magnitude to the one set via `.compute_wrt_scanner()` in fc0b6a5.

Also, I wrote some tests for `mrfilter gradient`. I noticed that there are 4 files (mrfilter/out8-12.mif) that are not used for any test. Does anyone use those or shall I get rid of them?

PR #751, opened 2016-08-06
Transform file handling

### LINKED ISSUE #718: mrtransform: -replace option & matrix import
1. Since the move to Eigen, `mrinfo -transform` produces a 3x4 matrix text file. But `mrtransform -linear` expects a 4x4. We should probably be able to import the transformations we're exporting.

2. I don't like how replacing the header transformation is done using `mrtransform -linear <text file> -replace`. I'd rather the `-replace` option be used independently of `-linear`. It would also then be possible for this option to take as its argument _either_ a matrix text file, _or_ a template image from which to take the header transform.

For reference, error is arising [here](https://github.com/MRtrix3/mrtrix3/blob/master/scripts/dwipreproc#L380).

Want people's opinions on this, since `mrtransform` is usually pretty contentious.

Lestropie:
Requires examination and testing, particularly by @draffelt and @maxpietsch, since this is not my area of expertise.

Went with the solutions in the `transform_file_handling` branch rather than the conflicting ones in 7be6580 since the error messages are more precise and the `dwipreproc` script fix is more direct.

Whether or not `mrinfo` should output a 3x4 or 4x4 matrix is up for debate.

Closes #718.


maxpietsch:
I think the changes are fine except for maybe one thing: `mrinfo -transform` uses commas as delimiter, `math.h`'s `save_transform` uses space. `mrinfo file.mif` also uses space. Import and export works fine with both but I would keep it consistent for the sake of easier interfacing with scripts. I would keep mrinfo output a 4x4 matrix for the same reason.

So, thumbs up from my side for merge (and feel free to revert 7d28777 if you prefer comma delimited output).

maxpietsch:
If anybody has any objections speak now or forever hold your peace ... ;-)

draffelt:
Seems reasonable to me

PR #753, opened 2016-08-10
dwipreproc: check gradient output files

maxpietsch:
`-export_grad_mrtrix` and `-export_grad_fsl` output locations are unchecked and script fails if they exist


Lestropie:
... Is it just me, or are you running `checkOutputFile()` on the _input_ options `-grad` and `-fslgrad`, rather than the _output_ options `-export_grad_*`?

maxpietsch:
It is not just you...  Sorry about that.

Issue #221, opened 2015-04-17
sinc interpolation mirrors instead of out of bounds value


thijsdhollander:
![screenshot0000](https://cloud.githubusercontent.com/assets/11952694/7199200/16f495b2-e536-11e4-89c8-90eaf2aef075.png)

Just as an example: this is an ADC map of a diffusion dataset with a "limited" field of view, regridded to the grid/field of view of another, larger, image.  The field of view of the ADC image was also tilted in comparison to the field of view of the larger image.  Sinc interpolation mirrors the image around the boundaries of the ADC image's field of view, instead of using the out of bounds value (be it 0, NaN, or anything else).

@Lestropie is aware of the issue (that's a fact, not a bug ;-) ).

PR #762, opened 2016-08-23
Testing: Check absolute or fractional differences

### LINKED ISSUE #720: testing_diff_*: Option for relative tolerance
As raised in #712.

Add option to toggle between absolute and relative tolerance in testing applications, and modify tests to use the latter where deemed appropriate.

Lestropie:
Implementing a fractional difference test actually provides some insight into the relative precision of different commands, based on the floating-point operations they perform, that weren't evident from the absolute difference tests. Adding this as a pull request since there's some testing results that individual devs may want to comment on or look into: we need to decide for each of these whether to ease the testing threshold, re-generate the target testing data using current code, or delve into where the variability has arisen from.

Current failures:

* `amp2sh`: > 1e-3 fractional difference

* `dwi2fod`: > 1e-3 fractional difference

* `dwidenoise`: >> 1e-3 fractional difference (definitely not just the 32-bit float image output write as suggested in #712)

* `fod2fixel`: Factor of 2 difference (due to me changing the scaling factor in 0.3.15), plus a fair bit of additional variability, probably due to the new numerical integration weighting.

* `mrresize`: Some failures at 1e-3 fractional difference. One may be due to alteration of sinc interpolation in #221, but the precision of all outputs are worse than I would expect given the underlying operations.

* `sh2amp`: > 1e-3 fractional difference. (Maybe this and `amp2sh` are just differences in the transform matrix inversion from GSL?)

* `sh2power`: > 1e-3 fractional difference - even though the target image is generated as part of the test and therefore not influenced by old code.

* `tckmap`: DEC and TOD outputs varying by more than I would expect.

Arose from the discussion in #712 (hence the merge targeting the `port_to_eigen_3.3` branch).

Closes #720.


jdtournier:
So I'm not sure this is necessarily as as simple as it sounds. In #712, the discussion was prompted by rounding errors in the binary to ASCII conversion, in which case the precision that matters is indeed fractional. But in many cases, operations involve a lot more data than a single value. In the list you have, most of the commands perform some kind of matrix-vector operation per voxel (or patch of voxels). The precision of any one value will depend on all the other values too, so the precision should probably be relative to the mean or max value in the vector of per-voxel outputs - but not always, for instance in `dwidenoise`, the relevant vector of outputs is the whole 5×5×5 patch that was used in the PCA. So I don't think it's all that obvious that testing for fractional error will really help. 

The current approach relies on setting the absolute precision, and that has to be based on the expected size of the signal. If you expect the signals to average ~200 (e.g. the DW signals), then an appropriate level of precision might be 1e-5 that value, i.e. around 2e-3 - depending of course on the particular operation performed and its stability. It might be too relaxed a condition for many voxels where the signal is much lower, but at least it's the same value for all the values in each operation, which I think is important. 

So I'd probably vote for the status quo. Just my 2 cents...

Lestropie:
> So I'd probably vote for the status quo. 

Or you could just use the `-abs` option to test the absolute difference in those few apps where you think that's more appropriate?

That list is just the ones that didn't behave well in these tests. In everything else, I either used `-abs` because it made perfect sense, or I used `-frac` and it made a helluva lot more sense than the old absolute threshold: previously every test had it own testing threshold depending on the data it was testing, and now the fractional threshold is more-or-less equivalent across tests.

jdtournier:
Quick thought: might be worth having an option to test for fractional differences where the threshold is set with respect to the average amplitude across volumes? This would obviously apply to 4D data only, but essentially for each voxel we'd take the mean (or max) amplitude of the image intensities, then set the tolerance threshold on the difference relative to that value. That should make the tests much better behaved for apps that operate on a vector of values per voxel, where the fractional difference might be huge for values that come out close to zero, since the noise level will be a function of the _overall_ amplitude of the inputs. This seems to apply to most of the tests mentioned above as problematic. What do you think?

Lestropie:
Yep that'll deal with most of that list.

Lestropie:
Merged partly because the test data modifications would make tests fail on master. However there's still a couple that should be looked into. `sh2power` test has been disabled because I was getting > 10% variation. `fod2fixel` test has been disabled also: I suspect something in the libraries is making the results differ between OS's.

PR #767, opened 2016-08-29
in place gaussian filter

maxpietsch:
- reduced memory consumption: does not copy the input image but smoothes it in place
- uses eigen for product of kernel and image values
- close to 2x speedup on hcp data

So far only used in the registration to reduce the memory consumption of mrregister ([forum entry](http://community.mrtrix.org/t/mrregister-error-error-allocating-memory-for-scratch-buffer-while-executing-population-template-command)). 

Replacing the default Gaussian smoothing filter in `mrfilter` with the in place filter turns out to be up to twice as fast as the previous Gaussian smoothing (8s vs 4s) for large 4D images but slightly slower for very small images (difference in the lower 10ms). 

The catch is that the smoothing dimension of the in place Gaussian adapter needs to be the same as the inner stride of the threaded loop. This makes the code a bit more verbose.

Possible improvements:
- Implement a 4D version that uses the `image.row()` method and Eigen matrix products.
- Smaller buffer size and fill it up on the fly 

Any thoughts on whether I shall replace the Gaussian adapter or on the implementation?


jdtournier:
A few comments from a quick look through:

- you seem to be running the in-place smoothing by doing a copy from an smoothing adapter onto the same image that was passed to the adapter? I can't see how this can work, you'll be modifying the values used for the next voxel along for any extent larger than 1. Am I reading this right?!?

- assuming the above is right, what I had in mind when we discussed this was that we would read all the values for one line into a pre-allocated `Eigen::VectorXd`, do the smoothing, saving the output into a _separate_ (also pre-allocated) `Eigen::VectorXd`, then write those values back into the image. That way, only two lines of values per thread needs to be loaded at any one time, rather than allocating a whole new image as is the case currently.

- the above would make the smoothing adapter redundant. Should we get rid of it altogether...? I don't think I'd recommend its use in any context, given the changes we're discussing here.

- We could potentially avoid one data copy by using the `.row()` method, but that could then only be used for direct IO images. I don't think there'sn much to lose from pre-loading into a dedicated vector anyway, performance-wise it makes sense: all the data get preloaded into L1 cache, ready for immediate use in the processing itself; and it guarantees that the subsequent operation can be vectorised with maximum efficiency by ensuring proper memory alignment (a big deal for Eigen 3.3 beta, but probably also for Eigen 3.2). 

- the above makes me wonder whether the `.row()` method is actually a good thing. It does make the code cleaner, but it will prevent some optimisations that the CPU might otherwise be able to perform. The main one is the use of vectorised instructions: at the moment, the `.row()` method will return a vector view starting from some given arbitrary memory address. There is no way for Eigen to guarantee memory alignment at compile-time, so it'll probably fall back to non-vectorised instructions to avoid the potential for runtime failures. That means a matrix-vector product operating on the output of the `.row()` method might potentially be 2-4 times slower than it would otherwise, had the row data been copied into an `Eigen::Vector` explicitly. And these explicit copies don't have much impact on the efficiency of the code anyway, since they essentially preload the data onto the CPU's closest and fastest memory bank (the L1 cache). In fact, that's exactly what most of the operations in Eigen will do anyway: preload block of data into chunks that are known to fit into L1 cache, process them, and write back. I think we need to seriously reconsider our use of `.row()`...

Over for now.
Greetings from Greece!

maxpietsch:
> you seem to be running the in-place smoothing by doing a copy from an smoothing adapter onto the same image that was passed to the adapter? I can't see how this can work, you'll be modifying the values used for the next voxel along for any extent larger than 1. Am I reading this right?!?

The idea is to copy the image line along which the smoothing happens into a [buffer](https://github.com/MRtrix3/mrtrix3/pull/767/commits/b78c7c23d76d8bdd7b1c6feac977bf08c407f76d#diff-6b33d563d3c11d7093c18214cf6da415R65). After that these voxels will be overwritten one by one by the (Threaded-)Loop. They are not read again as the smoothing is only along one dimension which has to be the same as the dimension along which the loop inner loop traverses. So no, this is also safe for smoothing with arbitrary extent. 

> values for one line into a pre-allocated Eigen::VectorXd, do the smoothing, saving the output into a separate (also pre-allocated) Eigen::VectorXd, then write those values back into the image.

This is what I do but I write into the input image avoiding the second copy into a vector. I do not allocate any image. The only allocation is a single line of voxels (buffer) once per thread.

> the above would make the smoothing adapter redundant. Should we get rid of it altogether...? I don't think I'd recommend its use in any context, given the changes we're discussing here. 

I included asserts that detect if the loop dimension is not the same as the smoothing dimension so I think it is safe to use the proposed smoothing adapter but at the expense of having to make sure the Loop is performed in the [correct dimension](https://github.com/MRtrix3/mrtrix3/pull/767/commits/b78c7c23d76d8bdd7b1c6feac977bf08c407f76d#diff-c12b7db079cffcebb210a071b705b9feR176).

> makes me wonder whether the .row() method is actually a good thing.

I was thinking of preloading all values in the 4th dimension (n volumes) via the .row()  method along a line in the direction of the smoothing (m voxels). This which would give a matrix with buffered values M nxm entries where smoothing could be performed as a matrix product of a local subset of M with the kernel. But I am not sure if preloading all volumes for a single line is a good idea as this could easily exceed the L1 cache.

jdtournier:
Ah, OK. So your Adapter is not a pure Adapter in the sense that it can be used like any other image - it relies on the order of traversal matching that smoothing axis - all makes sense now. 

Personally, I think that's probably too narrow a scope to genuinely be called an Adapter. I'd rather see all of this move directly into the Filter, which would then give you a bit more freedom in how you'd handle these issues. You could then just pre-load the line of data to be smoothed with an explicit for loop, and write back the values as you go, very much like you do now. I think that would make much more sense than this buffered adapter class. There's no need to use an Adapter for this, I think in this case it either hinders performance, and/or complicates the invoking code (i.e. the Filter). 

> I was thinking of preloading all values in the 4th dimension (n volumes) via the .row() method along a line in the direction of the smoothing (m voxels). This which would give a matrix with buffered values M nxm entries where smoothing could be performed as a matrix product of a local subset of M with the kernel. But I am not sure if preloading all volumes for a single line is a good idea as this could easily exceed the L1 cache.

So this might be a good idea, but we'd need to be really careful to make sure the potential benefits are genuinely realised. The preloading is unlikely to be a major bottleneck even if the data exceeds L1 cache - as long as the subsequent operations are performed using Eigen and all of its optimised blocking strategies. 

One thing to note is that this can be done along any of the non-smoothed directions: if smoothing along the x-axis, you could load a plane of data in the x-y plane, and perform the smoothing on a subset of that matrix, just as you were suggesting. The advantage of using this flexibility is that you could ensure the other axis used for this is that with the shortest stride, which would help a bit with throughput. 

I was originally thinking that this would be unlikely to boost performance all that much since smoothing is really basically a bunch of (short) vector-vector operations. Moving to a matrix-vector operation doesn't generally help all that much (the vector can remain in L1 cache while successive rows of the matrix are streamed in - they're only needed once). But it might very well be a significant boost in this case since the vectors are short (the kernel extent). For instance, with a kernel extent of 3, it's likely that Eigen's codepath would avoid vectorisation altogether - it's just too short an operation. But if you now make the operation a matrix-vector product, with one of the matrix dimensions being large, there's a chance it might allow Eigen to vectorise along the other (much larger) dimension of the matrix. It's not obvious that this would actually happen though, I'd expect Eigen would try to use a vectorised multiply-add, which would only make sense when applied along the (short) kernel dimension - it all depends on exactly how Eigen implements this. So it's worth a shot, but be prepared for slim to no benefits... 

Where things might get interesting is if we now try to expand out the operation as a full matrix-matrix product: one would be the plane of data, the other a band-diagonal matrix with replicates of the kernel all the way down its diagonal. That would allow the full range of optimisations that Eigen is capable of - but only if we manage to preserve the sparsity of the smoothing matrix (which will be zero away from the diagonals). We could try using Eigen's sparse matrix class to store the kernel (also using its symmetry for further compression), and see whether that gives any further performance improvements... Interested...?

thijsdhollander:
> Ah, OK. So your Adapter is not a pure Adapter in the sense that it can be used like any other image - it relies on the order of traversal matching that smoothing axis - all makes sense now.
Personally, I think that's probably too narrow a scope to genuinely be called an Adapter. I'd rather see all of this move directly into the Filter...

Yep, I agree as well with this logic... an Adapter definitely should be able to be used like any other image, without any preconditions.  A Filter, on the other hand, can take care of these things internally anyway.

maxpietsch:
So, in an attempt to clean this up, I removed the in place smoothing adapter and implemented a 1D smoothing functor in the filter class. Nothing fancy: It preloads the line in the dimension the smoothing is applied and applies the smoothing kernel to that while writing back to the input image. I found that on clinical resolution FOD images (15 vols) smoothing is about 20% faster using a threaded copy of the input to the output and then smoothing the output in place. So I modified mrfilter to do that but it might be worth changing the smoothing filter that takes in and output images to do this instead of using the Gaussian adapter. 

Also, I have not yet changed any other code where images could be smoothed in place instead of copying them (for instance in registration).

Happy with the changes?

draffelt:
:+1:

maxpietsch:
The `mrfilter gradient` checks fail due to differences in the 5th and 6th digit. I reckon this is fine due to the changed order of traversal through the image. Shall I change the tests or is this a matter for investigation?

jdtournier:
Not sure. Has the order of traversal really changed? I thought it would be identical, just using a temporary vector rather than in place. I wouldn't expect the order to have changed for this reason alone? Are you copying into a double precision vector, maybe? That might make a difference to the output precision...?

maxpietsch:
Yes, I do copy into a double eigen vector which is then [multiplied](https://github.com/MRtrix3/mrtrix3/pull/767/commits/1b7c96f6d99d493d17cfe433130a805ea9f4e3d0#diff-6b33d563d3c11d7093c18214cf6da415L82) with the double eigen kernel. The smoothing [adapter](https://github.com/MRtrix3/mrtrix3/blob/master/lib/adapter/gaussian1D.h#L71) multiplies the image value in it's original type with a double kernel which I think upgrades the image value to double precision. The result is then stored it in the input image's datatype.

I am pretty sure that the order of traversal has changed because the new in place smoothing filter has to use the dimension of smoothing as the inner loop stride to be able to buffer the data without overwriting it in another thread. The smoothing adapter uses the same strides for all smoothing dimensions.

jdtournier:
I'll not sure that the order of traversal has changed. Sure, the strides were changed from input to output to maximise throughout for the next stage, but each line of data still ought to be processed in the same order, right? That's the only thing that matters, since each line gets processed independently - unless I'm missing something...?

I think the promotion to double is the most likely source of difference. In which case, personally, I'd say the test data ought to be replaced, since the new version is going to be more accurate... You'd need to double-check this very carefully before doing that...

jdtournier:
Sorry, just reread your post more carefully. I agree that if the kernel is stored in double precision, the images values ought to be promoted to double precision too. So you're right, this is unlikely to be the issue...

But order of traversal for a single pass is unlikely to have any effect, certainly not to this level of precision. However, I _think_ maybe you meant that the axes get looped over in a different order? As in, smoothing happens along x, then y, then z in one case, and maybe along y, then z, then x in the other? That would potentially introduce differences, since intermediate results will be stored in whatever precision was for the original image. Is this what you meant...?

maxpietsch:
Ok, turns out, Donald's hunch about Eigen's dot product using different operations than manually looping over the array was correct. I get zero differences between new and old smoothing filter output if I set [this](https://github.com/MRtrix3/mrtrix3/commit/1b7c96f6d99d493d17cfe433130a805ea9f4e3d0#diff-c12b7db079cffcebb210a071b705b9feR259) line to nan and let the manual loop do it's thing.

maxpietsch:
I updated the test data. As an example, the difference for the first mrfilter test (`mrfilter dwi.mif smooth`) looks like this:

out1_diff0000

There seems to be a slight bias:
mrstats min, max, mean output of (new image  - old image):

    -3.05176e-05	6.10352e-05	6.76402e-06
    -6.10352e-05	6.10352e-05	6.35783e-06
    -7.62939e-06	7.62939e-06	1.09937e-06
    -3.8147e-06	1.14441e-05	8.23427e-07
    -3.8147e-06	7.62939e-06	1.09054e-06
    -3.8147e-06	5.72205e-06	6.53444e-07
    -3.8147e-06	1.14441e-05	9.64712e-07
    -3.8147e-06	7.62939e-06	1.08613e-06
    -3.8147e-06	1.14441e-05	1.04419e-06
    -3.8147e-06	7.62939e-06	7.92521e-07
    -3.8147e-06	7.62939e-06	8.36673e-07
    -7.62939e-06	7.62939e-06	1.32455e-06
    -6.10352e-05	9.15527e-05	7.39098e-06
    -3.8147e-06	7.62939e-06	1.11041e-06
    -3.8147e-06	1.14441e-05	1.05743e-06
    -3.8147e-06	7.62939e-06	8.91862e-07
    -7.62939e-06	7.62939e-06	7.96936e-07
    -3.8147e-06	5.72205e-06	7.61615e-07
    -3.8147e-06	7.62939e-06	9.29391e-07
    -3.8147e-06	7.62939e-06	8.3888e-07
    -3.8147e-06	7.62939e-06	9.7575e-07
    ...

Shall I merge these changes?

draffelt:
Happy with this. Are you able to merge it into tag 0.3.16?

maxpietsch:
Closing this as I merged it into tag 0.3.16 8afddf4194c3b4255b.

Issue #615, opened 2016-05-17
dwi2mask improvements


Lestropie:
A lot of people still encounter grief at this step; particularly as they fail to check their data.

One option that @draffelt posed was to do a non-linear registration to a manually masked template _b=0_ image, then propagate the mask back to subject space. Would take longer (seem to recall ~30s being quoted), and would only work for human images, but might be fairly robust.

There are various other ideas that could be tried, but this one could potentially be done for release.

thijsdhollander:
I actually find it works fairly well, but provided the data have been corrected for bias fields (sometimes makes quite a big difference).  Overall, I never really see gaps in or bites out of the intended mask.  The only "fails" are mostly added pieces (but only when there's an artefact outside of the brain having high intensities) or an eye (or two :-)).

The thing with the registration approach is, for it to actually work robustly (with this very application in mind), you need... a decent brain mask. *rimshot* :smile:

I still think, all in all, the current approach is most likely to give the best boundary definition for most of the brain's circumference...  But maybe we need to document the fact that bias field correction is of key importance, and those occasional added pieces are a thing.  For group wise analyses, using the cross section of masks (when building templates, but also for other steps thereafter) basically tackles this particular issue.

Actually, now that I'm thinking of it, to get rid of the bulk of those annoying added pieces, that are connected via that typically "thin bridge" to the actual mask, couldn't we do something like (everyone switch on your sense of appreciation for ad hoc strategies at this stage):

1. erode the mask by an **extent**
2. select all components but the biggest one
3. dilate these components again by the **extent** of (1)
4. remove the voxels outputted by step (3) from the *original* mask
5. select the biggest component from the output of step (4) ... this because the erosion followed by dilation is still a morphological opening operator, and will leave a bit of extra voxels.

The challenge here is a good default for "**extent**".  Probably a very small number is more than enough.  If the bridge is too wide, it also indicates a wide bridge of high intensities, so the registration strategy would probably fail miserably anyway, due to the lack of a (good) mask.

I might give this a shot when I find a bit of time... I'll self-assign for now.

thijsdhollander:
Ok, I was feeling rather inventive today, so I modified and extended elements of the aforementioned approach to a fast iterative multi-bridge multi-crap-to-bridge-ratio multi-crap-scale crap-removal algorithm.

[...]

PR #769, opened 2016-09-01
New mask cleaning filter

thijsdhollander:
Implementation of a strategy akin to that shown in #615 .

### DIFF
--- a/cmd/dwi2mask.cpp
+++ b/cmd/dwi2mask.cpp
@@ -17,19 +17,25 @@
 #include "command.h"
 #include "image.h"
 #include "filter/dwi_brain_mask.h"
-
-
+#include "filter/mask_clean.h"
 
 using namespace MR;
 using namespace App;
 
+#define DEFAULT_CLEAN_SCALE 2
+
 void usage () {
-  AUTHOR = "David Raffelt (david.raffelt@florey.edu.au)";
+  AUTHOR = "David Raffelt (david.raffelt@florey.edu.au) and Thijs Dhollander (thijs.dhollander@gmail.com)";
 
 DESCRIPTION
-  + "Generates an whole brain mask from a DWI image."
+  + "Generates an whole brain mask from a DWI image. "
     "All diffusion weighted and b=0 volumes are used to "
-    "obtain a mask that includes both brain tissue and CSF.";
+    "obtain a mask that includes both brain tissue and CSF. "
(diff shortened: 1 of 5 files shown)


draffelt:
Thijs, just wondering if this is robust enough to have it performed automatically at the end of dwi2mask? Or have you not tested it on enough data to know?

thijsdhollander:
(so, for reference, as we just discussed: the answer seems to be a quite safe "yes"; I'll wait for a little more testing and if positive, add it at the end of `dwi2mask` too)

jdtournier:
> just wondering if this is robust enough to have it performed automatically at the end of dwi2mask? 

I'm assuming this approach would only work for human (adult) brains? Should we provide this as an optional step? Or (probably better) provide an option to opt out? Or probably better again, provide an option to provide a custom template brain & mask so the approach could be trivially extended to other species or neonates, etc?

thijsdhollander:
Slight confusion here: so it's the technique I described a bit further in that issue (not the initial proposed registration-to-template approach).  It should generalise for a range of non-adult and/or non-human cases; but not perse all exotic cases that one may run into.  For any case in the realm of the exotic stuff, one should of course consider whether the basic current `dwi2mask` approach holds up any way.

But apart from that: even if we would add the step as a default extra step, there will definitely be an opt-out option (the one that will provide/change the default parameter of the method; if set to 0, it would be opt-out).

jdtournier:
Ah OK, this is purely about the removing crap connected via thin bridges approach. All good, sounds perfectly sensible to me!

thijsdhollander:
I've done quite a bit more testing on a wide range of data today; all looks good, and setting the parameter by default to 2 seems to be pretty safe for typical resolutions.  I've gone ahead and added the proposed changes to `dwi2mask` too; but as mentioned, there's an easy opt-out way as well.

It's in principle ready to be merged.  Feel free to test in the mean time; if no distressing findings pop up, I'll eventually merge it. :-)

draffelt:
:+1:

thijsdhollander:
Ok, did some final testing; nothing distressing so far for human subjects.  If (not) needed for other/specific applications, there's always opt-out.  I'll merge for now.  If something weird does turn up for anyone, we can always make it opt-in (rather than opt-out) in the future; it would only be a matter of changing the default value "2" in `dwi2mask` to "0".

PR #770, opened 2016-09-01
Add Jelle's new paper to command docs and website.

dchristiaens:
This updates the references in `dwidenoise` and adds them to the website.


dchristiaens:
Merged since it's only docs...

PR #772, opened 2016-09-04
New dwi2response algo

thijsdhollander:
Addition of...

> Dhollander, T.; Raffelt, D. & Connelly, A. Unsupervised 3-tissue response function estimation from single-shell or multi-shell diffusion MR data without a co-registered T1 image. ISMRM Workshop on Breaking the Barriers of Diffusion MRI, 2016, 5

...to `dwi2response`.  Got quite a few different people who were already eager to start using it for a range of different reasons.

(took the opportunity to fix a typo, and email address and commit documentation updates after running the doc script while I was at it too)


thijsdhollander:
We also took the opportunity to describe Dave's basic `dwi2mask` strategy while we were at that abstract; so as to have a reference and description for it out there.

Lestropie:
@thijsdhollander Add to [docs](http://mrtrix.readthedocs.io/en/latest/concepts/response_function_estimation.html)?

thijsdhollander:
Yep, it was definitely already on my todo-list!  I'm still being slightly careful about certain aspects in it; at the moment I see it as being in a very advanced beta-stage and I'm still getting input from users here and there on particular kinds of data.  But I'll definitely add a good description soon.

Issue #375, opened 2015-10-06
mrfilter removes gradient table


bjeurissen:
Filtering a DWI image removes the gradient table from the header which prevents subsequent analysis.

jdtournier:
I'm assuming this is using the version on `master`?  Chances are the `updated_syntax` version will leave it as-is (since there is no implicit parsing of the DW gradient table any more, it's just treated as a generic header entry). If not, then we'd need to check which `Header` is used as the template for the output image.

bjeurissen:
Indeed using `master`

jdtournier:
Sorry, could you confirm whether `updated_syntax` is affected...?

thijsdhollander:
Slightly in this context: some commands that take a DWI dataset and output "something else" (tensors, FODs, ...) do keep the gradient table.  Does this make sense?

thijsdhollander:
> Sorry, could you confirm whether `updated_syntax` is affected...?

Did a quick check using `mrfilter` with a smoothing filter and all default values.  The gradient table is preserved in `updated_syntax` in this scenario.

Lestropie:
> Slightly in this context: some commands that take a DWI dataset and output "something else" (tensors, FODs, ...) do keep the gradient table. Does this make sense?

We had this argument somewhere (can't find it). I thought not. I lost :-(

draffelt:
> Slightly in this context: some commands that take a DWI dataset and output "something else" (tensors, FODs, ...) do keep the gradient table. Does this make sense?

My preference would be to not include the gradient table in the FODs or tensor files. Mainly because I don't see any benefit of having them there, but also because it gets a little confusing performing downstream operations. 

For example, what if you want to truncate the SH series? Currently `mrconvert -coord 3 0:14` will automatically truncate the DW directions (assuming it is a DWI). 

Also, what do we do when transforming FOD images. Apply the same linear affine transform (which will be the wrong thing to do if the transform is non-rigid)? Remove the gradients directions since they are definitely not related to the reoriented FOD images? Or just ignore them.

jdtournier:
I have a feeling I'm about to lose again - not sure I like this democracy thing...

[...]

PR #774, opened 2016-09-05
New function DWI::stash_DW_scheme()

Lestropie:
Finishing off #375 

Took the liberty of ensuring that it's the gradient table _used_ in processing that gets captured in the header, not just what was stored in the header (which could have been either absent or over-ruled by providing such information at the command-line).

Alternative function / keyval naming suggestions welcome.

### DIFF
--- a/cmd/amp2sh.cpp
+++ b/cmd/amp2sh.cpp
@@ -229,6 +229,8 @@ void run ()
         dirs(i/2, 0) = dir_vector[i];
         dirs(i/2, 1) = dir_vector[i+1];
       }
+      header.keyval()["basis_directions"] = hit->second;
+      header.keyval().erase (hit);
     } 
     else {
       auto grad = DWI::get_valid_DW_scheme (amp);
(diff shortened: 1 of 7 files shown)


jdtournier:
Sounds good. Not sure about the key `basis_dw_scheme` though: the word 'basis' has other connotations, and doesn't really convey the sense that these were the directions used when processing the images. I'd rather use something like `dw_scheme_as_processed` or similar...

thijsdhollander:
`original_dw_scheme`, or maybe `acquired_dw_scheme` ?  I agree with "basis" potentially being confusing.  I may be missing the message/point of `dw_scheme_as_processed` as well though: that sounds as if the DW scheme has been processed...? :confused:

Lestropie:
`utilized_dw_scheme`? Or I'm happy with `original_dw_scheme`.

jdtournier:
Yes, `dw_scheme_as_processed` might give slightly the wrong impression. I was shooting for `dw_scheme_as_used_in_processing`, but that's getting a little long-winded... Rob's `utilised_dw_scheme` is closer to the mark. How about `dw_scheme_used`...?

Lestropie:
`prior_dw_scheme`?

jdtournier:
Happy with `prior_dw_scheme`.

thijsdhollander:
Sound good to me too. :+1:

PR #775, opened 2016-09-05
More verbose error for unknown image format on creation

Lestropie:
Following comment [here](http://community.mrtrix.org/t/mrconvert-error-unknown-format-for-image-home-output-template-wm-mask/424/3).

Any complaints?


jdtournier:
Nope, go for it.

PR #776, opened 2016-09-06
Explicit system signal handlers

### LINKED ISSUE #758: Catching GLIBC errors in scripts
Leading on from posts [here](http://community.mrtrix.org/t/error-response-function/393) and [here](http://community.mrtrix.org/t/error-of-running-population-template/395).

On running out of disk space (`Bus Error (core dumped)`), `libc` writes the error to the current `tty` rather than `stderr`; consequently `subprocess` can't catch it, the error disappears into the ether and people get confused.

Unfortunately the [reported solution](https://sourceware.org/ml/libc-help/2014-07/msg00008.html) doesn't work on my system: Confirmed with `VAR (std::getenv ("LIBC_FATAL_STDERR_"));` within `mrconvert` that the variable is being set, but the error still doesn't go to `stderr`. Set both at compile and run time. Baffled as to why this isn't working.

Lestropie:
Awaiting confirmation from the supreme allied commander @jdtournier that we're going to go with this solution, and testing on Mac.

Unfortunately the handling doesn't work on my Windows laptop; hence testing the capability in `configure` and only enabling the feature if an environment variable is explicitly set (so compilation doesn't fail on such systems if the update is pulled but `configure` is not re-run).

Closes #758.


jdtournier:
I'll take a look when I get the chance, but it might be a while before I
find the time. On my way back from Greece right now, and I'll have a fair
backlog to get through once back at my desk...

Can you explain what's not working on Windows? Does it fail to compile, or
does it fail to handle signals as expected at runtime?

And why the need to re-configure? Is it just to enable the feature, or is
there some actual configuration required?

Lestropie:
> Can you explain what's not working on Windows?

It claims that some of the structures /  functions used in constructing the handlers are not defined, despite having found the appropriate system header file. But there has been other weirdness associated with my Windows install so who knows if it's just me.

> And why the need to re-configure?

If certain systems (like mine) don't provide adequate support for constructing the signal handlers, then pulling such an update would cause compilation to fail, which would result in a lot of angry discussion forum posts. This is avoided by testing for support of the functionality in `configure`, and (if successful) adding a new environment variable `MRTRIX_SIGNAL_HANDLING` to the config file that triggers compilation of the relevant code. So if you `git pull` but don't re-run `configure`, this variable won't be defined, the relevant code won't be included in compilation, and the default handlers will remain in place.

jdtournier:
OK, sure, it it fails to compile on certain platforms, you'd need to deal with that. But wouldn't it be possible to prevent the use of signal handlers on Windows while we figure out how to get it working, and assume it's always supported on Linux and MacOSX? In which case all you'd need is the already-defined `MRTRIX_WINDOWS` macro...?

Lestropie:
Possibly. But I don't know whether or not it works on different versions of Windows (I already know that the same MSYS installer file results in different installations on Windows 7 and Windows 8.1), it hasn't been tested on Mac, so I went for the direct testing approach without assumptions. If you'd rather do more widespread testing in the hope of not depending on a `configure` re-run we can do that.

jdtournier:
Yes, it would be worth checking on MacOSX - I don't expect any problems particularly, it's a UNIX system. Windows on the other hand will rely on some serious MinGW boilerplate code...

Lestropie:
Did some more tweaking on this since I want to include it in the upcoming tag. No longer requires re-running `configure`. The `#ifdef` usage may be rather liberal, but I'd rather avoid the risk of breaking compilation for someone because of a missing symbol.

Remaining question is to do with standalone installations. While the signal symbol names should be consistent between systems, the integer codes aren't guaranteed to be. Therefore, if compiling on one system and running on another, incorrect error messages may be displayed. I could either add an environment variable flag to `configure` to indicate to not use signal handlers, or just add a warning to the standalone installation docs that the warnings may be wrong (they do also provide the integer signal code).

jdtournier:
OK, finally had a look into this. Had to fix the [flag check in the handler method](https://github.com/MRtrix3/mrtrix3/blob/152bbfcf3c808a561c141adf6955156ee6805902/lib/signal_handler.cpp#L172), since that only allowed execution if the flag was already set - meant the handler would only respond if it received a signal _twice_. Everything else is cosmetic, just trying to keep things clean and simple. See what you think.

Still needs testing on MacOSX and Windows. @maxpietsch, @thijsdhollander, or @chunhungyeh, would be good to have the OK for MacOSX if you get the chance to try it out. Just need to know that it compiles, catches a keyboard interrupt and in that case appropriately deletes temporary files - basically run something like `mrconvert -debug bigfile.mig -` and Ctrl-C halfway through, then check that the temporary file has indeed been deleted. @Lestropie, same for Windows I guess. 

One final comment:

> Remaining question is to do with standalone installations. While the signal symbol names should be consistent between systems, the integer codes aren't guaranteed to be. Therefore, if compiling on one system and running on another, incorrect error messages may be displayed. 

Nothing to be worried about there. If an executable can run on a different system, that means the system should behave identically - there's no way they can change things as fundamental as signal codes without breaking a whole lot of stuff... For instance, the executable wouldn't know whether it had just received a keyboard interrupt or a segfault - and it may chose to handle these very differently. While I agree these signal may differ across different OS's, they'll be invariant across all _flavours_ of the same OS.

Lestropie:
All looks good. Compiles on Windows, but catching Ctrl-C is more difficult on Windows since it's handled completely differently, and I'm not yet sure how much a MinGW toolchain alters that.

I thought I'd closed this PR: plan is to merge as part of #785, since it'll alter the 'interface' (even if it's just error reporting). So if you're happy it can be merged back across to there.

jdtournier:
Sure, happy to merge onto the `0.3.16` branch rather than master if you prefer. Let's wait until it's been tested on MacOSX first, then we can merge across. Not sure what you mean about it changing the interface though...?

Lestropie:
> Not sure what you mean about it changing the interface though...?

As in, if the user's going to start getting different error messages in their terminal when they Ctrl-C, I think that belongs in a major update where they're aware that there may be visible changes in the software, as opposed to 'sneaking in' during a seemingly innocuous update.

jdtournier:
> As in, if the user's going to start getting different error messages in their terminal when they Ctrl-C, I think that belongs in a major update where they're aware that there may be visible changes in the software, as opposed to 'sneaking in' during a seemingly innocuous update.

Ah, OK. I don't see that as an interface change... All it changes is one message in the rare instance when users interrupt their command or some other issue occurs. We routinely 'sneak in' much bigger changes than that... As far as I'm concerned, this is a totally innocuous update - the additional code only gets run when an otherwise fatal interrupt is caught. There is some scope for issues at initialisation when setting up the handlers, and when adding temp files to the to-be-deleted list, but these are very well-defined and I can't see them causing any particular trouble...

jdtournier:
OK, just had a go on Windows. It compiles fine, but doesn't catch a Ctrl-C interrupt on my system. Not sure why. But at least it doesn't interfere with normal function. We could disable this on Windows, leave it as-is with the hope that this is a temporary glitch in MinGW that will eventually get fixed by itself, or try to use a more [Windows-specific handling mechanism like SEH](http://www.programmingunlimited.net/siteexec/content.cgi?page=mingw-seh)...

Lestropie:
Doubt it'll get 'fixed' based on the reading I did at the time. Full port to e.g. SEH is probably overkill, and may limit installation options if people go for a non-MSYS2 route. It _should_ only be a subset of signals that aren't currently caught on Windows. For Ctrl-C specifically, I reckon [this](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686016(v=vs.85).aspx) might be the way to go.

jdtournier:
Just thought I'd update this thread with what I've been trying. The hurdle right now is getting this to work on Windows. It's not a deal-breaker since it doesn't interfere with current behaviour, and basically doesn't change anything - temporaries don't get cleaned just as is the case now... But it would be nice to get it to work. 

So I've tried @Lestropie's [`SetConsoleCtrlHandler`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686016%28v=vs.85%29.aspx) suggestion, but it still doesn't work... I've written a small example and [posted that on the MSYS2 users mailing list](https://sourceforge.net/p/msys2/mailman/message/35525378/), see whether there is a known solution for this. 

Otherwise, we're still waiting on confirmation that this works on MacOSX... Anyone....? :wink:

Lestropie:
Sounds like on Windows it's the [terminal emulator gobbling up the signal](https://sourceforge.net/p/msys2/tickets/135/), so not something we're likely to have a quick solution to. But the code might as well sit there just in case.

I might merge this into `tag_0.3.16`, and that way it'll get tested on Mac.

jdtournier:
Not sure it's simple as the terminal eating it up. Git manages to catch the interrupt (says `Killed by signal 2.` on the terminal). A quick google search shows [this is how they handle things](https://git.kaarsemaker.net/git/commit/176478a8bd028a3877d2be49a1e7a49386c095ce/). A bit of a hack, but it seems to work... Worth investigating?

Lestropie:
I'd seen more abstract descriptions of that type of solution. But because we're not waiting on user input via `fgetc()` or similar, you have to pull some trick whereby you spawn a new thread to run your actual program, and have the primary process just monitoring keyboard inputs, with appropriate setup such that Ctrl-C is sent as keyboard input rather than a signal. Sounds to me like too much work for something that's a new and non-critical feature.

P.S. When you say "Git catches the interrupt", do you mean `git`, or the "Git for Bash" terminal? My link suggests that it's an issue that's common to a lot of terminal emulators built around Cygwin, but isn't necessarily guaranteed to be an issue for all of them - hence why nobody's taken responsibility and hunted down the source of the issue.

jdtournier:
OK, clearly you've had a good look through this, it's a lot more complicated than it looks... Let's shelve and let them sort this out in due course. 

> P.S. When you say "Git catches the interrupt", do you mean git, or the "Git for Bash" terminal?

That was `git` within the MSYS2 default terminal. Haven't checked the 'Git for Bash' (did you mean 'Git for Windows'?) one, but I'd expect it would behave just the same. My little test command did anyway, in that it didn't catch the interrupt in either terminals...

Anyway, let's call it a day for now...

Lestropie:
Ah. I think the download is called "Git for Windows", but the shortcut you get in the Start Menu gets called "Git Bash". Whatevs.

Issue #757, opened 2016-08-17
Indiced beyond range provided to mrconvert -coord result in segfault


thijsdhollander:
Just ran into this in the morning: when running `mrconvert` with `-coord axis index/indices`, if the specified index (or one of the indices) is out of range, this seems to not be checked for.  This results with very high probability in a segfault (or maybe otherwise, unexpected behaviour).

jdtournier:
:grimacing: ahem, very shoddy programming indeed...

thijsdhollander:
Resolved by #789 ?

jdtournier:
Yes, sorry, got distracted just at the wrong moment...

jdtournier:
Sorry, thought this was a different issue! Far too early in the morning to be making sensible decisions...

PR #789, opened 2016-10-06
mrconvert: fix segfault for out of range coord

maxpietsch:
check if maximum coordinate position is within range


Lestropie:
Since the coordinates are a sequence of ints (and therefore non-negativity can't be tested at the command-line parsing stage), maybe test for non-negativity here too?

maxpietsch:
Oops, wrong button... I'll add a non-negativity test too.

thijsdhollander:
This resolves #757 right?  If so, you can happily close that one! :+1:

PR #792, opened 2016-10-07
mrmath: fix bug in handling of cross-image operations

### LINKED ISSUE #790: mrmath with multiple inputs has issues
As posted on the community forum in http://community.mrtrix.org/t/intersection-mask-using-mrmath-min/493 :

> I am performing the fixel-based analysis for my dataset, but have a question for calculating the intersection mask...
On the information page the following mrmath command is mentioned with 'min' option:
mrmath input_all_warped_masks_multiple_inputs min output_template_mask_intersection
However, if I inspect this intersection image, it rather looks like a sum image (the 'sum' option creates exactly the same image by the way)...
Whereas if I use the option 'product' instead of 'min', this rather looks like the intersecting mask that is appropriate in all subjects... 

I tested the same scenario and can confirm this.  This is especially problematic in the context of (fixel-based) analysis, where people seem to use this construct for intersecting masks.  (I hope we might not have been doing that ourselves potentially...?)

Unrelated to this bug, but related to the community post, @draffelt : it seems people do still run into some confusions related to crucial aspects of correctly analysing FD/FC/FDC.  Maybe worthwhile checking if some of those aspects can be documented better; so people aren't making crucial mistakes. :-)

jdtournier:
Fixes #790.

### DIFF
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -416,12 +416,13 @@ void run ()
       case 2:  kernel.reset (new ImageKernel<Sum>     (header)); break;
       case 3:  kernel.reset (new ImageKernel<Product> (header)); break;
       case 4:  kernel.reset (new ImageKernel<RMS>     (header)); break;
-      case 5:  kernel.reset (new ImageKernel<Var>     (header)); break;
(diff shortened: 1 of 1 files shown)


maxpietsch:
Mea culpa...

PR #795, opened 2016-10-12
derefence symbolic links in script input

maxpietsch:
Changed copy commands for `dwiintensitynorm` and `blend` from cp to cp -L as copying symbolic relative links breaks them.

### DIFF
--- a/scripts/blend
+++ b/scripts/blend
@@ -24,13 +24,13 @@ total_num_output_frames = len(file_list1) + len(file_list2) - num_overlap
 for i in range(total_num_output_frames):
     file_name = 'frame' + '%0*d' % (5, i) + '.png'
     if (i <= (len(file_list1) - num_overlap)):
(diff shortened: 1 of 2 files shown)


Lestropie:
Bit late to the party, but for Python scripts I still find seeing `runCommand('cp ...')` quite unusual. Using e.g. `shutil` feels more Pythonic; but then again, it means that `-continue` won't pick it up...

Alternatively, there is the possibility for providing wrappers to things like file move / copy in `scripts/lib/`, so that they can be logged / printed to terminal / used to trigger `-continue`. Think I was going to list it as an issue at one point but ... didn't. Thoughts anyone?

draffelt:
In the `fixel_format` branch I've recently made population_template cross platform by converting copies and moves to use shutil (in addition to using os.path.join for all paths). I didn't think about the `-continue` issue. I guess wrappers in the library would be handy.

PR #799, opened 2016-10-17
Stats field exception

Lestropie:
If `mrstats` is run using the `-output` option with an empty mask (e.g. using `getImageStat()` in a Python script), throw an exception rather than returning with no error code and an empty `stdout`.

### DIFF
--- a/docs/installation/windows_install.rst
+++ b/docs/installation/windows_install.rst
@@ -73,15 +73,34 @@ Install *MRtrix3* dependencies
 
 1. From the **'MinGW-w64 Win64 Shell'** run:
 
-   ::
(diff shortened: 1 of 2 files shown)


thijsdhollander:
The exception is indeed welcome for all but the `count` options, but I wonder about the `count` option also returning an empty `stdout`: wouldn't it make more sense to explicitly return 0 to `stdout` here?  It's basically the only statistic that does have a sensible definition in this context (and we could make good use of that in the scripts).

Lestropie:
Yeah good call.

thijsdhollander:
:+1:

Issue #805, opened 2016-10-28
-march not a flag for PPC arch


maedoc:
I'm trying to compile MRtrix3 0.3.15 on a PowerPC architecture, but GCC doesn't accept a `-march` flag for that architecture (cf. `man g++`), so `configure` fails. I've commented out the relevant section, but it might be nice to allow user to opt out of the architecture spec by setting the env var to empty string, e.g.

``` bash
ARCH="" ./configure -nogui
```

If desired I can resubmit this change as a PR.

maedoc:
Also, there seems to be a curious hard dependency on x86 intrinsics in `lib/command.h`.

jdtournier:
Wow, last time I came across one of these must be almost 10 years ago... Had no idea they were still going!

> If desired I can resubmit this change as a PR.

Sounds perfectly sensible, go for it.

> Also, there seems to be a curious hard dependency on x86 intrinsics in lib/command.h.

No need to worry about this, it's currently unused, and would only get included when debugging strange issues related to [denormal operations](http://stackoverflow.com/questions/9314534/why-does-changing-0-1f-to-0-slow-down-performance-by-10x)...

maedoc:
It's a POWER8, which is ppc64le. Thanks for the other info.

PR #807, opened 2016-11-01
no march flag if ARCH is empty string

maedoc:
See my complaint in #805, this is a workaround.

### DIFF
--- a/configure
+++ b/configure
@@ -124,7 +124,8 @@ LDLIB_ARGS      The arguments expected by the linker for generating a shared lib
                 "-shared LDLIB_FLAGS OBJECTS -o LIB"
 
 ARCH            the specific CPU architecture to compile for. This variable
-                will be passed to the compiler using -march=$ARCH.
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -516,8 +517,11 @@ elif system == 'darwin':
 
 if 'ARCH' in os.environ.keys():
   march = os.environ['ARCH']
-  report ('Machine architecture set by ARCH environment variable to: ' + march + '\n')
-  cpp_flags += [ '-march='+march ]
+  if march:
```
jdtournier: OK, seems fine [according to the python docs](https://docs.python.org/3/library/stdtypes.html#truth-value-testing). Let's merge.

maedoc: If `ARCH` isn't in `os.environ` dictionary, the `else` branch executes. If it is, but is an empty string, `bool('') -> False`. This is usually idiomatic except for NumPy arrays.

PR #823, opened 2016-11-16
New fixel format

draffelt:
The new fixel format code is nearly ready to merge with the tag_0.3.16 branch. Once I finish documenting the how the new format works, and updating/improving the FBA tutorial we are pretty much good to go. 

One thing I'm in the middle of is standardising all the documentation and code to have consistent terminology. I've moved the old *.msf sparse format from `/lib/sparse/` to `/lib/sparse/legacy/`.  Then started renaming all code related to the new format, which currently uses the term "fixel" throughout, to use "sparse" instead. 

One thing I've been thinking about is the use of the term "sparse". While I appreciate the new directory-based format is generic and may be used for other non-fixel applications later on, I'm beginning to wonder if by using this name in the documentation, help pages and code we are making it unnecessarily  complicated for our 99% of our user base who uses MRtrix to study diffusion MRI. I've had a few discussions with other non-diffusion people (both here and at Stanford) and I'm yet to find someone who has been able to suggest any other non-fixel uses of a sparse image format. 

For example, it might be simpler for (particularly new) users if in the fixel command help pages and FBA documentation, we were able to describe an argument as the "output fixel directory",  then in the description link to the "fixel image format documentation", where we use the term fixel throughout (instead of sparse). Having a command called `fod2fixel`, then telling people the argument is the "output sparse image directory", is a bit confusing. Using the word "sparse" is just one more piece of non-standard terminology to understand, with little benefit for most of our users in the short term.  Fixel is already non-standard, so if we want to lower the entry barrier for new users to MRtrix, we may be better off simplifying the documentation as much as possible. 

Note that if in 10 years time someone decides to use MRtrix to code up a command that generates non-fixel sparse data using the same format, then we can always change the terminology in the documentation and code then. With the new format there is no specific file extension that we need to get right from the start.

On a related note, given my exit from Alan's lab early next year, I'm unlikely to have the time required to herd diffusion MRI researchers to agree on a standard for DWI models and publish a paper. I'm happy if someone else wants to take this on. Alternatively we could submit a short communication in a software journal describing the format with only MRtrix contributors as authors, and see if it takes off on it's own (something I would have time to do). My concern with opening it up for others to involved is that people will inevitably want to change it, and we may not want to budge given how long its taken to agree on the current format. Another solution would be to only involve key people who have been working on fixel models and the analysis of (e.g. Benoit).


draffelt:
I've been thinking about this some more. The term sparse is really something that makes sense from a technical/developer point of view. Most users would not even consider that when storing fixel data as a 4D file it's sparse. 

As computer scientists we are used to the concept of data being sparse, and being able to store (dense) representations of sparse data using libraries like Matlab's sparse matrix etc. But to expect non-technical users to understand what sparse means in the context of images, and why fixel data is sparse compared to the way normal images are stored (a single vector with, image size and strides - which most users would not understand anyway), so that they are not confused when we use the words fixel directory and sparse directory interchangeably is not ideal.

Obviously we still need to use the word sparse when outlining the motivation for the fixel file format in the documentation. By my preference is to not use it in the name of the format, command help, or API. 

I'm keen to resolve this soon so I can move onto other stuff. So speak up if you have any objections.

Lestropie:
We tried the generic 'sparse' name, and I agree it's not working. But I don't like having `lib/fixel_format` either: I think it's the only directory name with an underscore. I'd be happy with `lib/fixel/` (or even `lib/formats/fixel/`?), and the old format currently in `lib/sparse` being moved to `lib/fixel/legacy/`. I also don't like 'fixel folder', since 'folder' is very much a Windows concept/definition; but it's hard to refer to it as a 'fixel image' when really there's multiple constituent 'images'. 'Fixel directory' is probably safest even if it doesn't roll off the tongue.

Thinking about it now, moving `lib/sparse` is going to prevent opening existing old-format images, since it'll change the constituent class name key. Given usage of the format never extended beyond the first simple `FixelMetric` class (to my knowledge), might need to just disable that name key check until the format suffers an unceremonious death.

draffelt:
Ok great, unless anyone else has any issues I will make the change. I like the idea of format/fixel/ instead of fixel_format.  I've already changed the word "folder" to directory everywhere. 

I'll disable the check for the old format class name too.

thijsdhollander:
I agree 100% on the "sparse" naming being confusing for 99% of users.  With respect to naming in a context of describing/referencing the format, I'd go for "fixel format.  With respect to the actual "folder", i.e. in a context of manuals for commands etc.., I'd actually avoid both "folder" as well as "directory": they both refer to a concept that operating systems happen to implement, and that we happen to exploit for this purpose.  Not sure if you'll all be a fan of this, but let's give it a shot: what about "fixel dataset"?  Note that for images, we also don't typically say things like "FOD file", but always use the word "image" instead of file.

thijsdhollander:
Small addition: "dataset" also nicely implies that it's a set of data (well, it literally says so); i.e., that we (can) store different pieces of information (per fixel) in it.

draffelt:
>  With respect to the actual "folder", i.e. in a context of manuals for commands etc.., I'd actually avoid both "folder" as well as "directory": they both refer to a concept that operating systems happen to implement, and that we happen to exploit for this purpose. Not sure if you'll all be a fan of this, but let's give it a shot: what about "fixel dataset"? Note that for images, we also don't typically say things like "FOD file", but always use the word "image" instead of file.

While I like the idea of calling it a dataset to describe it's a collection of related files, I'm not sure it's completely unambiguous. A single image file could also be thought of as a dataset (i.e. a collection of voxel data).  

There is not the same issue with FOD _images_ vs FOD _files_, since users expect files as arguments to programs, so calling it a FOD _image_ just helps to distinguish it from other file types. Where as in the case of the fixel format, it's likely to be more clear if we explicitly tell people the argument is a fixel directory. Especially when some commands have arguments that are fixel data files within the directory.

thijsdhollander:
Yep, definitely agree: no choice is optimal I think, and directory will probably work for most people, so certainly ok with that.  The other choice is indeed important too: are you thinking of going for "fixel data files" (rather than "images")?  Images in this context may confuse users who try to open them directly in the viewer.

draffelt:
Yes, they are described as data files

jdtournier:
I agree with the above, up to a point. There's two things I'm uneasy about - but note I don't have particularly strong opinions about this:

- I'm not sure that we need to change the API itself to match the docs. It feels a bit like saying that the name of the `.mif` format is a problem when discussing DWI analysis. We should be able to discuss the steps without getting sidetracked by what specific format is being used. I appreciate things are a nowhere near that simple here, given the tight interaction between the interface that these commands provide and the specifics of the fixel/sparse format... But I see no reason why we can't simply mention somewhere that the data are stored using a specific format that we have called X, and then talk about fixel data, fixel directories, etc throughout. I'm all in favour of the docs using the word 'fixel' exclusively, never referring to the 'sparse' format, but I don't think we therefore need to also change the API itself to match?

- As to the name at the API/backend level, I think a large part of the problem here is the word 'sparse': it not a very accurate description of what it's about. The appeal is the ability to store a variable number of items of data per voxel, and to easily extend the amount and type of such data. So while I fully support dropping the word 'sparse', I think it would be better to settle on a more overt naming that more closely matches what it does, without tying it completely to fixel-based analysis. The problem is that I can't think of anything that does this particularly succinctly - apart from maybe _multi-item per-voxel (mixel)_ storage ? 😜 

As I said, I don't hold a particularly strong opinion on this, and the suggestions so far are probably just fine. I just think changing the whole API is unnecessary. However, I reckon there is some mileage in the _mixel_ suggestion. It allows the format to remain generic, it's a closer match to what it's about, and it's so close to _fixel_ that it's unlikely to create too much confusion. So while I did suggest this tongue-in-cheek, I think it is worth thinking about - if only so you can vote me down in a democratic showdown that will put Brexit and the US elections to shame...

draffelt:
Agreed that the API does not need to have the same name as what we use in the help.

In terms of mixel, I'm not sure it really is any more clear. Ball and stick has multi-items per voxel....just the same in every voxel. How about Different number of Items per voXEL (dixel). Oh, damn that's taken...this namespace is going crazy.  

For now I've just stuck to fixel in the API too, however if we decide on a better name I'm happy to change it.

jdtournier:
Yes, I agree nothing really seems to be quite right here. Let's stick to fixel for now, at least there's no scope for confusion that way...

thijsdhollander:
Even if you'd still consider it: sorry guys, but the fun police tells me "mixel" is already taken by FSL. :disappointed: 

...but Totally Random Items per voXEL (trixel) isn't yet! :raised_hands:

jdtournier:
:smile: 'totally random' sounds too much like :poop:...

How about _Variable number of Items per voXEL_ - vixel? If you say it fast enough, it even sounds like fixel... :wink:

And this seems like a pertinent time to [remind everyone of our true responsibilities](https://goo.gl/images/18dusj)...

thijsdhollander:
All the naming silliness aside: I've checked out the branch and gave it a go in a range of simple scenarios.  I must say it all feels really good.  The fixel folders are nice and organised, and the interaction via `mrview` couldn't be simpler.  You can literally load anything in the fixel tool, and you've got access to all data files to visualise stuff.  This should make certain work flows much more easy!

However, testing this, I was reminded again by an awkward aspect that we may consider fixing altogether with this pull request: `fod2fixel -afd` seems to write half of the integral... i.e., in a voxel where the DWI signal is exactly the same (in shape and size) as the response, after CSD you expect an AFD of 1 (since the "unit" is the response function); but currently this ends up being 0.5.  This is apparent when setting the windowing of the colour scale, and also after e.g. `fixel2voxel`.  This really jars when trying to explain people the interpretation of e.g. their effect sizes in terms of the model (the response function).  Also, if you do have a compartment model with "real" fractions that sum to 1, this is off-putting and confusing.

Lestropie:
> `fod2fixel -afd` seems to write half of the integral

Normalisation of the numerical integration moved from [here](https://github.com/MRtrix3/mrtrix3/commit/883edf637cc1cd88e619f314020da2f52865698a#diff-acbc601238ab089b171563ac3a851715L124) to [here](https://github.com/MRtrix3/mrtrix3/commit/883edf637cc1cd88e619f314020da2f52865698a#diff-b67e007a657ec5d1e424e202562ceff8R149). Looks like I then changed the scaling [here](https://github.com/MRtrix3/mrtrix3/commit/9546a576c3c259a35a60158d680bccffe7836641#diff-b67e007a657ec5d1e424e202562ceff8): from memory this was to make the integral numbers close (but not precisely equivalent) to the values in the test data that were generated before the weighted integral calculation was put in place. So maybe the scaling was wrong before that.

This is being pushed as a pretty big change, so mixing of data processing within a cohort would be highly discouraged, and therefore we should be able to pick what's 'right' rather than what's consistent. If you're confident that the numbers need to be doubled, I've got no problem with that going out as part of 0.3.16.

thijsdhollander:
> This is being pushed as a pretty big change, so mixing of data processing within a cohort would be highly discouraged, and therefore we should be able to pick what's 'right' rather than what's consistent. If you're confident that the numbers need to be doubled, I've got no problem with that going out as part of 0.3.16.

Definitely confident that the "right(-er)" thing is the current output times two: the fixel should have the AFD of the total fibre population, which only has an orientation (not a direction).

This is indeed why (after being reminded of it), I brought it up now: it's quite unlikely that people would be even trying to mix data processing involving fixels before and after introducing the changes in this branch, even just because of the practicalities related to the format change. (and of course, we'd discourage it, just to be clear)

thijsdhollander:
...as I was typing the above, I realised something else: shouldn't we be naming the orientations file `orientations.mif` instead of `directions.mif`?

Apart from being more correct, this may also help avoiding ambiguity when people (on the forum) will start talking about "fixel dirs" as a short for "fixel directories".

thijsdhollander:
Ok, so I've updated that fixel integral, and cleaned up / corrected some comments (thanks @Lestropie for pointing out where the relevant code was).  The only remaining comment line I'm not sure about is [here](https://github.com/MRtrix3/mrtrix3/blob/6f44eb3c80c13990fdfed72b5f11a7cf215b3f4e/src/dwi/fmls.h#L123): is this also no longer relevant...?  I'm not 100% sure what the comment exactly refers to.

Lestropie:
> The only remaining comment line I'm not sure about ...

No, still want that. The 'mean' lobe direction (as opposed to the peak direction) is initialised [here](https://github.com/MRtrix3/mrtrix3/blob/6f44eb3c80c13990fdfed72b5f11a7cf215b3f4e/src/dwi/fmls.h#L75) and updated [here](https://github.com/MRtrix3/mrtrix3/blob/6f44eb3c80c13990fdfed72b5f11a7cf215b3f4e/src/dwi/fmls.h#L100): it's basically summing vectors corresponding to the (weighted) samples within that lobe to estimate the 'mean' direction of that lobe; but we want a unit vector once we're done. More relevant comments [here](https://github.com/MRtrix3/mrtrix3/blob/6f44eb3c80c13990fdfed72b5f11a7cf215b3f4e/src/dwi/fmls.h#L39-L44) and [here](https://github.com/MRtrix3/mrtrix3/blob/6f44eb3c80c13990fdfed72b5f11a7cf215b3f4e/src/dwi/fmls.cpp#L387-L390) if you're interested.

thijsdhollander:
Ah yes, now I see how it refers to that mean orientation vector (and is not related to the integral computation), and hence the line of code just after the comment.  Makes full sense; thanks!

Issue #477, opened 2016-03-18
Global tractography threading race condition.


dchristiaens:
As reported recently, multi-threaded `tckglobal` produces funny output on some systems, that occurs if the `SpatialLock` on a proposal is not properly released. As a result, a cubical region around its position is excluded for further processing, resulting in "dark" blocks of low particle density. It only seems to occur on some systems and seems to depend on the parameter settings. In any case, the single-threaded code (enforced with option `-nthreads 1`) works fine without issues.

I'm filing a proper issue here until I can implement a fix.

Lestropie:
If I had to guess, I'd say there's a small discrepancy in the implementation of `std::set`. Your `SpatialLock::unlock()` call is relying on finding equivalence of a floating-point 3-vector to delete an entry for the set, which is best avoided - and the `erase()` call won't flag an error if it fails to delete an entry. You could either round the positions to integers and use that in the set, or you could generate a unique integer for each thread and use that in a `std::map<int, point_type>`.

bjeurissen:
I gave it a spin on my linux system, following the instructions in the workflow manual, and it does not lock but segfaults when using multi-threading (pretty much immediately after the data has been preloaded). Using single-threading it seems to run fine.

dchristiaens:
Hmm, that's not good. So there's definitely something wrong with the multithreading on non-MacOS X systems. I'm a little busy these days and won't be able to investigate this properly before ISMRM, so maybe I should add a warning in the user guide until I do?

@Lestropie, you're right, it might be more safe to unlock any point within the thresholded range. However, the points used in the locking are never subject to any calculations or changes, so I don't think it would make a difference. Or do you reckon the float could be altered upon a simple copy to the set?

Lestropie:
I certainly wouldn't expect the values to change. It's just generally considered best practise to avoid testing equality of two floating-point numbers. There's a chance that precisely how the `std::set` is implemented, in terms of sorting / detection of duplicated entries, may differ between systems, and this could be why it works on OSX but not others. For instance: Whether the manually-defined comparison operator is expected to return true or false for equivalent entries.

It should be pretty easy to see when this is failing: Just test the output of the `erase()` call at `spatiallock.h:85`, and if it's zero, print out the position and the set contents.

dchristiaens:
[...]

PR #824, opened 2016-11-16
Fix multithreading issues in global tractography

dchristiaens:
These fixes should address issue #477 and the segfault that occurred on some systems. I think this is now ready to be merged into master, although it might be good to test on windows and other systems too.


dchristiaens:
I've tested extensively on my new linux system. @maxpietsch confirmed that this compiles on macOS. Do I wait for someone to check if it compiles on windows too or do we just merge it in hope for the best?

jdtournier:
Give me a minute, I'll give it a crack on my system...

dchristiaens:
Perfect, thanks. (Fingers crossed)

jdtournier:
OK, compiles fine on Windows 10. Haven't actually tried running the command, do you need me to do that? It'll be a while before I find the time. 

This may be a good time to add some tests...? :wink:

dchristiaens:
Good, thanks for that. I mainly wanted to make sure it compiles cross-platform before merging.

I did 3 runs, each with 1 billion iterations on 20 threads, all successful (only took half an hour by the way). In these MCMC algorithms, anything that can go wrong will eventually happen. So I'm fairly confident that no more segfaults or race conditions occur in this version. Even if a user does find something, this version is definitely an improvement over the current master, so I think we can hit merge.

I would also like to have tests in TravisCI, but I don't know how to deal with the random nature of the algorithm. Even if I allow to fix the seed for the random generators, there are quite a few of them and their interaction is quite complicated.

dchristiaens:
I've merged these changes. I'll think about adding tests when I find some time.

Issue #828, opened 2016-11-22
tckgen: Early abort mechanism


Lestropie:
Need this for my own work, but opening it up for comments given #809.

What I'm thinking is to look at the ratio of maximum number of attempts to maximum number of tracks, and if it's looking increasingly unlikely that the requested number of tracks is not going to be reached, abort `tckgen` rather than waiting until a massive number of attempts have been made.

* I think this would need to be an explicit command-line option: e.g. if the user manually specifies both maximum seed and maximum streamline counts, they may well be happy to proceed until one of the two is satisfied.

* Anyone know a good statistic to use in this scenario? Preferably one that handles the stochastic nature of the process, and would appropriately handle the case where streamline count remains at zero? There's scope for a Bayesian predictor but I'm unsure of the distributions.

jdtournier:
That's not a bad idea! The distribution in this case would be a [Poisson distribution](https://en.wikipedia.org/wiki/Poisson_distribution) - nuclear decay being the typical example. Basically, the probability of observing an event (i.e. a successful streamline) is a constant per unit time (iteration, or in our case, seeding attempt). If we know that probability, we can figure out how many seeds we need, or conversely, how likely we are to achieve the desired number within a fixed number of attempts. In our case, we'd need to figure out lambda and its confidence interval from the existing data, and check that against the minimum lambda needed to reach the desired number of streamlines. I _think_ that can be done using the equations on that wikipedia page and maybe [this post](http://stats.stackexchange.com/questions/15371/how-to-calculate-a-confidence-level-for-a-poisson-distribution)...?

Lestropie:
I initially shied away from Poisson because it's probability of 'a given number of events' per unit 'time'; it's not bounded above, whereas here we can't have more tracks than attempts. Here it's more the probability of any one event (streamline attempt) being successful; hence I was mucking about with binomial ([incomplete beta functions](https://en.wikipedia.org/wiki/Beta_function#Incomplete_beta_function)! Yay!) & approximations thereof. But maybe Poisson might behave better when the mean's closer to zero, I don't know. Or maybe I just derped somewhere in my math.

PR #832, opened 2016-11-23
tckgen: Premature exit mechanism

Lestropie:
OK @jdtournier, this is what I've ended up with for #828. I'm pretty sure my conditional probability is right but my priors were wrong, so I've simplified and am just testing that conditional directly.

Logic is as follows:

Requested maximum number of streamlines _N_, maximum number of attempts _A_ (whether set by default to _100N_ or set manually). Hypothesis _H_ is that the fixed probability of any generated streamline being accepted is greater than _N/A_, so we should reach the requested number of streamlines before we reach the maximum number of attempts.

If _H_ is true: what is the probability that, after _a_ streamline attempts, you will observe no more than _n_ streamlines (where _n_ and _a_ are the running values for streamline number and attempts respectively)?

If _n_ is so low that observing such after _a_ under _H_ is highly unlikely, then it's highly unlikely that _n_ will reach _N_ by the time _a_ reaches _A_.

By discarding the priors, I'm basically saying _P(H|n,a) = P(n,a|H)_.

Open to discuss, modify, reject, belittle.

### DIFF
--- a/docs/reference/commands/maskdump.rst
+++ b/docs/reference/commands/maskdump.rst
@@ -8,14 +8,15 @@ Synopsis
 
 ::
 
-    maskdump [ options ]  input
+    maskdump [ options ]  input[ output ]
 
 -  *input*: the input image.
+-  *output*: The (optional) output text file.
 
 Description
 -----------
 
-Print out the locations of all non-zero voxels in a mask image
+Print out the locations of all non-zero voxels in a mask image. If no destination file is specified, the voxel locations will be printed to stdout.
 
 Options
 -------
--- a/docs/reference/commands/mredit.rst
+++ b/docs/reference/commands/mredit.rst
@@ -16,12 +16,18 @@ Synopsis
 Description
 -----------
 
-Edit the intensities within an image from the command-line. If only one image path is provided, the image will be edited in-place. If input and output image paths are provided, the original image will not be modified.
+Directly edit the intensities within an image from the command-line. A range of options are provided to enable direct editing of voxel intensities based on voxel / real-space coordinates. If only one image path is provided, the image will be edited in-place (use at own risk); if input and output image paths are provided, the output will contain the edited image, and the original image will not be modified in any way.
 
 Options
 -------
 
--  **-voxel position value** Change the intensity of a single voxel
+-  **-plane axis coord value** fill one or more planes on a particular image axis
+
+-  **-sphere position radius value** draw a sphere with radius in mm
+
+-  **-voxel position value** change the image value within a single voxel
+
+-  **-scanner** indicate that coordinates are specified in scanner space, rather than as voxel coordinates
 
 Standard options
 ^^^^^^^^^^^^^^^^
--- a/docs/reference/commands/mrinfo.rst
+++ b/docs/reference/commands/mrinfo.rst
@@ -44,7 +44,7 @@ Options
 
 -  **-transform** the voxel to image transformation
 
--  **-norealign** do not realign transform to near-default RAS coordinate system (the default behaviour on image load). This is useful to inspect the transform and strides as they are actually stored in the header, rather than as MRtrix interprets them.
+-  **-norealign** do not realign transform to near-default RAS coordinate system (the default behaviour on image load). This is useful to inspect the image and/or header contents as they are actually stored in the header, rather than as MRtrix interprets them.
 
 -  **-property key** any text properties embedded in the image header under the specified key (use 'all' to list all keys found)
 
--- a/docs/reference/commands/mrview.rst
+++ b/docs/reference/commands/mrview.rst
@@ -69,6 +69,11 @@ Debugging options
 
 -  **-fps** Display frames per second, averaged over the last 10 frames. The maximum over the last 3 seconds is also displayed.
 
+Other options
+^^^^^^^^^^^^^
+
+-  **-norealign** do not realign transform to near-default RAS coordinate system (the default behaviour on image load). This is useful to inspect the image and/or header contents as they are actually stored in the header, rather than as MRtrix interprets them.
+
 Overlay tool options
 ^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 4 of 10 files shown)


jdtournier:
OK, I've had a bit more of a thought about this, I think your initial idea of looking at it as a binomial distribution is actually the right way to do it. I've had a quick look through this commit, it looks like you're basically falling back to a heuristic...? Not sure... 

In any case, I reckon we can approach this as a regular hypothesis test, using the Normal approximation to the binomial distribution for large N. Here goes:

First, some definitions:

- _N_: maximum number of seeding attempts
- _S_: desired number of streamlines
- _n_: current number of seeding attempts
- _s_: current number of streamlines after _n_ attempts
- _a_: actual probability of streamline acceptance

We form a null hypothesis H0 that states that the true acceptance rate is _S_/_N_. The alternative H1 is that the true acceptance rate is less than _S_/_N_. 

For large enough _n_, [the binomial distribution can be approximated as a Normal distribution](https://en.wikipedia.org/wiki/Binomial_distribution#Normal_approximation) with mean _np_ and variance _np_(1-_p_). We can reframe this as a function of the current proportion _p_=_s_/_n_, with distribution _N_(_p_, _p_(1-_p_)/_√n_). 

Under the null, _p_ in the equation above is _S_/_N_, and the confidence interval around that is that given by a Normal distribution: _S_/_N_ ± _z_(α) × _p_(1-_p_)/_√n_, where _z_(α) is the z-score for a significance value of α. 

So this is what this looks like for a nominal acceptance rate of 1% (the default) and 100,000 seeding attempts (the default on `tckgen`), when the actual probability of acceptance _a_ = 0.005, testing at the α=1e-6 level:

early_termination

I've also displayed the upper bound, although that is never actually going to be used in the testing... But you can see that the test is never going to fail for low numbers (reassuring since we need large numbers of the Normal approximation to hold), and gradually tightens up as the numbers come in. We only need to know about the distribution under H0, and in that regime the distribution should be Normal for any reasonable number of tracks. We need to set a _very_ conservative significance level α to make sure we don't abort runs needlessly. Note that in practice, the actual significance level is around 5×α (checked with simulations), presumably because we don't account for the multiple comparisons performed at each iteration. This is difficult to account for, these are not going to be independent at all - but as long as we pick a very conservative value, I don't think this is anything to worry about.

What do you reckon?

jdtournier:
OK, had a chance to look through your commit a bit more thoroughly, I'd missed the use of the incomplete beta distribution (never heard of it... 😕). Your commit is probably right, but it would take me a while to wrap my head around it... I expect both approaches will converge on the same result for large _n_, and in our case we have large _n_: the confidence interval derived from the normal approximation encompasses zero for any number <500-1000 (depending on the significance level and the minimum rate of acceptance), so nothing will get rejected until we have enough samples to make the call. Importantly, all of these parameters are known _a priori_ (minimum acceptance rate to reach target, significance level), so H0 is independent of the _actual_ probability of acceptance - this means we can rely on the Normal approximation and expect it to be very accurate for the numbers we have, even if the _actual_ rate of acceptance is zero (which would otherwise be problematic). 

So not sure whether you want to stick with what you have (but I can't verify the maths right this minute), or try using the Normal approximation (might be cheaper to compute?). I'll try to wrap my head around this incomplete beta distribution when I have a minute...

jdtournier:
Also, given TravisCI tests are failing due to the lack of the `unsupported/Eigen/SpecialFunctions` header, we might need to go for the Normal approximation purely for pragmatic reasons...?

Lestropie:
Just to be clear: When I sign off a post using self-deprecating verbs, that's my inferiority complex, not a set of instructions :-/

Fixed my priors and therefore used Bayes' Theorem.

Added the normal approximation approach for support with Eigen < 3.3.0. The normal approximation to the binomial distribution is notoriously bad when the sample probability is near one of the two extrema (which it is in our case); and the closer you are to the extremum, the larger quantities you need for the normal approximation to become reasonable. You'll need to make sure it's implemented / it behaves as you want.

jdtournier:
> Just to be clear: When I sign off a post using self-deprecating verbs, that's my inferiority complex, not a set of instructions :-/

:laughing: Sorry, couldn't resist such an open invitation...

> The normal approximation to the binomial distribution is notoriously bad when the sample probability is near one of the two extrema (which it is in our case); and the closer you are to the extremum, the larger quantities you need for the normal approximation to become reasonable. 

Agreed. I tried to address this concern earlier, but it might not have been all that clear. All that matters is how close we are to the extrema _under the null hypothesis_ - we are not trying to estimate the true probability of acceptance (in which case this really would be a genuine concern), we are making the assumption that it is at least 1%, and won't be making any decisions until we have a large enough number of observations anyway. In that regime, I'm pretty confident the Normal approximation will be just fine as an approximation for the null distribution. All we're doing then is looking at the probability of having that many acceptances if the acceptance rate genuinely was what we'd assumed under the null. The actual probability of acceptance has no bearing on the validity of the null hypothesis. 

What is more of a concern in both the Normal approximation and the proper incomplete beta case is the issue of multiple comparisons. These tests are valid for a single test at a given number of attempts. If we're going to test at every iteration, our actual p-value is going to be affected. Not massively since each test is clearly very strongly correlated to the previous one, but the test at iteration 500 is going to be fairly independent of the test at iteration 5000. I'm not sure how easy it might be too deal with this, we might just have to accept that our p-value is not that stated, and use a very low uncorrected p-value. In which case the validity of the Normal approximation is the least of our concerns...

Lestropie:
I'm not sure precisely where it sits in the whole 'multiple comparisons' family, since it's always re-using all existing data and simply concatenating additional data. There's probably some kind of hierarchical Bayes thing going on that's going to be beyond me for now.

Figured I'd merge this (to `tag_0.3.16`), but disable the mechanism by default unless a particular config file entry is present. That way it can be mucked around with without going stale or requiring git checkouts. Hopefully if we're content with the behaviour it can be activated by default down the line.

Issue #793, opened 2016-10-10
mrview - sometimes segfault upon loading data, sometimes not  on ubuntu 16.04


samuelstjean:
Hello,

I think this one is likely related to bad libraries/config stuff, but I mostly want to see if anyone else has the same issue. Basically, whenever I try to load an image into mrview, sometime it works, sometimes I get a segfault for exactly the same image, 2-3 times in a row. It actually used to work before on the same computer which was running ubuntu 14.04, and I can't get it to work reliably on 16.04.

I of course pulled the latest master this morning, did a build clean, configure, build again. I can start mrview, but not view stuff in it. As it used to work before, I suspect bad lib/compiler combination, here is what I got

~~~bash
samuel ~/git/mrtrix3 $ [mrtrix3 (master)] ./configure 

MRtrix build type requested: release

Detecting OS: linux
Checking for C++11 compliant compiler [g++]: 5.4.0 - tested ok
Detecting pointer size: 64 bit
Detecting byte order: little-endian
Checking for variable-length array support: yes
Checking for non-POD variable-length array support: yes
Checking for zlib compression library: 1.2.8
Checking for Eigen 3 library: 3.2.92
Checking shared library generation: yes
Checking for Qt moc: moc (version 5.5.1)
Checking for Qt qmake: qmake (version 5.5.1)
Checking for Qt rcc: rcc (version 5.5.1)
Checking for Qt: 5.5.1
writing configuration to file './release/config': ok
~~~

And for the weird part, it just seems to work sometimes and sometimes not. This is one that segfaults.

~~~bash
samuel my_secret_folder_path_which_contains_too_much_information $ [DICOM_gab] /home/samuel/git/mrtrix3/release/bin/mrview 2.nii.gz -debug
mrview: [DEBUG] No config file found at "/etc/mrtrix.conf"
mrview: [DEBUG] No config file found at ".mrtrix.conf"
mrview: [INFO] GL renderer:  Quadro K2200/PCIe/SSE2
mrview: [INFO] GL version:   3.3.0 NVIDIA 367.48
mrview: [INFO] GL vendor:    NVIDIA Corporation
mrview: [DEBUG] loading font into OpenGL texture...
mrview: [DEBUG] compiling OpenGL vertex shader:
#version 330 core
layout(location = 0) in vec2 pos;
layout(location = 1) in vec2 font_pos;
uniform float scale_x;
uniform float scale_y;
out vec2 tex_coord;
void main () {
[...]

Issue #579, opened 2016-04-28
Compatibility with Eigen 3.3


Lestropie:
Although we recommend that people use the stable release version of Eigen, ultimately some people will still install the beta, and eventually 3.3 will become their stable release. So we should get MRtrix3 compatible with it at some stage.

For now, should also alter the documentation to explicitly recommend download / installation of the stable release rather than the beta.

Edit: Following on from #578.

jdtournier:
This issue is still very much current: e.g. #793.

jdtournier:
OK folks, this is now critical: [Eigen 3.3 is now the stable version](http://eigen.tuxfamily.org/index.php?title=Main_Page)...

jdtournier:
Right, a bit of testing suggests that on my system, adding the `-mno-avx` flag prevents the need for 32 byte alignment, reverting back to 16 byte alignment as in Eigen 3.2. I think the quickest fix for now is simply to add that flag to the defaults, which will at least allow us to use Eigen 3.3 without crashes. 

Here's a quick snippet of code to demonstrate. Note the gymnastic involved in trying to figure out the memory allocator's alignment guarantees...

```C++
// To compile :
//
// with AVX enabled:
//
//   g++ -march=native -O2 -DNDEBUG -isystem eigen-3.3 test_alignment.cpp && ./a.out
//
// or with AVX explicitly disabled:
//
//   g++ -march=native -mno-avx -O2 -DNDEBUG -isystem eigen-3.3 test_alignment.cpp && ./a.out


#include <iostream>
#include <Eigen/Core>

#ifndef EIGEN_DEFAULT_ALIGN_BYTES
// If not specified, assume 16 byte alignment (it's hard-coded as such in Eigen 3.2):
# define EIGEN_DEFAULT_ALIGN_BYTES 16
#endif

// guess default alignment:
inline size_t default_alignment ()
{
  size_t default_align = std::numeric_limits<size_t>::max();
  for (size_t n = 0; n < 100; ++n) {
    struct alignas(64) X { char c; };
    X* x = new X;
    size_t align = reinterpret_cast<size_t>(x) & 63U;
    if (align && align < default_align)
      default_align = align;
  }
  return default_align;
}


int main (int argc, char* argv[])
{
  std::cerr << "allocator alignment detected to be: " << default_alignment() << " bytes\n";
[...]

PR #833, opened 2016-11-24
Proposed solution for Eigen memory alignment issue 

jdtournier:
See previous discussions in #578, #579, and #793 - amongst others...

I'll add further details about these changes in a minute. For now, you can probably get most of the information by looking through the commit log for 1ddefacd...


jdtournier:
As if things weren't complicated enough...

I thought I'd have a look into what kind of performance benefits we might get using the AVX instructions with Eigen 3.3. The results weren't quite what I was expecting...

- I got a tiny performance improvement for a straight CSD (7.5s down to 6.7s) - not enough to get excited about...

- with `dwidenoise`, I got a huge _loss_ of performance: ~30s _up_ to ~4 _minutes_ - that's not what was supposed to happen! 

- I looked into this further, it turns out the performance drop has nothing to do with AVX: it's the same story using the `master` branch with AVX disabled. If you configure to compile against Eigen 3.3, you get terrible performance for `dwidenoise`, despite the fact that AVX is disabled on `master`. Can someone replicate and verify that this is indeed the case? I find it very hard to believe that the new version of Eigen would ship with such a massive regression in performance, presumably in the eigen-decomposition, which I'd have thought would be something that would get quite a bit of scrutiny. Maybe I'm doing something wrong...?

thijsdhollander:
I'm not fully up to date with these changes and the implications, so this may be completely wrong, but: is there a possibility that the problems `dwidenoise` feeds to the eigen-decomposition are in a way of a very particular kind?  It sounds indeed very strange that Eigen itself would have introduced something that's such a huge regression in performance... unless it improves performance for a wide range of problems, but just not for a very particular kind... or we're feeding our problems in a very suboptimal way to Eigen...? :confused:

jdtournier:
This is what I don't understand. This was using the exact same command, on pretty run-of-the-mill data. Also, just to be clear: this has **nothing to do with these changes**, the performance degradation is on `master`, our code is strictly identical in the two cases: the only difference was the version of Eigen used. I just ran it again just to make sure:

    $ git checkout master
    $ ./configure && ./build release/bin/dwidenoise && time release/bin/dwidenoise dwi.mif out.mif -force
    ...
    
    real    0m30.758s
    user    6m4.347s
    sys     0m1.270s
    
    $ export EIGEN_CFLAGS="-isystem /home/jdt13/eigen-3.3/"
    $ ./configure && ./build release/bin/dwidenoise && time release/bin/dwidenoise dwi.mif out.mif -force
    ...
    
    real    4m26.919s
    user    52m56.493s
    sys     0m1.570s

So a massive loss of performance, but I can't see how what we're doing could possibly be suboptimal - we spent a fair of time making sure it was as lean as we could make it... The actual results are more or less identical as far as I can make out, although the differences are larger than I'd like (just shot noise though, nothing consistent that would raise suspicion - we know the EVD is not all that stable as a means of doing an SVD, especially when operating in single precision).

But I fully agree that a degradation in performance for something so commonplace as an EVD seems unlikely. Which is why I'd really like others to try the above and verify that I'm not going loopy...

Another option worth looking into potentially is the issue of [denormals](http://stackoverflow.com/questions/9314534/why-does-changing-0-1f-to-0-slow-down-performance-by-10x), which @maxpietsch has looked into a while back. Given that we end up with such small eigenvalues, this may indeed become an issue...?

jdtournier:
Just to clarify the above tests further: on `master`, the AVX instructions are disabled by the `-mno-avx` option, which I added to the defaults a few days ago. So it's not even that it's using a different instruction set...

thijsdhollander:
Nothing in http://eigen.tuxfamily.org/index.php?title=3.3#Dense_decompositions that explains it?

thijsdhollander:
Just randomly guessing, but maybe the *"GeneralizedEigenSolver now extracts the eigenvectors."* that causes extra work to be done compare to 3.2?

jdtournier:
Well, the [`GeneralizedEigenSolver`](http://eigen.tuxfamily.org/dox/classEigen_1_1GeneralizedEigenSolver.html) class performs a full-blown _generalised_ eigen-decomposition of two matrices, which is quite a rare operation. I think the docs are basically saying you can now get the eigenvectors with this approach too. In any case, in `dwidenoise` we use the [`SelfAdjointEigenSolver`](http://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html).

More to the point, we need to compute the eigenvectors in this case anyway so we can recombine those that are being retained. So it shouldn't entail any extra work at all. I really don't see why this would be a problem...

jdtournier:
Doing a bit more testing, just documenting as I go.

**Snippet:**
```C++
#include <Eigen/Dense>
#include <Eigen/Eigenvalues>

#define SIZE 200
using namespace Eigen;

int main (int argc, char* argv[])
{
  MatrixXf mat (SIZE,SIZE);
  SelfAdjointEigenSolver<MatrixXf> eig;

  for (int n = 0; n < 1000; ++n) {
    mat.triangularView<Lower>() = MatrixXf::Random(SIZE,SIZE);
    eig.compute (mat);
  }

  return 0;
}
```
**Results:**
```
$ g++ -march=native -O2 -DNDEBUG -isystem eigen-3.2.10 test_eigen_EVD.cpp -o test_eigen_EVD && time ./test_eigen_EVD

real    0m5.306s
user    0m5.307s
sys     0m0.000s
 $ g++ -march=native -O2 -DNDEBUG -isystem eigen-3.3.0 test_eigen_EVD.cpp -o test_eigen_EVD && time ./test_eigen_EVD

real    0m10.625s
user    0m10.623s
sys     0m0.003s
```

So the EVD is itself definitely slower - but not as much as suggested by the `dwidenoise` tests. I'll keep digging...

jdtournier:
OK, aside from the EVD itself taking twice as long, it turns out that [this line](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/dwidenoise.cpp#L152) also slows down the whole process by an unbelievable factor of ~5 - i.e. it takes ~10× longer than the EVD itself:

    152         X.col (n/2) = eig.eigenvectors() * s.asDiagonal() * eig.eigenvectors().adjoint() * X.col(n/2);

I have absolutely no idea why things would be this bad with Eigen 3.3... Maybe their evaluation pipeline is having trouble optimising this, or maybe they promote the `s.asDiagonal()` to a full dense matrix, but either way, it's _very_ hard to imagine any of this making that big of a difference in performance...

By, the way, I've tried to copy the eigenvectors and the X row vector into temporaries to allow Eigen to optimise things as much as it could, i.e.: 

    Eigen::MatrixXf V = eig.eigenvectors();
    Eigen::VectorXf D = X.col(n/2);
    X.col (n/2) = V * s.asDiagonal() * V.adjoint() * D;

but that made no difference...

Ideas, anyone?

dchristiaens:
Good catch @jdtournier. I'll investigate asap, I've been meaning to try and benchmark the new BDCSVD algorithm anyway.

> The actual results are more or less identical as far as I can make out, although the differences are larger than I'd like (just shot noise though, nothing consistent that would raise suspicion - we know the EVD is not all that stable as a means of doing an SVD, especially when operating in single precision).

My first thoughts are also along these lines: we've already established that the Eigen EVD had lower precision than its Matlab counterpart. Maybe the new version has improved precision this at the cost of increased computation time?

jdtournier:
OK, turns out adding brackets to force the order of operations solves the issues with the final recomposition of the eigenvectors. I'll commit the fix in a separate pull request.

We still have a factor 2 loss of performance in the EVD though... I've just [posted a question about this on StackoverFlow](http://stackoverflow.com/questions/40805386/performance-regression-with-eigen-3-3-0-vs-3-2-10), see whether there is a simple solution.

jdtournier:
> Maybe the new version has improved precision this at the cost of increased computation time?

That's also what I was hoping might be the case. Let's see what pops up on StackOverflow...

jdtournier:
I changed TravisCI to use Eigen 3.3 in its tests, and now a few tests are failing, same as on my system. These are all down to precision, and the chances are Eigen 3.3 is probably more accurate... In any case, we need to amend the tests to match. Is everybody OK with me reducing the precision for these tests? It will make them less sensitive, but also more tolerant of different architectures, etc. The difference in precision in most cases here will probably be down to the use of more advanced instruction sets, particularly those capable of fused multiply-adds (FMA) - which AVX provides, and that's used in Eigen 3.3. 

The alternative is to update the test data to match what Eigen 3.3 now produces, but that would make things fail when compiled against the older Eigen version, or potentially when compiled with AVX disabled. So I vote to increase the tolerance of the tests...

jdtournier:
You'll note that based on a recent discussion with @dchristiaens, I've now changed the configure script to default to `clang++`, falling back to `g++` if not found. This will probably be a good move on Linux and MacOSX (it's already default on MacOSX, and we use it for the tests on TravisCI). It turns out to be available on Windows also via MSYS2, although it requires removal of a few compiler command-line switches that `clang++` doesn't understand, many of which are actually provided by Qt5's config. I can remove them, and the whole thing compiles and tests just fine, but MRView crashed out pretty quickly... So it looks like it might be difficult to get Qt5 to play ball with `clang++` on Windows. I'll change the order of priority for Windows to be `g++` first, then `clang++` - but maybe eventually we can figure out how to get things working with `clang++` on Windows too.

thijsdhollander:
> The alternative is to update the test data to match what Eigen 3.3 now produces, but that would make things fail when compiled against the older Eigen version, or potentially when compiled with AVX disabled. So I vote to increase the tolerance of the tests...

Probably best to do both, right?  If we reason Eigen 3.3 now produces more accurate results, these should probably be the ones serving as the test data results.  But additionally, increased tolerance will still be needed for the reasons mentioned.

Lestropie:
^ Precisely my thinking.

jdtournier:
I'd agree with updating the test results too if I had solid independent evidence that 3.3 is more accurate, but at the moment it's just a working hypothesis. Not sure how we can go about testing this, or even whether it's worth doing. If we get different results on different architectures with different versions of Eigen, we want to be tolerant to that. What we're checking is whether our code is doing what it's supposed to within the environment it's operating. I'm not sure these tests need to be all that strict, we're trying to catch outright failures. Besides, even the same algorithm implemented subtly differently (maybe changing the order of operations) can introduce significant differences due to precision issues, even when the approach is mathematically equivalent. If we were testing the precision of our algorithms, then we'd need to worry about these issues, but right now all the tests can/should be used for is verifying that the output of our commands is as expected within the normal variation of the hardware and other software they're running under.

So to cut a long story short: happy to change the test data if we can verify that the new version produces more accurate results. Just not sure how to do that, and not sure the justification for it is that compelling...

thijsdhollander:
Yes, I do agree we can't be sure which outcome is more accurate, and most algorithms are indeed even just meant to be a numerical approximation to aims for "the real thing" they're after; they're "inaccurate" by default so to say.  From a pragmatic point of view, I'd say it simply depends on how different any current ("Eigen 3.3") outcome is to a previous one.  If the difference itself is of an order of magnitude worth worrying about, then probably that is in and of itself... well, worth worrying about.  In a way, I'd be uncomfortable going with any (either current or previous) outcome before we figure out what's exactly causing such changes.  But if there's no such changes to begin with, or all of them are within acceptable ranges oft he previous ones, then it's probably not worth losing sleep over.

In a way, this:

> all the tests can/should be used for is verifying that the output of our commands is as expected within the normal variation of the hardware and other software they're running under

... would actually argue for running them on the range of hardware/software that we deem "normal", and average the outcomes to generate the test data (and use the standard deviation to inspire the precision thresholds even).  But I agree that's quite a bit too much overkill. :wink: 

So pragmatic it is. :smile:  Are there actually any differences that are worth worrying about in any way currently?

jdtournier:
OK, I've relaxed the tolerance on the tests that were clearly purely related to precision (`dwidenoise`, `mrfilter` and `mrtransform`. @dchristiaens has just committed a fix for the `tckconvert` tests, which I've merged into this branch, and that fixes that issue. Only remaining one is `tckmap`. This is the failure (note the new output from `testing_diff_data`):

    # command: tckmap tckmap/in.tck -template dwi.mif -dec - | testing_diff_data - tckmap/tdi_color.mif.gz -abs 1.5 [ ERROR ]
    tckmap: mapping tracks to image...  [==================================================]
    testing_diff_data: [ERROR] images "/tmp/mrtrix-tmp-UNzOoh.mif" and "tckmap/tdi_color.mif.gz" do not match within absolute precision of 1.5 (198 voxels over threshold, maximum absolute difference = 14.327392578125)

That test is failing with a max abs difference of 14, which seems huge. I think this one does deserve a bit more investigation...

jdtournier:
One further issue with Eigen 3.3's memory alignment requirements is the use of classes that require 32 byte alignment in STL containers such as `std::vector`. I've just added a convenience class to deal with this, see commit message in fe40f04 for details. Bottom line is, replace all susceptible declarations of `std::vector` from:

    std::vector<MyClass> vec;
    
to:
    
    memalign_vector<MyClass>::type vec;

I have a feeling it'll take a while to go through and identify all such instances... It looks like the registration code in particular might be full of such instances. I've gone and cleaned up `mraverageheader`, but there'll no doubt be other cases. @maxpietsch, @draffelt: you might be in a better position fix these up...?

thijsdhollander:
Slightly off-topic, but slightly related as well: I just came across this when reminding myself again of how Eigen deals with aliasing:

> Moreover, starting in Eigen 3.3, aliasing is not assumed if the destination matrix is resized and the product is not directly assigned to the destination.

(in context of matrix multiplication; from http://eigen.tuxfamily.org/dox/group__TopicAliasing.html )

I'm not sure we'd have cases of this.  Obvious cases would make tests fail big time I'd say, but you never know...?

jdtournier:
> Slightly off-topic, but slightly related as well: I just came across this when reminding myself again of how Eigen deals with aliasing:

Worth keeping an eye on, but actually we've recently come across more worrying issues. The main one being that a matrix-matrix-vector product is _orders_ of magnitude slower in Eigen 3.3 than it was in Eigen 3.2 - as per the discussion above. That made a huge difference in the performance of `dwidenoise` for instance. But it's not unlikely to pop up elsewhere. [Here's one example](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/dwi2tensor.cpp#L118) just off the top of my head...

Identifying performance regressions is going to take some effort, since we don't currently benchmark these things. It would be quite an undertaking to do this, not sure it would ever be worth the time and effort. So unless this is fixed at source, we can't guarantee that all commands in MRtrix3 will have equivalent or better performance with Eigen 3.3...

For now though, we need to go through and replace all occurrences of `std::vector` where they might be a problem. Is anyone else starting to wonder whether this is really worth the extra effort...?

Lestropie:
> Identifying performance regressions is going to take some effort, since we don't currently benchmark these things.

That's what beta testers are for. :-D  If something's noticeably slower, someone is bound to complain.

> For now though, we need to go through and replace all occurrences of std::vector where they might be a problem.

I don't _think_ there should be too many instances of such things. In the registration code I'd expect there to be quite a few, but not so much outside of that.

> Is anyone else starting to wonder whether this is really worth the extra effort...?

Given it's now their recommended download, I think we have to. We can't trust users to follow our instructions to install an out-of-date version, no matter how bright the documentation warning is. Plus it would kind of go against our philosophy of being cutting-edge... aren't you an Arch user? :-P

chrisadamsonmcri:
I spent a bit of time looking into AVX optimizations with FSL, Freesurfer etc. I found that just enabling -mavx/-mavx2 didn't improve things much. Perhaps it is purely generating AVX instructions without attempting to vectorise, or at least no better than what it did for SSE. But if you use gcc6/g++6 with these new compiler flags
"-ftree-loop-vectorize -ffast-math -lm -mavx" this will utilise a vectorised version of libm. This seemed to deliver consistent improvements in FSL FEEDS. I would really like to try this for mrtrix.

jdtournier:
> I spent a bit of time looking into AVX optimizations with FSL, Freesurfer etc. I found that just enabling -mavx/-mavx2 didn't improve things much. Perhaps it is purely generating AVX instructions without attempting to vectorise, or at least no better than what it did for SSE. 

Yes, that's not too surprising. Vectorization only gets used by the compiler in very specific circumstances ([this page](https://monoinfinito.wordpress.com/series/vectorization-in-gcc/) explains what's going on pretty well). All it's telling the compiler is that it is allowed to use instructions from the AVX extension.

> But if you use gcc6/g++6 with these new compiler flags
"-ftree-loop-vectorize -ffast-math -lm -mavx" this will utilise a vectorised version of libm. This seemed to deliver consistent improvements in FSL FEEDS. I would really like to try this for mrtrix.

It's already there (apart from `-fast-math`), at least on this branch, since we're now using `-O3`, [which pulls in `-ftree-loop-vectorize`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html). The `-mavx` option is included for CPUs that support it via the (_MRtrix3_ default) `-march=native` flag. But if we passed `-mavx` explicitly, that would produce illegal instructions on older CPUs...

As to `-fast-math`, I'm not sure, I've never tried it. It may work just fine, but there are concerns with things like [checking for NaN and Inf](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1008), which we rely on quite heavily. It looks like these issues might be fixed by now, but they are compiler-dependent, so I'd be _very_ wary to enable this be default without running a fair few tests to verify that things like `isnan()` and `finite()` actually work as expected... If we can identify all the potential problems and write a short piece of code to check all this, I'm happy to include it in the `configure` script, and enable `-ffast-math` on this case. Would you be interested in giving that a shot?

In any case, nothing to stop you experimenting. All you need to do is run `./configure`, then you can edit the `release/config` file as needed before running `./build` as usual. That file is a dependency for everything else, by the way, so any change to that file will trigger a full recompile - no need to `./build clean` or anything.

> Could you simply replace eigen3 with something like blas, atlas, lapack or boost?

Not a chance. The code is _completely_ different. Switching to Eigen was a very major undertaking, taking several months (and many late nights) to complete. Besides, [Eigen performs better](http://eigen.tuxfamily.org/index.php?title=Benchmark) than any of these. We used to run with the GSL, which itself was a pretty thin veneer on top of blas. Runtime improved by a factor of between 2 & 3 for CSD when we switched to Eigen. Finally, it's the only solution that is genuinely guaranteed to provide close to optimal performance on any CPU: it's optimised at compile-time - most other implementations are provided as pre-compiled libraries, so will not necessarily provide all the optimisations possible on each CPU. And it provides a lot more than just matrix operations, we rely on a lot of its functionality throughout. So I'm not going back...

chrisadamsonmcri:
My point about the compiler flags was actually more about the gcc version you use. Only 6 seems to utilise the AVX optimised functions in libmvec without source code modification:
https://sourceware.org/glibc/wiki/libmvec
Unless you are using Ubuntu 16.10 you won't have gcc6 installed by default. Not sure about other distros.

Any time you build on an avx/avx2 cpu with -march=native it should enable all instruction sets available to it including FMA, BMI, F16C, which will also not run on an older cpu.

I would be happy giving the fast-math tests a shot. But the issue is that I can't use avx with eigen3.
I attempted to build eigen3 with avx support but mrtrix3 still didn't compile. Their claims of avx support are a little confusing.

dchristiaens:
> My point about the compiler flags was actually more about the gcc version you use. Only 6 seems to utilise the AVX optimised functions in libmvec without source code modification

I don't know about libmvec, but Eigen provides efficient vectorization at the syntax level and includes AVX support since 3.3. There is no mention on the Eigen website about this being compiler/version dependent, except for the required `-mavx` flags. If you notice a big performance improvement with GCC 6, let us know.

> Any time you build on an avx/avx2 cpu with -march=native it should enable all instruction sets available to it including FMA, BMI, F16C, which will also not run on an older cpu.

Of course. If you are compiling on one system and plan to run the build on another, then obviously `-march=native` won't work for older/different cpu architectures.

> I would be happy giving the fast-math tests a shot. But the issue is that I can't use avx with eigen3.
I attempted to build eigen3 with avx support but mrtrix3 still didn't compile. Their claims of avx support are a little confusing.

Since Eigen 3.3, AVX should be supported. The purpose of this pull request is to support Eigen 3.3, so if you check out this branch and you get compiler errors, post them here such that we can investigate.

> Could you simply replace eigen3 with something like blas, atlas, lapack or boost?

As @jdtournier already said, we're not going back. But if you really want to give it a try, since Eigen 3.3 you can use [Blas bindings](http://eigen.tuxfamily.org/dox-devel/TopicUsingBlasLapack.html) for dense matrix operations. With these, you could leverage e.g. the Intel MKL libraries without modifying the code. Again, if you want to benchmark these, go for it!

jdtournier:
OK, just to add a few comments on top of @dchristiaens's:

> My point about the compiler flags was actually more about the gcc version you use. Only 6 seems to utilise the AVX optimised functions in libmvec without source code modification:
https://sourceware.org/glibc/wiki/libmvec
Unless you are using Ubuntu 16.10 you won't have gcc6 installed by default. Not sure about other distros.

I'm not familiar with `libmvec` either, but in general, you don't want to rely on the compiler vectorising things for you, it's too complex a task, and the compiler needs to guarantee that the code produces exactly the same results as if vectorisation hadn't been used. These are very stringent constraints, and will often mean vectorisation won't used. The way Eigen handles vectorisation is that it includes code with [direct assembly instructions](https://eigen.tuxfamily.org/dox/dir_924cda28ed0b76bdcee5658eb3660dc4.html) that will get included if compiled with `-mavx` (or whatever other SIMD instructions are available), no matter what compiler is used. This doesn't rely on the compiler's optimisation strategy - or least not as much as it would have otherwise. There are certain issues with code efficiency when compiled with `gcc -O2` (which is the current default on `master`), but using `clang++ -O2` or `gcc -O3` seems to sort that out - and that's irrespective of whether it was compiled with AVX or not, so nothing to do with that. The issue we have is that to make things as fast as they can possibly be, the Eigen developers rely on the assumption that all the relevant objects are aligned on 32-byte boundaries, which avoids the need for unaligned loads. All of their objects are properly aligned - or at least they would be if the C++ standard had figured out a way to ensure that the `new` operator also respected the stated alignment requirements specified for these types...

> Any time you build on an avx/avx2 cpu with -march=native it should enable all instruction sets available to it including FMA, BMI, F16C, which will also not run on an older cpu.

> I would be happy giving the fast-math tests a shot. But the issue is that I can't use avx with eigen3.
I attempted to build eigen3 with avx support but mrtrix3 still didn't compile. Their claims of avx support are a little confusing.

- this branch does indeed fail to build since the last commit I made (fd96475) - I made some modifications that have repercussions elsewhere, and while it fixed the failure on `population_template` on my system, clearly I need to tidy a few things up. This is a work-in-progress after all... :wink: 

  Going back one commit should compile fine (although the tests will still fail for `tckmap` - we need to work that one out)

jdtournier:
OK, just to add a few comments on top of @dchristiaens's:

> My point about the compiler flags was actually more about the gcc version you use. Only 6 seems to utilise the AVX optimised functions in libmvec without source code modification:
https://sourceware.org/glibc/wiki/libmvec
Unless you are using Ubuntu 16.10 you won't have gcc6 installed by default. Not sure about other distros.

I'm not familiar with `libmvec` either, but in general, you don't want to rely on the compiler vectorising things for you, it's too complex a task, and the compiler needs to guarantee that the code produces exactly the same results as if vectorisation hadn't been used. These are very stringent constraints, and will often mean vectorisation won't be used - either because it's too complex to ensure identical operation, or because adding all the extra checks and balances would make the operation counter-productive. 

The way Eigen handles vectorisation is that it includes code with direct assembly instructions that will get included if compiled with -mavx (or whatever other SIMD instructions are available), no matter what compiler is used. This doesn't rely on the compiler's optimisation strategy - or least not as much as it would have otherwise. 

There are certain issues with code efficiency when compiled with `gcc -O2` (which is the current default on `master`), but using `clang++ -O2` or `gcc -O3` seems to sort that out - and that's irrespective of whether it was compiled with AVX or not, so nothing to do with that. 

The issue we have here is that to make things as fast as they can possibly be, the Eigen developers rely on the assumption that all the relevant objects are aligned on 32-byte boundaries, which avoids the need for unaligned loads. All of their objects are properly aligned - or at least they would be if the C++ standard had figured out a way to ensure that the `new` operator also respected the stated alignment requirements specified for these types... So we have problems with dynamic allocation in some very specific circumstances. This whole branch is about providing some compile-time guarantees that we have dealt with all of the instances where this might be a problem.

> Any time you build on an avx/avx2 cpu with -march=native it should enable all instruction sets available to it including FMA, BMI, F16C, which will also not run on an older cpu.

Absolutely. My point is that if we were to explicitly include `-mavx` by default, that definitely wouldn't run on an older CPU, even if the code had been compiled on that CPU with `-march=native`. This would definitely be a very bad outcome. I'm assuming that if you specify `-march=native -mavx` on a CPU without the AVX extensions, the compiler would nonetheless do what it was told and enable all the extensions available on the current CPU _and_ AVX - irrespective of whether AVX was indeed available or not.

> I would be happy giving the fast-math tests a shot. But the issue is that I can't use avx with eigen3.
I attempted to build eigen3 with avx support but mrtrix3 still didn't compile. Their claims of avx support are a little confusing.

So I'm not sure what you mean by _I attempted to build eigen3 with avx support_, but I assume you mean you tried to build this branch with Eigen 3.3...? I've explicitly disabled AVX on `master`, so you definitely won't get it there. This branch does indeed fail to build as of the last commit (fd96475) - I made some modifications that have repercussions elsewhere, and while it fixed the failure on `population_template` on my system, clearly I need to tidy a few things up. This is a work-in-progress after all... :wink:

Going back one commit should compile fine (although the tests will still fail for `tckmap` - we need to work that one out). If that's not what you meant, then you'll need to clarify - we've been able to build `master` against Eigen 3.3 with AVX enabled for quite some time - but it would routinely crash (plenty of examples of that on the community forum). I've only added the `-mnoavx` flag a few weeks ago to prevent these outright failures when I noticed that Eigen 3.3 had been released properly...

chrisadamsonmcri:
libmvec is just a vectorised version of libm, the math portion of the standard C library glibc.

What I meant by build eigen3 with avx support was I downloaded the eigen3 source code and explicitly enabled the AVX option in the cmake gui when I configured the cmake options. Didn't help when I tried to build mrtrix though.

As an aside, we are phasing out pre-avx machines so not having avx support is becoming a non-issue.

jdtournier:
> libmvec is just a vectorised version of libm, the math portion of the standard C library glibc.

OK, I reckon there'll be very few instances where that might matter in _MRtrix3_, anything heavy should use the Eigen3 functionality, which will probably vectorise better anyway since it doesn't have to be compatible with the standard libm, or any other constraints. Good to know though...

> What I meant by build eigen3 with avx support was I downloaded the eigen3 source code and explicitly enabled the AVX option in the cmake gui when I configured the cmake options. Didn't help when I tried to build mrtrix though.

I'm a bit confused here... As far as I can tell, CMake has no influence on anything other than the unit tests - at least that's what the [Eigen3 front page](http://eigen.tuxfamily.org/index.php?title=Main_Page#Requirements) says... AVX support just gets included if the compiler sets the requisite macros (see e.g. [this bit of code](https://bitbucket.org/eigen/eigen/src/0a9b971803ae56a4d0a3483dd7147c28cfff12c9/Eigen/Core?at=default&fileviewer=file-view-default#Core-123:156)), there really shouldn't be any need to mess around with CMake...?

> As an aside, we are phasing out pre-avx machines so not having avx support is becoming a non-issue.

For you maybe! :wink: We've recently had someone wanting to build _MRtrix3_ on a PowerPC... And my old laptop had a Core2 Duo CPU - no AVX (I've just upgraded though, so no longer a problem :grin:).

jdtournier:
I just merged with `master` to bring this up to date, with a view to merging with `tag_0.3.16` shortly. However, I decided to have a look at the reason for the failure of the tests for `tckmap` (and now also `mrtransform`. It seems only the DEC output is affected, and comparing the output maps against the ground truth, it looks like only the edge voxels are affected with quite big differences - for too big to be due to precision alone. I also have a look at what the equivalent code produced on `master`, and got the same answer as this branch... So it seems that the tests will fail using Eigen 3.3, whether using this branch or `master`, even with AVX disabled. So there must have been some difference introduced in there that gets triggered on the edge voxels. But I can't figure out what might be causing this - best guess is something like the handling of NaNs or something? Anyone willing to look into this...?

Otherwise, the last remaining thing to do here before merging proper, is to go through and identify any uses of `std::vector` that require 32-byte alignment, and replace them with `memalign_vector`...

Lestropie:
> It seems only the DEC output is affected, and comparing the output maps against the ground truth, it looks like only the edge voxels are affected with quite big differences - for too big to be due to precision alone. ... Anyone willing to look into this...?

I should have looked further into this *ages* ago. If it's edge voxels (= streamline endpoint voxels in the test data), I suspect it may have been a bug relating to how streamlines are temporarily extrapolated in order to enable hermite interpolation at the ends. I'm pretty sure it was fixed when the code moved from [here](https://github.com/MRtrix3/mrtrix3/commit/d008c122ee144ca06c9e1d13797806898c232fdd#diff-b727fe5619c17c07f2ef41e9e6fe870fL155) to [here](https://github.com/MRtrix3/mrtrix3/commit/d008c122ee144ca06c9e1d13797806898c232fdd#diff-9a31f5b0dffc20021e1e5b46972ba2caR75). So a test data regeneration is in order - possibly for all `tckmap` data, it's just that DEC is the only one exceeding the test threshold.

jdtournier:
> I should have looked further into this ages ago. If it's edge voxels (= streamline endpoint voxels in the test data), I suspect it may have been a bug relating to how streamlines are temporarily extrapolated in order to enable hermite interpolation at the ends. I'm pretty sure it was fixed when the code moved from here to here. So a test data regeneration is in order - possibly for all tckmap data, it's just that DEC is the only one exceeding the test threshold.

There's a few things that bug me about this. First is that the tests for the straight TDI (same command without the `-dec` option, as far as I can see) _don't_ fail. So the streamline mapping gets it right within a single streamline (threshold is 1.5). How can the DEC version fail with a whopping difference of 14 when the mapping itself is presumably using the same codepath? 

Also, looking at this on `master`, this checks out OK with Eigen 3.2.10, but not 3.3.0 (at least it seemed to when I looked into this briefly last night). So if it's due to a change in behaviour of the algorithm on our side, how come there's such a massive difference between Eigen versions?

Lestropie:
Straight TDI and DEC TDI follow two different code paths - though I don't think that's the source of the difference. If the extrapolated streamline trajectory goes awry, the 'mean direction' of the streamline within the end voxel may be correspondingly awry, leading to an incorrect DEC even though each streamline still contributes to each voxel it traverses with a norm of 1.0.

Alternatively, my initial hypothesis might be wrong.

jdtournier:
> If the extrapolated streamline trajectory goes awry, the 'mean direction' of the streamline within the end voxel may be correspondingly awry, leading to an incorrect DEC even though each streamline still contributes to each voxel it traverses with a norm of 1.0.

OK, that would make sense. Still, can't figure out why the results would differ between Eigen versions.

jdtournier:
OK, just replaced the slightly awkward `memalign_vector<X>::type` with the much simpler `vector<X>` (in `MR` namespace) - only snag was figuring out constructor inheritance for template classes, but it seems to work just fine. So all that was required after that was to replace all 1200+ occurrences of `std::vector` with `vector`... 

Thankfully this little gem did the trick for me: :wink:
```bash
find . -name '*.h' -or -name '*.cpp' -and -not -name '*_moc.cpp' | \
xargs sed -i'' -e 's|\bstd::vector\b|vector|g'
```
Only needed to fix up a couple of things after that. The main thing to watch out for is that it's no longer possible to write this:
```C++
class Myclass;
vector<Myclass> vec;
```
since the instantiation of `vec` needs to know about the alignment requirements of `Myclass`, which won't be known until it is itself fully defined. Which made [this line](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/shapes/sphere.h#L53) slightly awkward. Problem was basically:
```C++
class Myclass {
  void func (const vector<Myclass>& vec);
};
```
Problem there is that `vector<Myclass>` requires `Myclass` to be fully defined to even resolve its actual type (since it's a template specialisation based on the alignment requirement of `Myclass`), and `Myclass` won't be fully defined until the closing `};`. So I resorted to declaring `func()` as a template instead:
```C++
class Myclass {
  template <class X> void func (const X& vec);
};
```
Just in case anyone else comes across similar problems...

Otherwise, it's all good to go. As before, the only test currently failing is `tckmap -dec` - hopefully @Lestropie will have that fixed up soon :stuck_out_tongue_winking_eye: 

There's a script in there called `check_memalign` that will detect and flag any instances of classes or structs that haven't been declared `MEMALIGN()` or `NOMEMALIGN`, or any remaining instances of `std::vector`, and if any are found will cause the TravisCI tests to fail. That's the only way I could think of to provide some compile-time guarantee that the code was safe for Eigen 3.3 + AVX... 

Note if for some weird reason you really need to use a `std::vector<>`, write it as `::std::vector<>` - the script will allow these instances (if only so I can actually declare our new `vector<>` as deriving from `std::vector<>`...).

jdtournier:
OK, that last commit fails to compile with this message:
```
lib/adapter/reslice.cpp:23:23: error: default initialization of an object of const type 'const vector<int>' requires a user-provided default constructor
    const vector<int> AutoOverSample;
```
This is using `clang++` 3.4. I've had a quick look into this (e.g. [here](http://stackoverflow.com/questions/7411515/why-does-c-require-a-user-provided-default-constructor-to-default-construct-a)), and I reckon this version of clang is probably not following the standard. I guess it depends on what is meant by "user-provided default constructor" here, but since I've used constructor inheritance to inject all the `std::vector` constructors, I would have thought that would have been enough. It certainly compiles fine on `clang++` 3.9.1 (on my Arch system). Any chance we can update the version of clang used on TravisCI...?

jdtournier:
Couldn't figure out how to update `clang` on TravisCI - although I did come across a [blog about it](http://genbattle.bitbucket.org/blog/2016/01/17/c++-travis-ci/). Anyway, it looks like it's fine to inherit constructors _and_ add your own default constructor, and that seems to make clang happy - looks like all changes compile fine now. Only remaining issues here are tests failing for `tckmap` and `mrtransform`. I'll also have a go at merging with `tag_0.3.16` when I get the chance - I doubt we're going to be merging this monumental changeset directly to `master` anyway...

jdtournier:
Closing this pull request, I'll start another one to merge into `tag_0.3.16`.

PR #835, opened 2016-11-25
dwidenoise: add brackets to final recombine step

jdtournier:
This addresses the huge drop in performance observed with Eigen 3.3, as
discussed in #833.

### DIFF
--- a/cmd/dwidenoise.cpp
+++ b/cmd/dwidenoise.cpp
@@ -149,9 +149,9 @@ class DenoisingFunctor
       s.head (cutoff_p).setZero();
       s.tail (r-cutoff_p).setOnes();
       if (m <= n) 
-        X.col (n/2) = eig.eigenvectors() * s.asDiagonal() * eig.eigenvectors().adjoint() * X.col(n/2);
+        X.col (n/2) = eig.eigenvectors() * ( s.asDiagonal() * ( eig.eigenvectors().adjoint() * X.col(n/2) ));
       else 
-        X.col (n/2) = X * eig.eigenvectors() * s.asDiagonal() * eig.eigenvectors().adjoint().col(n/2);
+        X.col (n/2) = X * ( eig.eigenvectors() * ( s.asDiagonal() * eig.eigenvectors().adjoint().col(n/2) ));
     }
 
     // Store output


jdtournier:
OK, quick test to compare results before and after: it's pretty close to the same thing when using the same version of Eigen. Using Eigen 3.3 does produce quite different results, but not overtly wrong - there is a good chance it might just be more accurate. Runtimes are:

Eigen 3.2.10, current `master`: 30s
Eigen 3.2.10, this pull request: 28s
Eigen 3.3.0, current `master`: 4m30s
Eigen 3.3.0, this pull request: 48s

So this fix helps performance a bit anyway, but we still have an issue with the slower EVD. Not as big a showstopper as before, thankfully...

All in all, I guess this means this is good to merge, but it would be good to understand exactly why we get differences in the EVD with Eigen 3.3...

dchristiaens:
I agree about the brackets and that SO answer explains the need for them very well. They also advise to compile with `-O3` instead of `-O2`. Any particular reason why we don't do that already?

jdtournier:
I just answered on SO, I don't see _any_ performance improvements using `-O3`, or `clang++`. Maybe the poster is using a difference CPU architecture or something, but I'm not seeing the slightest difference on my Intel Core i7... 

Also, in my experience compiling with `-O3` introduces more problems with very little difference (if any) in performance. But I've certainly had code that crashed with `-O3`, and ran fine with `-O2`. Took a very long time to figure out that the optimisation was the problem... I also don't recall coming across any cases where performance was noticeably better using `-O3`. Maybe things have changed since then (that was maybe 5-10 years ago), but in my experience pushing the optimisation too hard only exposes you to subtle compiler bugs...

jdtournier:
On that note, @dchristiaens: can you check whether `-O3` makes any difference on your shiny Xeon...? :wink:

jdtournier:
Crap, I goofed on my tests... modified the wrong command-line in my little script, I was running with `g++ -O2` all along... The poster is spot on about the EVD: using `g++ -O3` does give us a speedup over `g++ -O2`, and using `clang++` also gives us good execution times regardless, slightly better with `-O3`, and better with Eigen 3.3. 

So this argues for turning on `-O3` in the config. I'm slightly uneasy about doing this given my past experiences, but I guess we won't know till we try...

dchristiaens:
If you're uneasy flagging `-O3`, how about switching to `clang++` instead? I have had nothing but good experiences with clang on Mac for years. Error messages are much more informative and warnings often alerted potential risks before they occurred. It's available on any modern UNIX system, not sure about Windows though.

jdtournier:
There's nothing to stop anyone from using `clang++`. I also use it from time to time on Arch Linux, it's the default on MacOSX, and we use it in the TravisCI tests, because it's faster and stricter than `g++`. Never tried on Windows, but I'm guessing it's unlikely to be available, since we rely on MinGW to provide the compiler and the whole POSIX compatibility layer needed to get everything running. I'd get happy to default to `clang++` everywhere if it's available, maybe we could change `configure` to try `clang++` first, and fall back to `g++` if it can't find it?

As to defaulting to `-O3`, well that depends on how stable this is these days. I'm not averse to giving it a shot, but we'll have to keep a close eye on things if we start getting strange reports from users...

I'll give this a shot tonight as part of #833...

PR #837, opened 2016-11-25
mrview: more config options for overlay

maxpietsch:
Switching off overlay elements via the config file came in handy for clean screenshots.

### DIFF
--- a/cmd/dwidenoise.cpp
+++ b/cmd/dwidenoise.cpp
@@ -149,9 +149,9 @@ class DenoisingFunctor
       s.head (cutoff_p).setZero();
       s.tail (r-cutoff_p).setOnes();
       if (m <= n) 
-        X.col (n/2) = eig.eigenvectors() * s.asDiagonal() * eig.eigenvectors().adjoint() * X.col(n/2);
+        X.col (n/2) = eig.eigenvectors() * ( s.asDiagonal() * ( eig.eigenvectors().adjoint() * X.col(n/2) ));
       else 
-        X.col (n/2) = X * eig.eigenvectors() * s.asDiagonal() * eig.eigenvectors().adjoint().col(n/2);
+        X.col (n/2) = X * ( eig.eigenvectors() * ( s.asDiagonal() * eig.eigenvectors().adjoint().col(n/2) ));
     }
 
     // Store output
--- a/cmd/mrtransform.cpp
+++ b/cmd/mrtransform.cpp
@@ -368,7 +368,7 @@ void run ()
 
   // Rotate/Flip gradient directions if present
   if (linear && input_header.ndim() == 4 && !warp && !fod_reorientation) {
-    Eigen::MatrixXd rotation = linear_transform.linear();
+    Eigen::MatrixXd rotation = linear_transform.linear().inverse();
     Eigen::MatrixXd test = rotation.transpose() * rotation;
     test = test.array() / test.diagonal().mean();
     if (!test.isIdentity (0.001)) {
--- a/cmd/mrtransform.cpp
+++ b/cmd/mrtransform.cpp
@@ -34,6 +34,7 @@
 #include "registration/warp/utils.h"
 #include "registration/warp/compose.h"
 #include "math/average_space.h"
+#include "math/SH.h"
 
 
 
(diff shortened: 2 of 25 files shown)


jdtournier:
Sounds useful. Is there also a command line switch for this? If not, it might be worth adding...

maxpietsch:
added the options `-focus_off -comments_off -voxelinfo_off -orientationlabel_off -colourbar_off` and removed the redundant `-nointerpolation` option.

jdtournier:
Just a few comments on this so you can close it off:

- I can't figure out where the interpolation is set from the config file. There is an entry for it, but it doesn't seem to be used anywhere...?

- it would be worth adding a single option to hide all of these elements. Maybe something like `-nohud` or `-bare`? Struggling a bit for inspiration here, but you get the gist...

- I notice you've gone for options that end in `_off`, rather than starting in `-no`. I prefer the latter, and you'll probably find it's the more common convention (that or `-with_`/`-without_`). For MRtrix in particular, the reason I think it's better is because it plays more nicely with option shortening: `-nointerp` is still clear as an abbreviation of `-nointerpolation`, but `-interp` is not so cool as an abbreviation of `-interpolation_off`... This might get worse if we add the reverse options too (see next point)

- we may need the reverse options to account for the case where off is the default, as per the config file.

- given the above, is it worth having these options take a bool argument? MRtrix will accept 0, false,  no & off or 1, true, yes & on, so it's quite easy to use: `-interp on`, etc.

- don't forget to update the docs using the `docs/generate_user_docs.sh` script when you're about to merge.

Opinions, anyone?

thijsdhollander:
> it would be worth adding a single option to hide all of these elements. Maybe something like -nohud or -bare? Struggling a bit for inspiration here, but you get the gist...

`-hide_annotations`, `-noannotations` or `-annotations_off`.  "annotations", because that's the same word used for the equivalent functionality in the `View` menu in the GUI. (i.e. the `Toggle all annotations` entry)

> I notice you've gone for options that end in _off, rather than starting in -no. I prefer the latter, and you'll probably find it's the more common convention (that or -with_/-without_). For MRtrix in particular, the reason I think it's better is because it plays more nicely with option shortening: -nointerp is still clear as an abbreviation of -nointerpolation, but -interp is not so cool as an abbreviation of -interpolation_off... This might get worse if we add the reverse options too (see next point)

Yep, I agree (I noticed the same, hence my multiple proposals for the annotation one above; just pick the one that goes with the "one" choice).

> we may need the reverse options to account for the case where off is the default, as per the config file.

Fully agree too.

> given the above, is it worth having these options take a bool argument? MRtrix will accept 0, false, no & off or 1, true, yes & on, so it's quite easy to use: -interp on, etc.

That's not a bad idea, although we don't use that all over the place yet for things that are essentially binary switches. Although, I reckon that's mostly because a lot of those cases are things that have a certain *fixed* default, and the switch only needs to be able to switch to the opposite choice.  Since we have the config file here, the switch needs to be able to be flicked both ways indeed...

maxpietsch:
> I can't figure out where the interpolation is set from the config file. There is an entry for it, but it doesn't seem to be used anywhere...?

[Here](https://github.com/MRtrix3/mrtrix3/blob/mrview_config_options/src/gui/mrview/window.cpp#L1764) is where the command line option and [here](https://github.com/MRtrix3/mrtrix3/blob/mrview_config_options/src/gui/mrview/window.cpp#L413) where the config file option is used.

I used `OPTION_off` as I wanted not to interfere with existing options such as `focus` or possible future options such as `interpolation`. I think the best is to add an option `-noannotations` and make the switches boolean switches. If we want to use those options for other things then we could just change the option parsing.

jdtournier:
> Here is where the command line option and here where the config file option is used.

Ah, OK. Sorry, it was just outside the 3-line context in the diffs, so wasn't showing. Forgot it was already in there...

> What do you think about changing -focus to work with one of x,y,z, 0 and 1?

I quite like this personally! Very simple to do with the [`to<bool>()`](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix.h#L319-L327) function (although I notice that it doesn't interpret yes/no as I'd stated earlier - easy fix though):
```C++
try {
  auto pos = parse_floats (opt[0]);
  if (pos.size() != 3) 
     throw Exception ("-focus option expects a comma-separated list of 3 floating-point values");
  set_focus (Eigen::Vector3f { float(pos[0]), float(pos[1]), float(pos[2]) });
}
catch (Exception& E) {
  try {
    show_crosshairs_action->setChecked (to<bool> (opt[0]));
  }
  catch (Exception& E2) {
    throw Exception ("-focus option expects a boolean or a comma-separated list of 3 floating-point values");
  }
}
```

or something like that... You'd need to change to `Option` specification to take an simple string `Argument`, but no big deal...

maxpietsch:
Done.  I use the toggle slot (`[space bar]`) for `-noannotations`. 

Hence, `-noannotations -comments true` shows the comments only whereas ` -comments true -noannotations` shows no overlay. Hope that is not confusing? The neat thing is that one can define what shows after pressing `[space bar]` again.

maxpietsch:
Just had someone request exaclty these features... Feel free to merge if you're happy with the changes.

PR #838, opened 2016-11-28
mrtransform: apply the inverse transformation to DW gradients

jdtournier:
As discussed on the [community forum](
http://community.mrtrix.org/t/gradient-orientation-after-alignment-of-nifti-dwi-data-to-structural-data/584/1).

I think this should fix the issue, but I haven't a chance to test. I'd really like people to had a good look at it to make sure it's actually doing what it's supposed to, so please give it a crack if you have a minute...

### DIFF
--- a/cmd/mrtransform.cpp
+++ b/cmd/mrtransform.cpp
@@ -368,7 +368,7 @@ void run ()
 
   // Rotate/Flip gradient directions if present
   if (linear && input_header.ndim() == 4 && !warp && !fod_reorientation) {
-    Eigen::MatrixXd rotation = linear_transform.linear();
+    Eigen::MatrixXd rotation = linear_transform.linear().inverse();
     Eigen::MatrixXd test = rotation.transpose() * rotation;
     test = test.array() / test.diagonal().mean();
     if (!test.isIdentity (0.001)) {


Lestropie:
Wondered if there was something funky going on here with my own work some time just last week. I don't have quite the same magnitude of rotations in my data; but for scan-rescan data of a single subject with reasonable rotations between scans, I found that after rotating gradient directions _with_ the fix, the resulting FODs appear more consistent & plausible to me.

Note that this will also affect the 'directions' entry in the header if present.

jdtournier:
OK, just verified with some large arbitrary rotation. This commit does fix the issue, both for the DW scheme and for the `directions` entries. 

For the DW scheme, tested with:

    mrtransform ~/test_data/dwi.mif -linear trans.txt - | dwi2tensor ^C- | tensor2metric - -vec pdd_rot.mif

and checking for alignment using MRView's vector plot tool. 

For the `directions` entries, tested with:

    mrtransform ~/test_data/dwi.mif -linear trans.txt - | amp2sh - - | sh2amp - ~/src/directions/dir100.txt amps_rot.mif

and checking for alignment in MRView's ODF display tool, loading as a Dixel image. 

I'll merge now, I think it all checks out.

PR #843, opened 2016-11-30
threads: fix std::bad_alloc bug when running with -nthread 0

jdtournier:


Lestropie:
How were you actually getting to that point? Anything run with `-nthreads 0` is supposed to [branch](https://github.com/MRtrix3/mrtrix3/blob/master/lib/thread_queue.h#L1033-L1039) without that class ever being instantiated.

(No problem with the fix; just might be worth looking one level up also)

jdtournier:
Only for the `run_queue` methods. Not if you [invoke Thread::run() directly](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/tckglobal.cpp#L318)...

jdtournier:
BTW, what do you mean by "looking one level up"...?

Lestropie:
The direct `Thread::run()` invocation would be "one level up".

I presume you're content in leaving such invocations as-is, rather than requiring developers to explicitly branch for the `-nthreads 0` case and avoid calling `Thread::multi()` like is done in `run_queue()`? Given direct access to `Thread::run()` is provided, I think that's the better option.

jdtournier:
> I presume you're content in leaving such invocations as-is, rather than requiring developers to explicitly branch for the -nthreads 0 case and avoid calling Thread::multi() like is done in run_queue()? Given direct access to Thread::run() is provided, I think that's the better option.

Yes, `Thread::run()` was always intended to be usable by devs as-is, in cases where the convenience loop and queue functions weren't appropriate. When a developer writes `Thread::run (Thread::multi (functor))`, that should always launch at least one thread, even if multi-threading has been explicitly disabled - there will be cases where the application would simply not work if we tried to perform the work within the current thread, since the invoking thread might need to interact with the newly-created threads later on. If developers had to handle this case explicitly and not launch _any_ threads, it could get very messy - and needlessly so. So the only safe thing to do here is to launch a single thread. Even though technically that hasn't disabled multi-threading as such, it's at least disabled multiple _worker_ threads...

PR #844, opened 2016-12-01
New script for per-shell mean, rms, etc.

dchristiaens:
This simple script applies `mrmath` to each shell in the input image. The output is concatenated in a 4-D image, where the 4th axis contains e.g. the per-shell mean in order of increasing b-value.

Example usage:
```
mathpershell dwi.mif mean dc.mif
```

### DIFF
--- a/scripts/mathpershell
+++ b/scripts/mathpershell
@@ -1,4 +1,26 @@
 #!/bin/bash
+
+if [ $# -eq 0 ]; then
+  cat <<HELP_PAGE
+This script applies mrmath to each shell in the input image. The output is 
+concatenated in a 4-D image, where the 4th axis contains e.g. the per-shell 
+mean in order of increasing b-value.
+
+USAGE: 
+  $ mathpershell <in.mif> <operation> <out.mif>
+
+  For example: 
+    $ mathpershell dwi.mif mean dc.mif
+  will output the mean diffusion-weighted volume per shell.
+
+AVAILABLE OPERATIONS: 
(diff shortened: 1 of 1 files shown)


Lestropie:
Would `shellmath` or `dwishellmath` be more consistent with the naming convention?

dchristiaens:
Perhaps they would. Do you have a preference? I liked the "pershell" bit as it clarifies that the script is "doing maths" _within each_ shell, not _with_ the shells, but I don't care too much...

jdtournier:
Not too fussed about the name either... I think `dwipershellmath` is probably the most explicit, but it's turning into a bit of mouthful... Up to you guys.

Just a quick note on the implementation though. Have you considered something like this instead of manually deleting the temporaries?
```ShellSession
tmpfiles=""
for b in $(mrinfo "$1" -shells); do
  tmpfiles="$tmpfiles "$(dwiextract "$1" -shell $b - | mrmath - $2 - )
done
mrcat -axis 3 $tmpfiles "$3"
```
That way, `mrcat` deals with the temporary files...

Also, note that I'm stuffing the arguments in quotes everywhere ("$1", etc) - this is to deal with the potential for spaces in the filenames.

dchristiaens:
Continued in #849.

Nice bash scripting @jdtournier, I didn't know that. I've updated the script accordingly.

I'm happy with @Lestropie's suggestion for `dwishellmath`, but I'd like to submit yet another candidate to this naming-brainstorming round: how about `msdwimath`? I think it pairs nicely with `mrmath` and most important it's short...

PR #849, opened 2016-12-02
New script for per-shell mean, rms, etc. (continued)

dchristiaens:
Continued from #844


dchristiaens:
Continued from #849.

Nice bash scripting @jdtournier, I didn't know that. I've updated the script accordingly.

I'm happy with @Lestropie's suggestion for `dwishellmath`, but I'd like to submit yet another candidate to this naming-brainstorming round: how about `msdwimath`? I think it pairs nicely with `mrmath` and most important it's short...

dchristiaens:
Tests are failing for `fod2dec`, although untouched in this branch...

Lestropie:
We kind of moved away from the `ms` prefix; also the script would quite happily work on non-'multi-shell' data. `dwimath` would still be too abstract I think, which is why I suggested `dwishellmath`.

dchristiaens:
`dwishellmath` it is.

thijsdhollander:
> Tests are failing for `fod2dec`, although untouched in this branch...

It seems [your initial commit](https://github.com/MRtrix3/mrtrix3/pull/849/commits/6041a9302331fe4535681bee1c6197db03709be1) changed the commit the test data submodule is referring to, to an earlier one where the `fod2dec` data wasn't present yet.  So the tests fail because they can't find the data.

thijsdhollander:
Waiting until the tests complete again, but [this](https://github.com/MRtrix3/mrtrix3/pull/849/commits/2a85f1b0bbdb9366b90b527f78054e0b8646ad76) should have fixed it again.

dchristiaens:
Thanks @thijsdhollander. You're right, I screwed up in that first commit.

Issue #848, opened 2016-12-02
mrview: add go to specific volume entry


maxpietsch:
It seems we don't look at a lot of fMRI data with thousands of volumes... Going to the last volume in the GUI can take a while, especially over slow connections. It would be nice to have a `go to volume...` menu entry and corresponding shortcut ("G"?).

jdtournier:
Not a bad idea. In the meantime, isn't there a control for that in the View tool...?

maxpietsch:
I didn't think about the View tool. Might be just me but I wouldn't search for it in a tool called "View options".

thijsdhollander:
As I mentioned in the pull request: I really like this.  Even though available in the view options, not everyone may easily discover it there indeed, but also the addition of a key combination to do this quickly is in an of itself a worthwhile thing, I reckon.  It's simple things like this that make the viewer a very powerful multi-purpose tool!

jdtournier:
I like the `G` / `Shift-G` shortcut idea for volume / volume group control. That was always a funny shortcut to set the background colour anyway... Not sure what to assign that action to now, maybe `U` for 'underlay colour'? We could also modify the text in the menu to match...?

jdtournier:
Crap, I guess this issue is obsolete now, should discuss in #852... Let's close this off.

PR #852, opened 2016-12-05
mrview: go to volume slot #848

maxpietsch:
Added "go to volume..." entry in the image menu. issue: #848

### DIFF
--- a/src/gui/mrview/window.cpp
+++ b/src/gui/mrview/window.cpp
@@ -354,6 +354,10 @@ namespace MR
           prev_image_volume_action->setShortcut (tr ("Left"));
           addAction (prev_image_volume_action);
 
+          goto_image_volume_action = image_menu->addAction (tr ("Go to volume..."), this, SLOT (image_goto_volume_slot()));
+          goto_image_volume_action->setShortcut (tr ("Shift+g"));
+          addAction (goto_image_volume_action);
+
           next_image_volume_group_action = image_menu->addAction (tr ("Next volume group"), this, SLOT (image_next_volume_group_slot()));
           next_image_volume_group_action->setShortcut (tr ("Shift+Right"));
           addAction (next_image_volume_group_action);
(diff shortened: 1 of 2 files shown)


thijsdhollander:
@maxpietsch I really like this addition, even just for the fact that it can be accessed with a shortcut key combination, and the number field gets selected automatically; so everything can be done keyboard-wise.  Awesome!

Only thing is the key combination itself: I appreciate you went for `Shift+G` because `G` was already taken by the background colour selection one.  But the latter is definitely a less basic operation (and people surely haven't got too used to the key for that one, I reckon), and one that you typically wouldn't do as frequently as switching volumes in a large dataset.  So what I'd suggest is:

- Tie the new "Go to volume..." to just `G`
- If the background one is going to become a `Shift+...` anyway, tie it to `Shift+B` (makes more sense than `G` even, but simple `B` in that case is already taken by the colour bar, which is still a more basic toggle than the background colour)

Just typing `G`, a number and hitting `Enter` would definitely make this a really handy thing that I'd use frequently, even in smaller datasets, to go to a specific volume across `mrview` sessions. :+1:

thijsdhollander:
Small addition: `Shift+G` could even become a "Go to volume group...`, would be very consistent with the keys for next/previous volume vs volume group.

jdtournier:
I like the G / Shift-G shortcut idea for volume / volume group control. That was always a funny shortcut to set the background colour anyway... Not sure what to assign that action to now, maybe U for 'underlay colour'? We could also modify the text in the menu to match...?

maxpietsch:
Happy with `G` and `shift + g`. Do we need a shortcut for the background colour? Doesn't strike me to be a thing that one would need constantly.

jdtournier:
> Do we need a shortcut for the background colour? Doesn't strike me to be a thing that one would need constantly.

I was having the same thought... Happy to drop the shortcut altogether if everyone is happy with that.

maxpietsch:
Done. Shortcuts: `g` and `shift+g`. for volume and volume group. Removed background colour shortcut.

maxpietsch:
In case anyone want the background shortcut back: background could become canvas: `Shift+c` is still free I think.

maxpietsch:
Feel free to press the button before master changes again :)

PR #858, opened 2016-12-09
Scripts signal handling

Lestropie:
Will catch Ctrl-C or anything else terminating a Python script, provide a terminal error message comparable to those given by the signal handler now implemented in the binaries, and sends `SIGTERM` to any underlying subprocesses so that they can clean up as best they can (including deleting any piped MRtrix images).

Remaining question is whether or not the deletion or non-deletion of the script temporary directory should be influenced by the fact that a signal has been received...?

### DIFF
--- a/scripts/lib/app.py
+++ b/scripts/lib/app.py
@@ -43,7 +43,8 @@ def addCitation(condition, reference, is_external):
 
 
 def initialise():
-  import os, sys
+  import os, signal, sys
+  import lib.signalHandler
   from lib.errorMessage          import errorMessage
   from lib.printMessage          import printMessage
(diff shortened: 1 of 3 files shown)


Lestropie:
Re-implemented in 4ca4247. Required some [trickery](https://github.com/MRtrix3/mrtrix3/pull/898/commits/4ca4247c3cd172254b2668e5010879e27c4eb9e6#diff-15ea8e46c5e64749b52b4dff228aec59R153) due to throwing `KeyboardInterrupt` on my Windows machine (I don't recall encountering this the first time around, maybe I was working on Linux that day?), so needs eyes/testing.

PR #862, opened 2016-12-13
Integrating amp2response in dwi2response scripts

thijsdhollander:
Additionally, default lmax=10 for anisotropic responses. Follows some discussion in #786.

### DIFF
--- a/cmd/amp2response.cpp
+++ b/cmd/amp2response.cpp
@@ -66,6 +66,8 @@ void usage ()
     + Argument ("response", "the output zonal spherical harmonic coefficients").type_file_out();
 
   OPTIONS
+    + Option ("isotropic", "estimate an isotropic response function (lmax=0 for all shells)")
+
     + Option ("noconstraint", "disable the non-negativity and monotonicity constraints")
 
     + Option ("directions", "provide an external text file containing the directions along which the amplitudes are sampled")
@@ -154,7 +156,11 @@ void run ()
   std::vector<int> lmax;
   int max_lmax = 0;
   opt = get_options ("lmax");
-  if (opt.size()) {
+  if (get_options("isotropic").size()) {
+    for (size_t i = 0; i != dirs_azel.size(); ++i)
+      lmax.push_back (0);
+    max_lmax = 0;
+  } else if (opt.size()) {
     lmax = parse_ints (opt[0][0]);
     if (lmax.size() != dirs_azel.size())
       throw Exception ("Number of lmax\'s specified (" + str(lmax.size()) + ") does not match number of b-value shells (" + str(dirs_azel.size()) + ")");
@@ -169,15 +175,15 @@ void run ()
     // Auto-fill lmax
     // Because the amp->SH transform doesn't need to be applied per voxel,
     //   lmax is effectively unconstrained. Therefore generate response at
-    //   lmax=8 regardless of number of input volumes.
+    //   lmax=10 regardless of number of input volumes.
     // - UNLESS it's b=0, in which case force lmax=0
     for (size_t i = 0; i != dirs_azel.size(); ++i) {
       if (!i && shells && shells->smallest().is_bzero())
         lmax.push_back (0);
       else
-        lmax.push_back (8);
+        lmax.push_back (10);
     }
-    max_lmax = (shells && shells->smallest().is_bzero() && lmax.size() == 1) ? 0 : 8;
+    max_lmax = (shells && shells->smallest().is_bzero() && lmax.size() == 1) ? 0 : 10;
   }
 
   auto image = header.get_image<float>();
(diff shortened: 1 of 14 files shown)


thijsdhollander:
So for reference: what I've done here is as a primary thing up the lmax to 10 and integrate `amp2response` in the `dwi2response` scripts.  However, as I was doing that, I came across a range of bugs, most of which I reason originated because we once considered single-shell as the default data that typically came in, as well as due to some stuff making other assumptions about the users' input.  For instance, `dwi2response tax` was actually grossly wrong when multi-shell data came in: the only way the data was processed when it came in, was (twice: once in the general script, and once in the tax code itself) a simple `dwiextract`.  All good for actual single-shell data, but for multi-shell data everything (but the bzeros) came in.  For most single-shell response algos, that was not a major issue (apart from wasting disk space and time), because the `amp2sh` took care of it; but e.g. in the tax algorithm, the amplitudes of the `dwi.mif` were also used, assuming they were a single shell (but they aren't in case of multi-shell data).

Anyways, that led me to reconsider some small aspects of the main `dwi2response` script, so single-shell algos would only get the real single-shell part.  However, that led me further to look at `dwiextract`, to see if useful functionality could be added to that one to make this all a bit easier (for scripts in general).  I thought that would be a simple one as well, as there already appeared to be such functionality in the `select_shells` method in `shells.cpp`.  So I added it, tested it, and then stumbled across a range of unexpected behaviours / bugs in `select_shells`.  To just name a couple (there were many more): selecting shells like 3000,0 under a single shell regime would fail as if it where multiple non-zero b-values.  Assumption made here was that the user would select 0 as the first one.  Things like 5,3000 would fail here too, because even though 5 would translate into a bzero, it wasn't detected at that check.  Things like 0,3000,3005 would go wrong often as well, again because checks where on what the user provided, but didn't take into account into what selection such things translate; so this may result in commands not actually getting the kind of data they desire, and the users potentially not being aware of this happening.  And if there was no "-shells" option, but the single-shell constraint was demanded, but not the bzero, then the bzero got removed if the data was already single-shell out of the box, but not if it was multi-shell out of the box (then suddenly the bzero was in).  And so on, and so on...

Long story short: I could actually see how some of these bugs had been introduced over time, and it was quite in line with an old single-shell logic that was migrated to a multi-shell one, but via smaller disconnected hacks over time.  The `select_shells` code was also not the easiest one to navigate through, due to lots of nested logic going on in there.  So I wrapped my mind around all of that, and rewrote the bits that had to do with these various optional constraints that you can ask from `select_shells`.  The bugs are ironed out (and lots of extra checks have been added), and the behaviour and functionality offered to developers should be more clear now.  Whatever was possible before, is still possible (and more), and apart from bugs, the behaviour of commands that uses this functionality is still the same (I took care of each separate occurrence to consider this well, and tested it extensively).  The relevant documentation that explains the behaviour (so future developers don't have to navigate the code of this nested logic to find out) was added [here](https://github.com/MRtrix3/mrtrix3/pull/862/commits/704cb75d5d348f354b40513d5fdbb8c5a4a41f2d).  This should give them all the tricks I can imagine would be requirements that they may have for incoming dMRI data.  Also no worries about the shell assignment logic: this part of it is unaffected.

Since having this functionality may also be useful when writing certain scripts, I basically offered it externally via the `dwiextract` interface as well; again no worries: everything is backwards compatible, no existing behaviours have been changed, and for a simple `dwiextract` or `dwiextract -bzero`, it still doesn't initialise the `Shells` class, as before. The new features are only relevant if people start to play around with shells, or are having shell requirements, anyway; so in a context where we're definitely dealing with shelled data.

Finally, there's also an added `-isotropic` option for `amp2response`.  It's a shortcut for `-lmax 0,0,0,..." essentially, but it's more easy to use, and independent of how many shells there are.  It's also just quite nice in context of scripts that e.g. get WM, GM and CSF responses, where the latter is/are isotropic; makes those scripts very readable as well.

All of this combined also resulted in a much smaller disk footprint of some `dwi2response` algos: no more SH images that have to be stored in between, single-shell algos only copying the part of the dMRI dataset they really need, etc...

thijsdhollander:
Maybe final finding I should mention: the output of all `dwi2response` algos will now of course not be exactly the same any more compared to before (and it's lmax=10 by default), but where this could've potentially have had a greater impact, was with iterative algorithms like the `tournier` one, as the output is used to initialise each next iteration.  However, I've tested this in a range of scenarios, and the output (and the final voxel selection) is very close to the same in all cases where the data quality is ok/fine.  In some cases, the voxel selection even ends up being exactly the same as before, which I reckon is a good thing.  The `tax` algorithm is now amazingly slow though, especially in the first or earlier iterations in general (but I haven't compared to before, it may already have been quite slow...?).  The huge number of voxels it feeds to `amp2response` early on (and even still quite a bit later on) is the main cause of this.  Oh yeah, and final thing now that I'm mentioning the `tax` algorithm: I also ported it to using the new fixel format while I was at it.  It seems @draffelt may have overlooked this one. :wink:

PR #866, opened 2016-12-21
allow setting of number of threads from environment variable

jdtournier:
This now also checks whether the MRTRIX_NTHREADS environment variable is
set, and if so will use that value instead of reading it from the config
file. However, the -nthreads option still take precedence.

Discussed on the forum:
http://community.mrtrix.org/t/control-multi-threading-on-a-cluster/632?u=jdtournier

### DIFF
--- a/lib/thread.cpp
+++ b/lib/thread.cpp
@@ -51,7 +51,7 @@ namespace MR
         return __number_of_threads;
       }
 
-      __number_of_threads = std::thread::hardware_concurrency();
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Sounds great, and should definitely help people setting up and managing MRtrix on a cluster. 👍

jdtournier:
Good thing I hadn't merged yet... All tested and seems to work fine now though - I'll merge.

PR #875, opened 2017-01-03
MRView: process command-line options after loading images

### LINKED ISSUE #871: mrview - segmentation fault with -overlay.load
Dear experts,

I encountered segmentation fault when running mrview with option -overlay.load.

The behaviour is very strange, the error occurs in only some precise conditions:
1. when main image is 4D dataset AND when overlay image is in .nii.gz format, or
2. when both main image and overlay image are 4D datasets.

AND

commandline is NOT using -load option, i.e. command is like:
mrview main_image -overlay.load overlay_image

In other cases when main image is 3D dataset (both main image and overlay image is 3D; main image is 3D, overlay image is 4D) it seems to work OK.

Loading the images with overlay tool in GUI or with specifying -load option works OK in all tested cases (i.e. also with images where loading via -overlay.load produces segmentation fault).

Also, when -load option is NOT specified, i.e. command is like:
mrview main_mage -nointerpolate
the -nointeropolate option is ignored, which is quite confusing.

My version of mrtrix is following:
0.3.15-370-gff066551

Regards,

Antonin Skoch

jdtournier:
Fixes #871

### DIFF
--- a/cmd/mrview.cpp
+++ b/cmd/mrview.cpp
@@ -64,6 +64,7 @@ void run ()
 {
   GUI::MRView::Window window;
   window.show();
+  qApp->processEvents();
 
   if (argument.size()) {
     std::vector<std::unique_ptr<MR::Header>> list;
@@ -80,6 +81,8 @@ void run ()
     if (list.size())
       window.add_images (list);
   }
+
+  window.process_commandline_options();
 
   if (qApp->exec())
     throw Exception ("error running Qt application");
--- a/src/gui/dialog/dicom.cpp
+++ b/src/gui/dialog/dicom.cpp
@@ -16,6 +16,7 @@
 #include "file/dicom/tree.h"
 #include "gui/dialog/list.h"
 #include "gui/dialog/dicom.h"
+#include "gui/gui.h"
 
 namespace MR
 {
@@ -150,7 +151,7 @@ namespace MR
         class DicomSelector : public QDialog
         {
           public:
-            DicomSelector (const Tree& tree) : QDialog (NULL) {
+            DicomSelector (const Tree& tree) : QDialog (GUI::App::main_window) {
               Model* model = new Model (this);
 
               Item* root = model->rootItem;
--- a/src/gui/dialog/file.cpp
+++ b/src/gui/dialog/file.cpp
@@ -39,7 +39,7 @@ namespace MR
 
 
 
-        std::string get_folder ( QWidget* parent, const std::string& caption, const std::string& folder) 
+        std::string get_folder (QWidget* parent, const std::string& caption, const std::string& folder) 
         {
           QString qstring = QFileDialog::getExistingDirectory (parent, caption.c_str(), folder.size() ? QString(folder.c_str()) : QString(), QFileDialog::ShowDirsOnly | FILE_DIALOG_OPTIONS);
           if (qstring.size()) {
--- a/src/gui/dialog/progress.cpp
+++ b/src/gui/dialog/progress.cpp
@@ -26,17 +26,30 @@ namespace MR
       namespace ProgressBar
       {
 
+        namespace {
+          QProgressDialog* progress_dialog = nullptr;
+        }
+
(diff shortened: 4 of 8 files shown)


jdtournier:
OK, wouldn't mind people double-checking that this doesn't introduce further issues, since I did also simplify the progress bar handling somewhat, which wasn't strictly required for this fix. For those interested: the [`invokeMethod()`](http://doc.qt.io/qt-5.7/qmetaobject.html#invokeMethod) call seemed redundant when used with `DirectConnection`, as far as I can tell it reduces to a direct invocation, so I basically got rid of that - I'd introduced it in the hope that I could get the main UI thread to handle the request, but that proved not to work. So I've cleaned that up. It works fine on my system, but there is a chance that it might have been required for different OS's or versions of Qt. 

@thijsdhollander, @Lestropie: I've assigned you two for reviews, since you seem to have a knack for making things fall over 😜. Also, I'd like this tested on MacOSX too, just to make sure...

maxpietsch:
Can't break it on OSX with combinations of -load -overlay.load and other command line options using large nii.gz files.

thijsdhollander:
Since it's merged in `master` and the 0.3.16 tag branch, I'll keep an eye on things and maybe push a bit here and there at times to see if stuff falls over. :wink:  ...but it seems good so far!

ansk:
I confirm, that #871 segfault does not appear anymore.

Just to comment on the current functionality:

1.
In case of loading of multiple images to `mrview` (without using `-overlay.load`), what is the difference between `mrview 1.nii.gz 2.nii.gz` and `mrview 1.nii.gz -load 2.nii.gz` ?

2.
As @maxpietsch pointed out in example

> Please note that -interpolation 0 is applied to only the active image. `mrview 1.nii.gz -interp 0 -load 1.mif` will cause `1.nii.gz` not to be interpolated and `1.mif` to be interpolated.

Is it possible to switch interpolation off both in `1.nii.gz` and `2.nii.gz` using command-line option?

maxpietsch:
> Is it possible to switch interpolation off both in 1.nii.gz and 2.nii.gz using command-line option?

No, `mrview` currently does not allow multiple options `-interpolate` because I wanted to avoid confusion:  If images are loaded via `-load` then using multiple `-interpolate` options makes sense. However, if images are loaded as arguments then they are processed before the command line options so only the last image is accessible. We could apply `-interpolate` to all open images as an alternative.

ansk:
I like the way how this is handled in `freeview` (freeSurfer). For each file loaded via command line option there is list of sub-options (delimited by `:` ). This allow more flexibility. For example, to make first image opaque and second half-transparent, the command line is:
`freeview -v image1.nii.gz:opacity=1 -v image2.nii.gz:opacity=0.5`
But this would require to completely change the design.

As I tested that, currently, when more images are specified as arguments, the `-interpolation 0` is applied only to the first (not last) one specified (which is also primarily shown in `mrview` start) which confused me somehow at start.

I achieved to switch interpolation off in all images by setting

`ImageInterpolation: false`
in  `.mrtrix.conf` file, so I am satisfied now :-)

jdtournier:
@maxpietsch is spot on: the reason for the difference in how images are handled when listed as standard command-line arguments as opposed to via the `-load` option is that these are processed at different times. It should however be relatively trivial to sort this out so that arguments are handled as if they had been listed with the `-load` option. For reference, it would require removing the argument handling from `cmd/mrview.cpp` ([these lines](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrview.cpp#L72-L83)), and modifying the looping over options that happens [here](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrview.cpp#L72-L83), so that it loops over all tokens (whether an option or an argument) and handles them appropriately ([much like the `mrcalc` does](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrcalc.cpp#L1062)). Should be a trivial change, if anyone wants to jump in...?

Issue #815, opened 2016-11-08
Image::row() with negative strides


Lestropie:
The code [here](https://github.com/MRtrix3/mrtrix3/blob/master/lib/image.h#L97-L104) obtaining a map for accessing all data along a particular axis of an image (e.g. data for all volumes for a particular voxel) fails an assertion in Eigen 3.2.93 if the axis of interest has a negative stride.

So we need to look into whether there's a way to 'trick' Eigen so that this functionality can still be used, or perhaps whether we can provide our own map wrapper class (albeit perhaps losing Eigen functionality / compatibility).

jdtournier:
I've been thinking about our increasingly heavy usable of the `row()` method and some of the problems it introduces. I think we need to rethink how it's implemented and used in the code, as things stand it's too fragile, and not necessarily as efficient as it might seem. 

- The first thing is that the image needs to be access with direct IO, which can already introduce issues when developers forget to specify this. It's easy to overlook in testing since MRtrix will try very hard to memory-map its input it at all possible, so will basically access the image with direct IO if it can. And our apps are typically geared to producing outputs that are suitable for direct mapping in downstream apps (e.g. `dwi2fod` will produce output with volume-consecutive strides in `Float32` format, which is what `tckgen` will expect). But things will fall over the minute someone else feeds in data that isn't directly suitable for it. It's easy to fix by adding the `.with_direct_IO()` call, but it feels wrong to have such a low-level call be so dependent on how the image was accessed in the first place. And also forces preloads in cases where they might not have been necessary (e.g. because the data type doesn't match, but it's still a single-pass application). 

- the way it's implemented, we have to set the returned vector (or rather `Eigen::Map`) as `Eigen::Unaligned`. This means it's _not_ suitable for vectorisation. So any performance-critical section that uses this vector (i.e. just about anything we do with it) will either be several times slower (no vectorisation) or will involve an on-the-fly copy to registers (an overhead that would hinder performance it the vector is used in multiple operations). 

- and finally, as @Lestropie points out, it's not even guaranteed to work properly when the relevant stride is negative. _MRtrix3_ won't bother preloading if the strides match _ignoring their sign_ - this is because performance won't be affected by this, the CPU's memory access prediction engine will be perfectly capable of dealing with this, and cache usage remains just as efficient. We could force a preload in these cases, but this feels a bit OTT, especially given all the concerns above. 

[...]

PR #882, opened 2017-01-18
New row method - take 2

jdtournier:
OK, had a go at this following #815. The approach taken is that summarised in [this post](https://github.com/MRtrix3/mrtrix3/issues/815#issuecomment-273100702). Basically, `.row(n)` now returns a lightweight proxy object to represent a row vector, but is not itself an Eigen object - so it can't be used directly in Eigen expressions. The advantage is that it can be used for _any_ image (whether direct IO or not), and isn't affected by the negative strides issue. 

It currently build off the `master` branch, but I think we'll probably want to merge into the `tag_0.3.16` branch given that the changes are fairly deep. I'll fix up any errors that occur in the process - chances are something will fall over... 

In any case, here's an overview of how it works:

There are methods provided to assign/modify it from Eigen objects (operators `=`, `+=` & `-=`):
```C++
// mat & vecs here are Eigen objects:
out.row(3) = mat * vec + vec2;
out.row(3) -= vec3;
```
Type-casting is done implicitly (no need to use `.cast<>()`:
```C++
Image<float> out;
out.row(3) = Eigen::Vector3d (1.0, 2.0, 3.0);
```
There are also methods to assign/modify from a single scalar (operators `=`, `+=`, `-=`, `*=` & `/=`):
```C++
out.row(3) = 0.0;
out.row(3) *= scale_factor*global_scale_factor;
```
and also methods to assign/modify from other instances of the `.row()` method:
```C++
out.row(3) = in.row(3);
```
again with implicit type-casting:
```C++
Image<int> in;
Image<float> out;
out.row(3) = in.row(3);
```

---

For convenience, I've also added `.size()` and `operator[]` methods (read-only at the moment):
```C++
Eigen::VectorXd vec (in.row(3).size()); // same as in.size(3)...
double val = in.row(3)[5];
```

---

Assignment from `.row()` to Eigen objects is made possible using [Eigen's plugin system](http://eigen.tuxfamily.org/dox/TopicCustomizing_Plugins.html). I've added a constructor for `Eigen::Matrix` (again with implicit type-casting):
```C++
Image<float> in;
Eigen::VectorXd vec (in.row(3));
```
and assign/modify operators (`=`, `+=` & `-=`) for general Matrices and Arrays:
```C++
Eigen::VectorXf vec;
vec = in.row(3);
vec += in2.row(3);

Eigen::MatrixXd mat;
mat.col(3) = in.row(3);
mat.diagonal() += in2.row(3);
```

---

For more complex expressions where you might want to mix the `.row()` method with Eigen expressions, things can get a little more complicated. When the size of the row is a compile-time constant, the best thing to do is to cast the `.row()` to e.g. an `Eigen::Vector3d` on the fly:
```C++
// relies on "in" being a 4D, 3-volume image:
out.row(3) = vec + mat * Eigen::Vector3d (in.row(3));
```
Otherwise, you'll need to cast or assign to a dynamically allocated vector. You can still do this on the fly:
```C++
out.row(3) += mat * Eigen::VectorXd (in.row(3));
```
but if this is in a performance-critical section of code, it's best to allocate this vector outside of the loop (e.g. as a class member), and assign to it beforehand:
```C++
Eigen::VectorXd vec_in (in.size(3));

...

// in loop:
  vec_in = in.row(3);
  out.row(3) += mat * vec_in;
```

---

So I think that more or less covers everything. I've gone through the code and modified everything - at least to the extent that the build complete without errors or warnings, and the tests run (at least those that weren't already failing - using Eigen 3.3.1 now). Feel free to have a play, comment, etc.


jdtournier:
Oh, forgot to add: this also brings in [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) for `ImageType` objects - including the main `Image` as well as `Adapters`. The base class implements the `.index()` and `.value()` methods, and invokes the derived class's `.get_index()` / `.move_index()` and `.get_value()` / `.set_value()` calls respectively to operate. All I had to do for the `.row()` method was to add it to the new `ImageBase` class... :+1:

jdtournier:
OK, having merged `tag_0.3.16` into this branch, the only tests failing now are for `mrfilter`, and they all seem to be precision-related. I guess I'll relax those tests and see what happens...

jdtournier:
I just took out the `-frac` option from the `mrfilter -gradient` tests - it was still failing even after relaxing the threshold. I think this option doesn't really work most of the time, not in the way it was intended, at any rate. In cases where multiple intensities from different sources are combined (multiple voxels in this case), the errors relate to the magnitude of the original intensities, not that of the final result. You end up with failures with apparently quite relaxed tolerances, because results computed from large values that happen to come out near zero have a much higher fractional error. I think most of the time, it's better to specify an absolute tolerance that's set relative to the expected overall intensities in the images involved in the computation, even if it requires a bit of tuning to the data being used in the testing.

So I'd argue that most of the tests currently running with `-frac` probably ought to use an appropriate absolute value instead...

Lestropie:
If it makes more sense to revert specific tests to absolute then by all means do so (I likely got a few decisions wrong when I went through the list), but please don't mass-revert. `-frac` makes sense in a number of instances, makes tests invariant to use of new test data where overall intensities may change, and means the test instructions make a lot more sense when you read them. But anything involving a subtraction will make `-frac` less stable.

Maybe a more direct solution: Calculate a 'tolerance image' by multiplying the input image by e.g. 1e-5, and feed *that* to the testing binary instead of a relative / absolute threshold.

jdtournier:
> If it makes more sense to revert specific tests to absolute then by all means do so (I likely got a few decisions wrong when I went through the list), but please don't mass-revert.

:rofl: no, wasn't planning on doing anything quite so extreme... I just think that `-frac` shouldn't be the default go-to option here, there's too many instances where it's not the right approach. These include any command where all volumes are processed together per-voxel (e.g. anything to do with DWI really, `dwi2fod`, etc), since e.g. some SH or DT coefficients might come out close to zero even though the data used in the computation is large. This is why there is now a `-voxel` option that basically does `-frac` relative to the max amplitude per-voxel across volumes. Not perfect by any stretch, but at least robust to the odd coefficient being near zero. 

The other case where this is a problem is for processing where neighbouring voxels are included in the computation - this is the case particularly for `mrfilter`. For many of the filters, it's not so much of a problem, since the output intensity is to some extent the average of the input values. But the gradient filter is an exception here, being based on differences rather than sums. The FFT filter would also be problematic here, I reckon. In any case, the `-frac` option is going to be troublesome in these cases also, since there's no guarantee that the error in the output relates to its intensity, it'll relate to the intensities of the _input_ data, which we don't currently look at.

And then you have gems like `dwidenoise` that operate along _all_ dimensions of the image, where `-frac` really won't make sense at all...

> Maybe a more direct solution: Calculate a 'tolerance image' by multiplying the input image by e.g. 1e-5, and feed that to the testing binary instead of a relative / absolute threshold.

Could do. Just have an additional argument to `-frac` to accept one of the input images, and use that to set the actual tolerance per data point. Still, seems a bit heavy to me, when a single scalar intensity representing the typical input value would probably be enough. Maybe what we could do is add a `-typical` option to state the expected input intensity, and then the tolerance would be set relative to that? So testing on commands that process the `dwi.mif` could test using `-typical 200 -abs 1e-5`? This would make the testing robust to small values, and still allow the user to supply a more interpretable tolerance value. I agree it would still not be invariant to the input data, but then the tests generally require a lot of double-checking as soon as you do change the input and test data, it wouldn't take much to just make sure that the ballpark intensity in the input matches what is stated in the test files. 

Other option is to use test data that has average intensity 1... Not advocating this though, it would involve overhauling all of the input and test data, which would take quite a bit of effort, and would need a lot of double-checking.

Anyway, that's just my 2 cents. Really, I think we're slightly over-thinking this, in practice the differences we're seeing between Eigen versions or between processors are large enough that there's little point in trying to be too precise about this... Like I said elsewhere, we just want to make sure the results are consistent with expectations given the variations in architecture and software version that we're likely to encounter in the wild.

Lestropie:
> Maybe what we could do is add a `-typical` option to state the expected input intensity, and then the tolerance would be set relative to that?

That's exactly where my train of thought started. But why require that value to be manually specified when you could instead e.g. run `mrstats -output mean` and feed that as the input to `-typical`? But then if you're going to go to that effort, it's still not going to perform ideally for non-homogeneous input image intensities; hence why not use the source image itself to set the threshold? I.e. the tolerance image idea.

jdtournier:
I guess the issue is about effort required vs. how big a problem this really is. When you say

> But then if you're going to go to that effort, it's still not going to perform ideally for non-homogeneous input image intensities

my take is that I don't think we need it to perform _ideally_.  It's a safeguard to catch bugs that affect the results to a noticeable extent. And machine precision is way below what's noticeable 99% of the time. Even if it's not perfect for non-homogeneous images, the intensities in the input data will only typically vary by a factor of 10, maybe 100 in extreme cases. So that would mean the tolerance would be too liberal in the low intensity voxels, but if it's set to e.g. 1e-5, then that would still be effectively 1e-3 for those regions, which I think is plenty enough to catch _visible_ differences. And it's a _lot_ easier to add that option to `testing_diff_data` (or is it `testing_diff_image` now?) and gradually update the tests as required, than to change all the tests over and either add commands to compute the tolerance image, or add precomputed tolerance images to the testing repo. 

Or to put it another way: realistically, how many additional _genuine_ regressions would we expect to catch by going for such an approach instead of sticking with an appropriately tuned absolute threshold?

Lestropie:
'Appropriately' tuning thresholds isn't *fun*. Getting data-driven thresholds from input images *is*. :-P

In other news: Main thing that caught my eye is [`Helper::ConstRow::image` being public](https://github.com/MRtrix3/mrtrix3/pull/882/files#diff-6afe43832f161b961e05fbcfe1cf9c6cR496). Could theoretically lead to very unusual usage. May not matter, but currently having a play trying to privatise it nonetheless.

jdtournier:
> Main thing that caught my eye is Helper::ConstRow::image being public. Could theoretically lead to very unusual usage. May not matter, but currently having a play trying to privatise it nonetheless.

Might be tricky: the [assignment operators that I added to the Eigen base classes](https://github.com/MRtrix3/mrtrix3/blob/9476fa0553ab5de2e2a316efb7e3cbe59015be8e/lib/eigen_plugins/dense_base.h#L4-L6) need access to these members...

jdtournier:
> Happy now that MR::Helper::ConstRow::image is protected.

Good for you. 

However, the [build now fails](https://travis-ci.org/MRtrix3/mrtrix3/builds/193382797): 
```
In file included from cmd/fod2dec.cpp:21:
In file included from ./lib/algo/threaded_loop.h:20:
In file included from ./lib/algo/loop.h:22:
./lib/image_helpers.h:499:11: error: too few template parameters in template redeclaration
          template <class MatrixType>     friend class Eigen::Matrix;
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
./lib/image_helpers.h:507:16: note: in instantiation of template class 'MR::Helper::ConstRow<MR::Image<float> >' requested here
        public ConstRow<ImageType> 
               ^
cmd/fod2dec.cpp:124:17: note: in instantiation of template class 'MR::Helper::Row<MR::Image<float> >' requested here
```

Problem being that [`Eigen::Matrix` takes _6_ templates arguments](http://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html#a8f7eef9d36f1057338309afb339c1661)... 

But really, I don't see that we need to do this. If this was Python, you wouldn't even have the option of protecting members like this, and no-one would be worried about it... It's not like the `image` member is somehow inaccessible otherwise and needs to be kept away from the developer, it's not an opaque detail of the implementation. I wouldn't worry about it, it's designed to be used as a temporary anyway, if people decide to do this:
```C++
auto row = image.row(3);
row.image.index(1) = x;
row = Eigen::vector3f (a,b,c);
``` 
or something, then that's just daft - but not actually a problem in terms of the operation of the row, it'll behaves exactly as expected. In fact, even if you do manage to protect the `image` member, this would still be perfectly legal and in my opinion more dangerous:
```C++
auto row = image.row(3);
image.index(0) += 10;
row = Eigen::Vector3f (a,b,c);
```
since the developer might have assumed that the row being accessed would what it was at the point when it was instantiated. But since it's just a thin reference to the original image, changing the position of the image directly like this means that the row assigned to will be different from what might be expected. 

So really, I think this would be better addressed with a bit of documentation explaining that it's designed to be used as a temporary (rvalue) only.

Lestropie:
I was expecting to have to define the Matrix friend class with 6 template parameters, but it compiled on my system so I left it :-/  I'll have another go and if it still doesn't play ball I'll just revert.

Agree it's not the end of the world if it has to stay public. But you could make the same argument for every single protected / private class member in the entire repo. If you weren't familiar with Eigen plugins etc., you would expect such a member to not be accessible if it's not supposed to be part of the interface for that class.

Lestropie:
`clang` and `gcc` now both happy.

jdtournier:
Glad you got it working. Feel free to merge when you guys are happy with these changes, I reckon it's good to go?

jdtournier:
Actually, don't bother, I'll merge this into the `eigen_memalign` branch, which is now in sync with `tag_0.3.16`. It resolves some of the failing tests on that branch, and it's quite a fiddly merge (having just resolved all conflicts on my system). I'll assume you're all happy with this, merge this with `eigen_memalign`, and then we can figure out the remaining test failures before merging the lot with `tag_0.3.16`.

PR #884, opened 2017-01-22
Mrview cmdline parsing

jdtournier:
This implements the changes I'd [suggested here](https://github.com/MRtrix3/mrtrix3/pull/875#issuecomment-271538907), in response to a discussion with @ansk. Basically, images provided as regular arguments will now be opened as if they had been supplied with the `-load` option, meaning any following options apply to the expected image, etc. 

One thing this does change is that the image initially displayed when the command is issued is the _last_ one specified, whereas currently this would be the first one. Not sure whether this is a big deal, but I don't think there's any way to avoid this given how we want this to work.

Also, to make something like this actually work:

    mrview a.mif -interp off b.mif -interp off

I needed to add `.allow_multiple()` to the `-interpolation` option. So I've now gone and made most options `.allow_multiple()` in the main viewer window (i.e. those defined in `src/gui/mrview/window.cpp`, but I reckon it might make sense to allow just about all options to appear multiple times from all the other tools too. Would be great if everyone could have a look at their tools and add `.allow_multiple()` to all options where it might make sense. If we do this, this could allow quite complex operations, like multiple screen captures all on the same command-line, etc. 

As always, feel free to make any suggestion, comment, etc.

### DIFF
--- a/cmd/mrview.cpp
+++ b/cmd/mrview.cpp
@@ -64,24 +64,6 @@ void run ()
 {
   GUI::MRView::Window window;
   window.show();
-  qApp->processEvents();
-
-  if (argument.size()) {
-    std::vector<std::unique_ptr<MR::Header>> list;
-
-    for (size_t n = 0; n < argument.size(); ++n) {
-      try {
-        list.push_back (std::unique_ptr<MR::Header> (new MR::Header (MR::Header::open (argument[n]))));
-      }
-      catch (Exception& e) {
-        e.display();
-      }
-    }
-
-    if (list.size())
-      window.add_images (list);
-  }
-
   window.process_commandline_options();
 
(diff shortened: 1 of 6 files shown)


jdtournier:
OK, just about all options are `.allow_multiple()` now - only exception is the connectome tool. @Lestropie, maybe you can make a decision on that one, if it makes sense to allow those options to be specified multiple times. Otherwise, if everyone is happy with the changes overall, it should be good to merge.

Lestropie:
> OK, just about all options are .allow_multiple() now - only exception is the connectome tool. @Lestropie, maybe you can make a decision on that one, if it makes sense to allow those options to be specified multiple times.

Done. Fine to import multiple matrices, but can only use one node image for initialisation.

Issue #619, opened 2016-05-18
Facilities for multi-shell responses


thijsdhollander:
Now that multi-shell response function files are becoming (or will quickly become) relatively common, we maybe need to make sure some of the obvious scripts/apps interacting with responses don't trip up.  The most obvious ones that come to mind are `shview` and the `average_response` script.  Sadly, I don't have the time to look into it any time soon, but just putting it here for reference...

jdtournier:
Well, `shview` already does support multi-shell responses - always did... You can supply a matrix instead of a row vector, and each row will be interpreted as a vector of coefficients. Use the arrows to navigate. 

The one thing I would like to change in `shview` though is the normalisation option. Currently, normalisation is either on or off, and when it's on, each ODF/response is normalised to 'unit' l=0. I think it would be better to have a 'rescale' action (Esc key for example) that would apply that normalisation on the currently displayed ODF/response, and the scaling would be maintained between ODF/responses. I think this would make more sense in general...

I guess the `average_response` script might need to be modified - but we have discussed in the past the possibility of adding an app to handle matrix operations, this might be a good use case for this...

thijsdhollander:
Hmm, just tested a multi-shell response with `shview`: it just showed the first shell (which happened to be the b=0 one), and the arrows didn't seem to have an effect.  I also noticed I'm still missing the menu bar in `shview` (reported somewhere in a much older issue, but I can't seem to find it at the moment).  Also, it seems for something like a CSF or other isotropic multi-shell response, `shview` interprets the file as a column vector (as if it were just a single shell with as many coefficients as there are actual shells).

jdtournier:
Ok, the left/you arrows not working is most likely related to the menubar not showing. That's been an issue on 14.04 for a while. But only on Qt 5, right?

The interpretation of a column vector as a single response is an issue though. The vector is actually explicitly transposed if it's a row vector, to allow more flexibility in how the response is specified. I'm happy to get rid of that, and mandate that the response must be a row vector. 

To display multi-shell, l=0 responses sensibly though, we'll really need to modify the way the normalisation is handled as I'd suggested above. Otherwise all shells will come out looking identical by default...

thijsdhollander:
So, related to #272 : upgrading to Ubuntu 16.04 also fixed the left/right arrows, so that specific aspect is no longer an issue.

thijsdhollander:
[...]

PR #885, opened 2017-01-23
shview fixes for multi-shell isotropic case

thijsdhollander:
This ended up actually being a removal of an old feature, which supposedly intentionally allowed a column vector.  In the single-shell days, the only reasonable interpretation was to treat the column vector as a row.  The fix introduces a segfault upon loading a multi-shell isotropic response (so not ready to merge yet); I'll quickly check if I can spot why...

(for reference: this is about one of the issues mentioned in #619)

### DIFF
--- a/cmd/shview.cpp
+++ b/cmd/shview.cpp
@@ -35,14 +35,14 @@ void usage ()
 
   ARGUMENTS
   + Argument ("coefs",
-              "a text file containing the even spherical harmonics coefficients to display.")
+              "a text file containing the even order spherical harmonics coefficients to display.")
   .optional()
   .type_file_in();
 
   OPTIONS
   + Option ("response",
-            "assume SH coefficients file only contains even, m=0 terms. Used to "
-            "display the response function as produced by estimate_response");
+            "assume SH coefficients file only contains m=0 terms (zonal harmonics). "
+            "Used to display the response function as produced by estimate_response");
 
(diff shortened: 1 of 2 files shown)


thijsdhollander:
Ok, obviously the reason is simply that there is no lmax=0 menu item yet... but all the hard-coded number of coefficients to lmax to menu-item-id calculations are a bit of a nightmare. :stuck_out_tongue:
[This line](https://github.com/MRtrix3/mrtrix3/blob/46555f6da3f55b3e4673d3efb4a48dbd39921811/src/gui/shview/render_window.cpp#L342) got me stumped for a bit: the naming `int lmax` seemed to suggest an lmax was being computed, and what followed of course didn't make sense in that context.

Completely separate from this issue (so whatever I might be about to change, isn't the cause of it :smile: ): there seems to be another (real) bug in `shview`.  If you load, e.g., a response with lmax=10, then what gets initially rendered is only up to lmax=8, even though the menu item "10" is checked.  The only way to get the lmax=10 rendering, is to then select another lmax, and switch back to 10.  I suppose this may also be the case for other lmaxes, but for those below 8 one would of course not (visually) notice the bug.

thijsdhollander:
That should be both of those bugs swatted.

jdtournier:
Looks good to me. Haven't had a chance to test yet, but everything looks sensible. Glad you figured out how my warped mind works eventually...

A few issues remaining to sort out, but these are best discussed over on #619.

PR #886, opened 2017-01-23
Further shview enhancements

jdtournier:
Following #885, this adds modifications suggested in #619:

- scale is now set once at load time, and only modified on demand by pressing the Esc key (or using the 'reset scaling' menu entry). This allows comparison of ODFs / responses across different rows (useful for multi-shell handling, etc). 

- Scaling is set according to norm of SH coefficient vector of currently displayed ODF, rather than _l_=0 term as previously (should be more robust to different types of ODFs). 

- Font rendering on viewport is now possible. The text "no data loaded" is shown initially, and set to "row _n_" once data are shown. Should be possible to set the actual b-value if we eventually decide to support that. 

- initial view is now oblique: 45° to X-Y, 30° above transverse plane. Makes it easier to appreciate response functions without needing to interact with the projection. 

I've not tested whether MRView will still work correctly with these changes - some of these changes involve the `DWI::renderer` and `DWI::render_frame` code, which is common to both `shview` and the `mrview` ODF tool. I don't anticipate any particular issues, but I would appreciate thorough testing before merging...  :wink:

### DIFF
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -34,8 +34,6 @@ namespace MR
 
         constexpr float DistDefault = 0.3f;
         constexpr float DistInc = 0.005f;
-        constexpr float DistMin = 0.1f;
-        constexpr float DistMax = 10.0f;
 
         constexpr float ScaleInc = 1.05f;
 
@@ -315,8 +313,6 @@ namespace MR
           }
           else if (event->buttons() == Qt::RightButton) {
             distance *= 1.0 - DistInc*dy;
-            if (distance < DistMin) distance = DistMin;
-            if (distance > DistMax) distance = DistMax;
             update();
           }
         }
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -189,12 +189,8 @@ namespace MR
           if (std::isfinite (values[0])) {
             gl::Disable (gl::BLEND);
 
-            if (!std::isfinite (scale)) {
-              if (values[0] != 0.0)
-                scale = 2.0f / values[0];
-              else 
-                scale = 2.0f / values.norm();
-            }
+            if (!std::isfinite (scale)) 
+              scale = 2.0f / values.norm();
 
             renderer.set_mode (mode);
 
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -56,7 +56,7 @@ namespace MR
 
       RenderFrame::RenderFrame (QWidget* parent) :
         GL::Area (parent),
-        view_angle (AngleDefault), distance (DistDefault), line_width (1.0), scale (NaN), 
+        view_angle (AngleDefault), distance (DistDefault), scale (NaN), 
         lmax_computed (0), lod_computed (0), mode (mode_t::SH), recompute_mesh (true), recompute_amplitudes (true),
         show_axes (true), hide_neg_values (true), color_by_dir (true), use_lighting (true),
         glfont (get_font (parent)), projection (this, glfont),
@@ -239,7 +239,6 @@ namespace MR
         }
 
         if (show_axes) {
-          gl::LineWidth (line_width);
           gl::BlendFunc (gl::SRC_ALPHA, gl::ONE_MINUS_SRC_ALPHA);
           gl::Enable (gl::BLEND);
           gl::Enable (gl::LINE_SMOOTH);
--- a/src/gui/dwi/render_frame.cpp
+++ b/src/gui/dwi/render_frame.cpp
@@ -44,10 +44,10 @@ namespace MR
 
       RenderFrame::RenderFrame (QWidget* parent) :
         GL::Area (parent),
-        view_angle (40.0), distance (0.3), line_width (1.0), scale (1.0), 
+        view_angle (40.0), distance (0.3), line_width (1.0), scale (NaN), 
         lmax_computed (0), lod_computed (0), mode (mode_t::SH), recompute_mesh (true), recompute_amplitudes (true),
         show_axes (true), hide_neg_values (true), color_by_dir (true), use_lighting (true),
-        normalise (false), font (parent->font()), projection (this, font),
+        font (parent->font()), projection (this, font),
         orientation (Math::Versorf::unit()),
         focus (0.0, 0.0, 0.0), OS (0), OS_x (0), OS_y (0),
         renderer ((QGLWidget*)this)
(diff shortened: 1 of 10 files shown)


Lestropie:
Just some tiny tweaks from me.

> initial view is now oblique: 45° to X-Y, 30° above transverse plane. Makes it easier to appreciate response functions without needing to interact with the projection.

Good idea :+1: Now annoyed I didn't think of that at some point.

thijsdhollander:
Took it for a spin, and I do like the rescale logic (it's basically the `mrview` windowing logic, which should be intuitive to most users).  I did stumble upon a few bugs though:

- Using the scroll wheel to rescale, it seems there's still some capping to a certain scale range being applied.  For reproducibility, I used this response (it basically just has to have large enough l=0 coefficients):

```
4156.5354 0 0 0 0
2142.200636 -917.3129077 207.5541769 0 0
1439.698345 -883.1213623 350.7187295 -94.67385459 0
1191.675628 -806.8944097 408.7715543 -150.9711246 43.26948986
```
Go, e.g., to the last shell here, and start scrolling; you'll immediately see it'll go to a certain minimal scale.

- There's a bug with the level of detail, similar to the lmax one I encountered/fixed before (haven't checked, but this must be on master still as well; so not introduced by the commits in this branch): the initial level of detail that is rendered, seems to not match with the initial selection in the menu.

- If [this commit](https://github.com/MRtrix3/mrtrix3/pull/886/commits/b1bea6a62e976a1fbc48bbe6dd037d9bfc56d8c1) was meant to reset to the initial oblique angle upon hitting `Esc`, then that seems to not be working in practice at the moment... but I'm actually not sure if I'd prefer this behaviour.  This kind of "resetting view" action is also a distinct one in `mrview` (i.e., the `r` key as opposed to `Esc`).

Apart from that, maybe a general suggestion: I'd let the "row x" (as well as the number in the title bar) actually start counting from 1 rather than 0, since this is essentially referring to a line number in a text file.  Most text editors that (allow to) show line numbers also start from 1.  This may be useful if investigating a larger file with many rows.

thijsdhollander:
> There's a bug with the level of detail, similar to the lmax one I encountered/fixed before (haven't checked, but this must be on master still as well; so not introduced by the commits in this branch): the initial level of detail that is rendered, seems to not match with the initial selection in the menu.

This actually makes me wonder whether the level of detail controls are relevant for `shview` at all: the default that is actually rendered (but wrongly indicated in the menu) is the maximal level of detail.  Since `shview` only renders a single SH plot at once, I can't see this maximal level of detail would be a problem for anyone...?

thijsdhollander:
> Apart from that, maybe a general suggestion: I'd let the "row x" (as well as the number in the title bar) actually start counting from 1 rather than 0, since this is essentially referring to a line number in a text file. Most text editors that (allow to) show line numbers also start from 1. This may be useful if investigating a larger file with many rows.

Ok, so this is tackled by @Lestropie's [commit ](https://github.com/MRtrix3/mrtrix3/pull/886/commits/31cbe99a9f815f2f97bf76c5aba7b27ff0c13cd5)above.

jdtournier:
> I'd let the "row x" (as well as the number in the title bar) actually start counting from 1 rather than 0

Yes, I did wonder about that. I think your line number argument is pretty convincing. Happy with that.

About the reset view thing: this is actually one of my magic hidden features. Reset the viewing direction with a regular double click, and reset the focus position with a middle button double click. Apologies, this dates back from around 2009, hasn't changed since... I'm thinking of adding a 'reset view' action with a Home button shortcut. Other option is to remap the 'treat as response' shortcut to Z (for zonal harmonics), and use R for the 'reset view' shortcut. What do you think?

Otherwise, I'll fix up the level of detail thing, should be trivial. But the initial level of detail shouldn't be the maximum possible in `shview`: the max is set much higher here than in MRView (since in MRView the rendering happens over the whole image). I'll have a look anyway.

And I'll get rid of the limits on the scale. There's no need for them now with the 'reset scale' action.

jdtournier:
OK, I think that just about fixes everything. I've also changed the default lighting parameters, since I found the ODFs looked a bit too saturated - see what you think. I've also made the reset view shortcut 'R', as mentioned previously, which meant re-mapping the 'treat as response' action to 'Z'. Maybe we should also consider renaming that action to 'treat as Zonal Harmonics' or something...? Or would that be too cryptic...?

thijsdhollander:
> I've also made the reset view shortcut 'R', as mentioned previously, which meant re-mapping the 'treat as response' action to 'Z'. Maybe we should also consider renaming that action to 'treat as Zonal Harmonics' or something...? Or would that be too cryptic...?

Makes sense, no worries about the "treat as response" shortcut I'd say: I think it's quite unlikely anyone was already using that frequently (or will ever need to).  Just naming it "zonal harmonics" would indeed be too cryptic, I reckon.  For your average user, `shview` still ends up being "that viewer for *response functions*" in practice.  I've added a "(zonal SH)" after the "treat as response", so it at least inherently explains the shortcut key, but without doing away with the clarity.

> I've also changed the default lighting parameters, since I found the ODFs looked a bit too saturated - see what you think.

:dizzy_face: my poor eyes; so much ambient / diffuse that all visual cues on shape are almost entirely gone.  I agree though that they used to be too bright / oversaturated though, but that's in the end more due to too much overall light (i.e. the sum of all light mechanisms) and a too diffuse (low) specular exponent; the latter then adding way too much light for little visual information (i.e. a waste of light).  The new settings were still okay-ish with directional colouring (but still a lot of shape information disappeared into a diffuse haze), but without directional colouring, everything became really flat.  I've re-tuned the parameters to another balance, and then found that that was (accidentally) pretty close to having all sliders set about halfway, so setting them exactly halfway may look like an appealing default.  This setting is in overall brightness (and hence saturation) very close to what you proposed, but the balance is different: a bit less ambient, more specular, but also a higher specular exponent (which makes the highlight smaller).  This works quite well for both directional and non-directional colouring (which I now happen to use quite a lot as well):

screenshot from 2017-01-25 14-40-25

screenshot from 2017-01-25 14-40-32

The nice thing about setting them halfway is that allows the users to change according to preference in both ways ("equally"; even though e.g. the exponent range is a bit arbitrary) ...and of course there's still the configuration file for personal preferences or specific environments.  But I agree this may still be mostly a (personal) preference thing. (although I wouldn't underestimate the shape information encoded in a specular highlight and naturally interpreted by our brains :slightly_smiling_face: )

jdtournier:
All good. Are we happy to merge this? Should we merge to `tag_0.3.16` rather than `master`?

thijsdhollander:
I'll create another pull request to merge to `tag_0.3.16`.

PR #888, opened 2017-01-24
dwiextract: preserve DW scheme information exactly as per input

jdtournier:
As discussed in #878. Basically, get a processed version of the DW info to allow the shell selection to work as expected, but write back the unmodified version of the relevant DW info to the output.


jdtournier:
Any objections to this change? Everyone happy to merge with `tag_0.3.16`?

PR #889, opened 2017-01-24
mrview: fix mouse wheel event handling

jdtournier:
Problem was introduced by 05e3e6542 to handle scrolling with touchpads,
but resulted in ~4x the step when using regular mouse wheel. This
affected scrolling through slices and zooming.


jdtournier:
:rofl: glad to hear it!

PR #890, opened 2017-01-24
Documentation restructure

draffelt:
Hi Guys,
Check out the new documentation on the 'doc_restructure' branch on [read the docs](http://mrtrix.readthedocs.io/en/doc_restructure/).

Some things to note:

* Removal of basic tutorials and advanced workflows section. Replaced with dedicated sections for features. 
* Updated description of the [main MRtrix functionality](http://mrtrix.readthedocs.io/en/doc_restructure/), as I mentioned before, I know it's not quite comprehensive but to not make it too long, I've only included features that have their own section in the docs. Feel free to modify
* Because the [list of MRtrix commands](http://mrtrix.readthedocs.io/en/doc_restructure/reference/commands_list.html) is way down the bottom, I've added [this small page](http://mrtrix.readthedocs.io/en/doc_restructure/getting_started/commands_and_scripts.html) to aid discover-ability of commands for new users. Note that it would be nice eventually to have a table with a one line description for all commands (pulled automatically from the command synopsis.). 
* [Fixel Image format docs](http://mrtrix.readthedocs.io/en/doc_restructure/getting_started/image_data.html#fixel-image-format). Would be great to have another set of eyes over this
* Changed DWI Modelling section to CSD. On the assumption that currently we would not really recommend doing anything thing else, so it's unlikey that we will make tutorials for other models. We can always change this if we do. Note I've moved the lmax stuff from the FAQ into this section. 
* Troubleshooting has been moved from getting stated to the end. With some grouping of display, and hanging related issues.
* FAQ currently in the troubleshooting section. @thijsdhollander unfortunately if I put it in its own section, then the section caption is not linkable, so it needs to be listed again below the section caption. 
* [Global intensity normalisation](http://mrtrix.readthedocs.io/en/doc_restructure/concepts/global_intensity_normalisation.html) has a dedicated page in concepts
* Tips and Tricks section. I've currently add a tutorial for [batch processing with the foreach script](http://mrtrix.readthedocs.io/en/doc_restructure/tips_and_tricks/batch_processing_with_foreach.html) (since this is now heavily used in the FBA tutorials). 
* FBA instructions for single-shell, multi-tissue CSD, as well as some basics on how to display FBA results with streamlines, and compute effect sizes. Note that the FBA instructions are much more detailed, and have more explicit steps. Users can now copy and paste from start to finish, providing they start off with the recommend layout of files. Basically the old instructions only gave vague descriptions of the inputs and outputs, and not explicit file names. This got tricky when trying to explain the steps working on the new fixel format, so I opted for more explicit instructions. 
* [Added NOTE](http://mrtrix.readthedocs.io/en/doc_restructure/installation/before_install.html) about citation MRtrix to before you install
* @Lestropie , [this](http://mrtrix.readthedocs.io/en/doc_restructure/quantitative_structural_connectivity/labelconvert.html) and [this](http://mrtrix.readthedocs.io/en/doc_restructure/quantitative_structural_connectivity/structural_connectome.html) tutorial have fairly similar names. Not sure if you want to do anything about this.


Lestropie:
> Fixel Image format docs. Would be great to have another set of eyes over this

Done; have a look. You didn't actually state anywhere that the indices for all fixels within a voxel must be / are inferred to be sequential. :-P

Still concerned that this is not sufficiently compatible with models generated through bootstrapping (and therefore FMRIB will reject the idea). For instance:

* *Easily identified as a data file type because the size of the image is 1 in the 3rd dimension*: If a model has *p* parameters per fixel but also *b* bootstraps, they might have an image that's n x p x b.
* *All DWI models must specify the direction of each fixel.*: With bootstrapping, each fixel doesn't have a 'fixed' direction; there's a large number of separate realisations of the fixel model, and each has a unique direction for each fixel. `bedpostx` does generate images for mean azimuth and mean elevation of each fixel, but technically to do this properly you can't just average across bootstrap realisations, you need to do some form of fixel correspondence, since the ordering isn't guaranteed to be equivalent. I see this as a requirement for many *MRtrix3* commands, including the current state of the `mrview` fixel tool, but I'm not sure if it should be a 100% hard rule.

> Added NOTE about citation MRtrix to before you install

I've been telling people to *not* cite the 2012 paper and just give the website / GitHub. For the analyses most people are doing, none of it will actually be described in that paper; the only thing in common is the name. Yes it means less citations for Donald, but that's just how I feel about it.

> @Lestropie , this and this tutorial have fairly similar names. Not sure if you want to do anything about this.

Done.

draffelt:
> Done; have a look.

Thanks, looks good

> Easily identified as a data file type because the size of the image is 1 in the 3rd dimension: If a model has p parameters per fixel but also b bootstraps, they might have an image that's n x p x b.

I guess they could have 1 file per bootstrap...bit silly I know. I guess not all packages/applications have to take this format on board. Personally I'm not going to have any time to drive this as a standard (I start my new job in 2.5 weeks). So I guess it's up to everyone else to decide on if they want to push it to be standardised and written up. 

> All DWI models must specify the direction of each fixel.: With bootstrapping, each fixel doesn't have a 'fixed' direction.

Clearly this format won't work for "All DWI models". Think I'll rename it to "fixel-based DWI models". 

> Has the ISMRM HCP connectome tutorial disappeared?

No, I just forgot to add the .rst extension when I renamed it. It's back now.

draffelt:
@jdtournier I'm assuming you are happy with the restructure? Will merge it into tag_0.3.16 if so

PR #891, opened 2017-01-25
Copyright for 2017 in headers

thijsdhollander:
Happy New Year!  I'll check some of the other occurrences too.  At the moment, the header of these has been updated:
```
find . -name '*.h' -or -name '*.cpp' -and -not -name '*_moc.cpp' -and -not -name 'nifti1.h' -and -not -name 'json.h'
```
... to take into account accidental mishaps like [this one](https://github.com/MRtrix3/mrtrix3/commit/e24892e8f8cb48e37c719ac0736075d10bb72d6d).

Is everything else in `lib/file` ours?


thijsdhollander:
(by the way, next year we'll get to celebrate 10 years of MRtrix)

thijsdhollander:
Ok, now the header files are done too, as well as the 2 bits that generate copyright info for the `__print_usage_rst_` output of commands and scripts for documentation generation purposes.

thijsdhollander:
In summary:
- all `.h` and `.cpp` files that already had our copyright header
- excluding `nifti1.h` and `json.h` because they're not ours
- the bits that generate the `__print_usage_rst__` output
- the script and command docs, as regenerated by `./build doc`

I'll merge into `tag_0.3.16`, so users are not bothered; the moment `tag_0.3.16` gets merged into `master`, they'll get many changes anyway.  If anyone can think of something that may have been overlooked, don't hesitate to update and commit to `tag_0.3.16`.

jdtournier:
Just a thought: @Lestropie, is the `lib/file/mgh.h` header ours...?

Lestropie:
I *believe* so; I seem to have a memory of generating it from the webpage linked in the code comment (the comments certainly match), I definitely added the `#pragma pack` myself (since it took a little while to figure out what was going wrong), and I reckon I recall having to add the `#ifdef __cplusplus`. If I'm right, the `#define _MGH_HEADER_` could be updated.

jdtournier:
Yes, I remember those discussions about the `#pragma pack`. Just had a look at their webpage, it does indeed look like you might have copy/pasted their _description_ of the contents of the header, but not their code as such. In fact, I reckon the only safe way to read this, given the need for this `#pragma pack` hack, is using old-fashioned `fread()` or `fstream.read()` rather than memory-mapping. So I think we should be fine here.

PR #894, opened 2017-01-27
Shview enhancements

thijsdhollander:
Original discussion in #886 .


thijsdhollander:
(for reference: @Lestropie just showed me that we don't have to create a new pull request if we change our mind about which branch we'd like to merge into: you can just click the "edit" button on the top right for unmerged pull requests and simply change that)

jdtournier:
> (for reference: @Lestropie just showed me that we don't have to create a new pull request if we change our mind about which branch we'd like to merge into: you can just click the "edit" button on the top right for unmerged pull requests and simply change that)

Good to know - the last time I looked into how to do this (about 6 months ago), it was highly requested feature, but not actually possible. Good to see GitHub does respond to user feedback. :grin:

PR #898, opened 2017-01-30
Script redesign

Lestropie:
Related to #864, #896.

So this ended up being a bit of a mishmash: I needed to modify the work I had done in the `package_scripts` branch, both to remove the actual `package_scripts` script & related functionality in order to still use the other changes I'd made in that branch, but also to do the required modifications to bring the script library in line with the FHS restructure. It looked like it was going to be more effort to split up the two jobs, so I've done both in the one go. Diff is a bit evil (a lot of files are flagged as new & delete rather than a move), so you might have a more fun time just looking at the branch itself.

Open to comments not only on the FHS restructure, but also the 'module-ing' of the Python code.

Only thing left in the `scripts/` directory is `generate_mrtrix_completion.py`: @rtabbara This probably needs to be modified to bring it up to date. It should perhaps also go into the root directory now?

**Edit**: Got `dwi2response` working, but all of the scripts need to be re-tested when I'm not on a Windows machine.


jdtournier:
Just had a look at this, and merged with `repo_fhs_restructure` to get a clearer picture. This looks good,  just a few suggestions:

- I'd merge the modules `app`, `message`, `cmdlineParser` and `progressBar` into a single `app` module - they're all likely to be used together in any given script, and all relate to how an app is structured and interacts with the user. There are also existing interactions, for instance the `addCitation()` function in `app` interacts with `print_help()` in `cmdlineParsing`, and the `message` functions are used in `app` and `progressBar`, etc. I think it makes sense to have them all in the same module...

- I'd merge the modules `file` and `path` together. I can see the distinction, but it's close enough that I'd expect them to be in the same module. 

- I'd merge the `misc`, `mrtrix`, `fsl`, and `run` modules into a single `run` module. Again, the current organisation is perfectly sensible, but these all contains related functionality. 

I reckon that would clean things up further...

I also removed one import line that caused an error when trying to use `dwi2response`, but it looked redundant. Works just fine without it, ran a `dwi2response tournier` to completion just fine. 

All good otherwise!

Lestropie:
With regards to merging of files / 'modules', I'm a bit conflicted as to how far to go. There's definitely some sense to Daan's suggestion in thread #896 (although I'd be tempted to just put it all together into a single MRtrix module if we were to go that far), and some of Donald's suggestions are definitely logical, but I'm not sure whether or not to adopt all of them. One thing I found when doing these changes is that a lot of function names could be shortened because a part of the context / purpose of the function became encapsulated within the module name; if you merge too many things, you lose this sense of 'namespace' / hierarchical functionality. I suppose I'll just have a go at some point and see where I end up.

jdtournier:
> With regards to merging of files / 'modules', I'm a bit conflicted as to how far to go.

There's no need to get too worried about this, the scripts work fine as-is, the changes @dchristiaens & I have suggested are purely cosmetic - just to try to make the scripts a little bit more 'pythonic'. But given that they're not really designed for 'external' use (even though there's nothing stopping anyone from using them as such), and not very 'pythonic' anyway given the way we locate the modules and the way the modules locate the binaries, I really don't think this is anything we need to worry about. My suggestion would reduce the number of modules to 6, by my count. So the functionality still remains pretty spread out. I think your logic of shortening the function names using their module prefix is very much the right way to go. So take whichever bits of our suggestions you think make sense and ignore the rest, no-one will be offended... :wink:

draffelt:
Looks good to me too.

thijsdhollander:
I'm good with the restructuring of the "modules", but I'm just slightly concerned (or probably just confused :confused:) about the `hasattr()` additions in various places.  I think I'm probably not getting the purpose...?  See for instance [here](https://github.com/MRtrix3/mrtrix3/pull/898/commits/95486d689d01208c005b6e8956d0548e714e9475#diff-a0ea946f466732ee40679d44d73ed2f5R61): that's definitely not the indented behaviour (the default here is actually 3, but if the attribute is *missing* I wouldn't want to silently proceed, in this case with non-default behaviour).

But more generally, as I was trying to wrap my mind around the purpose of these additions, [I stumbled upon this](https://hynek.me/articles/hasattr/)... which had me slightly more worried about the (different) behaviours this may lead to... what do you think?

Lestropie:
> ... I'm just slightly concerned (or probably just confused 😕) about the hasattr() additions in various places

Unnecessary safety measure put in place because of a different problem. I'll be going back and removing them all again, especially since it's warned against. I've gotten caught a couple of times thinking that not all versions of `argparse` are guaranteed to create an empty variable for options not specified by the user, and hence if you try to access the variable without first checking for its existence it throw an exception. But pretty sure it's just been the fact that getting the parsing for algorithm-based scripts (`5ttgen` and `dwi2response`) working is *really* finicky, and when it doesn't quite work properly this tends to be the first symptom.

Lestropie:
@draffelt or @maxpietsch: Can we do something about [these](https://github.com/MRtrix3/mrtrix3/blob/master/scripts/population_template#L9-L16)? I'd rather have such functionality within [`path.py`](https://github.com/MRtrix3/mrtrix3/blob/repo_fhs_restructure/lib/mrtrix3/path.py), preferably with some comments about how / why they're used, but I didn't want to touch them in case there was some trick regarding softlink handling or the like.

maxpietsch:
Correct me if I am wrong but I don't think there is any trickery going on, they are more convenience functions. Do you want to replace them with something else or just move them over to path.py? Symlinks were an issue using `cp` (5f87070c63a72c1341c47cf). I have a script that tests various options of `population_template` so happy to test any changes.

More important are the runFunction calls like [this](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/scripts/population_template#L85) to ensure that the `-continue` option works. (You are linking to the master branch. I have rewritten parts of `population_template` for tag_0.3.16.)

PR #900, opened 2017-02-01
dwi2response tournier: Fix for zero crossings

Lestropie:
Hopefully fixes for issues reported [here](http://community.mrtrix.org/t/dwi2response-error-mrconvert-coord-option-is-out-of-range-of-input-image/688); awaiting testing.

### DIFF
--- a/scripts/lib/app.py
+++ b/scripts/lib/app.py
@@ -188,12 +188,12 @@ def complete():
   elif tempDir:
     # This needs to be printed even if the -quiet option is used
     if os.path.isfile(os.path.join(tempDir, 'error.txt')):
-      with open(os.path.join(tempDir, 'error.txt'),'rb') as errortext:
-        sys.stdout.write(os.path.basename(sys.argv[0]) + ': ' + colourWarn + 'Script failed while executing the command: ' + errortext.readline().rstrip() + colourClear + '\n')
-      sys.stdout.write(os.path.basename(sys.argv[0]) + ': ' + colourWarn + 'For debugging, inspect contents of temporary directory: ' + tempDir + colourClear + '\n')
+      with open(os.path.join(tempDir, 'error.txt'),'r') as errortext:
+        sys.stderr.write(os.path.basename(sys.argv[0]) + ': ' + colourWarn + 'Script failed while executing the command: ' + errortext.readline().rstrip() + colourClear + '\n')
+      sys.stderr.write(os.path.basename(sys.argv[0]) + ': ' + colourWarn + 'For debugging, inspect contents of temporary directory: ' + tempDir + colourClear + '\n')
     else:
-      sys.stdout.write(os.path.basename(sys.argv[0]) + ': ' + colourPrint + 'Contents of temporary directory kept, location: ' + tempDir + colourClear + '\n')
-    sys.stdout.flush()
+      sys.stderr.write(os.path.basename(sys.argv[0]) + ': ' + colourPrint + 'Contents of temporary directory kept, location: ' + tempDir + colourClear + '\n')
+    sys.stderr.flush()
 
 
 
--- a/scripts/src/dwi2response/tournier.py
+++ b/scripts/src/dwi2response/tournier.py
@@ -29,11 +29,12 @@ def getInputFiles():
 def execute():
   import os, shutil
   import lib.app
-  from lib.delFile      import delFile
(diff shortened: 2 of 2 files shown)


thijsdhollander:
Ok, so I may not have fully thought this through, so happily correct me if I'm wrong, but: shouldn't the algorithm just continue in that case (but there should indeed be something in place to make sure those second peaks are accessed as if they were zero)?  Since the candidate voxels (and their neighbours) is a different (larger) set than the voxels that were used in the previous iteration to get the response, the voxels that would get selected in this new iteration (via the metric) could still be a different set, right?  The metric essentially becomes sqrt(peak1), so just selects the top largest peaked voxels; but the selected set could become different, and the next iteration could even actually have crossing voxels pop up in the candidate voxels (and their neighbours).

Lestropie:
Theoretically sure, the set could change. I don't think it's a given that the set of 300 voxels selected from that would be 'better' than the set of 300 voxels from the previous iteration: depends on whether the amount of signal going to sub-threshold crossing fibres varies between the two, and whether one prioritises large signal over 'single-fibre-ness'. I'm content with either. Yours is probably closer in spirit to the algorithm as it stands though.

thijsdhollander:
Well, it's indeed a very exceptional scenario anyway; but who knows how close the algorithm may actually have gotten to it, or even gets to in in practice on many occasions in the later iterations.  I wouldn't be surprised that the 300 voxels in many cases in the later iterations are all single fibre (i.e. peak2 == 0 voxels), and the 301st voxel probably is too.  So ultimately, the algorithm just ends up selecting the larger peaked single fibre voxels (even when peak2 != 0, there's an intended bias towards that built in).  It's indeed debatable whether those are better, but it is still the metric that we built in to quantify what's better.  :man_shrugging: 

But my main motivation, apart from keeping it in spirit, would be that it's just simpler (a bit more clear in the code?) and avoids all too cryptic errors/warnings that will at some stage probably end up on the forum again. :slightly_smiling_face:

jdtournier:
I was initially a bit puzzled as to how this could come about, since [`sh2amp` has a `-num` option](http://mrtrix.readthedocs.io/en/latest/reference/commands/sh2peaks.html#options) that should ensure the same number of peaks are extracted regardless. But then I remembered that this implementation uses `fod2fixel | fixel2voxel`. Would the simplest fix not be to add a similar option to `fixel2voxel` to extract the largest _N_ peaks, and pad with zero as necessary?

Lestropie:
> Would the simplest fix not be to add a similar option to fixel2voxel to extract the largest N peaks ...

I left a comment in the code some time ago for myself to do this. Such an option wouldn't pad with zeros though, given it's a sparse image output; it'd just limit the maximal number of fixels. `fixel2voxel` is responsible for [the zero-padding](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/fixel2voxel.cpp#L603). I suppose that `fixel2voxel` could have an option to specify the number of fixels used per voxel, and extend that functionality to not only limit the number of fixels but also pad appropriately.

Looking again at the code, the user's result is exceptionally unusual: The `fod2fixel` call [uses the `-fmls_no_thresholds` option](https://github.com/MRtrix3/mrtrix3/blob/master/scripts/src/dwi2response/tournier.py#L67), which means that for all 3,000 voxels tested, the FOD must be strictly monotonically decreasing from the peak. My guess is that the data have 6 directions.

jdtournier:
> Looking again at the code, the user's result is exceptionally unusual: The fod2fixel call uses the -fmls_no_thresholds option, which means that for all 3,000 voxels tested, the FOD must be strictly monotonically decreasing from the peak. My guess is that the data have 6 directions.

Yes, I have to admit, that's really unexpected. Would be good to get the data from the user to check, or at least get them to post the relevant info about the dataset, e.g. the DW gradient table.

thijsdhollander:
> My guess is that the data have 6 directions.

If all those FODs are strictly decreasing from the peak, there's no other realistic possibility (I can think of) than those FODs being represented with lmax=2, which can indeed only be explained by a very low number of directions.

> Yes, fixel2voxel could get an option to use the first N fixels, and either truncate or pad accordingly.

That's probably the cleanest solution, which would leave the original script mostly untouched (and its code much more readable).

thijsdhollander:
Simpler solution as discussed available in #923.

PR #901, opened 2017-02-01
Fix mrview cmdline processing

jdtournier:
This fixes issues when trying to run a complex set of operations in MRView from the command-line, and adds a new option `-volume` to set the volume currently shown. The main problem was that screen updates were not guaranteed to happen after each option was processed, leading to funny issues for later options, outdated/blank screenshots, etc. 

This came about because I was trying to do this:
```ShellSession
mrview -capture.folder . -capture.prefix screenshot- image.mif -size 256,256 -noannot -fov 120 \
    -vol 0 -autoscale -capture.grab \
    -vol 1 -autoscale -capture.grab \
    -vol 2 -autoscale -capture.grab \
    -vol 3 -autoscale -capture.grab \
    -vol 4 -autoscale -capture.grab \
    -vol 5 -autoscale -capture.grab \
    -exit
```
which now works with these changes. :grin:

One outstanding issue though is that any tool-specific option will cause the corresponding dock widget to show up, screwing up the window size. One way around it is to set the docks to floating in the config (i.e. add `MRViewDockFloating: 1` to the config file), but it's a bit ugly. It would be better to create the tool, but leave it hidden when started from the command-line...

### DIFF
--- a/cmd/mrview.cpp
+++ b/cmd/mrview.cpp
@@ -64,7 +64,6 @@ void run ()
 {
   GUI::MRView::Window window;
   window.show();
-  window.process_commandline_options();
 
   if (qApp->exec())
     throw Exception ("error running Qt application");
--- a/src/gui/mrview/window.cpp
+++ b/src/gui/mrview/window.cpp
@@ -231,7 +231,7 @@ namespace MR
         tool_has_focus (nullptr),
         best_FPS (NAN),
         show_FPS (false),
-        current_arg (1) {
+        current_arg (nullptr) {
           main = this;
           GUI::App::set_main_window (this);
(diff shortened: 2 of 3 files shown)


jdtournier:
OK, that last issue should be sorted with the latest commit. Seems all good.

jdtournier:
I might merge this into `tag_0.3.16` just to make sure everyone has a chance to test it on different platforms. Not too keen on merging to `master` only to find out it doesn't work on Qt4 or something...

jdtournier:
OK, looks like I'll be merging with `master` after all, given [this recent discussion](http://community.mrtrix.org/t/mrview-exit-option/705?u=jdtournier)...

PR #917, opened 2017-02-23
Investigate missing max align t

jdtournier:
Just to try to figure out these strange issues with `max_align_t` on `clang++` 3.4


configure
```diff
@@ -577,8 +577,42 @@ int main() {
   except:
     error ('unexpected exception!' + configure_log_hint)
   else:
-    break
+    max_align_t_defined = False
+    report ('Checking for ::max_align_t: ')
+    try:
+      compile ('''
+#include <cstddef>
```
maxpietsch: no using ::max_align_t?

maxpietsch:
Might have to do with this: https://stackoverflow.com/questions/5079325/should-i-include-stddef-h-or-cstddef-for-size-t

```
src/surface/polygon.h:20:#include <stddef.h>
src/gui/opengl/gl_core_3_3.cpp:17:#include <stddef.h>
src/gui/opengl/gl_core_3_3.h:87:#include <stddef.h>
src/min_mem_array.h:20:#include <stddef.h>
core/file/json.h:38:#include <cstddef>
core/types.h:18:#include <cstddef>
eigen/bench/tensors/benchmark.h:16:#include <stddef.h>
```

configure
jdtournier: Whoops... Must have forgotten to put it back in after testing on my system (since I can't get the tests to fail any other way here...).
Fix committed.

jdtournier:
Let's try using `<stddef.h>` instead then - I hadn't realised there was this distinction...

jdtournier:
OK, well that answered that question then: no sign of `max_align_t` either at global scope or within the `std` namespace... which matches your earlier testing, IIRC?

maxpietsch:
Hmm, I am stumped as well. Why did it compile the Eigen test if there is no ::max_align_t? We could just add it if none were found similarly to [here](https://reviews.llvm.org/rL201729#9c02349d)?

.travis.yml
```diff
@@ -7,7 +7,6 @@ addons:
     packages:
       - zlib1g-dev
       - libqt4-opengl-dev
-      - g++-4.8
```
maxpietsch: This might be why it is no longer in the global namespace

jdtournier:
> Why did it compile the Eigen test if there is no ::max_align_t? 

I think that's because we never actually used it till now. Eigen doesn't need it either (I checked, no sign of `max_align_t` in the entire source tree of either Eigen 3.2.10 or 3.3.3. I think the issue really is specific to your installation, where you have `clang++` 3.4 installed alongside `g++4.9`. 

> We could just add it if none were found similarly to here?

My thoughts exactly. Do you want to give that a shot...?

maxpietsch:
On it. That is going to be a massive change...

jdtournier:
@maxpietsch: just saw your latest commits, it's looking a little bit excessive...? I'd assumed all that was needed was an `#ifdef` in `core/types.h` that would define `max_align_t` if missing, just so that the `MEMALIGN` works. That's the only place that should need modifying, every other file in the source tree will need to include that header, since it'll need to use that macro.

maxpietsch:
Yes, excessive is the right word...

>  every other file in the source tree will need to include that header

Yes, but they don't. And they need to include it before any standard library header which they don't either.

So, inspired by Eigen's plugin system, I've created one for mrtrix.

It breaks memory alignment, though (using clang 3.4 and gcc > 4.9). Feel free to revert. But maybe wait an hour or so -- would feel too frustrating otherwise...

jdtournier:
> And they need to include it before any standard library /Eigen header which they don't either.

OK, I'm guessing you're trying to get this working on your system too? I was assuming we'd forget about that particular configuration, it is indeed going to be really hard to support. I was hoping we'd at least get the clang 3.4 with GCC < 4.9 setup working, since that's at least not outright buggy - we can't even include `<cstddef>` on your setup, that's just beyond salvage, I think...

If we restrict ourselves to fully C11/C++11 compliant setups, or those that simply fail to provide `max_align_t`, we shouldn't need to modify everything like this...(?)

What I had in mind is to test for both `::max_align_t` and `std::max_align_t` in `./configure` and set the appropriate macros (as currently), then define `:: max_align_t` with a `using std:: max_align_t;` if it's defined, otherwise fall back to the manual definition used in more recent versions of `clang`. What do you think?

jdtournier:
Ok, having waded through those commits to find the actually relevant bits, it looks like you're doing pretty much exactly what I had in mind (although I wasn't thinking about a separate header). Ignore my last comment.

jdtournier:
Feel free to add your new header to the [list of files ignored by check_memalign](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/check_memalign#L13), by the way... :wink:

maxpietsch:
Haha, thanks! I can't figure out why my memalign check goes so haywire on all classes but we could use that as a check in `configure` to bail out for configurations like mine.

maxpietsch:
If we keep this then we'll need to make sure that we include `__mrtrix_plugin.h` before any other header. Depending on our local compiler environment, we might not notice missing includes until we hit push. For now Travis should find them but if Travis gets updated then we might need to figure out a better way to check that `__mrtrix_plugin.h` is included first.

jdtournier:
It's looking good so far! :+1:

If the tests complete OK, let's try to distill everything on this branch to a smaller set of changes... I'm assuming it would be fine to include the contents of your header back into `core/types.h`, make sure `#define NOMEMALIGN` happens before that manual `max_align_t` definition, so you can actually use it for that class, and we can leave `check_memalign` as it was.

> If we keep this then we'll need to make sure that we include __mrtrix_plugin.h before any other header. 

So this is what I don't get. I can see that this would be needed on your setup, but I don't see why we'd need it otherwise. Does the TravisCI build fail if you don't do this...?

maxpietsch:
> So this is what I don't get. I can see that this would be needed on your setup, but I don't see why we'd need it otherwise. Does the TravisCI build fail if you don't do this...?

Hmmm, I think you're right. Maybe we should make a new branch and test it there. I've had enough for today though

jdtournier:
OK, I had a go at cleaning all this up. I think it's about as simple as I could make it - and seems to work fine on my system and on TravisCI. @maxpietsch, when you get the chance, it would be good to check that the compiler test fails on your problematic setup (due to the inclusion of the `<cstddef>` header in that test), just to polish that off. If it does fail - which I think is the correct failure mode in your case - then I guess it's good to merge. Assuming you're happy with the changes too, obviously...

maxpietsch:
Seems to work!  On my ubuntu, using clang 3.4 with new gcc, clang bails out and gcc works.
```
Looking for compiler [clang++]: Ubuntu clang version 3.4-1ubuntu3 (tags/RELEASE_34/final) (based on LLVM 3.4)
Checking for C++11 compliance: test failed (see configure.log for details)
Looking for compiler [g++]: g++ (Ubuntu 5.4.1-2ubuntu1~14.04) 5.4.1 20160904
Checking for C++11 compliance: ok
Checking for ::max_align_t: ok
Checking for std::max_align_t: ok
```

On my macbook, though, I get tons of warnings
```
In file included from core/phase_encoding.cpp:15:
In file included from ./core/phase_encoding.h:22:
./core/header.h:48:18: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:168:87: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void* operator new[] (std::size_t size) { return (alignof(__VA_ARGS__)>alignof(malloc_align)) ? __aligned_malloc (size) : ::operator new[] (size); } \
```

[This](http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20130128/072868.html)  claims that alignas applied to an expression is not standard conform.

thijsdhollander:
Very keen to see this merged with the tag branch as soon as you guys think it's reasonably ready.

A note (that I didn't mention yet) on why just upping the GCC requirement to 4.9, "since JSON requires that", was definitely not enough: on the systems of the people I quickly tried to help out, I first ran into the JSON test failing.  I quickly "fixed that" by just installing clang, since the JSON test requires either clang (version something something, don't recall by heart) **or** GCC 4.9+.  So my naive/lazy thinking (which I'm sure many other users would do as well) was to not bother with GCC and install clang to resolve.  But if we still have separate requirements about GCC even if clang is present, then that's another matter of course.  Note all of my aforementioned adventures documented in the tag pull request where on pretty standard Ubuntu 14.04 systems; which I reason realistically make up a large chunk of our user base.

jdtournier:
> This claims that alignas applied to an expression is not standard conform. Changing alignof(__VA_ARGS__) to std::alignment_of<__VA_ARGS__>::value fails to compile (without error message).

Yes, I saw those in one of your earlier posts, and assumed that was due to the missing `std::max_align_t`. I'm surprised about this warning, I do think it's standards compliant. The `__VA_ARGS__` is a [variadic macro](https://en.wikipedia.org/wiki/Variadic_macro), which is a C99 standard. It should be fully expanded by the preprocessor before further interpretation by the compiler, so that `__VA_ARGS__` shouldn't even be visible by this stage, it should have been substituted by the macro arguments. The reason for using __VA_ARGS__ is that the preprocessor is not smart enough to interpret a single argument like `Class<Type1, Type2>` as one argument, it will split it at the comma - exactly the same issue as used as an [example on that wikipedia page](https://en.wikipedia.org/wiki/Variadic_macro#Example).

Which version of clang are you using on your Mac? Is this the latest version? If so, I'm really surprised, I would have thought that was a bug... It looks like `clang` might be performing both preprocessor and compiler roles at the same time, and get itself mixed up. Unless I'm reading this all wrong...

Can I also check: all of these warning correspond to different locations, right? Or is there one in particular that triggers the warning?

jdtournier:
OK, I get the same warnings on my Arch desktop using `clang++` 3.9. If this actually relates to the fact that `alignof (Class<T>)` is an expression because `T` is not yet defined, then we're in serious trouble. Problem being that we can't figure out the alignment of the class until it's fully declared (so we know the alignment requirements of its contents), and we can't declare its `operator new` until we know whether it needs special alignment or not. It's a catch-22 situation... The current approach relies on the compiler delaying interpretation of `alignof(Class<T>)`, which is within the `operator new` methods, until instantiation of the class, when it does have all the information to do this. 

I tried all kinds of tricks to allow this to work. Problem is that `alignof` is a compile-time `constexpr`, and the compiler will need to evaluate it at some point. If it tried to do this before the information was available, we're in trouble. IIRC, I can't even do `alignof (decltype (*this))` instead, even though I think that would solve the problem - would be great in fact, we wouldn't need to add the class type in the MEMALIGN macro... But that didn't seem to work in my testing. I'll give it another shot. It might be that we'll have to override the global `operator new`, and align everything. Might be a cleaner solution all round, and may not waste all that much space, depending on how often our applications allocate small objects...

maxpietsch:
> Which version of clang are you using on your Mac?

The latest version of Xcode, [clang++]: Apple LLVM version 8.0.0 (clang-800.0.42.1)

There are no warnings in the current `tag_0.3.16` branch.

> Can I also check: all of these warning correspond to different locations, right? 

```
In file included from src/dwi/directions/file.cpp:15:
In file included from src/dwi/directions/file.h:18:
In file included from ./core/math/sphere.h:24:
In file included from ./core/math/math.h:21:
./core/types.h:202:40: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
    struct container_cast : public T { MEMALIGN(container_cast<T>)
                                       ^
./core/types.h:167:85: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void* operator new (std::size_t size) { return (alignof(__VA_ARGS__)>alignof(malloc_align)) ? __aligned_malloc (size) : ::operator new (size); } \
                                                                                    ^
./core/types.h:202:40: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:168:87: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void* operator new[] (std::size_t size) { return (alignof(__VA_ARGS__)>alignof(malloc_align)) ? __aligned_malloc (size) : ::operator new[] (size); } \
                                                                                      ^
./core/types.h:202:40: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:169:76: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void operator delete (void* ptr) { if (alignof(__VA_ARGS__)>alignof(malloc_align)) __aligned_free (ptr); else ::operator delete (ptr); } \
                                                                           ^
./core/types.h:202:40: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:170:78: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void operator delete[] (void* ptr) { if (alignof(__VA_ARGS__)>alignof(malloc_align)) __aligned_free (ptr); else ::operator delete[] (ptr); }
                                                                             ^
./core/types.h:229:43: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
  template <typename X, int N=(alignof(X)>alignof(malloc_align))>
                                          ^
5 warnings generated.

...

In file included from cmd/labelconvert.cpp:16:
./core/image.h:214:54: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:169:76: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void operator delete (void* ptr) { if (alignof(__VA_ARGS__)>alignof(malloc_align)) __aligned_free (ptr); else ::operator delete (ptr); } \
                                                                           ^
In file included from cmd/labelconvert.cpp:16:
./core/image.h:214:54: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
./core/types.h:170:78: note: expanded from macro 'MEMALIGN'
  FORCE_INLINE void operator delete[] (void* ptr) { if (alignof(__VA_ARGS__)>alignof(malloc_align)) __aligned_free (ptr); else ::operator delete[] (ptr); }
                                                                             ^
In file included from cmd/labelconvert.cpp:16:
./core/image.h:250:3: warning: 'alignof' applied to an expression is a GNU extension [-Wgnu-alignof-expression]
  CHECK_MEM_ALIGN (Image<float>::Buffer);
  ^

... many many more
```

jdtournier:
> There are no warnings in the current tag_0.3.16 branch.

Now that's a really good clue! Also explains why I hadn't come across this before - was wondering how I could have got so far without this issue cropping up. I'll look into that when I have a minute.

maxpietsch:
Not sure of that last commit makes sense but it seems to make clang happy.

jdtournier:
No, that last commit was not a good idea... But the good news is the fix is really simple - the error message was spot on, I had messed up. Commit on its way.

maxpietsch:
Shall we merge this?

PR #785, opened 2016-09-23
Tag 0.3.16

Lestropie:
OK, have started the branch that will be eventually merged to master to form tag 0.3.16, which _should_ be the last tag before release (I think I've said that before though...). It will be a bit of an agglomeration of all sorts of things, but the primary one is getting the new fixel format pushed out and tested.

Main changes already in this branch:
- Changes to mesh handling code: Not really anything new, just making it all more modular for future enhancements. Can now import FreeSurfer meshes (in code at least).
- Everything that was in the `stanford` branch. This includes:
  - Fixing all statistics code to use double precision (except for CFE connectivities), and getting some code into `.cpp` files to reduce RAM usage / increase parallelism in compilation. Also changes certain structures and whatnot to improve compatibility with non-CFE stats code usage. Supersedes #701.
  - `mrhistmatch`: Non-linear histogram matching (#671).
  - `connectomestats`: NBSE method (#723).
  - `vectorstats`: Basically does permutation testing on 1D data (#724).
  - NIfTI-2 support (#19).
  - JSON file support, as provided by JSON for Modern C++ (#725).
  - Get phase encoding information from DICOM and store in header (#725).
  - `dwipreproc`: Complete internal re-design, including using phase encoding information from image headers to set up `topup`/`eddy` (#747).

Things that still need to be merged in / could potentially be merged in / need development / testing:
- New fixel format (#620). Close from what I hear.
- System signal handlers (#776, #758): Awaiting confirmation that @jdtournier is content with the implementation. Would be nice to set up similar handlers for the Python scripts, but that can wait.
- `dwipreproc` needs testing - I've done what I can, it would be nice if others try to break it. Also need to pick a better name for `-topup_images` option; I'm thinking something like `se_epi_field_images` since dual gradient echo-based field images could theoretically be used, but wouldn't work with this particular option.
  Edit: Have changed it to simply `-se_epi`; complain if you don't like it.
- Need to test DICOM phase encoding info import: Acquire SE EPIs with axial / sagittal / coronal slices, all 4 possible phase encoding directions for each.
- Need those with non-Siemens data to figure out how other vendors provide the phase-encoding information in DICOMs and set up the import appropriately.
- 'Stashing' `dw_scheme` once it no longer applies to the image volumes (#774): Let's pick a name.
- `dwigradcheck` (#534): Since the problem has come up on the forum a few times, this would be useful to have. It's still a bit preliminary though. @bjeurissen Keen to put some effort into this?

PR #920, opened 2017-02-25
Scripts: Better handling of scripts invoking other scripts

Lestropie:
Some discussion already in #785.

Needed some gymnastics to get it working on Windows on Python2. Basically, even once you find the shebang, neither `/usr/bin/env` nor `/usr/bin/env.exe` works, and determining the MSYS2 root directory is not as easy as one would think (`os.path.abspath()` just inserts the *current* drive letter to the start of the string). So instead I just bypass `env`, and use `distutils.spawn.find_executable()` as is already used elsewhere.

Needs testing across OS's and Python2/3.

### DIFF
--- a/lib/mrtrix3/run.py
+++ b/lib/mrtrix3/run.py
@@ -72,6 +72,9 @@ def command(cmd, exitOnError=True):
       if is_mrtrix_exe:
         item = versionMatch(item)
       item = exeName(item)
+      shebang = _shebang(item)
+      if len(shebang):
+        new_cmdsplit.extend(shebang)
       next_is_exe = False
     if item == '|':
       if is_mrtrix_exe:
@@ -80,8 +83,8 @@ def command(cmd, exitOnError=True):
         # Get MRtrix3 binaries to output additional INFO-level information if running in debug mode
         if app._verbosity == 3:
           new_cmdsplit.append('-info')
-#        elif app._verbosity < 2:
-#          new_cmdsplit.append('-quiet')
+        elif not app._verbosity:
+          new_cmdsplit.append('-quiet')
       next_is_exe = True
     new_cmdsplit.append(item)
   if is_mrtrix_exe:
@@ -89,8 +92,8 @@ def command(cmd, exitOnError=True):
       new_cmdsplit.extend( [ '-nthreads', str(app._nthreads) ] )
     if app._verbosity == 3:
       new_cmdsplit.append('-info')
-#    elif app._verbosity < 2:
-#      new_cmdsplit.append('-quiet')
+    elif not app._verbosity:
+      new_cmdsplit.append('-quiet')
   cmdsplit = new_cmdsplit
 
   # If the piping symbol appears anywhere, we need to split this into multiple commands and execute them separately
@@ -314,7 +317,7 @@ def exeName(item):
 #   (e.g. C:\Windows\system32\mrinfo.exe; On Windows, subprocess uses CreateProcess(),
 #   which checks system32\ before PATH)
 def versionMatch(item):
-  import distutils, os, sys
+  import os
   from distutils.spawn import find_executable
   from mrtrix3 import app
   global _mrtrix_bin_path, _mrtrix_exe_list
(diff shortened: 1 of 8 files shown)


Lestropie:
OK, so I've verified this as working on Windows (Python 2 & 3) and Linux (Python 2.7.6, 2.7.13 and 3).

Example debugging output (on Linux) to show what it's doing:
```
dwi2response: Running 'tournier' algorithm to select 211 single-fibre WM voxels.
dwi2response: [DEBUG] versionMatch(): Version-matched executable for dwi2response: /software/mrtrix3/bin/dwi2response
dwi2response: [DEBUG] _shebang(): File "/software/mrtrix3/bin/dwi2response": string "#!/usr/bin/env python": ['/usr/bin/env', 'python']
Command:  dwi2response tournier dwi.mif _respsfwmss.txt -sf_voxels 211 -iter_voxels 2110 -mask refined_wm.mif -voxels voxels_sfwm.mif -tempdir /home/rob/tmp/dwi2response-tmp-5PX8B4/ -nocleanup
dwi2response: [DEBUG] command(): To execute: [['/usr/bin/env', 'python', '/software/mrtrix3/bin/dwi2response', 'tournier', 'dwi.mif', '_respsfwmss.txt', '-sf_voxels', '211', '-iter_voxels', '2110', '-mask', 'refined_wm.mif', '-voxels', 'voxels_sfwm.mif', '-tempdir', '/home/rob/tmp/dwi2response-tmp-5PX8B4/', '-nocleanup', '-info']]
```

Merging to `tag_0.3.16` so that hopefully it's tested on Mac.

It's not a terribly elegant solution, but I simply don't see an alternative that is clearly better. On Windows, once you're running within the Python environment you no longer know where the MSYS2 root path is, and therefore can't find `/usr/bin/env`. You could look for an executable in the path called `env` or `env.exe`, but there's an off chance that that could be some other command (bearing in mind that Windows has its own `mrinfo.exe`...). And using `shell=True` in `subprocess` (even if only for script targets) re-introduces a security hazard with malicious user input. If anybody has a better idea than the solution in 4d4e9f6, speak up.

jdtournier:
> using shell=True in subprocess (even if only for script targets) re-introduces a security hazard with malicious user input.

I appreciate the concern with security here, but I don't think this is really applicable here. According to the warning at the end of [this section in the Python docs](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments), the risk is that _if input comes from an untrusted source_, executing it directly may allow a malicious attacked to inject commands to be executed with the invoking script's permissions. This would be a huge concern for a network-enabled script, if input were genuinely received from a remote destination. It would also be a major concern for e.g. a system daemon, for the same reasons. But if the user themselves invokes the script, there's nothing the script can do that they couldn't do themselves already (and in a simpler way). There might be a concern with malformed input from the user causing unexpected consequences, but that would be very unlikely indeed - they'd need to type something like `dwi2response blah blah blah \; rm -rf important_stuff` or something to get in trouble - i.e. to escape the command delimiter (`;`), otherwise the shell itself will interpret the command delimiter and perform the second command itself anyway, although after script completion (no doubt there's other ways to supply malicious input, but the principle applies nonetheless). And that assumes that our script's own argument parsing hadn't already fallen over with the bad input. The only other way this could be a security risk would be if the user/sysadmin had installed the script with `setuid` permissions - in which case they'd deserve everything that's coming to them... Or the user invoked their bad command with `sudo` or something. So I don't think using `Shell=True` is a particular concern for us: if users _want_ to issue malicious code, they can do it already more easily without our use of `Shell=True`; and if we're worried about bad input, they'd really need to be typing a fantastically _bad_ command for this to actually do anything...

> It's not a terribly elegant solution, but I simply don't see an alternative that is clearly better.

Yes, I agree it's not cool. [This post](http://stackoverflow.com/a/912847) mentions a way of simply invoking the same executable as the current script, which would be much cleaner _provided we only invoke `python` scripts_. I think that would be an acceptable compromise, a script needing to invoke another script is fairly low use case, and something we have control over. It wouldn't prevent anyone from using `bash` scripts themselves. This would only mean that if one of our scripts needs to invoke someone else's script, that other script would need to also be a `python` script. 

The problem with this is that it would presumably break the use of e.g. FSL scripts within e.g. `dwipreproc`. I don't know whether we currently invoke any of their scripts, or only their executables, but if there are any calls to non-`python` scripts, this would break it. I'd be more in favour of running with `Shell=True` to keep things simple... 

Note that we can still use the shebang detection, and only run with `Shell=True` when we detect a script.

Lestropie:
> This post mentions a way of simply invoking the same executable as the current script, which would be much cleaner provided we only invoke python scripts. ... The problem with this is that it would presumably break the use of e.g. FSL scripts within e.g. dwipreproc

First thing I have to look at again in the morning is `5ttgen fsl`, which invokes FSL's `standard_space_roi`, which is `sh`. I can fix it, it just means resolving the absolute path to the script before `subprocess` invokes the interpreter. More generally though, I don't want an inadequacy in the library causing someone to realize halfway through script development that they can't use it. Plus, we haven't even got all of the *MRtrix3* scripts over to Python yet, so technically we couldn't even invoke all of our own scripts...

jdtournier:
Agreed. Which is why I think `Shell=True` is the simplest workaround...

Lestropie:
Was tempted to go with `shell=True`; but [this](http://stackoverflow.com/a/3172488) may also cause an issue for Windows... Going to keep trying with what I have (#922), but if I hit a dead end I'll give `shell=True` a go.

lib/mrtrix3/run.py
```diff
@@ -335,3 +338,52 @@ def versionMatch(item):
...
+      path = find_executable(exeName(item))
+  if not path:
+    app.debug('File \"' + item + '\": Could not find file to query')
+    return []
+  # Read the first 1024 bytes of the file
+  with open(path, 'rb') as f:
+    data = f.read(1024)
+  # List of permissible text characters
+  textchars = bytearray({7,8,9,10,12,13,27} | set(range(0x20, 0x100)) - {0x7f})
+  # Are there any non-text characters? If so, it's a binary file, so don't go looking for a shebang
```
jdtournier: I probably wouldn't bother checking so hard. There is an outside chance that the script might have been written with a comment in some non-English language in UTF-8 or something, which will cause this fail. I would just stick with detecting a `#!`: no binary executable will start with that. On Linux, we'd have ELF format binaries, which will start with (0x45)ELF, it'll be something equally obscure on Windows. I think it would be sufficient to assume that a file starting with `#!` is a script, without checking whether the rest is interpretable as text.

lib/mrtrix3/run.py
```diff
@@ -335,3 +338,52 @@ def versionMatch(item):
...
+    return []
+  # Try to find the shebang line
+  for line in data.splitlines():
+    line = line.strip()
+    if len(line) > 2 and line[0:2] == '#!':
+      shebang = line[2:].split(' ')
+      if app.isWindows():
+        # On Windows, /usr/bin/env can't be easily found, and any direct interpreter path will have a similar issue.
+        #   Instead, manually find the right interpreter to call using distutils
+        if shebang[0] == '/usr/bin/env':
```
jdtournier: Might even go for `os.path.basename(shebang[0]) == 'env'` here...

lib/mrtrix3/run.py
Lestropie: This would be vestigial code from other things I was trying re: detecting which files in `bin/` are binaries v.s. scripts now that they're all in the same directory. As long as binaries are guaranteed to have some deterministic code at the start, sure this can go.

jdtournier: > As long as binaries are guaranteed to have some deterministic code at the start, sure this can go.

I think this is a safe assumption, although the one case that troubles me is static binaries. I'm not sure whether they have a fixed 'magic number' at the start. If they really just consist of raw binary without any form of lead-in, then there is an outside chance that static executables might start with that combination - the chance is only ~1 in 65,000, but...

jdtournier: Just checked, on Linux static binaries still start with the same magic number as dynamic ones (0x7f ELF - got it wrong last time). Given that this is likely to be the only platform where static binaries are likely to be produced, I think we're safe...

jdtournier: Also checked for Windows: they use the [portable executable format](https://en.wikipedia.org/wiki/Portable_Executable), which has 0x5A4D as the signature (magic number).

lib/mrtrix3/run.py
Lestropie: Yes, I suppose if somebody adds a shebang pointing explicitly to some other executable called '`env`' they probably deserve what's coming to them.

PR #921, opened 2017-02-27
tckgen interface changes

thijsdhollander:
In line with discussions in #809.  All interface changes as mentioned in [this](https://github.com/MRtrix3/mrtrix3/pull/809#issuecomment-261190421) and [this](https://github.com/MRtrix3/mrtrix3/pull/809#issuecomment-261850167) post, *and then some*.  The easiest diff to get an idea of the changes would be the docs of `tckgen` (i.e. [tckgen.rst](https://github.com/MRtrix3/mrtrix3/pull/921/files#diff-a74fa826d1903d31dd24055051094347)).  The way the `-select` option now ends up at the top of the list, and the fat chunk of explanatory text that goes with it, should IMHO guarantee that users spot it (which was an initial concern of @jdtournier upon proposing the `-select` name).  Existing users will of course get caught out at least once if they don't check the massive changelog that will come with the `tag_0.3.16`... but then again, they will be in for a shock on many other fronts too. :smile_cat: 

Beyond individual changes to the interface, the overarching idea was to make the most common use cases (i.e. integrate over all users' use cases, taking into account the frequencies of said use cases, and potentially slightly weighting in favour of newer users that rely more on documentation ordering to find what they need) the most obviously documented.  This led to reordering of the categories of options, as well as moving around some options (even between categories).  This worked out quite well, with some lesser used options pertaining to seeding/initialising tracks ending up in the seeding category... so they're actually "out of the way" in the main tractography options category.

So erase all your preconditioned minds, and give the `tckgen` interface a look. :pray:  (think about Tim Cook's statements about how we have to be courageous and all...)

TODO:
* The `-num_seeds` option [doesn't work yet](https://github.com/MRtrix3/mrtrix3/commit/849dd03d63f4099141b4475b36b031f5e6718bc3#diff-a2113f6dabd2a0bae734f3ad8aa287aeR156), it just sets the `exact_num_attempts` property.  I'll leave the implementation to a `tckgen` adept (@Lestropie, @jdtournier?), so it performs exactly as to the functionality and use case you guys had in mind.
* Changes will probably have to happen to some docs that have `tckgen` examples in them...?


jdtournier:
I'm not sure I'll have the time to look into this any time soon - it'll take a while to re-familiarise myself with this part of the code. The other outstanding issue was the suggestion that maybe some of the header key names should change to match. I'm guessing something like `count` → `num_selected`, `total_count` → `num_seeds`, or something similar. Just worried about breaking all manner of other commands and scripts that might rely on these labels... I think `count` can remain as-is, it's unambiguous as to what it means once you have the file. I can see that `total_count` on the other hand is fairly meaningless without an explanation, so that's the one I might be tempted to change. But I'm not sure how safe safe it is to rename it: it's explicitly mentioned in `tckedit` and in the `Tractography::FileBase` class, so at the very least we'd need to keep interpreting that key as a synonym for the new one, at least for the foreseeable future, if we want to maintain our ability to interpret existing track files. Gets very messy once we start doing this...

I also note that this value only makes sense for a file generated by `tckgen` directly - if it's been edited by e.g. `tckedit` or `tcksift`, it's no longer consistent with the file contents. Not sure whether this is something we need to do anything about...?

Lestropie:
- I would consider splitting "Tractography seeding options" to have seed *sources* in one group (maybe with "must provide at least one" in the option group name?), and *options* / *parameters* affecting seeding behaviour in another.

- I think the contrast in option name between `-max_seeds` and `-num_seeds` needs to be increased (will come back to this).

> I'm not sure I'll have the time to look into this any time soon - it'll take a while to re-familiarise myself with this part of the code.

I'm happy to do the coding aspect, as long as I'm convinced that I actually understand and can make work what people are asking for.

Currently the termination of `tckgen` is handled primarily by the class at the receiving end of the queue, which detects either the total number of streamline "attempts" (received streamlines that are empty or not; now `-maxnum`, proposed to be `-max_seeds`) or "selected" streamlines (non-empty streamlines; now `-number`, proposed to be `-select`). The other mechanism that can cause `tckgen` to stop is if a "finite" seed mechanism is used (`-seed_grid_per_voxel` or `-seed_random_per_voxel`), when worker threads are denied their requests for new seeds.

From the description for the new proposed `-num_seeds` option, I think the intended behaviour is to use a counter in the `SeedList` class, that triggers `tckgen` termination once a certain number of seeds have been drawn. This option is therefore essentially putting a finite limit on otherwise "infinite" seeding mechanisms, and cannot be used in conjunction with "finite" seeding mechanisms. This quantity is *not* influenced by the frequency with which the tracking algorithm rejects the seed at the `init()` step: The number of seeds that get through the `init()` step , and therefore that the algorithm actually "tracks from" (as opposed to "attempts to track from") should be reflected in `-maxnum` / `-max_seeds`.

If this is the case, the option names / descriptions could be tweaked a bit. But I'll wait to confirm that this is in fact correct.

Also regarding option naming: An alternative to throw out there might be "output" v.s. "generate"; the description would then clarify that "output" is all generated streamlines that satisfy all criteria imposed.

> I also note that this value only makes sense for a file generated by tckgen directly - if it's been edited by e.g. tckedit or tcksift, it's no longer consistent with the file contents. Not sure whether this is something we need to do anything about...?

FYI Generally anything that extracts a subset of tracks, `total_count` is hacked to correspond to the number of tracks at *input* (by nulling the track but still sending it to the writer class). It may or may not make sense, and it may or may not be useful, it's just making use of the functionality that's there.

jdtournier:
> FYI Generally anything that extracts a subset of tracks, total_count is hacked to correspond to the number of tracks at input (by nulling the track but still sending it to the writer class).

Cool, that makes sense anyway. In which case it's OK to leave the terminology as-is, since its meaning depends on how the file was derived. That's one issue we don't have to deal with then.

> An alternative to throw out there might be "output" v.s. "generate"; the description would then clarify that "output" is all generated streamlines that satisfy all criteria imposed.

Sure, I'd be happy with that. Slight issue with that though is that `-output` in this context might be misinterpreted as the output track file, but that's a minor issue. Same problem as with `-select` really: perfectly clear when looking at someone else's command (`-output 1M` /  `-select 200K`), but maybe not so clear in isolation. I _think_ I still prefer `-select`, but either will do.

> Edit: Just thought of another detail (which is kind of what spawned this whole thing): If you set -number 0 / -select 0, then tckgen will continue until the number of streamlines specified with -maxnum / -max_seeds have been generated. This should appear in the -number / -select description.

My understanding was that `-num_seeds` would override `-select` and `-maxseeds`. But based on your other comments about the distinction between `-num_seeds` and `-max_seeds`, I think we need to clarify these issues, I'm not sure I completely follow all of the issues here. So:

> From the description for the new proposed -num_seeds option, I think the intended behaviour is to use a counter in the SeedList class, that triggers tckgen termination once a certain number of seeds have been drawn. This option is therefore essentially putting a finite limit on otherwise "infinite" seeding mechanisms, and cannot be used in conjunction with "finite" seeding mechanisms. This quantity is not influenced by the frequency with which the tracking algorithm rejects the seed at the init() step: The number of seeds that get through the init() step , and therefore that the algorithm actually "tracks from" (as opposed to "attempts to track from") should be reflected in -maxnum / -max_seeds.

I'm just going to try to clarify what's going in my head, and give them labels so I can reason about these (I'm not suggesting these as _the_ terminology, just to clarify the different concepts here). I have a feeling this probably repeats a lot of what's already been said, but bear with me, I'm getting old (40 next week... 👴). So we have:

- _attempts_: attempt to start from a specific seed location along some orientation
- _seeds_: (exhaustive) attempt to start from a specific seed location
- _starts_: successful _initiation_ of a streamline from a seed
- _selections_: a streamline generated from a seed that meets the criteria and makes it to the output

So currently, as far as I can tell, for 'infinite' seeders, _attempts_ happen at random locations within the seed region, with random orientations (unless `-initdirection` is set), and if the algorithm fails to start along that orientation from that location (i.e. below threshold), then it moves on to a fresh _attempt_ at a different _seed_ location. There is a `max_attempt` member in the `Seeder::Base` class, which is stated to be the "maximum number of times the tracking algorithm should attempt to start from each provided seed point", but it's currently unused, as far as I can tell (judging from a quick `grep -r max_attempts src/ cmd/`) - I'm guessing this is works-in-progress.

So as currently implemented, we'll have a _lot_ more _attempts_ than _seeds_, even in regions where seeding should be no problem, since most of the random orientations won't be above threshold - what [this post](https://github.com/MRtrix3/mrtrix3/pull/809#issuecomment-264570346) was about. This isn't a problem if we just keep going and don't care about the number of attempts/seeds, but if that number is to be meaningful, I think we'd need to fix this to make sure that the terminology was consistent and well-defined, so that a _seed_ really involved trying a bit harder than just a single _attempt_. Seems a bit problematic for iFOD2 from what @Lestropie says, but I guess we can just ask it to try harder in this case. 

But this still leaves the distinction between _seeds_ and _starts_ (as I've defined them above). If we change behaviour so that _seeds_ genuinely are a bit more exhaustive, then I think it would make sense to have `-max_seeds` and `-num_seeds` both influence the number of _seeds_, even if the final number of actual _starts_ is then allowed to be lower than this. This I think would be consistent with the finite seeders, where any seeds that aren't successful just don't lead to _selections_, but get recorded in the statistics as `NO_PROPAGATION_FROM_SEED`.

In this case, I think we can still rely on using the receiver to terminate tracking, provided streamlines get passed down the queue even if seeding was not successful, and the statistics get recorded right - seems like a relatively trivial change as far as I can tell (?), provided we can sort out the 'more attempts per seed' problem. And in this case, it makes sense that these options would have no effect if finite seeders are used. 

Sorry for the duplication, but hopefully this matches what we all have in mind here...?

thijsdhollander:
I'll answer with a few separate posts, so things don't get too mixed up.  Here's number 1:

> I would consider splitting "Tractography seeding options" to have seed sources in one group (maybe with "must provide at least one" in the option group name?), and options / parameters affecting seeding behaviour in another.

Sounds like a good plan.  The reordering of categories and options within/between categories I already did was with the intention of having more commonly used stuff at the top.  If this split would happen, I'd list the "seed sources" group right after the "streamlines tractography options" category (which is now the first one, prominently featuring `-select` as the first option).  I reason "options/parameters affecting seeding behaviour" is definitely less used by a large chunk of users; these options go more towards specialised scenarios.  Note I've also swapped the `-seed_sphere` and `-seed_image` options: I reckon `-seed_image` is definitely much more used (and more usable, since one can just draw/save an ROI for this purpose, rather than noting coordinates and radii).

thijsdhollander:
Number 2:

@jdtournier , either I'm misinterpreting something, or there's an inconsistency in your explanation above...  Give your first statements:

> So currently, as far as I can tell, for 'infinite' seeders, attempts happen at random locations within the seed region, with random orientations (unless -initdirection is set), and if the algorithm fails to start along that orientation from that location (i.e. below threshold), then it moves on to **a fresh attempt at a different seed location**.

...I fail to combine these with this statement:

> So as currently implemented, we'll have **a lot more attempts than seeds**, ...

It seems to me that the first bit above implies that we have an *equal* amount of attempts and seeds (because a fresh attempt is at a *different* seed), but that we have a lot more attempts (and seeds) than *starts*.

So your goal, if I get it right, is to reduce the drastic difference in numbers between seeds and starts, by doing more attempts per seed.  Or semi-formally, we currently have:

`attempts == seeds >>> starts`

and the proposed changes would aim for:

`attempts >> seeds > starts`

Where my numbers of ">" symbols vaguely describe levels of "more than".

Apart from checking whether I got it right, I do fully agree with it though.  This would make the seeds concept (and number) much more meaningful, because, as you say, the number of attempts isn't really all that meaningful (it's just high because of most FODs typically being very sparse).  I also fully agree that we'd want to look at it from the point of view of *seeds*, and not *starts*.  Seeders that e.g. do a fixed number of seeds per voxel, say for example 1 seed for each voxel in the centre of the voxel, also implicitly state the number of *seeds* beforehand (e.g. the same as the number of voxels in the mask, for the example of 1 seed per voxel), but not the *starts* (which can end up to be less than the number of seeds).

thijsdhollander:
Finally, number 3:

In line with the above, I agree with

> But this still leaves the distinction between seeds and starts (as I've defined them above). If we change behaviour so that seeds genuinely are a bit more exhaustive, then I think **it would make sense to have -max_seeds and -num_seeds both influence the number of seeds, even if the final number of actual starts is then allowed to be lower than this**. This I think would be consistent with the finite seeders, where any seeds that aren't successful just don't lead to selections, but get recorded in the statistics as NO_PROPAGATION_FROM_SEED.

The number of starts, to me, is quite an artificial concept again (even though it exists programatically): if a track is started, but not selected, then it basically just fulfils *part of* the constraints, but not all.  To me, from a seed(-point, potentially with many attempts), you end up with a *selected* track, or with none at all.  A "started" track may just as well only have done one step and then failed to proceed due to the other constraints... how much more of a track is it then, compared to the one ("of length zero") that didn't succeed to start from the seed at all?

So building on this logic, I think the "generated" terminology may not be favourable: "generated" seems to refer to the "starts" concept, rather than the "seeds" one.  It's funny actually, we currently (even before this branch and pull request) already have the word "selected" used somewhere; in the command line output during a run, e.g.:

```
tckgen: [100%]     1177 generated,     1000 selected
```

But in the context of the above proposal, that "generated" would better become "seeds".  The "selected" might become "selected tracks", so it's a bit more explicit.

Finally as to `-output` versus `-select`; I think I also prefer `-select` just slightly more.  It may just be me (and @jdtournier), but I also get triggered by `-ouput` into thinking that it's an option to supply an output file name to.  `-select` feels, to me, more in line with the fact that it survived a set of constraints.  As evidenced by the above command line output, it's also slightly more in line with the terminology of "selected" we already have out there.

jdtournier:
About point 2: yes, we're on the same page. I guess my attempts at defining my terms failed to clarify the argument... Oh well. 

But yes, the point is that currently, with infinite seeders there are really no such thing as _seeds_ (as defined above), only _attempts_, since there is no proper effort to try to _start_ from each _seed_ location (only a single _attempt_). What I meant by my contradictory comment: 

> As currently implemented, we'll have a lot more attempts than seeds

is that very few of these _attempts_ will pass the `init()` stage, and hence be considered as genuine _seeds_ from which streamlines can be initiated - there's a difference in terminology here, which probably contributes to the confusion... 

I guess the big distinction is that currently, seed location/directions are drawn from the combined seed ROI / orientation domain by random rejection sampling, and seeds produced by this process are already guaranteed to have FOD amplitude above threshold along the corresponding direction, so are already valid _start_ points too - means there is no distinction between a _seed_ and a _start_. What we're proposing these seeders should do is a more exhaustive search (lots of _attempts_) for a suitable initial direction of tracking from each _seed_, and if successful, then that _seed_ is also a _start_ - otherwise it remains just a (failed) _seed_. To use your notation:

- Currently: `attempts >>> seeds == starts > selected`
- Proposed: `attempts >> seeds > starts > selected`

Note here that the exact definition of the term 'seed' differs between the two versions, which is probably what's causing the apparent contradiction here...

jdtournier:
About point 3:

> But in the context of the above proposal, that "generated" would better become "seeds". The "selected" might become "selected tracks", so it's a bit more explicit.

Sounds sensible to me, happy with that.

jdtournier:
More about point 3:

> The "selected" might become "selected tracks", so it's a bit more explicit.

Not altogether against the idea, but I'd rather steer well clear of the word 'track' - as I've already discussed in [this recent forum post](http://community.mrtrix.org/t/beginner-dwi-preprocessing-for-tracking-trackvis/588/16?u=jdtournier)...

jdtournier:
> Damn you and your timing @jdtournier...

??? Aren't you supposed to be in bed already...? :wink: 

> The fact that there's four items in this list, but we're discussing three command-line options, demonstrates the source of the confusion.

Good observation. But I thought what we were proposing was supposed to cut the list down to three...? I think where we are still not quite on the same page is:

> This block defines, for each "type" of seeder, how many times (by default) the algorithm should try to find a valid initial tracking direction from a provided point. This is in fact mostly based on the computational expense associated with each "type" of seeder: if generating a seed point is expensive, you don't want to discard a seed point after just one direction sample.

What I had in mind what to get rid of the _attempts_ (my definition) as something that can influence outcome: it would all be about _seeds_ (my definition again). To do this, we need to have a mechanism that always performs an exhaustive search for a suitable seed direction from a given seed location (i.e. a sufficient number of _attempts_ per seed location that the actual maximum number allowed is to all intents and purposes irrelevant). It just needs to be set high enough that the chances of that seed location being suitable for initiating a streamline despite failing to find a suitable direction are very low. Would be better yet to have more foolproof mechanisms, but that's probably the easiest to implement, given that it basically already is implemented. 

So the idea then would be that _seeds_ are genuinely just candidate locations from which to start, and these would be the statistics reported in the output, etc. The idea is to make the finite and non-finite seeding strategies consistent (I'm not talking about implementation in the code at this point, this is purely about UX). In finite seeders, seed points are predetermined, and those from which streamlines cannot be started are not replaced by others - there are simply no streamlines starting from that point, not sure how that is captured in the output (given your comment about `NO_PROPAGATION_FROM_SEED` - more on that later). If we ensure that each seed location is genuinely given a fair go (by exhaustively searching for a suitable direction from that seed), then we can treat seeds in non-finite seeding approaches in the same way, and the number of seeds reported in the output would be a more informative metric (for instance, the _seeds_ / _starts_ ratio would give an idea of the seedable fraction of the seed ROI). 

> It might be better if I hack Donald's labels (sorry... but I do think it will actually help):

No problem, but I think your labels actually help to clarify _current_ behaviour. My labels were more about trying to define how we'd like things to work. I agree _candidate seeds_ is probably clearer than _attempts_, but the intent is the same (as far as I can see?). Where we differ is in the distinction between _actual seeds_ and _generated tracks_ (I'd consider both to be _starts_ using my labels): a failure to propagate even a single step from the seed presumably happened for the very same reason a termination or rejection would happen anywhere along the streamline, so there is nothing special in principle about it happening at the very first step (unless I'm missing something?). So I'd class that as a successful _start_, but a failed _selection_ - I don't find the distinction between _actual seed_ and _generated track_ to be useful here. I think it would be much more useful to reason about seed locations from which streamlines can or can't be generated (given the criteria for streamline initiation), which is reflected in the number of _seeds_ that are also _starts_. 

In terms of implementation, I'm guessing what this means is that the algorithm should try the `init()` call from each seed a large-ish number of times before calling it a failure. I'm assuming these macros you point to earlier (`MAX_TRACKING_SEED_ATTEMPTS_DYNAMIC`, etc) to just that? I can see they set the [seeder's `max_attempts` variable](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/seeding/basic.h#L45), which I note then goes on to [set the Shared class's `max_seed_attempts`](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/tracking/shared.h#L95), which already sets the [number of attempts per seed location](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/algorithms/iFOD1.h#L143-L149). So the mechanism is already in place. I guess I'm just not sure that the 'best' number here should be dependent on the seeding mechanism - I can see that as implemented, some seeders really do need to be more exhaustive than others, but the way I'm thinking about this suggests that it should just be large regardless, so that the probability of misclassifying a _start_-able seed as non-startable is low, but this has to be balanced against the computational overhead of spending too long trying in regions that are simply not suitable for seeding. 

> No, there might be a misunderstanding there. NO_PROPAGATION_FROM_SEED purely catches the case where the method's init() function returns true (i.e. it finds a direction that's above threshold), but next() fails in both directions. This is independent of the seeding mechanism.

OK, yes that was a misunderstanding. I didn't realise this was a new addition, by the way... But like I said above, I think now that you've added it, it would be better to use that to indicate a failure to _start_ (no successful `init()` call after lots of _attempts_). 


>> So building on this logic, I think the "generated" terminology may not be favourable: "generated" seems to refer to the "starts" concept, rather than the "seeds" one.
>
> Agreed. That's what I was aiming for: a discrimination between "generated tracks" and "seeds". I wasn't advocating that the former is preferable in any way.

Yep, this is something we do all agree on. 

>> Proposed: attempts >> seeds > starts > selected
>
> That's fine; but note that seeds and starts will be almost equal.

Well, in the current implementation, maybe, since seeds are only considered _actual seeds_ if `init()` returns `true`, which already signals that it's a suitable location for starting. What I'm proposing is that the word _seed_ take on a subtly different meaning, which would actually make it consistent with the finite seeders. 

So I think the discussion is getting a little messy because there's confusion as to whether we're talking about the current implementation or what we'd like it to do. I think we should focus exclusively on what we'd like it to do, and only bring up issues of implementation if we can already foresee that the things being discussed would be difficult to implement. Once we are all happy with what we think the ideal behaviour should be, then we can focus on what might be needed to make it happen.

As a brief summary of how I envisage things to work:

- the Seeder suggests seed locations, and we try hard to find a suitable direction of tracking from that, whether by random rejection sampling, or via a user-supplied initial direction. 
- if no such direction is found, the seed is marked as a failure, and recorded in the stats. No streamline passed down the queue.
- if a direction is found, tracking starts
- if the streamline is rejected, it is recorded as a successful start, but is not selected (currently, this is handled by passing an empty streamline down the queue, I think). 
- if the streamline is selected, it is written to the output as expected

I'm not sure how that works out in terms of implementation. I notice there already is a `max_seed_attempts` variable to implement exactly what I'm talking about... So presumably it's more about us agreeing on ideal behaviour and associated terminology...?

Lestropie:
> Aren't you supposed to be in bed already...? 😉

No rest for the wicked :rage4: 

> What I had in mind was to get rid of the attempts (my definition) as something that can influence outcome: it would all be about seeds (my definition again). To do this, we need to have a mechanism that always performs an exhaustive search for a suitable seed direction from a given seed location (i.e. a sufficient number of attempts per seed location that the actual maximum number allowed is to all intents and purposes irrelevant). It just needs to be set high enough that the chances of that seed location being suitable for initiating a streamline despite failing to find a suitable direction are very low. Would be better yet to have more foolproof mechanisms, but that's probably the easiest to implement, given that it basically already is implemented.

OK, so you're boosting the number of attempts per candidate seed point, such that the number of "attempts" is meaningless / not worth reporting / controlling. You could do that by basically boosting the values I linked to, which is I presume what you're referring to. Note however that this will incur a significant performance penalty in some instances: The seeding mechanisms and tracking method are independent, so a provided candidate seed is not guaranteed to have *any* valid initial tracking direction: the larger the number, the longer a thread can hang. You comment on this later:

> ... and the number of seeds reported in the output would be a more informative metric (for instance, the seeds / starts ratio would give an idea of the seedable fraction of the seed ROI).

, but I'm not sure we want to intentionally slow down tracking considerably in order to get that ratio reliably (for the infinite seeders at least; the finite ones essentially *force* you to get that ratio). So we may want a compromise here in terms of the magnitude / interpretation of those numbers. E.g. On a branch on GitLab I started working on rejection sampling for seeding; use the same calibration / sampling / threshold as what is used for tracking, and I reckon the compromise would be pretty logical.

> So the idea then would be that seeds are genuinely just candidate locations from which to start, and these would be the statistics reported in the output, etc. The idea is to make the finite and non-finite seeding strategies consistent (I'm not talking about implementation in the code at this point, this is purely about UX). In finite seeders, seed points are predetermined, and those from which streamlines cannot be started are not replaced by others - there are simply no streamlines starting from that point, not sure how that is captured in the output (given your comment about `NO_PROPAGATION_FROM_SEED` - more on that later). If we ensure that each seed location is genuinely given a fair go (by exhaustively searching for a suitable direction from that seed), then we can treat seeds in non-finite seeding approaches in the same way, and the number of seeds reported in the output would be a more informative metric (for instance, the seeds / starts ratio would give an idea of the seedable fraction of the seed ROI).

I'm increasingly thinking that much of this issue would be resolved by simply renaming "generated" to "seeds".

- For finite seeders, these are already equivalent: In [this block](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/tracking/exec.h#L180-L185), the thread aborts once the seeder(s) run out of seeds; an empty track is sent down the queue if either:

  - the seed fails to initialise;

  - there's no propagation from the seed;

  - the track is rejected.

- For infinite seeders, they're almost equivalent currently. The only issue is that if [a candidate seed is generated, but the method fails to initialise](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/tracking/exec.h#L189-L192), nothing is sent to the `WriteKernel` class. Change that block so that un-successful candidate seeds result in an empty track being sent down the queue, and bump up the number of attempts from each candidate seed, and I reckon everybody's happy.

Though we may still disagree on which three of the four elements we're making available at the command-line: The number of seeds, and the number of nonzero-length streamlines, are either equivalent or very close to it. I would think that if there were three parameters available, it would be:

- The number of attempts per candidate seed point (for which the default would be increased);

- The number of "seeds" (which is almost equivalent to what's currently available as the number of streamlines before criteria are applied, but not quite equivalent);

- The number of selected streamlines.

> I agree candidate seeds is probably clearer than attempts, but the intent is the same (as far as I can see?).

Yes; but "attempts" currently refers to when `init()` succeeds, so they are *tracking* attempts, not *seeding* attempts, which is only going to lead to confusion. Naming them *candidate seeds*, which we then *attempt* to turn into *actual seeds*, reduces the chance of mis-interpretation.

> Where we differ is in the distinction between actual seeds and generated tracks (I'd consider both to be starts using my labels): a failure to propagate even a single step from the seed presumably happened for the very same reason a termination or rejection would happen anywhere along the streamline, so there is nothing special in principle about it happening at the very first step (unless I'm missing something?). So I'd class that as a successful start, but a failed selection - I don't find the distinction between actual seed and generated track to be useful here.

Correct, nothing particularly special; that's precisely why I didn't discriminate between this and a streamline being rejected for being shorter than the minimum length until very recently, when a user asked for it. Even then, it's still classified as a "generated" track in the output file; the only difference is in how it's reported in the `-info` statistics. (Is this where some false contention is coming from?)

> In terms of implementation, I'm guessing what this means is that the algorithm should try different directions from each seed a large-ish number of times before calling it a failure. I'm assuming these macros you point to earlier (`MAX_TRACKING_SEED_ATTEMPTS_DYNAMIC`, etc) do just that? 

Yep.

> So the mechanism is already in place. I guess I'm just not sure that the 'best' number here should be dependent on the seeding mechanism - I can see that as implemented, some seeders really do need to be more exhaustive than others, but the way I'm thinking about this suggests that it should just be large regardless, so that the probability of mis-classifying a start-able seed as non-startable is low, but this has to be balanced against the computational overhead of spending too long trying in regions that are simply not suitable for seeding.

I'm fine with that to an extent. I'm just trying to remember whether or not it was you who wanted one attempt per seed point for basic seeding mechanisms... And if users don't like the arbitrary value, there's a command-line option there to change it. I reckon it should be larger for finite seeding mechanisms, since you want to be *really* sure you can't track from that point.

> But like I said above, I think now that you've added it, it would be better to use that to indicate a failure to start (no successful init() call after lots of attempts).

I don't think you'd retain the same name. You'd want to somehow report the difference between seeds and generated tracks, sure; but `NO_PROPAGATION_FROM_SEED` is very specific in *differentiating* cases where a seed is initialised but then no propagation occurs, from cases where the seed never initialised. But that's just a naming issue.

> if no such direction is found, the seed is marked as a failure, and recorded in the stats. No streamline passed down the queue.

That takes a bit of change, because the progress bar is generated based on what makes it to the `WriteKernel` class. If you want the terminal to present the number of *seeds* and *selected tracks*, rather than *generated tracks* and *selected tracks*, then information regarding the number of seeds (both successful and failed) needs to be sent down the queue.

A tempting option is to e.g. send an empty track for a failed seed, just the seed point for a rejected track, and the full track for a selected track. The progress bar would report "seeds" as the sum of all three. Would need special handling for the `SeedTest` algorithm (since it outputs single-point "tracks").

Alternatively, just send an empty track down the queue regardless of whether the seed failed to initialise, or the streamline was rejected. Whether or not this is an option might depend on whether you want all three of these parameters to go into the track file header, or whether you're willing to simply substitute the existing "generated" / `total_count` parameter with "seeds" (both successful and otherwise).

Lestropie:
Forward warning: I might have a go at modifying some code tonight. Maybe if we discuss what needs to happen there rather than trying to align our brains, our keyboards will be spared.

jdtournier:
> Forward warning: I might have a go at modifying some code tonight. Maybe if we discuss what needs to happen there rather than trying to align our brains, our keyboards will be spared.

Too late. I'd already written most of the crap below before your post... 

---

OK, I think we're reaching some kind of consensus :champagne: 

Just a few remaining points of contention for me though: 

> I'm increasingly thinking that much of this issue would be resolved by simply renaming "generated" to "seeds".

As currently implemented, that would make sense, in that with infinite seeders, seeds _are_ generated tracks. But I think we have the opportunity to make the terminology consistent between finite and non-finite seeding strategies, so that in both cases, seeds refer to points where an honest attempt at tracking was made. If for instance we decided to output a map of failed seeds, with infinite seeders and the current strategy, the map would cover most of the brain - and users would rightly be pretty confused about why regions with lots of streamlines through them would also have a high density of failed seeds. 

So on this issue, I think we should try to distinguish between seeds, starts, and selections. I'm more than happy to concede that the terminology can be improved, I put forward these labels purely to have short & punchy handles to discuss these concepts, I'm not otherwise emotionally attached to them. But I think it would useful for the progressbar to report on all three of these, e.g.:
```
tckgen: [ 54%]     18423002 seeds,    1394832 streamlines,     541319 selected
```
or whatever labels we feel are appropriate.

> I'm just trying to remember whether or not it was you who wanted one attempt per seed point for basic seeding mechanisms...

Most likely that was me... But I'm allowed to change my mind too. :innocent: 

> Though we may still disagree on which three of the four elements we're making available at the command-line: The number of seeds, and the number of nonzero-length streamlines, are either equivalent or very close to it. I would think that if there were three parameters available, it would be:
> 
> - The number of attempts per candidate seed point (for which the default would be increased);
> 
> - The number of "seeds" (which is almost equivalent to what's currently available as the number of streamlines before selection criteria are applied, but not quite equivalent);
> 
> - The number of selected streamlines.

I think I'd be happy with that. We might disagree on what constitutes a 'seed' though - as above. But I agree with this list either way. 

> You'd want to somehow report the difference between seeds and generated tracks, sure; but NO_PROPAGATION_FROM_SEED is very specific in differentiating cases where a seed is initialised but then no propagation occurs, from cases where the seed never initialised. But that's just a naming issue.

What I'm not sure I understand is why this particular case differs from the general early termination case (i.e. only a couple of points in the streamline, not selected because shorter than `min_length`)? And why would this need to be reported separately? I guess this user you refer to might have had a specific use case, but I really struggle to understand what that might have been... 

In any case, yes it's an easy change if you want to report these differently. 

> OK, so you're boosting the number of attempts per candidate seed point, such that the number of seed "attempts" is meaningless / not worth reporting / controlling. You could do that by basically boosting the values I linked to, which is I presume what you're referring to. 

Correct.

> Note however that this will incur a significant performance penalty in some instances: The seeding mechanisms and tracking method are independent, so a provided candidate seed is not guaranteed to have any valid initial tracking direction: the larger the number, the longer a thread can hang. 

Absolutely. I notice you set that number to 5,000 for the finite seeders, which is way higher than I was anticipating. But it might be that this is what's required to provide sufficient guarantees here. Another strategy is, as you mention, using the calibration idea from the iFOD2 algorithm: compute the amplitude along a sufficiently dense set of directions, find the max, adjust based on a heuristically determined factor, and if that is below threshold, then there's no point even trying. That might at least prevent most threads from locking up in no-hope regions... 

But otherwise, I agree the simplest strategy is probably to reduce the number of directions samples for the infinite seeders, especially if performance becomes a concern here. And make it a command-line option for those who really care about this. 

> That takes a bit of change, because the progress bar is generated based on what makes it to the WriteKernel class. If you want the terminal to present the number of seeds and selected tracks, rather than generated tracks and selected tracks, then information regarding the number of seeds (both successful and failed) needs to be sent down the queue.

So there's a couple of options here. One is to use atomic variable to store the statistics, so they can be accessed and modified from all threads concurrently. But probably a simpler and more flexible approach is to send specially formatted streamlines down the queue, so that the writer can be modified later if need be to e.g. output all failed seeds, etc. All we'd need is e.g. a triplet of `Inf`'s to signal a failed seed, followed by the seed location itself; a triplet of `NaN`s to signal a non-selected streamline, followed by the seed location. Or the full streamline with such a 'code' appended at the end, we just need to check for it in the writer. e.g. add a final vertex with `[ NaN NaN code ]`, where `code` is some value to indicate the status of the streamline. We could also add the seed location after that, so it can be written out if need be. Lots of possibilities there if we want to play with them... 

But by that stage, we might as well modify the streamlines class with a more explicit representation, and send that down the queue...

Lestropie:
> Too late. I'd already written most of the crap below before your post...

That makes it 1 - 1 then :-P

> What I'm not sure I understand is why this particular case differs from the general early termination case (i.e. only a couple of points in the streamline, not selected because shorter than min_length)? And why would this need to be reported separately? I guess this user you refer to might have had a specific use case, but I really struggle to understand what that might have been...

#734; couldn't find the post. It might have had something to do with the number of reported seeds and the number of terminations not matching? Also not fussed on discriminating this from the minimum length criterion.

> Another strategy is, as you mention, using the calibration idea from the iFOD2 algorithm

I think I mentioned elsewhere that I started working on this (and some other fundamental tracking stuff) at some point on a GitLab branch. Needs more work though.

> That takes a bit of change, because the progress bar is generated based on what makes it to the WriteKernel class. ...

Derp. It's already using a [custom Streamline class](https://github.com/MRtrix3/mrtrix3/blob/tag_0.3.16/src/dwi/tractography/tracking/generated_track.h). Just patch whatever variables are needed onto that.

Lestropie:
OK, so, some confusion that maybe I should have picked up on earlier:

> `-max_seeds`: set the maximum number of seeds that tckgen will attempt to track from. This is used to prevent the program from running indefinitely when no streamlines can be found that match the selection criteria. By default, this is set to 100× the number of selected streamlines. Set to zero to disable, which will result in streamlines being generated until the number specified by -select has been reached.

> `-num_seeds`: set the number of seeds that tckgen will attempt to track from. This overrides both -select and -max_seeds. Use this option if you genuinely need a constant number of seeds rather than selected streamlines. However, note that in most cases, the -seed_random_per_voxel or -seed_grid_per_voxel options are likely to be more appropriate.

A value of zero can be used for the `-select` option just as it can be used for `-max_seeds`. Therefore, the functionality that's proposed here as "`-num_seeds`" can actually be achieved by providing `-select 0`, and setting the desired number of seeds via the "`-max_seeds`" option. The functionality in "`-num_seeds`" is therefore redundant; unless I'm still missing something?

jdtournier:
> A value of zero can be used for the -select option just as it can be used for -max_seeds. Therefore, the functionality that's proposed here as "-num_seeds" can actually be achieved by providing -select 0, and setting the desired number of seeds via the "-max_seeds" option. The functionality in "-num_seeds" is therefore redundant; unless I'm still missing something?

No, you're not missing anything here. There is redundancy in that if you allow `-select 0`, then the same result as `-num_seeds N` can be achieved with `-select 0 -max_seeds N`. The only reason I'd included that option in the proposal was that it would be quite a bit more obvious for users... But its effect would indeed be identical.

This does beg the question as to whether we should allow `-select 0` and `-max_seeds 0` to effectively disable these limits. It's not immediately intuitive - although not a uncommon idiom, and is documented. But in practice the only reason users might want to use `-select 0` is to allow `-max_seeds` to be reached, in order to achieve what `-num_seeds` would do. And in the case of `-max_seeds`, it's to prevent the algorithm from running indefinitely without any chance to reaching the target number. In some way, it might be clearer to handle these cases in a slightly more obvious way, which the `-num_seeds` would do. I can't think of use cases where `-select 0` would be used in conjunction with `-max_seeds 0`, although it _might_ make sense in conjunction with streamline piping if we get round to implementing that (not sure what for, but at least it won't use up all available storage). 

For the case of `-max_seeds`, we might be better off providing a different 'abort' mechanism, more in line with the reasoning in #832, and get rid of `-max_seeds` altogether. All we'd need then is a means of allowing the user to specify how hard they want `tckgen` to work to get their tracks out - something like their lowest acceptable success rate, which could default at 0.1% (I think @thijsdhollander already suggested that too), and then we can use the premature exit strategy to give up if we're never going to make it. That way, the user would be in control of the number of _seeds_, or the number of _selections_ with the ability to decide what is an acceptable failure rate. We could also print out the current success rate as tracking progresses, which would give the user immediate feedback into what's going on and why their tracking run might have stopped... 

Anyway, food for thought - I think there's still a few things worth discussing here...

Lestropie:
> The only reason I'd included that option in the proposal was that it would be quite a bit more obvious for users... But its effect would indeed be identical.

> But in practice the only reason users might want to use -select 0 is to allow -max_seeds to be reached, in order to achieve what -num_seeds would do. And in the case of -max_seeds, it's to prevent the algorithm from running indefinitely without any chance to reaching the target number. In some way, it might be clearer to handle these cases in a slightly more obvious way, which the -num_seeds would do.

Even going through this whole discussion, I was constantly second-guessing what the actual purpose was for having separate `-num_seeds` and `-max_seeds` options, and how they were supposed to behave; it's only just clicked now. So I'm not at all convinced that having two separate options makes it "more obvious" for users, even if the option helps are verbose.

> That way, the user would be in control of the number of seeds, or the number of selections with the ability to decide what is an acceptable failure rate. 

With what's implemented currently, the user can set the number of seeds with `-seeds X -select 0`, or the number of selected streamlines with a specified failure rate with `-seeds X -select Y`. I'm not sure that the user needs to be able to control both of the above *at the same time*: it'd be a fair bit of command-line and coding gymnastics to support such an unusual use case.

> All we'd need then is a means of allowing the user to specify how hard they want tckgen to work to get their tracks out - something like their lowest acceptable success rate, which could default at 0.1% (I think @thijsdhollander already suggested that too), and then we can use the premature exit strategy to give up if we're never going to make it. 

If people are adamant that they want separate command-line controls for selecting the number of seeds v.s. setting "something that controls the permissible failure rate", I'd be in favour of providing an option like described here rather than "`-max_seeds`", which to me is too difficult to discriminate from "`-num_seeds`".

jdtournier:
> Even going through this whole discussion, I was constantly second-guessing what the actual purpose was for having separate -num_seeds and -max_seeds options, and how they were supposed to behave; it's only just clicked now. So I'm not at all convinced that having two separate options makes it "more obvious" for users, even if the option helps are verbose.

Yes, I think I was having similar thoughts myself while writing my previous post, which is why I eventually ended up with the idea of ditching `-max_seeds` in favour of a setting the success rate more explicitly... 

> With what's implemented currently, the user can set the number of seeds with -seeds X -select 0, or the number of selected streamlines with a specified failure rate with -seeds X -select Y. I'm not sure that the user needs to be able to control both of the above at the same time: it'd be a fair bit of command-line and coding gymnastics to support such an unusual use case.

You're totally right, I can't think of any sensible case where the user would need to control both the number of seeds and the failure rate. You're right that there is no need for separate `-max_seeds` and `-num_seeds`. 

As far as I can tell, there's really only 3 cases that the inifinite seeders need to be able to handle:

1. the user wants _N_ selected streamlines: `-select N`

2. the user wants _N_ selected streamlines, but the success rate is really poor: `-select N -seeds huge_N`. At the end of my previous post, I was proposing to change that to e.g. `-select N -min_success 1e-6` (or something equivalent).

3. the user wants _N_ seed points: `-seeds N` (implies `-select 0` unless set explicitly?)

If we add a `-min_success` option, then `-select` and `-seeds` should be mutually exclusive. But I agree that it is redundant, the same outcome can be achieved by setting the number of seeds to _N_/acceptable_success_rate (which is exactly how the default is set currently). It's purely about whether this is likely to make more sense to users or not... 

> If people are adamant that they want separate command-line controls for selecting the number of seeds v.s. setting "something that controls the permissible failure rate", I'd be in favour of providing an option like described here rather than "-max_seeds", which to me is too difficult to discriminate from "-num_seeds".

Well, they can already control this with the existing options, as you mentioned. I just think it might be a more obvious solution for users when faced with problem (2) above. I have to say I'm not all that convinced it's going to make a great deal of difference, what do you reckon?

Lestropie:
> the user wants *N* seed points: `-seeds N` (implies -select 0 unless set explicitly?)

That's precisely the behaviour I was advocating for in #809. As is currently implemented: No, if the user sets `-seeds N`, the number of selected streamlines remains as the default (1,000). I would prefer that `-select 0` is indeed implied if the user provides `-seeds` but not `-select`.

> Well, they can already control this with the existing options, as you mentioned. I just think it might be a more obvious solution for users when faced with problem (2) above. I have to say I'm not all that convinced it's going to make a great deal of difference, what do you reckon?

Personally I would prefer just using `-seeds` and `-select` to control the failure rate. I'm imagining trying to write the option help for `-min_success`, explaining how it's setting the ratio between seeds and selected streamlines; but you're explicitly forbidden from setting both the number of seeds and selected streamlines, and instead have to use this option, which is calculated based on what would have been the ratio between those two values (if you were allowed to set them both; which you're not). All a bit backwards for my liking.

---------------

Other thing that needs to be finalized is the [values](https://github.com/MRtrix3/mrtrix3/pull/921/files#diff-b9c71cd67db8628819b6656e0b04b5ea) for the number of attempts to initialise from any given seed point. Currently 100 for infinite seeders, 1,000 for finite seeders.

jdtournier:
> That's precisely the behaviour I was advocating for in #809. 

Yes, I realised... And I'm happy with this if the `-maxnum` gets renamed to `-seeds`. The issue was that the idea behind `-maxnum` was really to set the failure rate, rather than the number of seeds - although it eventually became clear it could be used to set the number of seeds. I think if we clarify the terminology as we've been discussing, this is no longer an issue, and makes a lot more sense. And it should already be relatively obvious that the failure rate can be adjusted using a larger number of seeds, even for new users. 

> Personally I would prefer just using -seeds and -select to control the failure rate. 

Yes, I think you're probably right. It's already pretty self-evident with just `-select` and `-seeds`, let's keep it simple.

> Other thing that needs to be finalized is the values for the number of attempts to initialise from any given seed point. Currently 100 for infinite seeders, 1,000 for finite seeders.

I'm not sure what the 'best' values should be here - but after all these discussions, my gut feeling here is that we should use the same large values, and set the number of attempts per seed to a blanket 1,000... I appreciate there may be cases where this will significantly affect performance, but one thing that I reckon might help feed this back to the user is to have the progressbar update the number of _seeds_, _starts_ and _selections_, rather than just _starts_ and _selections_ as currently - something I'd suggested in an earlier post. If you see the number of _seeds_ going up with no or very few _starts_, let alone _selections_,  that should give you a pretty good idea of why things aren't progressing... The current behaviour when the seed ROI is unsuitable for starting is that the whole thing seems to crawl along (although it does abort if there genuinely haven't been any starts after a large number of attempts).

Lestropie:
> And I'm happy with this if the `-maxnum` gets renamed to `-seeds`.

Done.

> I'm not sure what the 'best' values should be here - but after all these discussions, my gut feeling here is that we should use the same large values, and set the number of attempts per seed to a blanket 1,000

Done.

> I appreciate there may be cases where this will significantly affect performance, but one thing that I reckon might help feed this back to the user is to have the progressbar update the number of seeds, starts and selections, rather than just starts and selections as currently - something I'd suggested in an earlier post.

[Already done](https://github.com/MRtrix3/mrtrix3/pull/921/files#diff-e57344c0ff314328a031f7797ad9894cR44); though I went with "streamlines" rather than "starts".

thijsdhollander:
> I'm happy with this if the -maxnum gets renamed to -seeds. The issue was that the idea behind -maxnum was really to set the failure rate, rather than the number of seeds - although it eventually became clear it could be used to set the number of seeds. I think if we clarify the terminology as we've been discussing, this is no longer an issue, and makes a lot more sense.

Yep, fully agree with this now too; for the exact same reasons.  The new terminology should really help a lot to get this logic across.

> It's already pretty self-evident with just -select and -seeds, let's keep it simple.

Yep again, very happy with keeping it simple.  In line with what Rob mentioned, I think having them state a ratio rather than absolute number is actually harder to explain (and understand, I reckon).  I can see people actually calculating the absolute number from the ratio they would state in such a scenario.  Absolute numbers are the way to go. :+1: 

> set the number of attempts per seed to a blanket 1,000

Fully agree too, if only for clarity.  I've just added this to the documentation too.  It's much easier to say "default: 1000" than to say "default: long conditional text goes here".  For most users, I reckon this will remain an obscure option anyway (that most would also not have to change, so that's good).

I've updated some other docs, and even a functional reference to an old option (property) in `tckgen.cpp`.  I've also slightly changed the name of `max_seed_attempts` to `max_attempts_per_seed`: the former sounded very much like equivalent of the old `-maxnum` that became the current `-seeds`; i.e. how many seeds will be attempted, rather than how many attempt will be made for each seed.  The downside, as always, is that the option name is a bit longer... but I reckon this is not a problem, since the option should probably not be used/changed in most "default" scenarios.  We should probably keep an eye on things ourselves though, to find out if 1000 is indeed a good default across a wide range of "typical" or "realistic" scenarios.

It's probably a good idea to get yet another pair of eyes to finally check that we've not overlooked any other terminology changes, and the whole beast is still/again self-consistent ( @jdtournier ? :grin: ).  I think most of the external / UI logic sits entirely in...

```
cmd/tckgen.cpp
src/dwi/tractography/tracking/tractography.cpp
src/dwi/tractography/seeding/seeding.cpp
```

...right?

thijsdhollander:
I've updated the docs for examples of `tckgen`.  It's surprising how the old `-maxnum` option / new `-seeds` option appears nowhere (except in the reference of `tckgen` itself of course).  Well, actually not that surprising really: it shows that these options are certainly beyond the default scenarios *most of the time*.

thijsdhollander:
Ok, finally (as I think all the rest is good now), while we're changing interface as well as behaviour here anyway: would it be worthwhile to "upgrade" the default number of streamlines to be selected to something bigger than 1000?  My reasoning is that, since the introduction of that default long ago, the average user's hardware is probably a bit more capable nowadays.  A bigger number, e.g. 5000 or maybe even 10000, better shows of the "default" capabilities of a probabilistic tractography algorithm such as iFOD2.  Every once in a while, there still seems to be a user on the forum who is a bit unimpressed with the "sparsity of our tractograms compared to other-diffusion-package-X".  Of course, in any realistic scenario, a user should *use* the `-select` option; but so then again, should the default not be a better initial showcase?

Well, just an idea of course; happy to hear opinions. :-)

jdtournier:
So I've had a quick look, and updated the doc for the `-seeds` option - I think it was describing the old intent, rather than what we'd like it to be now, see what you think. 

Otherwise, it all looks good, and seems to work as expected - at least from my very limited testing. A couple of points though:

- `NO_PROPAGATION_FROM_SEED` is still there - we had loosely discussed maybe labelling it as `TRACK_TOO_SHORT`? No big deal, but if it stays it would be good to figure out what its purpose was... (?)

- We had [discussed raising the default](https://github.com/MRtrix3/mrtrix3/pull/809#issuecomment-261850167) for `TCKGEN_DEFAULT_SEED_TO_SELECT_RATIO` (currently 100). I'd be happy to raise it to 1,000, it might indeed help for those hard to track structures (optic radiations are a prime example). 

- I'm not entirely convinced about raising the default for `-select`, but I'm not against it either. The point of 1000 is that it's a sensible default for the number of _selections_ for your average tract when doing targetted seeding - e.g. to delineate the CST, optic radiations, SLF, etc. It will undoubtedly be too low for anything approaching whole-brain, but then it's difficult to see how to set a default that would be appropriate in all cases. It's also worth bearing in mind that by default, this will lead to a maximum of 100K streamlines being seeded (1M if we increase that default), which in cases where the tractography has a very low success rate, may result in very long runs even for that number. For example, tracking the optic radiation from LGN to V1 using a simple `-seed` & `-include` strategy, using `-seeds 1M -select 1K` (which would be the default if we go with the above) can take a while on my system:
    ```
    tckgen: [100%]   640256 seeds,   637366 streamlines,     1000 selected
    real	0m43.283s
    user	8m34.143s
    sys	0m0.477s
    ```
  And looking at the reconstruction obtained with this, it looks dense enough, certainly for display purposes:
    screenshot from 2017-03-20 14-47-06
  So I'm not altogether convinced it needs changing. Perhaps you can give us examples of use cases where it's not sufficient?

Lestropie:
> updated the doc for the -seeds option

Happy.

> `NO_PROPAGATION_FROM_SEED` is still there

It's still there, in part because it's there and it'd be more work to revert it back so that these get reported as `TRACK_TOO_SHORT`, in part because I don't recall / can't find where that functionality was requested but if it were removed there'd be at least one person out there unhappy with the decision.

> We had discussed raising the default for TCKGEN_DEFAULT_SEED_TO_SELECT_RATIO (currently 100). I'd be happy to raise it to 1,000, it might indeed help for those hard to track structures (optic radiations are a prime example).

Not against it either. But for both here and for lengthy instructions for `-select` / `-number`: It might be worth adding some test to the `DESCRIPTION` field regarding how these behave; that'll be ore likely to be spotted by new users than specific options (even though each is now at the head of their relevant option group).

> I'm not entirely convinced about raising the default for -select, but I'm not against it either. 

Same. The intent is just to "do some tracking" if neither is specified. Maybe a warning-level message in this instance would be better (e.g. `[WARNING] Defaulting to 1,000 selected streamlines in absence of -select / -seeds options`)?

jdtournier:
>> NO_PROPAGATION_FROM_SEED is still there
>
> It's still there, in part because it's there and it'd be more work to revert it back so that these get reported as TRACK_TOO_SHORT, in part because I don't recall / can't find where that functionality was requested but if it were removed there'd be at least one person out there unhappy with the decision.

I assumed this was the case. I'm not too bothered either, but if users rely on `TRACK_TOO_SHORT` for whatever reason, they might need to think a bit about what to do with `NO_PROPAGATION_FROM_SEED`: should it be included in their total or not...? No big deal either way, but it might lead to confusion - especially if we ourselves can't figure out why it exists in the first place. 

> It might be worth adding some test to the DESCRIPTION field regarding how these behave;

Good idea. A little paragraph with a brief overview might do the trick. I'll see if I can draft something this evening. 

> Maybe a warning-level message in this instance would be better (e.g. [WARNING] Defaulting to 1,000 selected streamlines in absence of -select / -seeds options)?

I'm not sure that's warranted. Maybe as an info-level message, but I'd reserve warnings for conditions that are dangerous and/or likely to result in incorrect output. Besides, the progressbar essentially says the same thing already - although I agree it won't say _why_. But given that basically the first thing they'll come across when looking at the help page will relate to this issue, any potential confusion should be short-lived.

Lestropie:
> I assumed this was the case. I'm not too bothered either, but if users rely on `TRACK_TOO_SHORT` for whatever reason, they might need to think a bit about what to do with `NO_PROPAGATION_FROM_SEED`: should it be included in their total or not...? No big deal either way, but it might lead to confusion - especially if we ourselves can't figure out why it exists in the first place.

Looks like I earlier missed the hint that left myself in #734:

> ... (since [the minimum length] may indeed be set to 0mm, but is forced to be at least two points)

So clearly someone was confused as to the streamline rejection (reporting) behaviour when setting `-minlength 0` (still can't find the post though...). I [reject streamlines with 1 point only](https://github.com/MRtrix3/mrtrix3/blob/tckgen_interface_changes/src/dwi/tractography/tracking/exec.h#L387-L391) (as long as it's not the `seedtest` algorithm), whereas regardless of user input I set the minimum number of points to be [at least 2](https://github.com/MRtrix3/mrtrix3/blob/tckgen_interface_changes/src/dwi/tractography/tracking/shared.h#L244) (in order to constitute a "track"). So I think I'd rather keep as-is; could just document somewhere if necessary.

thijsdhollander:
Sorry for not chipping in earlier, because this pull request should be about finished now.  A few quick things:

1. Happy with the increase of seed-to-select ratio to 1000!  I think I brought that up in the previous thread / pull request somewhere, but the rationale was basically to decrease the chances of (with the default settings) hit the maximum number of seeds before the tracks to select target had been reached.  So more users would find themselves in more scenarios actually obtaining the number of tracks requested via `-select`.  The default `-seeds` number can of course still be hit, but that would indicate an "exceptionally" hard to track bundle, where it's more legitimately up to the user to take action (and increase `-seeds` manually).

2. About...

> So clearly someone was confused as to the streamline rejection (reporting) behaviour when setting -minlength 0 (still can't find the post though...). I reject streamlines with 1 point only (as long as it's not the seedtest algorithm), whereas regardless of user input I set the minimum number of points to be at least 2 (in order to constitute a "track"). So I think I'd rather keep as-is; could just document somewhere if necessary.

As far as I understand this correctly, it also seems to make sense me to keep this behaviour.  Just wondering in what scenario(s) streamlines with 1 point only occur.  This would be a successful seed, so it found at the seed location an orientation of the FOD above the cutoff (otherwise it would not go on to become a "streamline", wrt our current progress output, right?).  Having established that, it'll take 1 tracking step and find itself e.g. outside of the mask, or in a spatial location where the FOD within the angular threshold has no amplitudes above the cutoff.  Am I correct that this second location (after the first tracking step) is hence not a part of the streamline, so it ends up having only 1 point and be rejected (regardless of `-minlength 0`)?  It makes sense to reject it on grounds of "not being a streamline", but does it get added to the number of "streamlines" in the progress output?  It sounds like it should not get added to the "streamlines" count, if it didn't successfully get it second point yet, and has "a priori" (i.e. even when `-minlength 0`) no chances to end up in the "selected" category.

3. About the default for `-select`...

> I'm not entirely convinced about raising the default for -select, but I'm not against it either.

I'm certainly on the same general front here, so I just wondered if there could be a lot "against it", and if not, whether it would be ok to raise it anyway.  I agree there is no perfect default that'll strike a balance between the most compact of bundles on the one hand, and a whole-brain tractogram on the other hand.  But even for a compact bundle, more tracks (within a reasonable number of course) doesn't hurt, while more tracks may help for other more extended bundles -- especially those that include fanning, and even more so if said fanning may be more challenging in some orientations than others.  So my logic is along the lines of: what number can we (potentially) reasonably raise it to, so we at least extend the applicability to more bundles, without making the number too high for the simple intention of "just doing some tracking" (which I agree this default is about indeed).

The example of the optic radiation is a pretty compact bundle (at least in the brain of "Tournier Donald" :stuck_out_tongue_winking_eye: ), I'd say.  1000 indeed works ok for that one.  But my inspiration to consider a small increase to, let's say, 5000, came from examples like this:

Simple whole brain tractography with minimal input to `tckgen`, i.e. only the WM FODs and providing a whole brain mask to `-seed_image`:

`-select 1000`:
screenshot0000

`-select 5000`:
screenshot0001

Simplest attempt (I can imagine) at the CST, seeding from the spine, excluding the midsagittal plane from about above the pons, including an axial plane around the height of the upper half of the thalamus:

`-select 1000`:
screenshot0002

`-select 5000`:
screenshot0003

My finding being: with 5000 (versus 1000), there's definitely a *much* better impression as to what the (in this case) iFOD2 algorithm will ultimately explore.  Even for the whole brain tractogram, it gives a pretty decent idea, whereas the 1000 case is still severely lacking to provide that impression.  In a case like the CST and tracking just in one direction (from the spine up), any algorithm is often faced with the much easier "trackability" of the straight path up versus the lateral projections.  Also, with 5000 streamlines, I could much more easily appreciate where my naive approach still lacks the most (in this case tracks diverging in e.g. the SLF, and then heading towards unrelated lower bits of the brain again), and eventually go back and fine-tune the criteria more.

So long story short, I often find myself using something like 5000 for the purpose of "just doing some (initial) tracking", across a very wide range of scenarios, if not almost all realistic ones.  The ultimate question is then if we can "safely" increase the default to that of course.  Beyond 5000 definitely doesn't seem necessary for this particular purpose.

Lestropie:
> Just wondering in what scenario(s) streamlines with 1 point only occur.

Method's `init()` function succeeds, but `next()` fails at the first iteration (in both directions if bidirectional tracking). Probably unlikely to occur for any algorithm other than iFOD2: The `init()` function only looks for a direction at the seed point that is above the FOD amplitude threshold, whereas the first call to `next()` needs to find an *arc segment* that satisfies the FOD amplitude threshold.

I suppose one could argue whether or not iFOD2 should actually be generating that first segment as part of the `init()` call... but even if that were changed, it's still conceivable that some future tracking algorithm could exhibit similar behaviour, potentially even for a different reason.

> Having established that, it'll take 1 tracking step and find itself e.g. outside of the mask, or in a spatial location where the FOD within the angular threshold has no amplitudes above the cutoff. Am I correct that this second location (after the first tracking step) is hence not a part of the streamline, so it ends up having only 1 point and be rejected (regardless of -minlength 0)?

If you were to ignore the above, and use unidirectional tracking, and the first step (second streamline point) terminated for one of the reasons flagged as 0 in [this array](https://github.com/MRtrix3/mrtrix3/blob/tckgen_interface_changes/src/dwi/tractography/tracking/types.h#L40) but not be rejected, then yes, that second point would be omitted from the streamline and hence the "streamline" would consist of only one point. 

> It makes sense to reject it on grounds of "not being a streamline", but does it get added to the number of "streamlines" in the progress output?

Currently yes. It just goes down as a "[rejected streamline]"(https://github.com/MRtrix3/mrtrix3/blob/tckgen_interface_changes/src/dwi/tractography/tracking/exec.h#L126-L128) and contributes to the "streamline count" accordingly in the progress message. You *could* separate it, essentially logging a single-point streamline as a "rejected seed" from within the [`track_rejected()`](https://github.com/MRtrix3/mrtrix3/blob/tckgen_interface_changes/src/dwi/tractography/tracking/exec.h#L381) function; but if we were to do that for the command-line logging, we might as well make a similar change to the internal track rejection logging, removing "`NO_PROPAGATION_FROM_SEED`" and recording a zero-length streamline as "`INVALID_SEED`" (even though *technically* the seed was perfectly valid according to the method's `init()` function). I don't think I'd modify one and not the other.

But as I've said previously, these are really quite rare, so I'm not convinced it's worth the gymnastics.

thijsdhollander:
> But as I've said previously, these are really quite rare, so I'm not convinced it's worth the gymnastics.

Hmm, yes, what I overlooked is that the default is of course bidirectional tracking, in which case these almost certainly become very rare indeed.  I fully agree that it ends up not being worth such big changes in the  back-end at this stage just for these rare cases.  Maybe just something to take note of for the future; in case anyone wants to refactor the tracking code for other (more compelling) reasons, it could be taken on board at that stage.  In the end, the command line output not meant to be much more of a way to keep track of general progress, and give a sense of insight into why progress (accepted streamlines) may be slow in some scenarios.

So this pull request should basically be finished now (right?).  There's still my outstanding screenshot'ed case above to increase the default selection goal from 1000 to 5000 (or anything in between really, e.g. 2000, whatever people may find themselves comfortable with :wink: )... but if there's still too much controversy about this (and that's ok), this is far from a priority with respect to merging the pull request.  It doesn't have major repercussions on the tckgen interface anyway, which is what this pull request was essentially about.

Lestropie:
I'm happy to boost the default `-select` to 5,000 and merge.

thijsdhollander:
:+1:  Sounds good to me!  @jdtournier : any objections?

thijsdhollander:
(just did the commit; will wait until confirmation before merging)

jdtournier:
Happy enough with a default `-select` of 5000 - go for it!

thijsdhollander:
👍

PR #922, opened 2017-02-28
More fixes for scripts run.command() in 0.3.16

Lestropie:
Following on from discussion in #920.

I thought that perhaps there was some sort of network file system race condition that was causing the annoying `standard_space_roi` missing file issue in `5ttgen fsl` (hence function `file.waitFor()`). It now appears as though it's `sh` not finding the script, yet giving a 0 return code. Differences in `subprocess` between Python versions may have contributed to the prior confusion in diagnosing - on top of the `sh` error message not being found until the `-debug` option was used.

Independently of the `standard_space_roi` issue, this function can also be used to wait for the `run_first_all` output rather than globally disabling SGE.

Need to confirm that this still works on Windows.


Lestropie:
Merged because `tag_0.3.16` had borked script functionality. Works on Windows, and fixes `5ttgen fsl` on a config that otherwise wouldn't run it at all. But if any issues arise post them straight away.

PR #923, opened 2017-02-28
fixel2voxel: New option -number

Lestropie:
Alternative solution to #900.

Forgot I did this at one point; figured I'd list it to remind myself to have another look, check my laptop and make sure there weren't any other commits. Also whether `LoopFixelsInVoxelWithMax` should remain here or move into the headers.

### DIFF
--- a/cmd/fixel2voxel.cpp
+++ b/cmd/fixel2voxel.cpp
@@ -12,12 +12,14 @@
  */
 
 
+#include "apply.h"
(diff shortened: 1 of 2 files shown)


thijsdhollander:
:+1: I've merged manually.  Was slightly awkward since the `scripts` folder is already gone on `tag_0.3.16`, but I think I've done it correctly.

Issue #910, opened 2017-02-10
mrview argument parsing performace


maxpietsch:
`mrview` in the current tag_0.3.16 branch, `mrview` opens the first and then loops through all images until it reaches the last one if multiple images are supplied via command line:

For instance images im0.mif im1.mif im2.mif, using `mrview im?.mif`, mrview flashes up im0.mif and im1.mif to then show im2.mif.

I guess that this is not intended?

jdtournier:
Actually, that is (unfortunately) the expected behaviour now, as of [this merge](https://github.com/MRtrix3/mrtrix3/pull/884). As mentioned there, to be able to treat the `-load` option in the same way as providing the image as a regular argument, so that a command like this:

    mrview a.mif -interp 0 b.mif -interp 0

behaves as expected, with the same result as:

    mrview -load a.mif -interp 0 -load b.mif -interp 0

This does mean that listing the images on the command line now has the same effect as opening these images one after the other via the File->Open dialog. 

I had mentioned at the time that this would change behaviour, in that the last image on the command-line would now be the one displayed, rather than the first as was the case before. But the only way around this is to treat images provided as regular arguments differently from those opened via `-load`, by loading them before any command-line options are actually processed, but this leads to the unexpected behaviour that prompted this change in the first place...

Not saying I know what the right thing to do is here, but the current behaviour is probably the most sane, even if it is a bit awkward when just listing a few images on the command-line. Open to suggestions as to potential ways around it...

maxpietsch:
Not a big deal if the images reside locally but it can be a bit annoying when browsing a large network repository with lots of large images. 

I think the easiest thing to do might be to parse all options sequentially but don't update the GUI if there are more images to load and no screenshot option requested. For that, we just would need to have a check if an image in the argument list is intended to be rendered and adjust the existing image load actions to have an option to not load and render the images but only add them to the image queue.

But, when using `mrview` interactively, I think it makes sense to expect to see the first image in the argument list, not the last. Ideally, we could associate options with images and store them in the image class and apply them when they are looked at. That way, we could parse all options, apply them but only show the first image in the queue with its options - if there is no screenshot option set. However, this gets complicated or ambiguous when tool options or global options are also supplied.

Not sure what is best or whether it is worth the hassle.

jdtournier:
[...]

PR #924, opened 2017-03-01
Mrview load images before cmdline options

jdtournier:
As discussed in #910. Probably requires testing on various platforms, and I wouldn't mind getting @ansk's opinion on this, since this largely reverts changes introduced in #884 following discussions on #875.

### DIFF
--- a/cmd/mrview.cpp
+++ b/cmd/mrview.cpp
@@ -42,14 +42,14 @@ void usage ()
     "command-line) are opened before the GUI is shown, subsequent actions to be "
     "performed via the various command-line options must appear after the last "
     "argument. This is to avoid confusion about which option will apply to which "
-    "image. If you need fine control over this, please use the -load or -selectimage "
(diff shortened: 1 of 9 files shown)


ansk:
@jdtournier , I did not have time to test this practically (yet), but from my opinion the current update is perfectly OK. I do not mind to have different behavior between specifying images as command line arguments and specified via -load option. I admit that some confusion can arise, but I think that currently updated documentation can make things clear at least for majority of users.

jdtournier:
> rather than quitting in a clean fashion, a segfault follows

Yes, I'd noticed that too. I'm not sure what that is. I'll try to compile with debug + assert, see if that catches anything. I have a feeling it could be due to Qt not being a fan of exceptions - but this occurs outside the event loop, so I'm not sure that this is really the issue here.

jdtournier:
OK, that last commit seems to fix the remaining segfault.

PR #926, opened 2017-03-02
sh2amp: add missing direct io.

dchristiaens:


jdtournier:
After discussions with @dchristiaens, we decided this won't be merged. These issues shouldn't happen on #785 anyway, and forcing direct IO in this command is likely to slow processing down substantially.

Issue #916, opened 2017-02-23
Docs: include section about DW gradient handling


jdtournier:
As prompted by [this discussion](http://community.mrtrix.org/t/dwi2response-failed/444/24?u=jdtournier).

Lestropie:
[Bump](http://community.mrtrix.org/t/after-the-command-mrconvert-why-changed-the-precision-of-gradient-table/767).

jdtournier:
Yes, I suppose I should just bite the bullet... :weary:

jdtournier:
closed by #932

PR #932, opened 2017-03-07
docs: add page explaining DW gradient handling

jdtournier:
Added to the 'Concepts' section. Happy to reorganise if anyone feels it belongs elsewhere. 
Addresses #916.


jdtournier:
The formatted page is available [here](http://mrtrix.readthedocs.io/en/doctest/concepts/dw_scheme.html), for those who'd like to give it a quick once-over... 😙

jdtournier:
> an image would be beneficial, showing difference between scanner and image space directions

It would. But I'm not sure I'll have the time to do this justice at the moment. Let's merge for now, we can revisit later. At least this'll provide enough information for most users... 

Like your changes. I've fixed up a few other things, I'll merge once the tests pass.

Issue #904, opened 2017-02-08
usage(): Split SYNOPSIS and DESCRIPTION


Lestropie:
Pretty much every command starts the `DESCRIPTION` field with a single sentence describing the purpose of the command; some then go on to give further details. If this were instead split into two fields, e.g. `SYNOPSIS` for the first line and `DESCRIPTION` for anything subsequent to that, then a functionality like `__print_synopsis__` could be added, and used to populate the [command list documentation page](http://mrtrix.readthedocs.io/en/latest/reference/commands_list.html) with a small amount of information on each command.

draffelt:
Wound be very useful

jdtournier:
Had a go at this with while I was hacking away at  #932 - see #933. I didn't add an explicit `SYNOPSIS` field, but decided to simply take the first entry in the description to populate the list. We just need to communicate to app developers that this first line needs to be kept short and punchy. You may want to tweak the scripts' handling of the description field in some way to match, but currently none of the descriptions are that long anyway...

thijsdhollander:
This should probably have been closed upon [this commit](https://github.com/MRtrix3/mrtrix3/commit/a12939fff61a84cd63d1e0a6558778eb391aee0f) (see above).  But there's a little :information_source: icon next to the commit's listing in this issue that says "This issue will close once commit a12939f is merged into the 'master' branch.".

Long story short: we need to close issues that get addressed by a commit that doesn't immediately end up on `master` (e.g. all work on `tag_0.3.16`) manually...

thijsdhollander:
(the :information_source: icon is gone now that I have closed the issue manually)

Issue #881, opened 2017-01-15
New MR::App extern: DOCUMENTATION


Lestropie:
Since the documentation is fairly mature / stable now, and dedicated documentation pages are likely to become more common with new commands as analyses become increasingly complex, I had the idea to add a `DOCUMENTATION` string field that can be set in `void usage()`, providing links to online documentation pages where appropriate; e.g. as suggested [here](http://community.mrtrix.org/t/does-mrtrix-have-water-compartment-modeling-similar-to-noddi/655/8).

draffelt:
Sounds sensible to me. Would be great to have an `EXAMPLE USAGE` section too. I thought we added this as an issue a while back, but can't find it now.

jdtournier:
Both good ideas. In general though, I think the bigger problem is that the existing command docs are a little bit too brief anyway. You'll note that the [discussion that prompted this issue](http://community.mrtrix.org/t/does-mrtrix-have-water-compartment-modeling-similar-to-noddi/655/8) was really about clarifying which argument is an input and which is an output. This is all [stated in the doc](http://mrtrix.readthedocs.io/en/latest/reference/commands/dwi2fod.html) if you look in the right place, but blink and you'll miss it: it's only mentioned in the synopsis, in the description of the arguments:

> _response odf:_ pairs of input tissue response and output ODF images

Example usage would help clarify this (for instance, [`mrcalc` already does](http://mrtrix.readthedocs.io/en/latest/reference/commands/mrcalc.html#description)), as would a link to more detailed documentation where available, but ultimately this is a simple question about _usage_ and shouldn't require referring to external documents. It would be good to go through the existing command help pages and make sure that their usage is clear from that alone. As always though: who has the time to do this...?

Lestropie:
This should only be an issue for the minority of commands, where there isn't a 'fixed' set of arguments. So yes, `dwi2fod msmt_csd` should probably mimic `mrcalc` in that respect, but there shouldn't be many like that.

Another thing I might add here is that all commands currently have the first sentence give a quick synopsis of what the command does; some then have a greater block of text giving more detail on its usage, purpose or the like (or even storage convention of SH coefficients in some instances). This could be better reflected using e.g. `SYNOPSIS` and `DESCRIPTION` fields, and the 'synopsis' text could then appear right at the start of the online command doc page before the command-line usage, rather than simply being 'the first sentence'.

PR #933, opened 2017-03-09
docs: add first line of description to command list

jdtournier:
Addresses #904.


jdtournier:
As before, I used the `doctest` branch - results can be previewed [here](http://mrtrix.readthedocs.io/en/doctest/reference/commands_list.html).

thijsdhollander:
stop-the-presses

Didn't @Lestropie already tackle this in the `tag_0.3.16` branch with a proper `SYNOPSIS` field, as well as new `USAGE` and `DESCRIPTION` fields?  ...results of which can be viewed [here](http://mrtrix.readthedocs.io/en/tag_0.3.16/reference/commands_list.html)... as well as on individual commands' pages, e.g. [see here](http://mrtrix.readthedocs.io/en/tag_0.3.16/reference/commands/dwi2fod.html). (and the same for the list of scripts and individual scripts' pages)

jdtournier:
Oh dear... It does seem I have been wasting my time. It's high time we merged this `tag_0.3.15` Frankenstein already.

I'll can this pull request then...

Lestropie:
Yes, I mentioned the idea in #881 but never generated an issue for that alone. I thought I drew attention to it at some point when it was done; don't recall.

Given this and #934, there might need to be a little more discipline around issue listing / assignment / updates. I'm particularly guilty in that there were a lot of things I knew I would need to do for the Stanford coding sprint, many of which I listed explicitly but some things just kind of got done when they were needed; #934 was probably one of the latter. The period of time that `0.3.16` has been listed (opened September last year!) is contributing, as is the fact that issues closed via commit messages don't actually get closed until the relevant commit is merged to master, so there's a lot sitting there listed as "open" that are actually already done.

Open to ideas or suggestions of things we can try to do to avoid this sort of thing (without getting buried in micro-management). It really does suck when work gets unnecessarily duplicated.

jdtournier:
> Yes, I mentioned the idea in #881 but never generated an issue for that alone. 

Nope, you did: #904 ... which is why I thought I'd give it a crack. No harm done, it didn't take long. 

> It really does suck when work gets unnecessarily duplicated.

Sure, but it's not such a big deal for a small changeset like this. It's going to happen from time to time. It would be more important for larger changes, and then it would really pay to push an issue or pull request up on GitHub so others can raise the alarm (thanks @thijsdhollander :wink:).

PR #934, opened 2017-03-10
mrconvert: add options to modify generic header entries

jdtournier:
This has been a while coming... Should help with issues such as [anonymisation](http://community.mrtrix.org/t/anonymise-nifti/773?u=jdtournier).

### DIFF
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -77,6 +77,24 @@ void usage ()
             "effect for floating-point and binary images.")
   + Argument ("values").type_sequence_float()
 
+
+  + OptionGroup ("Modify generic header entries")
+
+  + Option ("clear_property", 
+            "remove the specified key from the image header altogether.").allow_multiple()
+  + Argument ("key").type_text()
+
+  + Option ("set_property", 
+            "set the value of the specified key in the image header.").allow_multiple()
+  + Argument ("key").type_text()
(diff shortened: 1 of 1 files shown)


Lestropie:
You're on a roll! 2dacd0b2

jdtournier:
Dude...

I assume this is on `tag_0.3.16` only currently then... OK, let's cherry-pick that onto master then, so we can [help Samuel out](http://community.mrtrix.org/t/anonymise-nifti/773?u=jdtournier).

jdtournier:
Actually, on second thoughts: I'm not sure I like the option names very much... Do you have any objections to what I've used here? Two reasons: 

- the name `property` matches the equivalent option in `mrinfo` to query these values.

- prefixing the action before the property allows option shortening to be used meaningfully. With an option like `-set_property`, I can write:
  ```ShellSession
  mrconvert in.nii -set comments anonymous out.nii
  ```
  and that command still makes sense. But that functionality is essentially no use if you have to write `-header_set`...

Lestropie:
No objections, that logic is fine by me. Just need to try to remember / find whether the functionality was actually used anywhere in order to update them in `tag_0.3.16`.

jdtournier:
OK, thanks. I'll merge this, and push the corresponding changes onto `tag_0.3.16` in a minute - after checking nothing is actually making use of that already ...

PR #938, opened 2017-03-28
Eigen OpenMP handling

dchristiaens:
As explained on [this page](https://eigen.tuxfamily.org/dox/TopicMultiThreading.html), Eigen handles multi-threading through OpenMP. This functionality can be disabled with macro `EIGEN_DONT_PARALLELIZE`, which we had activated in `core/types.h` because we typically have many, relatively small (voxelwise) operations that are more efficiently parallelized using image-level threading primitives.

However, there are two issues with this approach. First of all, `EIGEN_DONT_PARALLELIZE` is a one-off flag read only upon the first inclusion of `Eigen/Core`. The current implementation is therefore unpredictable because it depends on the header include order. For instance,
```
#include types.h
#include <Eigen/Core>
```
will deactivate Eigen parallelism, whereas
```
#include <Eigen/Core>
#include types.h
```
will not. It is therefore safer to define `EIGEN_DONT_PARALLELIZE` as a compiler flag, as done in 6df61fa. The reason why this hasn't been an issue so far, is because the OpenMP flags are not set in the default build configuration, hence making the `EIGEN_DONT_PARALLELIZE` macro mute anyway.

Secondly, in applications involving large matrices it may be useful to rely on the Eigen OpenMP parallelization. I have added the necessary compiler flag in 99c5aef, which can be activated with
```
./configure -openmp
```

However, this would then activate OpenMP for *all* commands, including the many operations that are already parallelized at the voxel or track level. Therefore, I have undefined the Eigen macro `EIGEN_HAS_OPENMP` by default. The developer can then de-/activate Eigen parallelism as desired using
```
#define EIGEN_HAS_OPENMP

// parallel Eigen code

#undef EIGEN_HAS_OPENMP
```

Before we merge, this needs to be properly tested on different systems. All works well on my Ubuntu 16.04; I'll test on my Mac later tonight.

### DIFF
--- a/configure
+++ b/configure
@@ -42,9 +42,9 @@ OPTIONS
 
     -verbose     enable more informative output.
 
-    -dev         enable the extended development build process
+    -dev         enable the extended development build process.
 
-    -R           used to generate an R module (implies -noshared)
+    -R           used to generate an R module (implies -noshared).
 
 
 ENVIRONMENT VARIABLES
@@ -776,6 +776,8 @@ except ( LinkError, RuntimeError ) as e:
   This shouldn't happen!''' + configure_log_hint)
 except:
   error ('unexpected exception!' + configure_log_hint)
+
+eigen_cflags += [ '-D'+'EIGEN_DONT_PARALLELIZE', ]
 
 
 
--- a/configure
+++ b/configure
@@ -816,6 +816,32 @@ except:
 
 
 
+# add openmp flags if required and available
+
+if openmp:
+  cpp_flags += [ '-fopenmp' ]
+  ld_flags  += [ '-fopenmp' ]
+  report ('Checking for OpenMP support: ')
+  try:
+    compile ('''
+    #include <Eigen/Core>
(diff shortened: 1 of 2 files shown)


dchristiaens:
The `./configure -openmp` option does not compile on macOS. Apparently XCode clang++ doesn't accept the `-fopenmp` flag. Can we ignore that for now, or should we better remove the option from the configure script and leave it to the user to set `CFLAGS` and `LD_FLAGS` to their needs?

jdtournier:
We could just catch that combination in the configure script, and simply tell the user they can't use that flag? Or maybe add an additional check for OpenMP...? Or just document that this option won't work on macOS...

dchristiaens:
Thanks for giving this a test drive Rob. I added a test program last Friday, but apparently forgot to push... It tries initialising multi-threading in Eigen, and then tests whether the number of threads matches the specified number. This has practically no overhead and runs fine on my system.

PR #939, opened 2017-03-29
mrview view tool: buttons to copy voxel and scanner location to clipboard

maxpietsch:
Added two buttons to the view tool that allow to copy x,y,z voxel or scanner coordinates to the clipboard and write them to stdout. Could be used as a poor man's annotation tool.

2017-03-29t10_14_35 080200 0100_hooowx

### DIFF
--- a/src/gui/mrview/tool/view.cpp
+++ b/src/gui/mrview/tool/view.cpp
@@ -82,22 +82,22 @@ namespace MR
               return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable;
             }
 
-            QModelIndex index (int row, int column, const QModelIndex& parent = QModelIndex()) const 
-            { 
+            QModelIndex index (int row, int column, const QModelIndex& parent = QModelIndex()) const
+            {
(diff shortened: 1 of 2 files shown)


jdtournier:
Good idea. You should be able to copy the text directly into the clipboard though, which might be more intuitive - and usable for those that start mrview without a terminal...

maxpietsch:
I [do](https://github.com/MRtrix3/mrtrix3/pull/939/commits/191783b822588256854f8ab42eb2570db16e902e#diff-946b56617a4ae8cfde2faecc0c250589R572). I only added the terminal output because it is useful if multiple landmarks need to be selected.

jdtournier:
So you do! Yet another example of me not reading people's posts carefully... :blush:

maxpietsch:
Shall I squash and merge?

jdtournier:
Sure - but why not a regular merge commit...?

maxpietsch:
Just wrote what the big green button said without thinking about it...

jdtournier:
Funny that -  on my system, the default is always a merge commit...

Issue #940, opened 2017-03-29
mrview: windowing: assert failure for switching between images


maxpietsch:
Assertion `std::isfinite (value_min)` fails in `src/gui/mrview/displayable.h` when I switch between images for the first time id the images don't overlap:

`mrview: src/gui/mrview/displayable.h:309: void MR::GUI::MRView::Displayable::update_levels(): Assertion `std::isfinite (value_min)' failed.`

This doesn't happen when snap to voxel grid is off. Happens in axial or ortho view.

Might be related: I get different windowing at different locations if snap to voxel grid is on. Windowing is the same for all locations if snap to voxel grid is off (even if the location is on voxel grid).

PR #943, opened 2017-03-30
mrview: outside FOV: set value_min and value_max to zero

maxpietsch:
Proposed fix for #940: `mrview: src/gui/mrview/displayable.h:309: void MR::GUI::MRView::Displayable::update_levels(): Assertionstd::isfinite (value_min)' failed.` 

I added `value_min = 0.0f;` and `value_max = 0.0f;` for all colourmaps in case the cursor is outside the FOV. This gives nan as lower and upper limit for the colourbar which I think is sensible.

### DIFF
--- a/src/gui/mrview/gui_image.cpp
+++ b/src/gui/mrview/gui_image.cpp
@@ -184,7 +184,10 @@ namespace MR
           format = gl::RGB;
           internal_format = gl::RGB32F;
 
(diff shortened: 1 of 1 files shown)


maxpietsch:
Any objections against merging?

jdtournier:
Haven't tested, but I can't see any issues with this. But since this isn't merging to master, go for it - it'll get more extensive testing in `tag_0.3.16`, and it's not going to break anything for users...

PR #947, opened 2017-04-03
Dwigradcheck

Lestropie:
New script that checks for erroneous permutations / flips in the gradient table.

Ported to 0.3.16 script library; might as well go out on the off chance it proves useful.

### DIFF
--- a/bin/dwigradcheck
+++ b/bin/dwigradcheck
@@ -71,7 +71,6 @@ if len(grad_mrtrix[-1]) <= 1:
   grad_mrtrix.pop()
 if len(grad_fsl[-1]) <= 1:
   grad_fsl.pop()
-print(grad_mrtrix)
 # Turn into float matrices
 grad_mrtrix = [ [ float(f) for f in line.split(' ') ] for line in grad_mrtrix ]
 grad_fsl = [ [ float(f) for f in line.split() ] for line in grad_fsl ]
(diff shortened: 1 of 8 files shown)


bjeurissen:
While making these final changes, I noticed two things:
- the order of the options is reversed in the help output, probably to ensure that the default options (which are added first) will come last in the help output. Would make sense to work around this in ```app.py``` and then let individual scripts list the options in the order of appearance in the help output.
- options that are not in an argument_group are ignored: is this intended?

bjeurissen:
Should be good to merge once the travis CI build completes. Have tested it with different gradient import and export options and the mask and number option.

We could maybe reduce the default number of tracks from 10000 to 1000 for the impatient.

Lestropie:
> the order of the options is reversed in the help output

Yes, it's a bit weird. I kind of left it as-is so that the "standard options" (which are added first) appear at the end of the list, and ordered everything accordingly. I suppose I could try only adding the standard options when the parsing begins, see if that works... It's not ideal, but I've already put so much effort into hacking the internals of `argparse` to make them work that I often wonder if I'd be better off starting from scratch...

> options that are not in an argument_group are ignored: is this intended?

No, but I can see how that could arise. Again, hacking the `argparse` internals as best I can with no documentation.

> We could maybe reduce the default number of tracks from 10000 to 1000 for the impatient.

I'd rather set it at a value that's reasonably reproducible; the impatient can use the command-line option if they want to.

PR #954, opened 2017-04-04
Feature/cpp modernize

WillBrennan:
Made changes to modernise aspects of the code as well as removed a comparison thats always true.

- replaced C Standard Library includes with C++ Standard Library includes.
- removed comparison that is always true (unsigned ints are always greater than or equal to zero)
- replaced typdef aliases which are a C++03 method with using aliases which are C++11 and above.

The last commit is a significant change, feel free to remove it.

### DIFF
--- a/cmd/afdconnectivity.cpp
+++ b/cmd/afdconnectivity.cpp
@@ -83,9 +83,9 @@ void usage ()
 }
 
 
-typedef float value_type;
-typedef DWI::Tractography::Mapping::SetDixel SetDixel;
-typedef DWI::Tractography::SIFT::FixelBase FixelBase;
+using value_type = float;
+using DWI::Tractography::Mapping::SetDixel;
+using DWI::Tractography::SIFT::FixelBase;
 
 
 
--- a/cmd/amp2sh.cpp
+++ b/cmd/amp2sh.cpp
@@ -87,7 +87,7 @@ void usage ()
 #define RICIAN_POWER 2.25
 
 
-typedef float value_type;
+using value_type = float;
 
 class Amp2SHCommon {
   public:
--- a/cmd/dirflip.cpp
+++ b/cmd/dirflip.cpp
@@ -52,10 +52,8 @@ void usage ()
 }
 
 
-typedef double value_type;
-typedef Eigen::Vector3d vector3_type;
-
-
+using value_type = double;
+using vector3_type = Eigen::Vector3d;
 
 
 class Shared {
--- a/cmd/dirgen.cpp
+++ b/cmd/dirgen.cpp
@@ -94,7 +94,7 @@ class Energy {
       return p == 1 ? x : fast_pow (x*x, p/2);
     }
 
-    typedef double value_type;
+    using value_type = double;
 
     size_t size () const { return 3 * ndir; }
 
--- a/cmd/dirmerge.cpp
+++ b/cmd/dirmerge.cpp
@@ -44,9 +44,9 @@ ARGUMENTS
 }
 
 
-typedef double value_type;
-typedef std::array<value_type,3> Direction;
-typedef std::vector<Direction> DirectionSet;
+using value_type = double;
+using Direction = std::array<value_type,3>;
+using DirectionSet = std::vector<Direction>;
 
 struct OutDir {
   Direction d;
--- a/cmd/dirorder.cpp
+++ b/cmd/dirorder.cpp
@@ -44,7 +44,7 @@ void usage ()
 }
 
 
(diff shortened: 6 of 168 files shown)


dchristiaens:
All seems sound to me. Thanks for contributing!

The new type aliasing syntax it is quite a significant change of our coding style. Personally, I'm not too fussed either way, but I can see the appeal of templated type aliases. No doubt @jdtournier will have something to say about this :smile:

WillBrennan:
dw, I understand, its a pretty big change. Just think it improves readability. 🙂

jdtournier:
Just a note about the first commit though: there are differences between the C style header and its C++ counterpart, which have caused odd issues, albeit on configurations​ known to be buggy (see #917). I'm not saying these changes aren't the right thing to do, but let's keep an eye out for related errors on the forum...

Also, this may end up being a messy merge to `tag_0.3.16`... I'm not sure I'll have the time to look into it any time soon, so if anyone else wants to have a go... 😁

dchristiaens:
> Also, this may end up being a messy merge to `tag_0.3.16` ...

It is :smile: I'm working on it though...

dchristiaens:
All done now.

jdtournier:
Thanks! Sounded like a load of fun. :wink:

dchristiaens:
It certainly showed how large our codebase has become.

I was quite excited about these templated typedefs, to do stuff like
```
template <T = MR::default_type> using MatrixX = Eigen::MatrixX<T, Eigen::Dynamic, Eigen::Dynamic>
```
Unfortunately, that means that you still need to write `MatrixX<>` in every instance where you want to fall back on the default type (simply `MatrixX` doesn't suffice), which isn't quite as elegant as I had hoped. Eigen's own solution with `MatrixXf` and `MatrixXd` then makes much more sense. If only they would default the other template arguments to dynamic... Anyway, no use case (or changes in the code) so far.

jdtournier:
That's exactly the kind of use case I was thinking might come in handy. Particularly for vectors, since supporting this kind of thing is a bit ugly currently:
```C++
template <typename T> void some_func() 
{
  Eigen::Matrix<T,Eigen::Dynamic,1> vector;     // yuk! I just wanted a vector of type T...
  ...
}
```
It would be cleaner IMO to be able to write:
```C++
// in lib/types.h:
namespace Eigen {
  template <typename T=MR::default_type> using Vector = Eigen::Matrix<T,Eigen::Dynamic,1>;
}

...

template <typename T> void some_func() 
{
  Eigen::Vector<T> vector;    // much more readable... 
  ...
}
```
It would also allow the definition of templated static types, which would also make their use in templated functions a lot more elegant:
```C++
template <typename T=MR::default_type> using Vector2 = Eigen::Matrix<T, 2, 1>;
template <typename T=MR::default_type> using Vector3 = Eigen::Matrix<T, 3, 1>;
template <typename T=MR::default_type> using Vector4 = Eigen::Matrix<T, 4, 1>;
```
You'll note that the [current kludge](https://github.com/MRtrix3/mrtrix3/blob/master/lib/types.h#L178-L181) doesn't provide the option of using anything but `MR::default_type`...

dchristiaens:
Funny, that's exactly what I tried today. You didn't hack into my computer, did you? 😉 

What I meant to say in my earlier comment was that, yes,
```
Eigen::Vector<T> vector;
```
is a lot cleaner than having to write
```
Eigen::Matrix<T, Eigen::Dynamic, 1> vector;
```
all the time. For vectors of dynamic length, this is perfectly possible as you also showed. Matrices are more tricky though, because the name `Eigen::Matrix` is reserved and it doesn't seem a good idea to me to override this. This is why I was complaining that our lives would involve a lot less typing if the Eigen developers would have defined `Eigen::Dynamic` as the default template argument for their `Matrix` class. That would give us the option to do
```
Eigen::Matrix<T> mat;
```
straight away. To my knowledge, C++ does not allow to override template default arguments once they've been set. Therefore, I went with the name `MatrixX` in my previous example (a `typename` keyword was missing in its definition).

A second issue, however, is that if a default type is set to `MR::default_type`, as in your vector example, it requires
```
Eigen::Vector<> vec;
```
in any instance where we want to fall back on the default datatype. Hence, if I redefine `Vector3` etc. in this way, all its uses throughout the code would need to be replaced with `Vector3<>`, and I had done just about enough refactoring for the day 😄

jdtournier:
> I had done just about enough refactoring for the day 😄

:+1: fully understand... 

And I agree with everything you say - not sure it is worth changing anything at this point either.

Issue #747, opened 2016-08-04
Global handling of phase-encode direction data


Lestropie:
Following on from #591, and something that has come up during the process of #725.

Providing an automated processing pipeline where you don't necessarily know _a priori_ what type of reversed phase-encode acquisition has taken place, or what direction(s) the phase encoding directions are in, or whether a fieldmap has instead been acquired. In such cases the current `dwipreproc` interface is limited.

I've started on code to extract the phase encode direction and readout bandwidth information from DICOMs; and this information is expected to be provided in BIDS compliant data. But given the number of possible combinations of such data, it may be preferable to instead embed this information within image headers, so that `dwipreproc` can figure out what it has to do (this would be in addition to the current 'manual' options for specifying phase-encoding information).

One option would be to embrace the approach in #591, adopting a wider 'standard' of having phase encoding information embedded within the '`dw_scheme`' header entry. Alternatively, particularly since the DWI scheme is now a plain key/value entry in the header rather than a dedicated field, we could add a '`pe_scheme`' entry. I think I'm in favour of the latter; not only for separating the two sources of information, but also the DW scheme is specified with respect to scanner space whereas the PE scheme is with respect to image axes, so combining them is a bit of a hack.

There's a semi-standardised header field '`"PhaseEncodingDirection`' (`i`, `i-`, `j`, `j-`, `k`, `k-`) that may be specified for any particular image file, but obviously this can't vary for different volumes in a 4D image. I could _technically_ get away with just using this information, but it would require a _lot_ of gymnastics in order to figure out how to invoke `dwipreproc`, and wouldn't be guaranteed to work for all cases. Plus, given whatever evil scheme @jdtournier has going that spawned #591 requires a PE direction per volume, I reckon it looks like a good idea to support that longer-term.

Opening up for discussion / comments. If I go for my preferred option ('`PhaseEncodingDirection`' for fixed PE direction, '`pe_scheme`' for variable), @jdtournier you'll need to warp your code around that accordingly.

Lestropie:
Notes:

* Would revert #591 and enforce 4 columns in the `dw_scheme` entry if going for the `pe_scheme` option.

* Commands such as `dwiextract`, `mrconvert`, `mrcat` would need to be aware of the `PhaseEncodingDirection` / `pe_scheme` header entries and behave accordingly.

* `dwiextract` could select and extract volumes based on phase encoding direction.

* Propagating / deleting these fields by certain operations would probably mimic the result of #375.

[...]

PR #956, opened 2017-04-06
MGH / MGZ formats: Always big-endian

### LINKED ISSUE #952: .mgz / .mgh files created by mrtrix tools are not correctly read by FreeSurfer's tools
Dear experts,
I found that the FreeSurfer tools are segfaulting when reading the `.mgz / .mgh` files created by mrtrix tools. I tested this on FreeSurfer 5.3 as well as on dev version of FreeSurfer from 2/2017. I am using most recent build of mrtrix on master branch: 0.3.15-522-gbc67f1a1

Example: `orig.mgz` is file created by freeSurfer.
```
mrconvert orig.mgz orig_mrtrix.mgz

mrinfo orig_mrtrix.mgz 
Image:               "orig_mrtrix.mgz"
  Dimensions:        256 x 256 x 256
  Voxel size:        1 x 1 x 1
  Data strides:      [ -1 3 -2 ]
  Format:            MGZ (compressed MGH)
  Data type:         unsigned 8 bit integer
  Transform:                    1   -1.11e-16   4.657e-10      -131.2
                        5.551e-17           1   7.276e-12      -114.8
                       -4.657e-10  -7.276e-12           1      -101.7

mri_info orig_mrtrix.mgz 
Floating point exception

mri_convert orig_mrtrix.mgz orig_mrtrix2.mgz

mri_convert.bin orig_mrtrix.mgz orig_mrtrix2.mgz 
$Id: mri_convert.c,v 1.226 2016/02/26 16:15:24 mreuter Exp $
reading from orig_mrtrix.mgz...
/hydra-db/hydra_io/vypocty/skoch/bin/freesurfer_dev_20170208/bin/mri_convert: line 3: 11714 Segmentation fault      mri_convert.bin "$@"

mri_binarize --i orig_mrtrix.mgz --min 1 --max 255 --o orig_mrtrix_binarized.mgz --count junk.txt

$Id: mri_binarize.c,v 1.43 2016/06/09 20:46:21 greve Exp $
cwd /hydradb/hydra_io/vypocty/freeSurfer/v6.1beta/CANEF/A_in_progress/skoch/ESO_P00007_20100611_1217_1_base/mri
cmdline mri_binarize.bin --i orig_mrtrix.mgz --min 1 --max 255 --o orig_mrtrix_binarized.mgz --count junk.txt 
sysname  Linux
hostname sup1.ad.nudz.cz
machine  x86_64
user     skoch

input      orig_mrtrix.mgz
frame      0
nErode3d   0
nErode2d   0
output     orig_mrtrix_binarized.mgz
Binarizing based on threshold
min        1
max        255
binval        1
binvalnot     0
/hydra-db/hydra_io/vypocty/skoch/bin/freesurfer_dev_20170208/bin/mri_binarize: line 3: 11922 Segmentation fault      mri_binarize.bin "$@"
```

Lestropie:
Creating .mgh / .mgz images with little-endian encoding causes faults in other softwares; therefore enforce big-endianness regardless of header datatype.

Closes #952.

### DIFF
--- a/lib/formats/mgh.cpp
+++ b/lib/formats/mgh.cpp
@@ -20,6 +20,7 @@
 #include "file/mgh_utils.h"
 #include "image_io/default.h"
 #include "formats/list.h"
+#include "raw.h"
 
 namespace MR
 {
@@ -43,8 +44,21 @@ namespace MR
         mgh_other MGHO;
         memcpy (&MGHO, fmap.address() + other_offset, other_floats_size);
         MGHO.tags.clear();
-        if (other_tags_offset < fmap.size()) {
 
+        uint8_t* p_current = fmap.address() + other_tags_offset;
+
+        while (p_current < fmap.address() + fmap.size()) {
+
+          int32_t tag = Raw::fetch_BE<int32_t> (p_current);
+          int64_t size = Raw::fetch_BE<int64_t> (p_current+4);
+          if (size) 
+            add_line (H.keyval()["comments"], "[TAG "+str(tag) + "]: "   + (const char*) (p_current+12));
+
+          p_current += 12 + size;
+        }
+
+        /*
+        if (other_tags_offset < fmap.size()) {
           // It's memory-mapped, so should be able to use memcpy to do the initial grab
           const size_t total_text_length = fmap.size() - other_tags_offset;
           char* const tags = new char [total_text_length];
@@ -60,8 +74,8 @@ namespace MR
           }
 
           delete[] tags;
-
         }
+*/
 
         File::MGH::read_other (H, MGHO, is_BE);
 
--- a/lib/formats/mgh.cpp
+++ b/lib/formats/mgh.cpp
@@ -1,16 +1,16 @@
 /*
  * Copyright (c) 2008-2016 the MRtrix3 contributors
- * 
+ *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
- * 
+ *
  * MRtrix is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
- * 
+ *
  * For more details, see www.mrtrix.org
- * 
+ *
  */
 
 #include "header.h"
@@ -32,7 +32,7 @@ namespace MR
       if (!Path::has_suffix (H.name(), ".mgh"))
         return std::unique_ptr<ImageIO::Base>();
       File::MMap fmap (H.name());
-      bool is_BE = File::MGH::read_header (H, * ( (const mgh_header*) fmap.address()));
+      File::MGH::read_header (H, * ( (const mgh_header*) fmap.address()));
 
       // Remaining header items appear AFTER the data
       // It's possible that these data may not even be there; need to make sure that we don't go over the file size
@@ -43,19 +43,16 @@ namespace MR
 
         mgh_other MGHO;
         memcpy (&MGHO, fmap.address() + other_offset, other_floats_size);
-        File::MGH::read_other (H, MGHO, is_BE);
+        File::MGH::read_other (H, MGHO);
 
-        MGHO.tags.clear();
         uint8_t* p_current = fmap.address() + other_tags_offset;
 
         while (p_current < fmap.address() + fmap.size()) {
-
-          int32_t tag = Raw::fetch_BE<int32_t> (p_current);
-          int64_t size = Raw::fetch_BE<int64_t> (p_current+4);
-          if (size & p_current[12]) 
-            add_line (H.keyval()["comments"], "[MGH TAG "+str(tag) + "]: "   + (const char*) (p_current+12));
-
-          p_current += 12 + size;
+          auto tag = File::MGH::prepare_tag (*reinterpret_cast<int32_t*>(p_current), *reinterpret_cast<int64_t*> (p_current+sizeof(int32_t)));
+          p_current += sizeof(int32_t)+sizeof(int64_t);
+          tag.content = std::string (reinterpret_cast<const char*> (p_current), tag.size);
+          File::MGH::read_tag (H, tag);
+          p_current += tag.size;
         }
 
 
@@ -113,9 +110,9 @@ namespace MR
       out.open (H.name(), std::ios_base::out | std::ios_base::app);
       out.write ((char*) &MGHO, 5 * sizeof (float));
       for (const auto& tag : MGHO.tags) {
-        out.write (reinterpret_cast<const char*> (&std::get<0>(tag)), sizeof(int32_t));
-        out.write (reinterpret_cast<const char*> (&std::get<1>(tag)), sizeof(int64_t));
-        out.write (std::get<2>(tag).c_str(), std::get<2>(tag).size());
+        out.write (reinterpret_cast<const char*> (&tag.id), sizeof(tag.id));
+        out.write (reinterpret_cast<const char*> (&tag.size), sizeof(tag.size));
+        out.write (tag.content.c_str(), tag.content.size());
       }
       out.close();
 
--- a/lib/formats/mgh.cpp
+++ b/lib/formats/mgh.cpp
@@ -43,41 +43,21 @@ namespace MR
 
         mgh_other MGHO;
         memcpy (&MGHO, fmap.address() + other_offset, other_floats_size);
-        MGHO.tags.clear();
+        File::MGH::read_other (H, MGHO, is_BE);
 
+        MGHO.tags.clear();
         uint8_t* p_current = fmap.address() + other_tags_offset;
 
         while (p_current < fmap.address() + fmap.size()) {
 
           int32_t tag = Raw::fetch_BE<int32_t> (p_current);
           int64_t size = Raw::fetch_BE<int64_t> (p_current+4);
-          if (size) 
-            add_line (H.keyval()["comments"], "[TAG "+str(tag) + "]: "   + (const char*) (p_current+12));
+          if (size & p_current[12]) 
+            add_line (H.keyval()["comments"], "[MGH TAG "+str(tag) + "]: "   + (const char*) (p_current+12));
 
           p_current += 12 + size;
         }
 
-        /*
-        if (other_tags_offset < fmap.size()) {
-          // It's memory-mapped, so should be able to use memcpy to do the initial grab
-          const size_t total_text_length = fmap.size() - other_tags_offset;
-          char* const tags = new char [total_text_length];
-          memcpy (tags, fmap.address() + other_tags_offset, total_text_length);
-
-          // Extract and separate null-terminated strings
-          size_t char_offset = 0;
-          while (char_offset < total_text_length) {
-            std::string line (tags + char_offset);
-            if (line.size())
-              MGHO.tags.push_back (line);
-            char_offset += line.size() + 1;
-          }
-
-          delete[] tags;
-        }
-*/
-
-        File::MGH::read_other (H, MGHO, is_BE);
 
       } // End reading other data beyond the end of the image data
 
--- a/lib/formats/mgh.cpp
+++ b/lib/formats/mgh.cpp
@@ -110,7 +110,14 @@ namespace MR
 
       File::resize (H.name(), MGH_DATA_OFFSET + footprint(H));
 
-      File::MGH::write_other_to_file (H.name(), MGHO);
+      out.open (H.name(), std::ios_base::out | std::ios_base::app);
+      out.write ((char*) &MGHO, 5 * sizeof (float));
+      for (const auto& tag : MGHO.tags) {
+        out.write (reinterpret_cast<const char*> (&std::get<0>(tag)), sizeof(int32_t));
+        out.write (reinterpret_cast<const char*> (&std::get<1>(tag)), sizeof(int64_t));
+        out.write (std::get<2>(tag).c_str(), std::get<2>(tag).size());
+      }
+      out.close();
 
       std::unique_ptr<ImageIO::Base> io_handler (new ImageIO::Default (H));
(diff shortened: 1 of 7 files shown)


jdtournier:
OK, had a look at this - not fun... 

First off, [the docs state clearly that it's supposed to be stored big-endian](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat#line-39) - I think we can safely skip all the endian-ness checks and just assume big-endian.

Otherwise, I had a go at understanding their storage conventions, and I think it's not too difficult. Obscure to figure out, but otherwise pretty easy. The relevant sources are their [tags.h](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat?action=AttachFile&do=view&target=tags.h) and [tags.c](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat?action=AttachFile&do=view&target=tags.c) files. But essentially, it boils down to: 

- a 32 bit int (signed as far as I can tell) with a tag 'ID' (3 corresponds to a command-line history entry, I guess, etc);
- a 64 bit int (also signed as best I can figure out), to store the size of the entry (no idea why they'd need a 64 bit integer here, seems a bit OTT...); 
- the entry itself (which it looks like we can safely interpret as a string).
- next tag starts at (entry contents offset) + size. 

I've had a quick go at getting this working for reading, check out branch `mgh_mgz_try_reading_other`, it seems to work fine for the dataset shared by @ansk (for reading `.mgh` only, this is just to verify operation). What I've done so far is store them all in comments with a `[TAG N]: ` prefix, but maybe they could be stored under their own keys. For the file we have, most tags are not listed in their `tags.h`, only the `CMDLINE` one. Maybe we could store that in the header under the key `CMD` or something. For the others, maybe a key name like `MGH_TAG_N` or similar would do the trick...?

As to supporting these entries for `.mgz`, I guess we could add a `lead_out` member variable to the [ImageIO::GZ handler](https://github.com/MRtrix3/mrtrix3/blob/master/lib/image_io/gz.h) - there's already a `lead_in` to store whatever comes before the data (i.e. the header itself), so it should be trivial to add the equivalent `lead_out` variant and write it out if non-empty. What do you think?

ansk:
Here is relevant up-to-date excerpt from `tags.h` containing definition of remaining tags:
```
#define TAG_OLD_COLORTABLE          1
#define TAG_OLD_USEREALRAS          2
#define TAG_CMDLINE                 3
#define TAG_USEREALRAS              4
#define TAG_COLORTABLE              5

#define TAG_GCAMORPH_GEOM           10
#define TAG_GCAMORPH_TYPE           11
#define TAG_GCAMORPH_LABELS         12

#define TAG_OLD_SURF_GEOM           20
#define TAG_SURF_GEOM               21

#define TAG_OLD_MGH_XFORM           30
#define TAG_MGH_XFORM               31
#define TAG_GROUP_AVG_SURFACE_AREA  32

#define TAG_AUTO_ALIGN              33

#define TAG_SCALAR_DOUBLE           40
#define TAG_PEDIR                   41
#define TAG_MRI_FRAME               42
#define TAG_FIELDSTRENGTH           43

```
PEDIR means phase encoding direction.

I have tested this with other `.mgh` files and it seems to work.

jdtournier:
OK, had a shot at getting full read/write support working for tags with both `.mgh` and `.mgz` formats. I've opted for keeping things as comments, to avoid any potential conflicts with keys in the header. I've also decided to skip tags that contain no data, even though they might have a non-zero size according to the tag size field. Finally, since we read the content of the tag as a string, these enormous tags will be shortened considerably, so the output file will not have bitwise identical content compared to the original. Would appreciate thorough review and testing before this gets merged in...

ansk:
I have tested this on several files and it worked OK: The tags have been correctly preserved and read by `mri_info`. The only think to come to my mind is whether it is optimal to preserve record of tag 3: Command line which produced the file. To preserve previous record could be misleading in this context. To achieve full-consistency with FreeSurfer behavior, this tag content should be maybe replaced by actual `mrtrix` command line which produced the file. Another option could be to preserve existing instances of tags 3 and append append additional instance of tag 3 with current command line. I have seen files with two instances of tag 3, not sure whether there is some limit of occurences of tag 3.

jdtournier:
>  I have seen files with two instances of tag 3, not sure whether there is some limit of occurences of tag 3.

Well, the file you sent through has 3 such tags... :wink:

> To achieve full-consistency with FreeSurfer behavior, this tag content should be maybe replaced by actual mrtrix command line which produced the file.

We have discussed including such an option within _MRtrix3_ quite a few times in the past. Maybe now is the time... Part of the reason we haven't been too bothered about it so far is that the command history won't survive conversion to other formats, notably NIfTI. So passing data through e.g. `dwipreproc`, which invokes FSL commands with data converted to NIfTI, will lose that history. But it does look like we might be able to preserve it across MRtrix and FreeSurfer formats, so it's probably worth doing. 

Note that if we do go along with it, it'll probably be as part of the next release, which shouldn't be too far now. In this case, we'd need to handle that MGH tag differently - not a problem, just needs to handled properly.

On the other hand, I'm not too sure how to handle that tag 41 `TAG_PEDIR` correctly. The example data you sent through just shows `UNKNOWN`, but we'd have to see what format the information might come in if present, and whether it's compatible with what @Lestropie has been up to on that front (#747). So for now, the safest thing is to keep them all as specially crafted comments, so that they can be pushed back to the output file without loss of information...

ansk:
> not sure whether there is some limit of occurences of tag 3

Looking at the code, the maximum number of command line tags limits `MAX_CMDS` constant, which is set to 1000.

> On the other hand, I'm not too sure how to handle that tag 41 TAG_PEDIR correctly. The example data you sent through just shows UNKNOWN, but we'd have to see what format the information might come in if present, and whether it's compatible with what @Lestropie has been up to on that front (#747)

`TAG_PEDIR` is not so informative as in #747 . As far as I can tell, the workflow of this tag content is following: During DICOM conversion the content of DICOM tag 0018,1312

http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,1312)

is copied into `TAG_PEDIR`. Therefore the possible values are only `ROW`, `COL` and `UNKNOWN` (the latter in case this info is not present or relevant). 

During `mri_convert -conform` which is very early step of `recon-all` the `TAG_PEDIR` is set to `UNKNOWN`.

https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllDevTable

Lestropie:
> So passing data through e.g. `dwipreproc`, which invokes FSL commands with data converted to NIfTI, will lose that history.

Such cases could (should!) be dealt with explicitly.

Ideally each script would grab the `command_history` contents of the *primary input file*, append a string for the *script invoked* (i.e. *not* include all of the *MRtrix3* commands invoked *within* the script, but the script invocation), and overwrite the `command_history` key of any output images with that. As it stands now, image outputs from these scripts would have a long command history corresponding to the internals of the Python script subsequent to the first `.nii` -> `.mif` conversion.

jdtournier:
@Lestropie: are you happy to merge this as-is to `master`? The discussion about command-line history handling can carry on in #962, which is due to merge into `tag_0.3.16`, not `master`. We'd need to modify the handling of this MGH tag anyway if we want to support it transparently alongside our own history, and that should probably be reserved for the next release anyway...

thijsdhollander:
Sounds good to me :+1: .  Administrative announcement: @Lestropie will probably not respond until after the long Easter weekend.  (he left camping with only pen and paper as a means to do science)

Lestropie:
Man, writing TCP-IP packets using pen and paper is hard... :-P
(But yes, I'm radio silent until probably Tuesday night; y'all go on without me)

Don't have time to muck around with this extensively. If the addition is to properly read the MGHO "tags" field, and it's tested for both `.mgh` and `.mgz`, and it's just reading them into header key-value fields / writing to output image file lead-out as-is, that's fine by me. Only things I would suggest:

- Double-check what's been committed to `mgh_mgz_try_reading_other` v.s. this branch. 

- Copy the `#define` table posted by @ansk and use that to define appropriate keys for the imported header entries, rather than just enumerating them?

jdtournier:
OK, had a shot at cleaning things up a bit more, should be good to go. Tags are now named according to the table supplied by @ansk, and convert back and forth OK as far as I can tell. The `mgh_mgz_try_reading_other` branch is gone, it got merged into this one quite early on anyway. Wouldn't mind another pair of eyes to make sure everyone does work as expected before merging. 

After that, I guess we can start thinking about #962...

jdtournier:
OK, I see you didn't like my `prepare_tag()` idea... Oh well.

> FreeSurfer will still not open files created by MRtrix3

That's odd, I thought this previously had been confirmed as working by @ansk? Maybe the latest commits have mangled things a bit? Does it not work with e19dabe?

Lestropie:
I got the idea of `prepare_tag()`, but in the process of trying to figure out what was going wrong, it disappeared. Partly because in `mgh_header` and `mgh_other`, the stored values always remain as big-endian and have to be read/written accordingly. It looks like you tried to do that, but then there were missing `ByteOrder::BE()` calls around the place, so I decided to remove that function and validate every read / write functionality individually.

Yes, it was confirmed working at one point; but there were changes subsequent to that. I've also been looking at null terminations (or lack thereof) wondering if that's causing issues. Also not sure if all combinations were tested, i.e. Freesurfer export to MRtrix import, MRtrix export to FreeSurfer import, MRtrix export to MRtrix import, zipped v.s. unzipped.

lib/formats/mgh.cpp
```diff
@@ -112,6 +122,7 @@ namespace MR
       for (const auto& tag : MGHO.tags) {
         out.write (reinterpret_cast<const char*> (&tag.id), sizeof(tag.id));
         out.write (reinterpret_cast<const char*> (&tag.size), sizeof(tag.size));
+        assert (size_t(ByteOrder::BE (tag.size)) == tag.content.size());
```
jdtournier: That assert will fail routinely, because of the null termination on the `tag.content` string. For example, in the data sent through by @ansk:
```
...
mrinfo [lib/formats/mgh.cpp: 63]: tag.id = 33
mrinfo [lib/formats/mgh.cpp: 64]: tag.size = 1600
mrinfo [lib/formats/mgh.cpp: 65]: tag.content.size() = 185
mrinfo [lib/formats/mgh.cpp: 66]: tag.content = AutoAlign   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000
...
```
Not sure whether this has anything to do with FreeSurfer not accepting these data, but that's how it worked before when it was confirmed as working. I'm guessing it'll be something else...

Lestropie: That specific case is a write, so it's only testing that MRtrix has set the tag.size value appropriately. Nothing to do with input data.

I'm pretty sure that FreeSurfer does *not* include null terminators (at least not always) in the length calculation: the phase encode direction tag gave a size of 7 for the string `UNKNOWN`.

jdtournier:
> I got the idea of prepare_tag(), but in the process of trying to figure out what was going wrong, it disappeared. Partly because in mgh_header and mgh_other, the stored values always remain as big-endian and have to be read/written accordingly. It looks like you tried to do that, but then there were missing ByteOrder::BE() calls around the place, so I decided to remove that function and validate every read / write functionality individually.

I figured it might have been something like that... 

My only concern with your changes is that it looks like you're keeping big-endian versions of the tags once read in RAM. I think it would be much cleaner to convert at the read or write time, and maintain native-endian versions in RAM at all times otherwise. I'll have a go at that. 

> Yes, it was confirmed working at one point; but there were changes subsequent to that. I've also been looking at null terminations (or lack thereof) wondering if that's causing issues. Also not sure if all combinations were tested, i.e. Freesurfer export to MRtrix import, MRtrix export to FreeSurfer import, MRtrix export to MRtrix import, zipped v.s. unzipped.

Pretty sure the testing wouldn't have been extensive, at least not on my side - I don't have FreeSurfer installed... It will need some more thorough testing before merging.

lib/formats/mgh.cpp
jdtournier: Ah, OK... please ignore me.

Lestropie:
> I think it would be much cleaner to convert at the read or write time, and maintain native-endian versions in RAM at all times otherwise.

Trouble then is that you have these tags being stored in RAM with native endianness, but the header & other structs being represented as big-endian at all times. Was just another source of confusion when I was looking through it all.

jdtournier:
> Trouble then is that you have these tags being stored in RAM with native endianness, but the header & other structs being represented as big-endian at all times.

OK, I see. It might be that we'd be better off not using memory-mapping for the read then. Would also avoid this issue with alignment and the `#pragma pack` directive...? Would be a bit more work, though...

jdtournier:
Canned in favour of #970

Lestropie:
@ansk: Where did you get that up-to-date `tags.h`? I've checked downloads of FreeSurfer 5.3 and 6.0 beta, and neither seem to come with the relevant source code. I'm pretty sure that the source files linked to on the [MGH format page](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat) are old; they contain less `#define`'s than what you provided, and I can see some potential bugs in the image file read/write functions that may have since been fixed. If I could get my hands on fully up-to-date source code, I might be able to mimic their functions to provide more reliable support in *MRtrix3*...

ansk:
@Lestropie,
I requested access to their git repository. But recently it is much easier since FreeSurfer sources are now also on github: https://github.com/freesurfer/freesurfer
So `tags.h` for development version of FreeSurfer is:
https://github.com/freesurfer/freesurfer/blob/dev/include/tags.h

Lestropie:
OK thanks @ansk; that looks more up-to-date than a forked version of the repository I found after sending that message.

For reference for future me: Implementation code is found in `/utils/` rather than `/include/`.

Lestropie:
@ansk (and anybody else interested): I've mimiced much of the FreeSurfer code in *MRtrix3* to further improve the handling of MGH / MGZ files in #970. If anybody has image files that have consistently caused problems in *MRtrix3*, it would be appreciated to test them against the `mgh_mgz_write_fix_take2` code branch before those changes get merged to master. Though it certainly worked with the example problematic files I've been provided with.

PR #961, opened 2017-04-10
Added command options for mrview:

LeeReidCSIRO:
tractography.load (loads a track file)
tractography.slab (slab thickness, or -1 for unlimited)
tractography.tsf (uses a tsf file for track node colours)
WARNING: tsf optional arguments do not appear to actually be optional due to pre-existing code enforcing 'optional' args (bug?).

Edits work, but may require a slight tidy to match expected patterns etc. I've never touched QT before!


maxpietsch:
Can confirm that `tractography.slab` works.
`tractography.tsf` works except for described problem with the default argument.

Arguments can be declared optiona; but, as far as I know, there is no way of specifying an Argument to an Option as optional. So I'd suggest using separate options instead:

```
+ Option ("tractography.tsf_load", "Load the track scalar file. Filename ").allow_multiple()
+ Argument("tsf").type_file_in()

+ Option ("tractography.tsf_range", "Set range for the track scalar file.  RangeMin,RangeMax").allow_multiple()
+ Argument("range").type_sequence_float()

+ Option ("tractography.tsf_thresh", "Set thresholds for the track scalar file.  ThresholdMin,ThesholdMax").allow_multiple()
+ Argument("thresh").type_sequence_float()
```

jdtournier:
OK, had a quick look into this. I agree with @maxpietsch that it would be best to split off the different settings into different options. And it's indeed not possible to have optional arguments to options, otherwise a command like this can't be parsed correctly:
```bash
$ command argument1 -option mandatory_argument optional_argument argument2 ...
```
since `optional_argument` could legitimately be interpreted as main argument 2 - so it's fundamentally ambiguous. 

Otherwise, I fixed up the indentation to use 2 white spaces rather than tabs - this is the standard we've settled on in _MRtrix3_. There might be a way to set your editor to do use this for this project. 

The main issue you were facing was the fact that no tractogram was selected when you want to load the TSF. This would be better fixed by ensuring that the [Tractography::tractogram_open_slot()](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/tractography/tractography.cpp#L288-L300) call actually selected the newly loaded tractograms. This should immediately fix that problem, and also ensure correct operation when you want to load a TSF file that doesn't correspond to the _last_ tractogram in the list (we would eventually need to add an option to choose which tractogram is selected). 

I'm surprised that you'd need to manually invoke `on_crop_to_slab_slot()` and `on_slab_thickness_slot()`. This shouldn't be required, not sure why it's not behaving as expected...

I won't have time to look into this a great deal more over the next few days, but if you're feeling motivated, hopefully the above will give you some inspiration... 😉

LeeReidCSIRO:
> This should immediately fix that problem, and also ensure correct operation when you want to load a TSF file that doesn't correspond to the last tractogram in the list (we would eventually need to add an option to choose which tractogram is selected).

The way it is set up at the moment means that the TSF file applies to the previously stated tractography file provided. This means you can do the following to apply two different tsf files to two different tck files:

`mrview -tractography.load a.tck -tractography.tsf_load a.tsf -tractography.load b.tck -tractography.tsf_load b.tsf `

I.E. a.tsf is applied to a.tck, and b.tsf is applied to b.tck. 
I figure that this is intuitive for the user?

Modifying `Tractography::tractogram_open_slot` will not affect the current behaviour because this method is not used (this slot launches an open file dialog). I've now made a method called `Tractography::select_last_added_tractogram` that does what it says. If you want to call this from `Tractography::tractogram_open_slot` at some point in the future, now you can. :)

> I'm surprised that you'd need to manually invoke on_crop_to_slab_slot() and on_slab_thickness_slot(). This shouldn't be required, not sure why it's not behaving as expected...

No idea! I was surprised too. Qt is foreign to me. Perhaps signals are suppressed by the class until things are drawn to screen or I made an error somewhere? Not critical at the moment.

> Arguments can be declared optiona; but, as far as I know, there is no way of specifying an Argument to an Option as optional. So I'd suggest using separate options instead

Edited as suggested

jdtournier:
> The way it is set up at the moment means that the TSF file applies to the previously stated tractography file provided.

I fully agree that it would work as expected, certainly as things stand. But I do think the correct solution here would be to fix the behaviour when a tractogram is added and make sure it is also selected at that point. It would make sense when interacting with the tool directly anyway. The Overlay tool does this, so it's probably a good example of how to get it to work. 

> Modifying Tractography::tractogram_open_slot will not affect the current behaviour because this method is not used (this slot launches an open file dialog). 

Of course, you're right. [The Overlay tool splits the open slot from the actual method that opens the images](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L198-L222), so that the latter can be used programmatically. Ultimately, that would probably be favoured solution... 

As to the issue with the update needed to be triggered manually, I can probably look into this when I have a minute (probably after the ISMRM though...).

LeeReidCSIRO:
I beg your pardon, I see what you were saying now. I'll have a go at refactoring it in the next few days
________________________________________

LeeReidCSIRO:
Sorry for the delay. I've refactored the code as suggested. Opening a tractogram will now result in it being selected.

There is a bug (in the current version of MRtrix), and I'm not whether I introduced it or not in my previous commit. To replicate:
1) Open MRview, open an image
2) Open the tractography tool side bar
3) Open a tractogram

The tractography file will load, but the side bar will not list the file until the window is resized.

LeeReidCSIRO:
Worth highlighting that the clang build is failing due to some change in the Eigen 3 library, not my code. 

> Checking memory alignment for Eigen 3.3 compatibility... FAIL

Compiling with g++-6 works fine

jdtournier:
@LeeReidCSIRO, I finally found a bit of time to look through your suggestions. I made the very minor edits required to get it to pass the Eigen compatibility tests (I've yet to document these changes, so it's no surprise that it would catch people out...). The other changes are basically purely cosmetic. Everything seems to work fine from my side. The issue you mentioned with the sidebar not updating after tractography load until a resize isn't a problem for me. What version of Qt are you using...?

maxpietsch:
> There is a bug (in the current version of MRtrix), and I'm not whether I introduced it or not in my previous commit. To replicate:
> 
> Open MRview, open an image
Open the tractography tool side bar
Open a tractogram
The tractography file will load, but the side bar will not list the file until the window is resized.

I can't reproduce this. Neither in our nor in LeeReidCSIRO's master branch. Loading a tractogram causes the toolbar to flicker so I think it gets updated.

LeeReidCSIRO:
Qt 4.8.6
Open GL 3.3
It's a Ubuntu 14.04 box.

LeeReidCSIRO:
@maxpietsch I think if it's working for you here, chances are it's a bug that's a little deeper than my commits.

LeeReidCSIRO:
Kerstin Pannek has just pointed the following:
Open Mrview with image -->  add tck file --> not shown in toolbar until resize
but
Open Mrview with image -->  maximise window -->  add tck file --> no problems

jdtournier:
I've just had another quick go, I can't reproduce the issues you're reporting using either Qt 5.8.0 or Qt 4.8.7 (on Arch Linux). I have no idea why this might be happening. Could be an interaction with your window manager? We use Gnome Shell typically (or i3 for @maxpietsch...), maybe this is an issue specific to Unity...?

jdtournier:
Let's merge this into the `dev` branch with a view to making it part of the next tag release. I don't see any issues with it, other than the problems reported by @LeeReidCSIRO and Kerstin, which we can't reproduce. But I don't think they're show-stoppers anyway...

LeeReidCSIRO:
That sounds good to me.

jdtournier:
OK, changed base branch to `tag_3.0_RC2` and updated. I'll merge once testing completes OK...

Issue #183, opened 2015-03-18
Save the command used to generate a file in the output header


draffelt:
We could even propagate the commands saved in the input images to the output images, such that each file has a history of where it came from.  

By saving absolute file paths in combination with the git hash (see #111), a record of exactly how the file was generated would exist. We could add a -history option to mrinfo to dump what is effectively a bash script, so then you could re-run the analysis if code or the input images are changed. Since some files may have common ancestors, we could even have a dedicated mrtrix command called 'regenerate', that takes multiple input files, reads their history, removes any duplicate commands, and re-runs the analysis.

This would also serve as a great tool for making backups of what you did, without having to store duplicate copies of all images.

As Thijs raised, the only concern with absolute paths is that sensitive information might be saved in files (such as a persons username or patient names if the parent folder is labelled as such)

Of course if files are moved after they are created then this will also be an issue. But the history 'script' could easily be modified.

draffelt:
I guess if we made it clear when storing the command used to generate the file what arguments were input files, then we could always write a script/app to work backwards build the complete history of commands on the fly, providing that the input files were not moved. This would avoid us having to store the complete history in each file (along with the git hash for each command).

Lestropie:
Probably wouldn't put this in the header storage; just store the generating command string as-is in the header, and e.g. parsing the output of `__print_full_usage__` for each command would allow a script to figure out what's input and what's output for each. Or implement an alternative command-line option that just gives the necessary information, maybe `__interface__` or the like, and have the set of Python scripts output something appropriate as well. Though I guess this wouldn't work for non-MRtrix stuff.

Also need to think about how piping might be handled in this context.

draffelt:
Good point, piping could be an issue

Lestropie:
Could potentially be done if the translation of the '-' were done immediately at argument/option parsing time rather than by the pipe image format/handler; that way both calls would store a path to the same (non-existent) temporary file.

jdtournier:
Yes, the idea of storing the command history in the header has come up quite a few times before, and I've always thought it would be worth doing. Problem is that doing it _right_ is actually very tricky...

First off, the command history is rarely linear. Many commands take multiple inputs, and each of these commands would have their own history. So we would need to store a history _tree_. Not impossible clearly, but not as neat and tidy as might appear at first sight.

[...]

PR #962, opened 2017-04-10
add current command-line to 'command_history' header entry

jdtournier:
as suggested in #956.

All up for discussion. It's a simple addition, so why not. But it's not perfect, that history won't survive conversion to other formats - although it might (with a few modifications) work well with the MGH formats.

### DIFF
--- a/core/header.cpp
+++ b/core/header.cpp
@@ -161,10 +161,6 @@ namespace MR
       throw Exception ("no name supplied to open image!");
 
     Header H (template_header);
-    std::string cmd = App::argv[0];
-    for (int n = 1; n < App::argc; ++n)
-      cmd += std::string(" \"") + App::argv[n] + "\"";
-    add_line (H.keyval()["command_history"], cmd);
 
     try {
       INFO ("creating image \"" + image_name + "\"...");
@@ -172,6 +168,15 @@ namespace MR
       H.keyval()["mrtrix_version"] = App::mrtrix_version;
       if (App::project_version)
         H.keyval()["project_version"] = App::project_version;
+
+      std::string cmd = App::argv[0];
+      for (int n = 1; n < App::argc; ++n)
+        cmd += std::string(" \"") + App::argv[n] + "\"";
+      cmd += std::string ("  (version=") + App::mrtrix_version;
+      if (App::project_version)
+        cmd += std::string (", project=") + App::project_version;
+      cmd += ")";
+      add_line (H.keyval()["command_history"], cmd);
 
       H.sanitise();
 
--- a/core/header.cpp
+++ b/core/header.cpp
@@ -126,7 +126,7 @@ namespace MR
       }
 
       H.sanitise();
-      if (!do_not_realign_transform) 
+      if (!do_not_realign_transform)
         H.realign_transform();
 
     }
@@ -139,14 +139,14 @@ namespace MR
 
 
   namespace {
-    inline bool check_strides_match (const vector<ssize_t>& a, const vector<ssize_t>& b) 
+    inline bool check_strides_match (const vector<ssize_t>& a, const vector<ssize_t>& b)
     {
       size_t n = 0;
-      for (; n < std::min (a.size(), b.size()); ++n) 
-        if (a[n] != b[n]) return false; 
-      for (size_t i = n; i < a.size(); ++i) 
+      for (; n < std::min (a.size(), b.size()); ++n)
+        if (a[n] != b[n]) return false;
+      for (size_t i = n; i < a.size(); ++i)
         if (a[i] > 1) return false;
-      for (size_t i = n; i < b.size(); ++i) 
+      for (size_t i = n; i < b.size(); ++i)
         if (b[i] > 1) return false;
       return true;
     }
@@ -161,6 +161,10 @@ namespace MR
       throw Exception ("no name supplied to open image!");
 
     Header H (template_header);
+    std::string cmd = App::argv[0];
+    for (int n = 1; n < App::argc; ++n)
+      cmd += std::string(" \"") + App::argv[n] + "\"";
+    add_line (H.keyval()["command_history"], cmd);
 
     try {
       INFO ("creating image \"" + image_name + "\"...");
@@ -226,7 +230,7 @@ namespace MR
         size_t axis = 0;
         while (axis < limits.size()) {
           pos[axis]++;
-          if (pos[axis] < limits[axis]) 
+          if (pos[axis] < limits[axis])
             return true;
           pos[axis] = 0;
           axis++;
@@ -255,7 +259,7 @@ namespace MR
         H.axes_.resize (n + Pdim.size());
 
         for (size_t i = 0; i < Pdim.size(); ++i) {
-          while (H.stride(a)) 
+          while (H.stride(a))
             ++a;
           H.size(a) = Pdim[i];
           H.stride(a) = ++next_stride;
@@ -281,14 +285,14 @@ namespace MR
 
 
 
-  Header Header::scratch (const Header& template_header, const std::string& label) 
+  Header Header::scratch (const Header& template_header, const std::string& label)
   {
     Header H (template_header);
     H.name() = label;
     H.reset_intensity_scaling();
     H.sanitise();
     H.format_ = "scratch image";
-    H.io = make_unique<ImageIO::Scratch> (H); 
+    H.io = make_unique<ImageIO::Scratch> (H);
     return H;
   }
 
@@ -299,14 +303,14 @@ namespace MR
 
 
 
-  std::ostream& operator<< (std::ostream& stream, const Header& H) 
+  std::ostream& operator<< (std::ostream& stream, const Header& H)
   {
     stream << "\"" << H.name() << "\", " << H.datatype().specifier() << ", size [ ";
     for (size_t n = 0; n < H.ndim(); ++n) stream << H.size(n) << " ";
     stream << "], voxel size [ ";
-    for (size_t n = 0; n < H.ndim(); ++n) stream << H.spacing(n) << " "; 
+    for (size_t n = 0; n < H.ndim(); ++n) stream << H.spacing(n) << " ";
     stream << "], strides [ ";
-    for (size_t n = 0; n < H.ndim(); ++n) stream << H.stride(n) << " "; 
+    for (size_t n = 0; n < H.ndim(); ++n) stream << H.stride(n) << " ";
     stream << "]";
     return stream;
   }
@@ -364,10 +368,10 @@ namespace MR
 
     for (const auto& p : keyval()) {
       std::string key = "  " + p.first + ": ";
-      if (key.size() < 21) 
+      if (key.size() < 21)
         key.resize (21, ' ');
       const auto entries = split_lines (p.second);
-      if (entries.size() > 5) 
+      if (entries.size() > 5)
         desc += key + "[ " + str (entries.size()) + " entries ]\n";
       else {
         for (const auto value : entries) {
@@ -426,13 +430,13 @@ namespace MR
       size_t num_valid_vox = 0;
       for (size_t i = 0; i < 3; ++i) {
         if (std::isfinite(spacing(i))) {
-          ++num_valid_vox; 
+          ++num_valid_vox;
           mean_vox_size += spacing(i);
         }
       }
       mean_vox_size /= num_valid_vox;
-      for (size_t i = 0; i < 3; ++i) 
-        if (!std::isfinite(spacing(i))) 
+      for (size_t i = 0; i < 3; ++i)
+        if (!std::isfinite(spacing(i)))
           spacing(i) = mean_vox_size;
     }
   }


Lestropie:
Originally raised in #183.

Thought: Should the most recent command be the *first* line in `command_history`, rather than the last? That way if there's a long history, e60e9ba will make sure it still appears in the `mrinfo` output.

thijsdhollander:
Makes sense that, if you were interested in one command, of all commands, that would indeed probably be the most recent one.  Another solution, rather than inverting the order (which has the potential to be non-intuitive in other scenarios), would be to let `mrinfo` show something like the first and last 3 lines (with a "..." in between) of any entry.  Potential bonus: in the absence of conversion to non-supporting formats, it would give a good impression of what "part of a pipeline" has been applied to an image (i.e. first few until last few commands)... if that makes sense.

jdtournier:
> Another solution, rather than inverting the order (which has the potential to be non-intuitive in other scenarios), would be to let mrinfo show something like the first and last 3 lines (with a "..." in between) of any entry. 

I like this idea! You've probably noticed that my recent commit on `tag_0.3.16` (e60e9ba) now prints the first 5 lines of any property, and cuts it short with a `[ + N more ]` line if it's longer than that (it also adds the `-all` option to prevent this truncation). It would make sense to instead show the first & last 2 or 3, with the `[ + N more ]` in the middle. I think that would address the issue? 

The other thing to worry about if we reverse the order of the history is to ensure it's still consistent with the MGH format. I assume these entries appear in chronological order in the file, but that is admittedly only an assumption...

jdtournier:
> You've probably noticed that my recent commit on tag_0.3.16 (e60e9ba) now prints the first 5 lines of any property

:grin: of course you already have... Sorry, should have re-read your post before jumping in.

ansk:
In MGH format the command line history is chronological, not reverse. See the example of `filled.mgz` file which has 9 command line history entries. The tags also contain other various useful information, such as software and platform version.

```$mrinfo filled.mgz -property comments
TR: 2300ms
Flip: 10deg
TE: 4.63ms
TI: 900ms
[MGH TAG 31]: /hydradb/hydra_io/vypocty/freeSurfer/v6.1beta/CANEF/A_in_progress/skoch/ESO_P00007_20100611_1217_1_base/mri/transforms/talairach.xfm
[MGH TAG 33]: AutoAlign   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000   0.000000   0.000000   0.000000   0.000000   1.000000
[MGH TAG 41]: UNKNOWN
[MGH TAG 3]: mri_convert /tmpcount/freeSurfer/A_input/skochv/export_20_4_2016/ESO_P00007_20100611_1217_1/0006_t1_sag_mpr_3Diso__1mm/MR.1.3.12.2.1107.5.2.32.35226.20100611124804148709128 /tmpcount/freeSurfer/A_results/A_in_progress/skochv/pat/ESO_P00007_20100611_1217_1/mri/orig/001.mgz ProgramVersion: $Name: stable5 $  TimeStamp: 2016/04/21-13:32:39-GMT  BuildTimeStamp: May 13 2013 16:24:28  CVS: $Id: mri_convert.c,v 1.179.2.7 2012/09/05 21:55:16 mreuter Exp $  User: skochv  Machine: matpcp.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.2.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_convert.bin /hydradb/hydra_io/vypocty/freeSurfer/v6.1beta/CANEF/A_in_progress/skoch/ESO_P00007_20100611_1217_1/mri/rawavg.mgz /hydradb/hydra_io/vypocty/freeSurfer/v6.1beta/CANEF/A_in_progress/skoch/ESO_P00007_20100611_1217_1/mri/orig.mgz --conform ProgramVersion: $Name:  $  TimeStamp: 2017/02/13-09:32:22-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_convert.c,v 1.226 2016/02/26 16:15:24 mreuter Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_ca_normalize -c ctrl_pts.mgz -mask brainmask.mgz nu.mgz /hydra-db/hydra_io/vypocty/skoch/bin/freesurfer_dev_20170208/average/RB_all_2016-05-10.vc700.gca transforms/talairach.lta norm.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/03/19-00:40:49-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_ca_normalize.c,v 1.69 2016/10/22 17:30:57 fischl Exp $  User: skoch  Machine: sup3hw.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_ca_normalize -c ctrl_pts.mgz -mask brainmask.mgz norm_template.mgz /hydra-db/hydra_io/vypocty/skoch/bin/freesurfer_dev_20170208/average/RB_all_2016-05-10.vc700.gca transforms/talairach.lta norm.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/04/05-09:40:16-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_ca_normalize.c,v 1.69 2016/10/22 17:30:57 fischl Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_normalize -mprage -aseg aseg.presurf.mgz -mask brainmask.mgz norm.mgz brain.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/04/05-13:18:28-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_normalize.c,v 1.92 2017/01/27 22:31:34 ah221 Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_segment -keep -mprage brain.mgz wm.seg.mgz 
[MGH TAG 3]: mri_edit_wm_with_aseg -keep-in wm.seg.mgz brain.mgz aseg.presurf.mgz wm.asegedit.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/04/05-13:23:06-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_edit_wm_with_aseg.c,v 1.25 2012/03/29 13:17:38 fischl Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_pretess -keep wm.asegedit.mgz wm norm.mgz wm.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/04/05-13:23:30-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_pretess.c,v 1.22 2013/08/30 18:12:25 mreuter Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400  
[MGH TAG 3]: mri_fill -a ../scripts/ponscc.cut.log -xform transforms/talairach.lta -segmentation aseg.auto_noCCseg.mgz wm.mgz filled.mgz ProgramVersion: $Name:  $  TimeStamp: 2017/04/05-13:23:34-GMT  BuildTimeStamp: Feb  6 2017 17:11:52  CVS: $Id: mri_fill.c,v 1.119 2011/10/25 14:09:58 fischl Exp $  User: skoch  Machine: sup4.ad.nudz.cz  Platform: Linux  PlatformVersion: 3.16.0-4-amd64  CompilerName: GCC  CompilerVersion: 40400```

**Comment by jdtournier** [comment]
OK, I've added the version of the executable to the command history, so we can track that through also. 

I'll merge this now, we can modify over the next few days if need be.

File: core/header.cpp [hunk 1]
```diff
@@ -169,6 +169,15 @@ namespace MR
       if (App::project_version)
         H.keyval()["project_version"] = App::project_version;
 
+      std::string cmd = App::argv[0];
+      for (int n = 1; n < App::argc; ++n)
+        cmd += std::string(" \"") + App::argv[n] + "\"";
```
Lestropie: @jdtournier Is it necessary to add double-quotes to every entry in `argv`? Could it not be reserved only for entries that contain whitespace?

jdtournier: Yes, it is a bit ugly. But white spaces aren't the only potential problem. We'd have to watch for at least these characters: ` []*?%#~&$(){}\t\\"';:<>` - i.e. absolutely _anything_ that the shell might conceivably interpret. If the command we received contained these characters, then that's what the user intended, and they must have supplied the argument in such as way as to avoid interpretation by the shell. So we should respect that - the arguments the executable received have already been interpreted by the shell, so we should ensure that a straight copy/paste of the command works without the shell re-interpreting anything incorrectly. 

Also, I think we should be using single quotes `'` to avoid further variable expansion, etc. - otherwise an argument like `stuff$n.mif` would be interpreted when it should be left alone (although I completely agree that users who actually names their files like that deserve everything they get...).

Lestropie: Reminds me of [here](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/path.py#L35-L37), where I wrap paths in quotation marks only when necessary for `shlex.split()`. Admittedly that maybe needs to check for more than just whitespace; but that's what triggered the implementation. Why not simply use a similar heuristic here, i.e. only wrap a string / path in quotation marks if it contains special characters (which will be the vast minority of the case)?

Lestropie: Minor addition to this: Could probably use `Path::basename(App::argv[0])`. I'm not sure that keeping track of the location of the binary on the system for each command in the history is worth the bloat for the very rare instance where it might provide useful information.

Lestropie:
Another follow-up on this:

If a command takes multiple input images, and each contains its own `command_history` entry, what should the output header contain? One option would be to timestamp each entry, and sort them on combination. Though it starts to sound like it would require its own dedicated handling in the `Header` class rather than being just another key-value pair...

jdtournier:
> If a command takes multiple input images, and each contains its own command_history entry, what should the output header contain? One option would be to timestamp each entry, and sort them on combination. Though it starts to sound like it would require its own dedicated handling in the Header class rather than being just another key-value pair...

This is an issue I'd [raised a while back](https://github.com/MRtrix3/mrtrix3/issues/183#issuecomment-82829944), and I'm not sure there's an obvious easy answer here... If we store a tree, it needs to be easily parsed by a human, I doubt we'll see these entries needing to be parsed by a computer any time soon (unless you guys have grand plans I'm not privy to). But you'd need to store a record of which input to the Nth command a particular history branch refers to, and ways of expressing this hierarchy. Maybe we need to prefix the command history entry with the filename of the file it refers to (as provided to the command storing the history), and some way of expressing the tree structure in a way that can easily be represented on a terminal screen. I'm not sure there's any worth in making the header structure any more complicated if it doesn't also make the display of this information any easier. From my point of view, if we can come up with a _textual_ representation of the hierarchy, there is nothing stopping us from just storing this in the header using the current _key: value_ format, and just dumping it out when needed, like we do the comments fields for instance. All it needs is for the formatting that carries the hierarchy information to be part of the text stored in these fields, for example:

    $ dwi2mask dwi.mif mask.mif
    $ dwi2tensor dwi.mif - | tensor2metric - -mask mask.mif -fa fa.mif

would need to store a structure like this:
~~~
[fa.mif] tensor2metric "-" "-mask" "mask.mif" "-fa" "fa.mif"  (version=3.0_RC1-186-g6034fc2f)
├── [-] dwi2tensor "dwi.mif" "-"  (version=3.0_RC1-186-g6034fc2f)
└── [mask.mif] dwi2mask "dwi.mif" "mask.mif"  (version=3.0_RC1-186-g6034fc2f)
~~~

There's no reason why we couldn't store that as-is in the header, simply prefixing each line with a `command_history:` key as we currently do. It might actually be relatively easy to manage appending to the history that way. For example, if the next command was:

    $ mrthreshold fa.mif -top 300 sf.mif

Then the history would become:
~~~
[sf.mif] mrthreshold "fa.mif" "-top" "300" "sf.mif"   (version=3.0_RC1-186-g6034fc2f)
└── [fa.mif] tensor2metric "-" "-mask" "mask.mif" "-fa" "fa.mif"  (version=3.0_RC1-186-g6034fc2f)
    ├── [-] dwi2tensor "dwi.mif" "-"  (version=3.0_RC1-186-g6034fc2f)
    └── [mask.mif] dwi2mask "dwi.mif" "mask.mif"  (version=3.0_RC1-186-g6034fc2f)
~~~
which should be relatively easy to manage: the whole history for the arguments gets indented by 4 or so spaces, the leading line prefixed by `└── `, and that whole chunk gets added to the history for each of the image arguments (ideally including arguments to options). So I think we could deal with this. 

A bigger issue though is how to handle this transparently. At the moment, the history update is handled by the `Header::create()` call, based on the information already present in the incoming template `Header`. It works since the incoming template `Header` is itself typically read from (one of) the input image. But if we want to inject the history of _all_ the inputs to the current command, we need access to the contents of all of the `Header`s at the point where any output image gets created. This doesn't match the way this is handled currently. The most obvious way to handle this seamlessly would be to have the command-line parsing backend load all image headers upfront before `run()` is invoked. This wouldn't be impossible to do, but there will be exceptions to this in cases where we can't determine whether a particular argument is going to be an image or something else (`mrcalc` immediately springs to mind). Relying on each command explicitly updating this field is not a robust solution. Maybe we could have a system whereby it is expected that all inputs have been opened before the first output is created, ensuring that the information has been gathered by that point - this may be the least intrusive way of doing this... assuming we want to go along with the whole idea, of course...

jdtournier:
By the way, I do think it would be worth removing the quotes when they're not required. We just need a semi-robust way of detecting characters that would require quoting (whitespaces probably aren't enough).

Also, I'd rather keep the full name of the `argv[0]` if possible. 99% of the time, this will only be the actual command basename anyway, since that's what the user typed. In those rare cases where it's not, it might actually prove important to know that the command wasn't in the default path... But admittedly the git history SHA1 is probably more important here.

thijsdhollander:
I'm not entirely convinced the benefits are worth the hassle, and the level of intrusion in commands (and potential performance).  As @dchristiaens brought up in that original discussion, I too (any others that I know) use non-MRtrix steps at times.  And then there's manual steps (e.g. ROI creation or editing) in many typical scenarios too...

For me, if any benefit at all, there's mostly some value into storing the last command an image was generated from, as some sort of automated note-taking functionality.  The current implementation already goes beyond that by keeping the (linear) history, but even there it creates such a big bulk of almost mess when outputting that, that I've actually never used it for any purpose.  I'd rather look at the 'mrinfo' of the command/image before and look at what produced that, if I'd ever need to.

Thinking about that though, this may be something: if an image just stores the last command that generated it (which has the source images file names in there too), and applies a convention as to what path these are relative to (or stores this (relative) path), then you could have an `mrshowhistory` command that goes and backtracks as far as it can in existing headers, and shows you this kind of history tree output.  It's of course not robust to files having moved, but then again if actual files have moved, one could question the value of a history (that will also have path issues) for the purposes of re-creating stuff as well.  As @jdtournier comments, I suppose for the foreseeable future the value is mostly in just showing this history to the user?

Lestropie:
> But if we want to inject the history of all the inputs to the current command, we need access to the contents of all of the Headers at the point where any output image gets created.

This doesn't just apply to `command-history`, but more generally to any header key-value entry in the event of a command that takes multiple input files. I've encountered this a couple of times recently (e.g. [this commit](https://github.com/MRtrix3/mrtrix3/commit/febc0a47c4194ed85f385a36223ac99b92003e62#diff-2afc70db0ec1201c11e0a41baf90d652)); I've also dealt with a similar issue in [`tckedit`](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/tckedit.cpp#L113-L149). There should perhaps be a helper function somewhere for generating the output header key/value set based on input headers? That would at least reduce this risk:
> Relying on each command explicitly updating this field is not a robust solution.

More generally, I think I'm in the camp of only storing the generating command, not a full command history. With the potential for non-*MRtrix3* commands & use of non-*MRtrix3* image formats in intermediate steps, potential non-linear paths, and no immediate need to be able to re-produce processing steps in an automated fashion using this functionality alone, it might be a rabbit hole best avoided. This would also be *much* more compatible with the Python scripts, since the "generating command" stored in the final output image header(s) could be over-ridden by the script library to represent the user's call to the script, rather than the internal `mrconvert` call that happens to be the last stage of the script.

thijsdhollander:
> More generally, I think I'm in the camp of only storing the generating command, not a full command history.

Yes, agreed.  That's probably the best (easiest) defined behaviour as well.  As I mentioned above, this still doesn't limit users from recreating a pretty detailed history in a well-controlled environment.

jdtournier:
>> More generally, I think I'm in the camp of only storing the generating command, not a full command history.
>
> Yes, agreed. That's probably the best (easiest) defined behaviour as well. As I mentioned above, this still doesn't limit users from recreating a pretty detailed history in a well-controlled environment.

Just thought I'd point out that this won't handle piped commands... There'd be no permanent record of how the temporary files were generated. 

But generally, I don't think we necessarily need to vastly improve on what we already have. It's a nice bonus, and if someone wants to improve it, they'd be more than welcome. But this has to remain a very low priority issue...

> This doesn't just apply to command-history, but more generally to any header key-value entry in the event of a command that takes multiple input files. I've encountered this a couple of times recently (e.g. this commit); I've also dealt with a similar issue in tckedit. There should perhaps be a helper function somewhere for generating the output header key/value set based on input headers? 

As far as I can tell, the correct action here is by necessity command-specific, so I don't think we can generalise enormously. Helper functions may help for common cases, like the one you highlighted in `mrcalc` - i.e. when multiple inputs are equivalent, you might elect to preserve the DW scheme or PE scheme if they match, and remove them otherwise. But you probably don't want to do that if one of the images is e.g. a mask. So for instance, as far as I can tell, your change to `mrcalc` would remove the DW scheme from a set of DW images if we did something innocuous-looking like:
~~~
$ mrcalc dwi.mif mask.mif -mult dwi_masked.mif
~~~
But then it's almost impossible to guarantee correct behaviour in general, since simply swapping the inputs would result in no DW scheme anyway if it wasn't already there in the `mask.mif` image... Personally I reckon I'd stick with the previous behaviour (I hadn't realised you'd made that change), and maybe document that most of the header information is gathered from the first image on the command-line - there's too much scope for unexpected problems with the change you made, I think...

Lestropie:
> Just thought I'd point out that this won't handle piped commands... There'd be no permanent record of how the temporary files were generated.

Not trivially anyway. Technically the temporary file's header entry should contain within it the segment of the piped command used to generate that temporary file, so theoretically it might be possible to correspondingly stitch the full piped command back together; just need to differentiate between input & output pipes. Even with [non-linear pipe arrangements](http://mrtrix.readthedocs.io/en/latest/getting_started/command_line.html#really-advanced-pipeline-usage), you could detect the fact that the command-line argument is a temporary file rather than a -`-`' and substitute accordingly.

> So for instance, as far as I can tell, your change to `mrcalc` would remove the DW scheme from a set of DW images ...

Nope: only erases if *both* images contain a DW scheme, and they don't match. A mask image shouldn't contain a DW scheme (it should be in `prior_dw_scheme`).

It does however mean that the DW scheme will be retained if the DWI is the first argument, whereas it'll be absent if the mask is the first argument. Hence the suggestion for more "advanced" handling of header "consistency / variance", that may be sufficiently complex to justify embedding in helper functions.

jdtournier:
> Technically the temporary file's header entry should contain within it the segment of the piped command used to generate that temporary file, so theoretically it might be possible to correspondingly stitch the full piped command back together; just need to differentiate between input & output pipes. Even with non-linear pipe arrangements, you could detect the fact that the command-line argument is a temporary file rather than a --' and substitute accordingly.

Yes, that might be possible. Just to make sure I understand this right, the idea would be to store the full multi-stage pipeline command in the header as a single `command_history` entry, so that for a set of commands like my previous example:
~~~ShellSession
$ dwi2mask dwi.mif mask.mif
$ dwi2tensor dwi.mif - | tensor2metric - -mask mask.mif -fa fa.mif
~~~
then `mask.mif` would have a single entry with

    dwi2mask dwi.mif mask.mif`  (version=3.0_RC1-186-g6034fc2f)

while `fa.mif` would have a more complex entry, ideally:

    dwi2tensor dwi.mif - | tensor2metric - -mask mask.mif -fa fa.mif  (version=3.0_RC1-186-g6034fc2f)

And this would be achieved by reading the history from the temporary file produced by `dwi2tensor`, detecting it as a temporary input, and prepending the corresponding entry to the `command_history` with a trailing `|`...?

So that would do-able, with a few gotchas. As you mention, handling multiple temporary inputs to the same command is going to be tricky (but then it's a _very_ rarely used feature - might be worth disabling altogether...?). Other minor issue is what to do if the different commands happen to have different version strings - maybe store them explicitly in order at the end, e.g. like this?

    dwi2tensor dwi.mif - | tensor2metric - -mask mask.mif -fa fa.mif  (version=3.0_RC1-185-g2b38f9a 3.0_RC1-186-g6034fc2f)

And ultimately, this still suffers from the problem I raised before: if we want this to happen transparently, we need to gather the information about all the input headers _before_ creating the output header. I reckon it would be sufficient to ensure all inputs are read before any output is created - maybe enforced through runtime checks just to make sure...

jdtournier:
>Nope: only erases if both images contain a DW scheme, and they don't match. A mask image shouldn't contain a DW scheme (it should be in prior_dw_scheme).

Neat. Hadn't picked up on that. All good.

> It does however mean that the DW scheme will be retained if the DWI is the first argument, whereas it'll be absent if the mask is the first argument. Hence the suggestion for more "advanced" handling of header "consistency / variance", that may be sufficiently complex to justify embedding in helper functions.

Yes, I can see what you mean. I agree that would be useful, particularly for multi-input single-output apps.

Lestropie:
> ... the idea would be to store the full multi-stage pipeline command in the header as a single `command_history` entry ...

Well, would probably remove the "`_history`" bit if it's only one entry :-P

> And this would be achieved by reading the history from the temporary file produced by dwi2tensor, detecting it as a temporary input, and prepending the corresponding entry to the command_history with a trailing |...?

:+1:

> As you mention, handling multiple temporary inputs to the same command is going to be tricky (but then it's a very rarely used feature - might be worth disabling altogether...?). 

Actually I don't think so. Let's use the example from the documentation:

`$ dwi2tensor /data/DICOM/ - | tensor2metric - -mask $(dwi2mask /data/DICOM/ - | maskfilter - erode -npass 3 - ) -vec ev.mif -fa - | mrthreshold - -top 300 highFA.mif`

The temporary image created by `dwi2tensor` contains:
`dwi2tensor /data/DICOM/ -`

There's then a second temporary image created by the combination of `tensor2metric` & `maskfilter`, which will contain:
`dwi2mask /data/DICOM/ - | maskfilter - erode -npass 3 - `

For `tensor2metric`, `argv` should then contain:
`tensor2metric - -mask mrtrix-tmp-######.mif -vec ev.mif -fa -`

The shell is responsible for taking the standard output of the call encapsulated within `$()` and placing it into the `argv` list. In the case of piped images, that's a temporary image name string. Therefore, the hypothetical code responsible for construction of the command string would be able to tell the difference between an input *piped* image "`-`" (in which case the command string should be prepended with that of the piped image, followed by a "`|`"), and an input *non-piped* temporary image (in which case the name of that file within the command string should be replaced with the command string of the temporary image, surrounded by "`$()`"). Differentiating between *input* piped images and *output* piped images can probably be done using relative position alone.

> Other minor issue is what to do if the different commands happen to have different version strings ...

In `tckedit` I write the value "`variable`" whenever the value of an entry is different between different input files. Realistically, if you're trying to debug an issue, and you see different *MRtrix3* code versions within the single-line command string, you wouldn't be debugging based on the different versions of different commands - you'll be telling the user they have a version mismatch, and to check their installation and then re-generate the result.

jdtournier:
>> As you mention, handling multiple temporary inputs to the same command is going to be tricky (but then it's a very rarely used feature - might be worth disabling altogether...?).
>
> Actually I don't think so. Let's use the example from the documentation:

Yes, I don't mean it's impossible, just messy. And given that I'm probably the only person to have ever used the feature, I think we could just decide to simplify and limit passing of temporary files to pipes only, which would then make this a lot simpler to manage...

Unless this happens to be something you guys use all the time...?

> Differentiating between input piped images and output piped images can probably be done using relative position alone.

That's not an issue - the code needs to know this to figure out whether to push the image name out to `stdout` or delete the file (see how this is done in the [ImageIO::Pipe handler](https://github.com/MRtrix3/mrtrix3/blob/master/core/image_io/pipe.cpp#L45-L58))

PR #963, opened 2017-04-11
mrinfo: -property: print key and indentation to stderr unless -quiet

maxpietsch:
printing multiple header entries via `-property` can be confusing as to which line belongs to which entry. So I've changed mrinfo to write the key to stderr and indent according to the length of the key. The downside is that when stdout is redirected but not stderr, this writes whitespaces and newlines to stderr. Printing to stderr  can be deactivated with `-quiet`.

I found it useful but not sure if you like it. Feel free to ditch it!

```
# human readable
mrinfo dwi.mif -property prior_dw_scheme  -property normalisation_scale_factor
prior_dw_scheme: 0.01986599705,0.4115229389,0.9111828647,0
                 -0.3703421358,-0.3132067544,0.8744988459,2600
                 -0.7907743375,0.3971774229,0.4657531985,1000
                 0.2141400482,0.9015290358,0.3760178684,400
...
                 -0.2827501725,-0.4725152915,-0.8347344723,2600
                 -0.2116865689,0.8488509743,0.4843973781,999.9999999

normalisation_scale_factor: 3.2808873653411865

# original behaviour
mrinfo dwi.mif -property prior_dw_scheme  -property normalisation_scale_factor -quiet
0.01986599705,0.4115229389,0.9111828647,0
-0.3703421358,-0.3132067544,0.8744988459,2600
-0.7907743375,0.3971774229,0.4657531985,1000
0.2141400482,0.9015290358,0.3760178684,400
...
-0.2827501725,-0.4725152915,-0.8347344723,2600
-0.2116865689,0.8488509743,0.4843973781,999.9999999
3.2808873653411865


# stdout redirect
mrinfo dwi.mif -property prior_dw_scheme  -property normalisation_scale_factor > test
prior_dw_scheme:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
normalisation_scale_factor:

head test
0.01986599705,0.4115229389,0.9111828647,0
-0.3703421358,-0.3132067544,0.8744988459,2600
-0.7907743375,0.3971774229,0.4657531985,1000
0.2141400482,0.9015290358,0.3760178684,400
0.4666931091,0.7922772138,0.3930577036,2600
0.1689515733,0.7379504547,-0.6533639815,1000
0.8713410126,-0.4567926614,0.1791795306,2600
...
```

### DIFF
--- a/cmd/mrinfo.cpp
+++ b/cmd/mrinfo.cpp
@@ -169,11 +169,18 @@ void print_properties (const Header& header, const std::string& key, const size_
     const auto values = header.keyval().find (key);
     if (values != header.keyval().end()) {
       auto lines = split (values->second, "\n");
+      const size_t indenterr (App::get_options ("quiet").size() ? 0 : key.size() + 2);
+      if (indenterr)
+        std::cerr << key << ": ";
       std::cout << lines[0] << "\n";
       for (size_t i = 1; i != lines.size(); ++i) {
         lines[i].insert (0, indent, ' ');
+        if (indenterr)
+          std::cerr << std::string(indenterr, ' ');
         std::cout << lines[i] << "\n";
       }
+      if (indenterr)
+        std::cerr << std::endl;
     } else {
       WARN ("no \"" + key + "\" entries found in \"" + header.name() + "\"");
     }
--- a/cmd/mrinfo.cpp
+++ b/cmd/mrinfo.cpp
@@ -169,18 +169,12 @@ void print_properties (const Header& header, const std::string& key, const size_
     const auto values = header.keyval().find (key);
     if (values != header.keyval().end()) {
       auto lines = split (values->second, "\n");
-      const size_t indenterr (App::get_options ("quiet").size() ? 0 : key.size() + 2);
-      if (indenterr)
-        std::cerr << key << ": ";
+      INFO(key + ": ");
       std::cout << lines[0] << "\n";
       for (size_t i = 1; i != lines.size(); ++i) {
         lines[i].insert (0, indent, ' ');
-        if (indenterr)
-          std::cerr << std::string(indenterr, ' ');
         std::cout << lines[i] << "\n";
       }
-      if (indenterr)
-        std::cerr << std::endl;
     } else {
       WARN ("no \"" + key + "\" entries found in \"" + header.name() + "\"");
     }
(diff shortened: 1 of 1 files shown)


jdtournier:
Not sure I'm all that keen on mixing `stdout` and `stderr`... For one, they're typically buffered differently on different systems. On Linux systems, `stdout` is typically line-buffered, whereas `stderr` is typically unbuffered. But this is different on Windows, and may also differ on macOS. But the main point is, it's probably a bit too fragile to be relied upon. 

However, you'll note that I've just added a `-all` option to `mrinfo` in `tag_0.3.16` (commit e60e9ba), which should allow users to see the full unabridged contents of the header, including properties that currently contain too many entries to be displayed by default (this is probably the main reason I use the `-property` option, to be honest). For use in scripts, I don't think we'll get around the fact that you'd need separate invocations to get separate properties. 

If we had to include a header for each property, I reckon we'd be better off pushing it as a comment line to `stdout`, e.g. your example would become:
```bash
$ mrinfo dwi.mif -property prior_dw_scheme  -property normalisation_scale_factor
# prior_dw_scheme
0.01986599705,0.4115229389,0.9111828647,0
-0.3703421358,-0.3132067544,0.8744988459,2600
-0.7907743375,0.3971774229,0.4657531985,1000
0.2141400482,0.9015290358,0.3760178684,400
...
-0.2827501725,-0.4725152915,-0.8347344723,2600
-0.2116865689,0.8488509743,0.4843973781,999.9999999
# normalisation_scale_factor
3.2808873653411865
```
That would still probably clash with use in scripts, so I'd disable that if only one property is being requested. If more than one is requested, it would make sense to enable it since scripts will also struggle to figure out which is which anyway... What do you think?

maxpietsch:
> Not sure I'm all that keen on mixing stdout and stderr...  it's probably a bit too fragile to be relied upon.

Fair enough! 

How about ditching the indentation and changing the stderr output to `INFO (key+":");`? That would not interfere with any existing scripts and is intuitive. It's not as clean as it also shows other INFO output such as reading config files and opening images.

jdtournier:
Sure! Although I'd probably change the message to 'showing property "key"', just in case stderr and stdout get out of sync...

PR #967, opened 2017-04-14
PLY writer for tractography "streamlines"

blezek:
Adds PLY writer to `tckconvert`.  PLY formats are commonly used in visualization software and 3d printing applications.

The conversion adds some command line parameters to control the generation of streamlines.  Of special interest are the `-sides` and `-increment` options to help reduce the number of points generated.  NB: it is very easy to generate **huge** .ply files!

```
  -radius radius
     radius of the streamlines (0.1 default)

  -sides sides
     number of sides for streamlines (9 default)

  -increment increment
     generate streamline points at every (increment) points (1 default)
```


blezek:
Would also like to add a binary version, but would like some guidance from the developers on how to write binary the `mrtrix` way.

jdtournier:
Thanks @blezek - really appreciate the contributions!

I've just got a few comments to make, I'll add those to the relevant lines in the code. I've also made minor edits to your additions, see what you think. 

As to how we write binary in MRtrix, well very much the standard C++ way... The main exception to that is that we often use memory-mapping to handle data files, so that reads & writes are then direct memory access operations (the [File::MMap](https://github.com/MRtrix3/mrtrix3/blob/master/lib/file/mmap.h#L31) class handles this aspect, if you're interested). 

When we don't use memory-mapping (which does turn out to be how the bulk of the binary IO is handled in MRtrix), we just use standard C++ IO streams. The one minor difference (which you've already figured out) is the use of the [File::OFStream](https://github.com/MRtrix3/mrtrix3/blob/master/lib/file/ofstream.h#L37) class to detect whether the file exists and handle the forced overwrite if the `-force` option is provided. 

You can have a look at the [DWI::Tractography::Writer](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/file.h#L352) class, which appends to file in a way that's likely to be much closer to what you need. Look in particular at the [commit()](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/file.h#L311-L325) method for example. 

The other big issue to worry about when writing binary is byte endian-ness. For this, I recommend you use the methods in the [ByteOrder ](https://github.com/MRtrix3/mrtrix3/blob/master/lib/raw.h#L44) namespace if you know the byte order at compile time (as used for example [here](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/file.h#L291-L297)), or the various `fetch()` & `store()` methods in the [Raw ](https://github.com/MRtrix3/mrtrix3/blob/master/lib/raw.h#L100) namespace in the same file. These are template methods, and perform the swapping as needed based on the system byte order and the desired/expected byte order on file. When using these methods, I strongly urge you to use the standard sized types in `<cstdint>` (already included from [types.h](https://github.com/MRtrix3/mrtrix3/blob/master/lib/types.h#L21)), using `uint32_t`, etc. or the MRtrix additions, [`float32`, and `float64`](https://github.com/MRtrix3/mrtrix3/blob/master/lib/types.h#L114-L115), just to make sure the variables sizes genuinely are the expected number of bits. 

I think that just about covers it. But if you have any questions or suggestions, don't hold back... :wink:

cmd/tckconvert.cpp
```diff
@@ -190,6 +200,294 @@ class ASCIIWriter: public WriterInterface<float>
...
+            "end_header\n";
+
+        std::ifstream vertexIF(vertexFilename);
+        out << vertexIF.rdbuf();
+        vertexIF.close();
+        std::ifstream faceIF(faceFilename);
+        out << faceIF.rdbuf();
+        faceIF.close();
+        std::remove ( vertexFilename.c_str() );
+        std::remove ( faceFilename.c_str() );
```
jdtournier: I'd probably use the _MRtrix3_ [File::unlink()](https://github.com/MRtrix3/mrtrix3/blob/master/lib/file/utils.h#L185-L189) call here, but it's not a big deal - this will work just as well.

blezek:
@jdtournier, wondering if you wanted additional changes to this branch before merging? or if it just got lost in the shuffle before ISMRM.  😉

jdtournier:
Ah, sorry, I'd assumed you wanted to also implement the binary version... Happy to merge these changes now if you have no intention of doing this at this stage.

I'll update the branch to match the new `master`, hopefully it won't conflict too badly...

blezek:
The ASCII writer is sufficient for our current needs.  If it changes, I'll add a binary writer.

Thanks!

PR #969, opened 2017-04-18
file.waitfor(): Make more robust

Lestropie:
Resurgence of [this post](http://community.mrtrix.org/t/5ttgen/133/17) made me have another look at the newer solution to the `5ttgen` `run_first_all` problem, and I suspect what I had wasn't good enough.

The theory was to wait until a valid file handle can be obtained for the file the script is waiting for. That way it would happily wait for any SGE jobs to complete, rather than trying to disable SGE altogether. Trouble is, I think that a read handle could be obtained even if some other process had not yet completed writing the file data.

New approach is based on [this post](https://www.calazan.com/how-to-check-if-a-file-is-locked-in-python/): Try to get a file handle that includes write permissions, since that shouldn't be permitted if another process has its own write file handle.

Another option I looked into is `fcntl.flock()`, but it looks like those are merely *advisory* locks...

Haven't tested; would very much appreciate a test from anyone who has SGE up and running on their system. Or if anybody has a better idea for how to achieve a similar result, since there's still an outside chance of a race condition with this one.

### DIFF
--- a/lib/mrtrix3/file.py
+++ b/lib/mrtrix3/file.py
@@ -75,16 +75,15 @@ def newTempFile(suffix):
 #   finished its work)
 # Using open() rather than os.path.exists() ensures that not only has the file
 #   appeared in the directory listing, but the data are there and a file handle
-#   can be obtained. Additionally, opening the file in append mode should fail if
-#   some other process still has that file open for writing.
-# There's a plausible race condition where if the file is deleted between
-#   os.path.exists() and open() then an empty file will be created, but this is
-#   hopefully unlikely to occur in realistic use.
+#   can be obtained. Additionally, an exclusive file lock is obtained.
+# 'r+' mode is used since it includes write access (which may fail if some other
+#   process also has write access to that file), but will _not_ create a new file
+#   if the file is somehow deleted between os.path.exists() and open().
 # Initially, checks for the file once every 1/1000th of a second; this gradually
 #   increases if the file still doesn't exist, until the program is only checking
 #   for the file once a minute.
 def waitFor(path):
-  import os, time
+  import fcntl, os, time
   from mrtrix3 import app
   if not os.path.exists(path):
     delay = 1.0/1024.0
@@ -96,9 +95,11 @@ def waitFor(path):
   if not os.access(path, os.W_OK):
     app.warn('User does not have write access to file \"' + path + '\"; unable to verify file is complete')
     return
-  flags = os.O_RDWR | os.O_EXLOCK
   try:
-    open(path, flags)
+    with open(path, 'rb+') as f:
+      fcntl.lockf(f, fcntl.LOCK_EX)
+      fcntl.lockf(f, fcntl.LOCK_UN)
+    app.debug('File \"' + path + '\" immediately ready')
     return
   except:
     pass
@@ -106,10 +107,12 @@ def waitFor(path):
   delay = 1.0/1024.0
   while True:
     try:
-      open(path, flags)
+      with open(path, 'rb+') as f:
+        fcntl.lockf(f, fcntl.LOCK_EX)
+        fcntl.lockf(f, fcntl.LOCK_UN)
       app.debug('File \"' + path + '\" appears to have been finalized')
       return
-    except IOError:
+    except (IOError, OSError):
       time.sleep(delay)
       delay = max(60.0, delay*2.0)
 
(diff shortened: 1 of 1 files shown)


jdtournier:
I'm not sure I understand the problem enough, but just a suggestion in case it helps: if the problem reduces to figuring out whether the file both exists and is no longer being accessed by any other process, you could use a combination of `os.path.isfile()` and a call to the `fuser` executable ([man page here](https://linux.die.net/man/1/fuser))? This will only work on Unix (looks like it queries the `/proc/` filesystem), but then you'd only expect this to be a problem on Unix anyway, right? I had a quick play with it, a command like `fuser -s thefile` will return a zero exit status if the file is opened by any process, and 1 otherwise (including if the file doesn't exist). Would that do the trick...?

Lestropie:
`fuser` might be a decent option, given `fcntl` isn't cross-platform (I'd need to find corresponding win32 calls to do equivalent), and it's still not clear to me the extent to which other processes will obey `fcntl`'s apparently "advisory" locks.

jdtournier:
Yes, I noticed `fnctl` isn't cross-platform either, but I don't think it's a big deal, since this is trying to deal with a situation that can only occur when FSL and SGE both available, and neither are on Windows (as far as I know...?). 

And as far as I can tell, [advisory locking only works for cooperating processes](https://unix.stackexchange.com/a/147393/166790). All commands that need access to the file need to query and respect the advisory lock. It's essentially useless in this context...

Lestropie:
> ... since this is trying to deal with a situation that can only occur when FSL and SGE are both available, and neither are on Windows ...

It still would have been nice for the library function that provides the solution to be robust outside of the one current use case ¯\_(ツ)_/¯

jdtournier:
And by the way, that trick of opening in append mode doesn't work on Unix (at least as far as I can tell). Multiple processes accessing the same file with write permissions is perfectly allowed on Unix, only on Windows is there a restriction on concurrent write access. I just had a go at this with two concurrent python sessions, having opened the file in append mode first (`f = open ('tmp', 'a')`): no errors, and `fuser -v tmp` happily reports:
```ShellSession
$ fuser -v tmp 
                     USER        PID ACCESS COMMAND
/home/jdt13/tmp:     jdt13      3975 F.... python
                     jdt13      3995 F.... python
```

Lestropie:
OK, had another go; see what you think of that one.

lib/mrtrix3/file.py
```diff
@@ -73,54 +73,69 @@ def newTempFile(suffix):
...
-    do_fcntl = True
-  except ImportError:
-    do_fcntl = False
   from mrtrix3 import app
+
+  def inUse(path):
+    import subprocess
+    from distutils.spawn import find_executable
+    if app.isWindows():
+      if not os.access(path, os.W_OK):
```
jdtournier: This test isn't Windows-specific - I'd run it before the `if app.isWindows():` line.

lib/mrtrix3/file.py
```diff
@@ -73,54 +73,69 @@ def newTempFile(suffix):
...
+        return None
+      try:
+        with open(path, 'rb+') as f:
+          pass
+        return False
+      except:
+        return True
+    if not find_executable('fuser'):
+      return None
+    # Apparently the 'F' for open with write access only appears in the verbose output
```
jdtournier: I'm not sure I'm comfortable looking through the verbose output, there's no guarantee that it won't change. I think it would be safer to rely purely on the exit code. You've already checked that the file exists and is accessible (assuming you go with my previous suggestion of running the `os.access()` check regardless of OS), so that ambiguity is not an issue. Checking the verbose output allows you to test whether the file is opened with write access, but I'm not sure that's necessarily a requirement. You don't expect the file to be opened read-only either, and if another process was accessing it read-only, you'd probably need to wonder why, right...?

lib/mrtrix3/file.py
Lestropie: The "Windows version" of the file completeness test is to try to gain write access to the file - if something else still has write access, it should fail. But if it's impossible to get write access due to permissions, there's no point sitting there in an infinite loop trying to get write access. For non-Windows, whether or not this user has write access to the file doesn't matter, since we never actually intend to write to it.

jdtournier: OK, I get it. 
Still, is it safe to assume that the file has already been tested for existence by this point? This is relevant for my other comment...

Lestropie: `inUse()` isn't called until after `os.path.exists()` has succeeded. So barring an annoying race condition where the file is deleted by some other process in between the two, yes it exists.

Lestropie:
Done run-throughs on Linux and Windows, `waitFor()` reported files being immediately ready, which at least means I didn't get my ones and zeroes mixed up. More thorough testing will need someone with SGE, which is unlikely to get tested here; therefore merging.

Issue #936, opened 2017-03-15
Compile on Linux with Anaconda Python distribution


lgianca:
This is not exactly a bug about MRtrix3 but rather an issue with the configuration script (at least in Linux) when trying to compile release/bin/mrview and release/bin/shview with the anaconda python libraries installed (with the python Qt package). What happens is that the the linker fails to generate mrview and shview due to multiple unresolved dependencies.

In my system (Ubuntu 16.04), the reason was that the configure script gave precedence to the Qt libraries in the anaconda directory rather than the system wide Qt libraries. 

A workaround that worked for me was to change 
```
export PATH="/home/lgianca/anaconda2/bin:$PATH"
```
to 
```
export PATH="$PATH:/home/lgianca/anaconda2/bin" 
```

I hope this helps other people running into the same problem.

jdtournier:
Yes, I had a [fair rant about this](http://community.mrtrix.org/t/error-processing-project-file-tmp-tmpexdmto-qt-pro/288/4?u=jdtournier) on the forum a while ago... Like you said, there's not much we can do from our side. However, what might be helpful is if you could post this information on that same thread on the community forum. I expect​ this will be the first port of call for most people, posting there would probably reach more users. I don't expect many users would want to go through the issues here, they tend to be a little too technical for most people...

lgianca:
Hi jdtournier,
Thank you for pointing out the existing rants about it.
I have posted my workaround to the same thread.

jdtournier:
Great, thanks!

PR #976, opened 2017-05-04
Remove Anaconda from PATH.

dchristiaens:
Let the configure script strip out any mentions of Anaconda python in the PATH. This should fix recurring installation issues on both macOS and Linux (cf. issue #936 and [this](http://community.mrtrix.org/t/compilation-on-linux/123), [this](http://community.mrtrix.org/t/loading-freesurfer-6-mgz-files-crashes-mrview-mrconvert/826/11), [this](http://community.mrtrix.org/t/permission-denied-when-using-mrconvert-and-dwi2response/873/8), or [this](http://community.mrtrix.org/t/error-processing-project-file-tmp-tmpexdmto-qt-pro/288/3) forum post)

### DIFF
--- a/configure
+++ b/configure
@@ -205,6 +205,23 @@ if nogui: report (' [command-line only]')
 report ('\n\n')
 
 
+# remove any mention of anaconda from PATH:
+path = os.environ['PATH']
+if path.endswith ('\\'):
+  path = path[:-1]
+cleanpath = [];
+oldpath = path.split(':')
+for entry in oldpath:
+  if 'anaconda' not in entry.lower():
+    cleanpath += [ entry ]
+if not oldpath == cleanpath:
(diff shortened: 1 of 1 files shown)


jdtournier:
Just in case anyone is wondering what this is for:
```
if path.endswith ('\\'):
  path = path[:-1]
```
See [issue 60 on the GitLab repo](https://gitlab.com/mrtrix/mrtrix3/issues/60) - requoting the relevant section here for reference:

> OK, this is due to the config being stored as a set of python instructions, which just set the relevant variables. When writing out Windows paths, these strings end up full of backslashes, some of which get interpreted by python (\n, \b, \v, ...) when executed. Works OK by making sure the path gets interpreted as a raw string. Also need to make sure the path doesn't end in a backslash since that would escape the final quote character...

PR #979, opened 2017-05-05
Prevent estimation of optimal threshold when user specifies value

blezek:
In the original code, the optimal threshold was always being calculated,
even when the user specified the threshold on the command line.  Now,
check that the threshold flag was not given before calculating the
optimal threshold.  Saves some processing time and prevents the user
from being worried that their value was not being honored.


blezek:
@Lestropie, sorry, emacs and ctags couldn't find `default_type` for me, so I assumed a float.  Fixed now.

PR #981, opened 2017-05-08
docs: removed dirty page from display issues

maxpietsch:
The section `Linux: very slow performance when writing large images` does not make sense in [troubleshooting/Display issues](https://mrtrix.readthedocs.io/en/latest/troubleshooting/display_issues.html) so I moved it to the FAQs.

Edit: Did the same for `Compiler error during build` and `Hanging on network file system when writing images`.

As a side node: I tried to edit it in the doctest branch but it seems to be way behind `master` and has a lot of conflicts after merging with master.


jdtournier:
Fixed `doctest` for you - forced update to your branch... :+1:

PR #982, opened 2017-05-08
Support Python3 scripts with -version option

bjeurissen:
Ensure Python3 compatibility of the new ``-version`` option in the script-API.


jdtournier:
Yes, typically I use `.decode (errors='ignore')` in `configure` and `build`. But I note all of the remaining Python code uses `decode ('utf-8')` - not sure what that means in terms of how tolerant the scripts are going to be with respect to encoding mismatches. I reckon that might be a bigger problem for systems not using either ASCII or UTF8 by default, i.e. European systems set to ISO-8859-1 / Windows-1252. 

But in this instance, we're parsing output from our own app in what we know will be straight ASCII, so there should never be an issue here. I'll merge.

jdtournier:
... any idea why GitHub tells me that 'This pull request is closed, but the `appVersion-py` branch has unmerged commits'...?

jdtournier:
Ignore me - pressed the wrong button...

PR #984, opened 2017-05-08
File::MMap: make sure destructor no longer throws

jdtournier:
As discussed on the forum:
http://community.mrtrix.org/t/permission-denied-when-using-mrconvert-and-dwi2response/873


Lestropie:
It might not be possible to test for non-zero `App::exit_error_code` from *any* code that creates a `File::MMap`, but just wondering if e.g. the `run_queue()` functions should check this variable once they're done? At least try to a reasonable extent to halt code execution on write failure.

jdtournier:
Yes, I was trying to figure out ways of handling this, I'm not sure what the best way forward is... One option might be to provide our own wrapper around `std::shared_ptr` that does allow throwing from the destructor. But somehow I think it would be best to find a way to deal with this that doesn't go against the strong general recommendations to _not_ throw from destructors... 

So this 'fix' catches the exception and prints out an error message, but doesn't otherwise allow the error to propagate back down to the invoking code. My addition of `App::exit_error_code` is an attempt to at least ensure that the next command (in a script, etc) will be aware that something went wrong. But As you figured, it won't stop the command from carrying on as if nothing had happened... The reason I went down that road was that there may be several files to write back, so an immediate abort is not great. I'm not sure how to deal with this. The general idea of [RAII](http://en.cppreference.com/w/cpp/language/raii)  is that the constructor may throw, but the destructor cannot. In our case, there's plenty of scope for things to go wrong in the `File::MMap` destructor, which we don't want to simply ignore. But we can't return anything to indicate failure, and we can't throw an exception... 

The general recommendation I guess with these kinds of issues is to have your resource releasing code encapsulated in a separate function, to be invoked explicitly by the code using that class, and removing any possibility of your destructor throwing. This strikes me as fundamentally unsafe - we can't guarantee that the cleanup function will be called every time, especially when we're already in the process of unwinding the stack while handling another exception. I'll need to think about this some more, see if there's a simple way to deal with this - doubt it somehow...

Lestropie:
Having another look at this as part of my effort to clean up the PR list.

@jdtournier: I also don't see a clean way of dealing with this issue more generally. I do however think we can make inroads by considering the `File::MMap` destructor to be an exceptional case, and working around it as best we can.

The logic in 89c0e44 is, if:

- There are locations in the code where destruction of a `File::MMap` is reasonably frequent; and:
- Those locations are *less likely* to be invoked during stack unwinding (e.g. in some class's destructor) than the `File::MMap` destructor itself,

Then checking the code flagged by an error in the `File::MMap` destructor and throwing an exception there will hopefully result in:

- More frequent termination of commands without necessarily running to completion / some other concomitant error if a `File::MMap` destruction flags an error but the command continues, compared to if this function were not used;
- Less frequent throwing of exceptions during stack unwinding than the current `master` / `dev` code; indeed, hopefully *never* throwing an exception in a destructor (just like this branch prior to my commit) if the locations where this function is invoked never appear in any destructor (as they shouldn't).

It's a bit wishy-washy, but maybe this is the best we can do.

If you want to move this along, I'll rebase to `dev` and resolve.

jdtournier:
Had a quick look into this issue earlier, can't really see anything else we can do... This is already better than what we have currently, so I think it should be merged into the next release.

cmd/tckglobal.cpp
```diff
@@ -75,7 +75,7 @@ void usage ()
   + "Example use: "
 
   + " $ tckglobal dwi.mif wmr.txt -riso csfr.txt -riso gmr.txt -mask mask.mif \n"
-    "   -niter 1e8 -fod fod.mif -fiso fiso.mif tracks.tck "
+    "   -niter 1e9 -fod fod.mif -fiso fiso.mif tracks.tck "
```
jdtournier: Good catch... :clap:

Lestropie: *pats TravisCI on back*

PR #988, opened 2017-05-10
package_mrtrix updated: stripped release, scripts-->bin.

maxpietsch:


package_mrtrix
```diff
@@ -62,7 +61,7 @@ EOF
     )
 
     echo - collating dependent system libraries...
-    LIBS=$(for n in release/bin/*; do ldd $n; done | sed 's/(.*)$//g' | sed 's/^.*=>//g' | sort | uniq | grep -v 'mrtrix\|vdso\|libGL.so\|libEGL.so\|nvidia\|fglrx' | xargs)
+    LIBS=$(for n in bin/*; do ldd $n; done | sed 's/(.*)$//g' | sed 's/^.*=>//g' | sort | uniq | grep -v 'mrtrix\|vdso\|libGL.so\|libEGL.so\|nvidia\|fglrx' | xargs)
```
maxpietsch: Not sure about this line. Not exactly intuitive...

jdtournier: :rofl: agreed...

What it does is list all dynamic dependencies (libraries) used by our executables (`ldd`), makes sure they're only listed once (`sort | uniq`), and strips out libraries that we shouldn't just blindly add (mostly all OpenGL/graphics-related libraries).

Not sure it would still work anyway since `bin/` now also contains scripts. Not sure whether `ldd` would produce output that might interfere in this case...

maxpietsch: For scripts, ldd produces the output "not a dynamic executable" which triggers the warning "not all executables have dynamic libraries. Possibly skipping dependent libraries for some commands. This is expected if you built with -static" 

I'd just get rid of the warning. Alternatively, we could use `file` to filter out scripts.

PR #990, opened 2017-05-11
Allow custom topup options as well

bjeurissen:
To address http://community.mrtrix.org/t/issue-with-epi-correction/887


bjeurissen:
@Lestropie, there is indeed no requirement for the intensities of the topup pair to match with those of the actual data, but I think the problem in http://community.mrtrix.org/t/issue-with-epi-correction/887 is not that the intensities don't match between the topup pair and the actual data (as hypothesised by the user), but that the intensities don't match within the topup pair.

jdtournier:
Pretty sure I'm with @lestropie on this one - prescan normalize is intended to correct for receive coil inhomogeneity...

PR #996, opened 2017-05-19
Remove #include <vector> usages

Lestropie:
Seeing da512aa pop up made me think to check for this.

No idea if it'll have any effect, but maybe there's an outside chance that if nothing has `#include <vector>` in it, trying to use a `std::vector` will cause a compiler error; or maybe one will be able to use `vector` rather than having to resort to `MR::vector`.


jdtournier:
> maybe there's an outside chance that if nothing has #include <vector> in it, trying to use a std::vector will cause a compiler error;

it won't, since our `MR::vector` class derives from `std::vector` directly, so `<vector>` will always be included anyway...

> maybe one will be able to use vector rather than having to resort to MR::vector.

You already can, these latest commits are probably overly cautious - it _should_ work with just `vector<type>` already, and you'll find plenty of examples of that elsewhere in the code. As long as there is no `using namespace std;` anywhere, there shouldn't be any confusion possible. I might remove these `MR::` prefixes if they're going to cause confusion...

jdtournier:
@Lestropie, I've merged this commit into #961 - I assume you're OK with that...

Lestropie:
> You already can, these latest commits are probably overly cautious - it should work with just vector<type> already, and you'll find plenty of examples of that elsewhere in the code. As long as there is no using namespace std; anywhere, there shouldn't be any confusion possible. I might remove these MR:: prefixes if they're going to cause confusion...

Cool, I had assumed that the explicit `MR::vector` namespace addition was necessary for whatever reason in that context. If it's just left over from bug hunting, probably removing them is the best bet.

The `using namespace std;` point is a good one though; `check_memalign` currently won't catch that (ideally it'd flag the presence of `using namespace std;` then check for corresponding `vector<>` usage rather than `MR::vector<>` in that file).

> I assume you're OK with that...

:+1:

jdtournier:
> The using namespace std; point is a good one though; check_memalign currently won't catch that (ideally it'd flag the presence of using namespace std; then check for corresponding vector<> usage rather than MR::vector<> in that file).

Yes, that check is really just a safeguard. To be bullet-proof would require essentially catching the fully-parsed output of the compiler... Not something we can possibly hope to achieve. We could simply catch any use of `using namespace std;` and fail outright though, I think it's fair to say that this is not encouraged within the codebase for basically this reason...

Issue #992, opened 2017-05-17
tckglobal Segmentation fault: Invalid memory access in MH Sampler


LeeReidCSIRO:
Potentially related to #477 
tckglobal using a HCP dataset is producing segmentation faults.

My call is as follows:
`tckglobal /somepath/diffusion/data.mif /somepath/diffusion/response_wm.txt -riso /somepath/diffusion/response_csf.txt -riso /somepath/diffusion/response_gm.txt -mask /somepath/diffusion/brainmask_diff_space.nii.gz -niter 1e9 -fod /somepath/diffusion/fod_global_tractography.mif -fiso fiso.mif -nthreads 8 /somepath/diffusion/tracks_global_tractography.tck`

The output is as follows:

> tckglobal: [100%] preloading data for "/somepath/diffusion/data.mif"
> tckglobal: [100%] uncompressing image "/somepath/diffusion/brainmask_diff_space.nii.gz"
> tckglobal: [  0%] running MH sampler...
> tckglobal: [SYSTEM FATAL CODE: SIGSEGV (11)] Segmentation fault: Invalid memory access

This 'running MH sampler' can get to anything between 0 - 100% before the crash occurs. In one instance the crash occurred at 100%, and an empty track file had been generated.

**With threading disabled (-nthreads = 0) the issue vanishes.** 

I am happy to supply images and the scripts used to generate them if required.

This is not an issue that is critical for me, as I'm returning to iFod2 for other reasons, but I thought I should flag it for the sake of others.

LeeReidCSIRO:
(Upon closer reading I realise that this is highly likely to be the same issue as #477 and so will request 477's reopening)

dchristiaens:
To be absolutely clear: this was a new issue, introduced when #785 was merged into `master`. The cause related to memory alignment and the new row method. This is now fixed in #998.

PR #998, opened 2017-05-24
Fix segfault in global tractography.

dchristiaens:
As reported in #992, the recent merge of #785 accidentally introduced a segfault in `tckglobal`. This pull requests fixes this.

The error was caused by two issues:
- The new `row` method for image access interfered with an out-of-bounds check on the third dimension (DWI volumes) in the data energy computation. This was easily fixed by limiting the bounds check to the spatial dimensions.
- The memory alignment issues required a dedicated aligned allocator in `std::deque` and `std::stack` members of the object pool used to store track segments. This was fixed in 860c474 by setting the appropriate template arguments.

Today's commits clean up the code without affecting performance.

### DIFF
--- a/cmd/tckglobal.cpp
+++ b/cmd/tckglobal.cpp
@@ -74,7 +74,7 @@ void usage ()
   + "Example use: "
   
   + " $ tckglobal dwi.mif wmr.txt -riso csfr.txt -riso gmr.txt -mask mask.mif \n"
-    "   -niter 1e8 -fod fod.mif -fiso fiso.mif tracks.tck "
+    "   -niter 1e9 -fod fod.mif -fiso fiso.mif tracks.tck "
   
   + "in which dwi.mif is the input image, wmr.txt is an anisotropic, multi-shell response function for WM, "
     "and csfr.txt and gmr.txt are isotropic response functions for CSF and GM. The output tractogram is "
--- a/src/dwi/tractography/GT/externalenergy.cpp
+++ b/src/dwi/tractography/GT/externalenergy.cpp
@@ -138,10 +138,14 @@ namespace MR {
         
         void ExternalEnergyComputer::acceptChanges()
         {
+          assert (changes_vox.size() == changes_tod.size());
+          assert (changes_vox.size() == changes_eext.size());
+          assert (!fiso.valid() || changes_vox.size() == changes_fiso.size());
+
           for (size_t k = 0; k != changes_vox.size(); ++k) 
           {
             assign_pos_of(changes_vox[k], 0, 3).to(tod, eext);
-            assert(!is_out_of_bounds(tod));
+            assert(!is_out_of_bounds(tod, 0, 3));
             tod.row(3) = changes_tod[k];
             eext.value() = changes_eext[k];
             if (fiso.valid()) {
@@ -196,7 +200,7 @@ namespace MR {
           if (w == 0.0)
             return;
           assign_pos_of(vox, 0, 3).to(tod);
-          if (is_out_of_bounds(tod))
+          if (is_out_of_bounds(tod, 0, 3))
             return;
           t = w * d;
           for (size_t k = 0; k != changes_vox.size(); ++k) {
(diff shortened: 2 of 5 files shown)


ansk:
I am not able to successfully build `mrtrix` after this merge:

```

ERROR: ( 76/495) [CC] tmp/src/dwi/tractography/GT/particlegrid.o

g++ -c -std=c++11 -pthread -fPIC -DMRTRIX_WORD64 -Wall -O3 -DNDEBUG -Isrc -I./core -Icmd -isystem /usr/include/eigen3 -DEIGEN_DONT_PARALLELIZE src/dwi/tractography/GT/particlegrid.cpp -o tmp/src/dwi/tractography/GT/particlegrid.o

failed with output

In file included from /usr/include/c++/4.9/deque:66:0,
                 from src/dwi/tractography/GT/particlepool.h:18,
                 from src/dwi/tractography/GT/particlegrid.h:26,
                 from src/dwi/tractography/GT/particlegrid.cpp:15:
/usr/include/c++/4.9/bits/deque.tcc: In instantiation of void std::deque<_Tp, _Alloc>::emplace_back(_Args&& ...) [with _Args = {const Eigen::Matrix<float, 3, 1, 0, 3, 1>&, const Eigen::Matrix<float, 3, 1, 0, 3, 1>&}; _Tp = MR::DWI::Tractography::GT::Particle; _Alloc = Eigen::aligned_allocator<MR::DWI::Tractography::GT::Particle>]:
src/dwi/tractography/GT/particlepool.h:51:41:   required from here
/usr/include/c++/4.9/bits/deque.tcc:137:6: error: no matching function for call to std::_Deque_base<MR::DWI::Tractography::GT::Particle, Eigen::aligned_allocator<MR::DWI::Tractography::GT::Particle> >::_Deque_impl::construct(MR::DWI::Tractography::GT::Particle*&, const Eigen::Matrix<float, 3, 1>&, const Eigen::Matrix<float, 3, 1>&)
      this->_M_impl.construct(this->_M_impl._M_finish._M_cur,
      ^
/usr/include/c++/4.9/bits/deque.tcc:137:6: note: candidate is:
In file included from /usr/include/eigen3/Eigen/Core:256:0,
                 from /usr/include/eigen3/Eigen/Geometry:4,
                 from ./core/types.h:63,
                 from ./core/mrtrix.h:36,
                 from ./core/cmdline_option.h:27,
                 from ./core/app.h:27,
                 from ./core/header.h:21,
                 from src/dwi/tractography/GT/particlegrid.h:20,
                 from src/dwi/tractography/GT/particlegrid.cpp:15:
/usr/include/eigen3/Eigen/src/Core/util/Memory.h:729:10: note: void Eigen::aligned_allocator<T>::construct(Eigen::aligned_allocator<T>::pointer, const T&) [with T = MR::DWI::Tractography::GT::Particle; Eigen::aligned_allocator<T>::pointer = MR::DWI::Tractography::GT::Particle*]
     void construct( pointer p, const T& value )
          ^
/usr/include/eigen3/Eigen/src/Core/util/Memory.h:729:10: note:   candidate expects 2 arguments, 3 provided
```

dchristiaens:
Strange. Do you also get the error when building against Eigen 3.3?

dchristiaens:
@ansk, does #999 fix it for you? It's cleaner anyway...

jdtournier:
OK, had a quick look into this. It seems the issue is that Eigen's allocator in the older version is not fully C++11 compliant. To expand:

- an [allocator object](http://en.cppreference.com/w/cpp/memory/allocator) (`Eigen::aligned_allocator` in this instance) needs to provide `allocate()` and `construct()` calls (although the latter is deprecated in C++17). The signature of the `construct()` call itself is however version-dependent (see [this page](http://en.cppreference.com/w/cpp/memory/allocator/construct)). The error you're seing is consistent with the older `construct()` method being used, yet we're compiling in C++11: 
```
/usr/include/eigen3/Eigen/src/Core/util/Memory.h:729:10: note: void Eigen::aligned_allocator<T>::construct(Eigen::aligned_allocator<T>::pointer, const T&) [with T = MR::DWI::Tractography::GT::Particle; Eigen::aligned_allocator<T>::pointer = MR::DWI::Tractography::GT::Particle*]
     void construct( pointer p, const T& value )
          ^
/usr/include/eigen3/Eigen/src/Core/util/Memory.h:729:10: note:   candidate expects 2 arguments, 3 provided
```
In C++11, this call would take a variadic argument passed directly through from the C++11 [emplace_back()](http://en.cppreference.com/w/cpp/container/deque/emplace_back) call, and all would be well (it's supposed to invoke the class's matching constructor, passing it the additional arguments). The older `construct()` call however expects to invoke the _copy_ constructor. So it seems to me that the simplest fix would be to change [this line](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/GT/particlepool.h#L51):
```C++
 51               pool.emplace_back(pos, dir);
```
to:
```C++
 51               pool.emplace_back (Particle (pos, dir));
```
which should work for both versions... (?) I don't think there'll be any performance penalty thanks to [copy elision](http://en.cppreference.com/w/cpp/language/copy_elision).

PR #999, opened 2017-05-24
Move memory aligned deque to core/types.h

dchristiaens:
Hopefully ensures Eigen 3.2 compatibility? (as discussed in #998)

### DIFF
--- a/core/types.h
+++ b/core/types.h
@@ -20,6 +20,7 @@
 #include <complex>
 #include <iostream>
 #include <vector>
+#include <deque>
 #include <cstddef>
 #include <memory>
 
@@ -242,6 +243,21 @@ namespace MR
       public:
         using ::std::vector<X>::vector;
         vector() { }
+    };
+
+  
+  template <typename X, int N=(alignof(X)>::MR::malloc_align)>
+    class deque : public ::std::deque<X, Eigen::aligned_allocator<X>> { NOMEMALIGN
+      public:
+        using ::std::deque<X,Eigen::aligned_allocator<X>>::deque;
+        deque() { }
+    };
+
+  template <typename X>
+    class deque<X,0> : public ::std::deque<X> { NOMEMALIGN
+      public:
+        using ::std::deque<X>::deque;
+        deque() { }
     };
 
 
--- a/src/dwi/tractography/GT/particle.h
+++ b/src/dwi/tractography/GT/particle.h
@@ -74,7 +74,15 @@ namespace MR {
               removeSuccessor();
             alive = false;
           }
+
+          // disable copy and assignment
+          Particle(const Particle&) = delete;
+          Particle& operator=(const Particle&) = delete;
           
+          // default move constructor and assignment
+          Particle(Particle&&) = default;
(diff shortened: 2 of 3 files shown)


ansk:
@dchristiaens, 
with Eigen 3.2.2 I am still getting the same error as in https://github.com/MRtrix3/mrtrix3/pull/998#issuecomment-303790071 . Build with Eigen 3.3.3 is without error.

dchristiaens:
@ansk OK, good to know and thanks for testing. So the issue is that you get a compiler error from the memory alignment handling in a `std::deque` in Eigen 3.2. This pull request makes sure to handle `deque` in completely the same way as `std::vector`, i.e., using a template specialisation that calls into `Eigen::aligned_allocator` when needed.

According to the Eigen docs, this allocator function can be used with any STL container. Hence, I don't understand why you can compile our `vector` classes without issues but not `deque`... 

@jdtournier, any ideas?

dchristiaens:
It may be related to [this](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1164) bug report, which was presumably fixed in 3.2.8, but not sure. It might be a good idea to upgrade your Eigen version anyways if at all possible; there's loads of cool new features in 3.3 that I'm keen to start using. (Although I admit that doesn't really answer your question.)

ansk:
The 3.2.2-3 is default version in Debian Jessie I am using. I downloaded 3.3.3 and compiled against it without problem. However, if requiring 3.3 version, the installation requirements should be updated since actual requirement is >=3.2. http://mrtrix.readthedocs.io/en/latest/installation/linux_install.html

dchristiaens:
You're right, of course, we will update the docs if we can't find a solution to make it work for Eigen 3.2. Like I was saying, I'd be happy if such a solution exists, but I'm out of ideas on this one given that the code is now exactly replicating the existing `vector` infrastructure...

thijsdhollander:
FYI: my current `master` branch (that has been updated this morning to include the commits from the other merged pull request) compiles without any problems.  My current Eigen version on this machine is reported as 3.2.92 in the configure output, which is essentially the current `libeigen3-dev` on Ubuntu 16.04.

I wouldn't start requiring versions beyond that at this point, as this would make installation for a large part of the user base much more challenging.

dchristiaens:
@ansk I've just tested against Eigen 3.2.2 and I can't reproduce your compiler error. I've added a commit to this branch to enforce move semantics for the objects in the pool, which I hope will force the compiler to call the right `construct()` signature. Would you mind giving this a shot? If this doesn't help, we can try @jdtournier's suggested fix in #998 (personally I just think it's ugly).

@thijsdhollander Eigen 3.2.92 is basically a release candidate for 3.3, with all its bells and whistles, and would indeed be perfectly fine for most users. Obviously it would be an easy thing to clarify this in the docs, although installing Eigen is really not that hard 😉

ansk:
@dchristiaens, this version now compiles against Eigen 3.2.2 in my system without error.

dchristiaens:
Excellent, thanks for giving it a try.

PR #1004, opened 2017-05-30
mtbin: fix isfinite check

maxpietsch:
fix error: 

`error candidate template ignored: disabled by 'enable_if' [with _A1=MR::Helper::Value<MR::Image<float> >]`

via cast to float.


thijsdhollander:
Nice catch!

PR #1005, opened 2017-06-02
Improve matlab import code

jdtournier:
Rather than ignoring unknown entries, these changes will now allow these fields to be included as additional fields in the output structure. Should avoid [confusion of this nature](http://community.mrtrix.org/t/matlab-read-mrtrix/874?u=jdtournier)...


bjeurissen:
Good idea! Should we also make it possible to reimport them again using mrtrix_write?

jdtournier:
Hadn't thought of that... I guess we should, do you want to handle this...? :grin:

Issue #1006, opened 2017-06-02
mrview image.mif -info segfaults


maxpietsch:
`mrview dwi_fod.mif -info` segfaults (using asserts):

```
gdb --args mrview dwi_fod.mif -info
GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.2) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from mrview...(no debugging symbols found)...done.
(gdb) r
Starting program: /home/mp14/mrtrix3/bin/mrview dwi_fod.mif -info
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffecad3700 (LWP 3248314)]
mrview: [INFO] reading config file "/etc/mrtrix.conf"...
mrview: [INFO] reading config file "/home/mp14/.mrtrix.conf"...
[New Thread 0x7fffe56a6700 (LWP 3248316)]
[Thread 0x7fffe56a6700 (LWP 3248316) exited]
[New Thread 0x7fffe56a6700 (LWP 3248317)]
mrview: [INFO] GL renderer:  GeForce GTX 760/PCIe/SSE2
mrview: [INFO] GL version:   3.3.0 NVIDIA 378.13
mrview: [INFO] GL vendor:    NVIDIA Corporation
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/debug/vector:409:
Error: attempt to subscript container with out-of-bounds index 1, but 
container only holds 1 elements.

Objects involved in the operation:
    sequence "this" @ 0x0xa53060 {
      type = 
Program received signal SIGSEGV, Segmentation fault.
[...]

PR #1007, opened 2017-06-02
Mrview fix command line parsing

maxpietsch:
fixes segfault and assert failure in command line parsing as described in #1006

### DIFF
--- a/src/gui/mrview/window.cpp
+++ b/src/gui/mrview/window.cpp
@@ -2012,7 +2012,7 @@ namespace MR
             return;
           }
 
-          assert ("shouldn't reach here!" && false);
+          assert (opt.opt->is ("info") or opt.opt->is ("debug") or ("shouldn't reach here!" && false));
         }
         catch (Exception& E) {
           E.display();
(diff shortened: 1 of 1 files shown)


src/gui/mrview/window.cpp
```diff
@@ -2012,7 +2013,7 @@ namespace MR
             return;
           }
 
-          assert ("shouldn't reach here!" && false);
+          assert (opt.opt->is ("info") or opt.opt->is ("debug") or ("shouldn't reach here!" && false));
```
jdtournier: Struggling a little to understand how this even compiles...? What are these 'or' statements doing in there? This is supposed to be C++, not python... 😕

Lestropie: http://en.cppreference.com/w/cpp/keyword/or

Is [C++98](http://www.bezem.de/pdf/ReservedWordsInC.pdf) onward apparently.

maxpietsch: [seems to work](https://qph.ec.quoracdn.net/main-qimg-d2762148238117d097dba8367d42bf33-c)

jdtournier: Wow! I never knew that... Guess you learn new things every day... :wink:

PR #1013, opened 2017-06-06
TW-dFC method

Lestropie:
Fernando wants this included as soon as the most recent implementation is tested and verified. Given it doesn't particularly overlap with existing code, I'm not really concerned about it breaking anything, so I don't think it needs to wait for a tag.

For now, have only gotten it compiling; awaiting testing.


Lestropie:
This is now finalised and tested by Fernando, and he's happy for it to be merged. `dev` is currently unlocked, so I could merge it across, but leaving some opportunity for any comments before I hit the button.

Possible point of contention is the interface, so will explain the logic behind it in case anyone cares:

*Both* the 'static' and 'dynamic' forms of TW-dFC are encapsulated within the one command, "`tckdfc`". Originally, the 'dynamic' version was implemented as command "`tckdynamicmap`" (note intended relation to `tckmap`), while the 'static' version was implemented as a track-wise statistic ("`ends_corr`" = "correlation at endpoints") in `tckmap`, requiring a 4D image as the input "associated scalar image". While this 'worked', it was going to conflict with future changes associated with fixel TWI. For instance: In that branch, responsibility for calculation of streamline TWI factors is wholly passed to `tcksample`, then `tckmap` receives these factors as a 1D vector of data and is responsible for the mapping to voxels only (i.e. does not handle track-wise statistics at all); this future change would however somewhat break the similarity between `tckmap` and `tckdynamicmap`. Having these two 'versions' of TW-dFC within the same command "`tckdfc`", rather than one of them 'hiding' in `tckmap`, will hopefully make it easier to find, means that features like the `-backtrack` option behave consistently across both versions, and it contributes to helping me reduce the scope creep and bloating of `tckmap`.

jdtournier:
OK, can I just check that you're talking about merging to `dev` at present? Or is the intention to merge to `master`? If the latter, I do think that's the kind of modification that should really be left for a tag - it's not a bug fix, and clearly a new feature. I guess if it's _guaranteed_ not to break anyone's workflow, then there's no harm in pushing it to `master`, but given that it won't be publicly announced, it probably won't see a great deal of uptake until we announce it as part of the next tag... (?)

Right now, we should be focusing our efforts on sorting out the distribution problems, and pushing for that full release ASAP... This could then be included in that. I'm hoping we can push the final 3.0 release within a matter of weeks... (or am I being overly optimistic yet again...?)

Lestropie:
This is going to `dev`. Changes to existing files are minimal, but non-zero. Mostly trying to clear my backlog of things that are basically done.

With regards to release, would really benefit from a TODO list. I'm sure I'm not the only one who's unsure about what's best to be doing.

PR #1016, opened 2017-06-08
mtbin: fade bias field to 1.0 far from refined mask

maxpietsch:
Fade the bias field to its mean value (1.0) starting at a certain fraction of the FOV away from the refined mask.

This is to counteract cases where the bias field's cubic function amplifies artifacts far away from the mask. The default value is that the bias field is untouched for voxels 25% of the FOV away from mtbin's mask.

The screenshot below shows the weighting of the bias field vs the mean value (1.0) for default values:

2017-06-08t15_54_45 012264 0100_b8dh5y

Here is the weighting applied with extrapolation further away than 1% of the FOV (producing nonsensical steps in the bias field). I warn if the range where the fading happens is smaller than 4 voxels but I do not check if holes in the refined brain mask are actually bridged:

2017-06-08t16_40_22 209244 0100_aubm4j

### DIFF
--- a/cmd/mtbin.cpp
+++ b/cmd/mtbin.cpp
@@ -22,16 +22,19 @@
 #include "transform.h"
 #include "math/least_squares.h"
 #include "algo/threaded_copy.h"
+#include "algo/threaded_loop.h"
+#include "filter/base.h"
 
 using namespace MR;
 using namespace App;
 
 #define DEFAULT_NORM_VALUE 0.282094
 #define DEFAULT_MAXITER_VALUE 100
+#define DEFAULT_EXTRAPOLATION 0.25
 
 void usage ()
 {
-  AUTHOR = "David Raffelt (david.raffelt@florey.edu.au), Rami Tabbara (rami.tabbara@florey.edu.au) and Thijs Dhollander (thijs.dhollander@gmail.com)";
+  AUTHOR = "David Raffelt (david.raffelt@florey.edu.au), Rami Tabbara (rami.tabbara@florey.edu.au), Max Pietsch (maximilian.pietsch@kcl.ac.uk) and Thijs Dhollander (thijs.dhollander@gmail.com)";
 
   SYNOPSIS = "Multi-Tissue Bias field correction and Intensity Normalisation (MTBIN)";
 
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Hmm, I'm having second thoughts already about my non-negativity proposal... okay, happily forget about that bit for now maybe.  It also amounts to a smoothness constraint that then depends on the field of view.  That's not per se a good thing.

The core of the issue of the bias field itself being so hard to control is actually the use of polynomial basis functions, rather than e.g. splines or something free-form that's iteratively smoothed by a Gaussian (similar to the deformation fields during registration; i.e. elastic regularization).   But it it's not local (yet global), it's going to have these issues.

Well, putting all my money on the population_template and mrregister changes I'm hacking together in my mind for now then. :neutral_face:

thijsdhollander:
Amazing how typing up stuff about this triggers new ideas.  I should really put rubber duckies on my desk or talk more often to my rubber batmans and deadpools.  I've got another idea that may make the mtbin output also much more robust, but I'll allow the idea to mature a bit more over the next week (we've got our retreat on Thursday/Friday/Saturday, which allows me a lot of time to meditate over these things).  So well, to be continued in a week or so.

maxpietsch:
> The core of the issue of the bias field itself being so hard to control is actually the use of polynomial basis functions, rather than e.g. splines or something free-form that's iteratively smoothed by a Gaussian (similar to the deformation fields during registration; i.e. elastic regularization). But if it's not local (yet global), it's going to have these issues.

I agree. I am not sure if it is wise performance-wise but a 3D Gaussian process regression with a radial basis function kernel should do all of that. 

> Well, putting all my money on the population_template and mrregister changes I'm hacking together in my mind for now then.

I think I have a working solution to robustify the centre of mass initialisation. It basically consists of creating a refined mask given a bad mask. I create the mask by calculating a robust z-statistic on voxel intensities in the initial mask followed by the exclusion of voxels with aberrant high or low intensity values (`|z|>2.5`). This is repeated a few times which makes the intensity distribution converge to foreground only:

image

I then remove all but the largest connected component of the mask and optionally fill the holes in the created mask. It seems pretty robust as long as the initial mask contains at least 50% brain and it works on low b-value diffusion weighted images or post-csd images. Here is an example showing the masked input image in the top row and the thresholded image in the bottom row:

image

I think the most sensible thing to do is to make this a mask filter and to add an option to mrregister to use a separate mask for initialisation. This mask could then be created using this algorithm or for the case of multi-tissue data, one could simply use the mtbin refined mask. This should be sufficient to allow a robust initialisation for multi-tissue and single tissue data.

I am also currently porting back some of my robust registration code into the mrreg branch but I need to test it further before I merge it.

thijsdhollander:
I'll close this, since `mtlognorm` solves the original issues.  Note that the normalisation ("bias") field can't be negative anymore, since the whole estimation is done in the log-domain.  This has an important (yet justified) impact on the relative spatial frequencies between lower and higher intensities that can be fit, but once you start to think log-wise, it makes all sense.  We're estimating a multiplicative effect, not an additive one.  The only way to make the problem additive is to make the transition to the log-domain.

PR #1019, opened 2017-06-19
mrconvert: -axes option now also swaps voxel sizes as expected

jdtournier:
Tiny fix...


thijsdhollander:
Looks all right to me.  Quite important though, so I'll merge it already.  It's funny how these things go unnoticed for so long. :open_mouth:

PR #1021, opened 2017-06-20
app.debug(): Enhance

Lestropie:
Little addition to `app.debug()` in the Python libraries to give better localisation of where the information is coming from. Came in handy during my own debugging so thought I'd push it.

Needs testing across OS's and Python versions.


Lestropie:
Tested as working on Linux & Windows, Python2 & Python3 on each. Unlikely to have any issue specific to Mac.

PR #1025, opened 2017-06-23
Tiff image format

jdtournier:
Implements rudimentary TIFF import support. Already described in some detail in #4, so I won't repeat. No export support as yet, but that can be added later.


jdtournier:
Cool, let's merge. I'll create the issue for write support now.

PR #1026, opened 2017-06-23
Tensor2metric sort by abs eigval

jdtournier:
This adds an option to sort eigenvalues by absolute size, even when eigenvalues are negative. Makes life easier for structure tensor analysis, if you must ask...

### DIFF
--- a/cmd/tensor2metric.cpp
+++ b/cmd/tensor2metric.cpp
@@ -80,6 +80,10 @@ void usage ()
             "principal (1) and minor (3) eigenvalues/eigenvectors (default = 1).")
   + Argument ("sequence").type_sequence_int()
 
+  + Option ("abs",
+            "sort eigenvalues in order of decreasing absolute value, rather than actual "
+            "value (the default). This interacts with the -num option above.")
+
   + Option ("modulate",
             "specify how to modulate the magnitude of the eigenvectors. Valid choices "
             "are: none, FA, eigval (default = FA).")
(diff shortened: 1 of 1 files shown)


jdtournier:
Anyone see any issues with merging this branch? It's a trivial addition, and unlikely to affect anyone - assuming it's bug-free, of course...

jdtournier:
> Is it absolutely necessary to have the option of sorting by value v.s. absolute value? Could the behaviour with the -abs option not simply be the one and only behaviour?

Not sure. In the context of DWI, we always expect positive eigenvalues, so that shouldn't matter. Of course, in practice things aren't that simple and we often end up with negative eigenvalues. But I'm not convinced that it would make a great deal of difference in any practical dMRI scenario whichever way we sort. I'd be happy to always sort by absolute eigenvalue, and maybe provide an option to disable if needed...? On the other hand, I'd left it the same as before to make sure this couldn't introduce any differences in anyone's results, and part of me feels that this is the right way to handle this... But realistically, I have to admit that I can't think of any tensor-derived measures in common use that would be affected by this (?).

> I'm not sure if it's clear enough that -abs also applies to the -vector option. Currently it says that -abs applies to -num, and -num says it applies to both eigenvalues and eigenvectors; but the fact that -vector -abs -num 1:3 will sort eigenvectors by absolute eigenvalues is maybe too complex an inference.

Well, maybe we can add a bit more info about that in the `-abs` option. But it's such a rare occurrence that this would make any difference for any reasonable DWI processing, I'm not sure it's worth worrying about...? Obviously this wouldn't matter if we decide to always sort by absolute value anyway...

Lestropie:
When I tested, it was tough to find voxels where `-abs` made a difference, and they were usually noisy / non-brain voxels anyways. Plus the whole point of a tag update is that we're announcing that results may change. Personally I'd simplify and treat it as a fix rather than an option.

jdtournier:
> Personally I'd simplify and treat it as a fix rather than an option.

OK, happy enough with that - I'll make it the normal behaviour and remove the option to modify it. Assuming no-one else objects?

PR #1030, opened 2017-06-27
fixelcfestats: New option -mask

### LINKED ISSUE #987: fixelcfestats: -mask option
Allow masking of which fixels within the template will be assessed; if implemented properly it should be applied before computing fixel-fixel connectivity, and at the subject data import stage.

As discussed [here](http://community.mrtrix.org/t/how-to-restrict-fixelcfestats-to-subset-of-tracts/845).

Lestropie:
Had a go since there were [multiple requests](http://community.mrtrix.org/t/how-to-restrict-fixelcfestats-to-subset-of-tracts/845). Untested, but it's easier for me to get it installed on a server to test if I push to the online repo, so might as well post here straight away in case anybody is curious.

Wasn't a single unique solution. What I've gone with is:

- Despite using `-mask` option, output fixel image will still contain the same set of fixels as the input fixel template.

- When constructing the fixel-fixel connectivity matrix, streamlines may still be assigned to a fixel that is outside the mask, if the streamline is most collinear with that fixel. However connectivity to that fixel will not contribute in any way to the analysis.

- Only fixels within the mask are involved in data smoothing, GLM, CFE.

Benefit of keeping all fixels is that you keep fixel correspondence across different experiments from the same template. You can still see those fixels, but they should disappear as soon as any quantitative value is loaded (should; need to check). Cropping the output would have made the set of fixels within the analysis mask more immediately obvious; but you can just `fixelcrop` after the fact if you want.

Closes #987.

### DIFF
--- a/cmd/fixelcfestats.cpp
+++ b/cmd/fixelcfestats.cpp
@@ -124,7 +124,7 @@ void write_fixel_output (const std::string& filename,
                          const vector<int32_t>& fixel2row,
                          const Header& header)
 {
-  assert (header.size(0) == fixel2row.size());
+  assert (size_t(header.size(0)) == fixel2row.size());
   auto output = Image<float>::create (filename, header);
   for (uint32_t i = 0; i != fixel2row.size(); ++i) {
     output.index(0) = i;
@@ -153,13 +153,15 @@ void run() {
   const std::string input_fixel_directory = argument[0];
   Header index_header = Fixel::find_index_header (input_fixel_directory);
   auto index_image = index_header.get_image<uint32_t>();
+
   const uint32_t num_fixels = Fixel::get_number_of_fixels (index_header);
   CONSOLE ("number of fixels: " + str(num_fixels));
 
   Image<bool> mask;
   uint32_t mask_fixels;
-  // Lookup tables that will map from input fixel index to row number, and
-  //   from row number to fixel index
+  // Lookup table that maps from input fixel index to row number
+  // Note that if a fixel is masked out, it will have a value of -1
+  //   in this array
   vector<int32_t> fixel2row (num_fixels);
   opt = get_options ("mask");
   if (opt.size()) {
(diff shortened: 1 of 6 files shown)


Lestropie:
OK, looks like this is now working as it should. I'll see if I can get one of the users to test as well. Then need to decide whether to go to `master` or a generic `dev` branch (which I've been meaning to create for other things anyway).

Comes with a nice execution speedup thanks to better cache hitrate: Independently of masking, my test dataset went from 8h to 2h :smiling_imp:

jdtournier:
These changes sound good - but I'm not sure I'm in a position to review them properly at the moment... I've added a couple of other reviewers, hopefully someone can just have a look and approve these changes more rapidly than me. 

Impressed and somewhat puzzled by this:

> Independently of masking, my test dataset went from 8h to 2h

Not sure I fully understand how that worked out...

Lestropie:
> Not sure I fully understand how that worked out...

The connectivity matrix is built using a `std::map<uint32_t, float>` per fixel, so that existing connections can be found efficiently. But if `std::map` is implemented using a binary search tree, that's 2 64-bit pointers per 8-byte connection stored; and the memory allocations could end up scattered, depending on how STL manages memory allocation for a `std::map` (or ~500,000 of them at once).

With these changes, during the normalisation / thresholding step the connectivity matrix is converted to a `vector<std::pair<uint32_t, float>>` per fixel, making the data per fixel contiguous and reducing its total size. For the statistical enhancement you only need to loop over the data per fixel; no need for efficient lookup of a particular fixel index.

Time will tell if the speedup is reproducible, I'm only reporting a single timed run; but it makes sense.

jdtournier:
Good point re the tests - this is a pretty central command, it really does need to be protected with a few tests...

jdtournier:
@draffelt - you might want to actually approve these changes if you're happy with them, otherwise Rob can't merge...

draffelt:
I'll approve for now. But was thinking it might be good to check we at least get the same result without the mask as current master.

draffelt:
Or have you done that already Rob?

That is an impressive speed up.

jdtournier:
@draffelt: OK, gotcha. That's a really good suggestion, we should set up some testing now to verify equivalent results pre & post. @Lestropie, any chance you can find the time to generate a few tests?

Lestropie:
I did a pre & post without mask, it wasn't bitwise identical (presumably due to random permutations) but it was only edge fixels here and there that changed for p<0.05. I could run again with pre-generated permutations if you want.

One thing I did change was [this hard-coded value](https://github.com/MRtrix3/mrtrix3/pull/1030/files#diff-72473e45445eec1a3b304d902e7dab86L284), to use the same threshold for the smoothing matrix as the connectivity matrix. For default parameters the change has no net effect though; it just made sense to me that if the user expands or contracts the matrix by tuning the connectivity threshold up or down, the extent of the smoothing matrix should do the same thing.

Lestropie:
With pre-calculated permutations, `fwe_pvalue` images vary by no more than one permutation's worth; p<0.05 masks are bitwise equivalent.

```
$ time /software/mrtrix3/bin/fixelcfestats ...
real    518m12.199s
user    13251m53.299s
sys     195m55.055s

$ time /software/mrtrix_testing/bin/fixelcfestats ...
real    137m26.918s
user    1562m37.928s
sys     83m32.612s
```
:muscle:

I'm veering away from pushing anything other than fixes to `master`, but if we're tagging `3.0_RC2` for #1036 (which I think we should), this can possibly be carried along with that?

jdtournier:
Good to hear. Any chance of running tests to verify operation that don't take 1½ hours...?

> I'm veering away from pushing anything other than fixes to master, but if we're tagging 3.0_RC2 for #1036 (which I think we should), this can possibly be carried along with that?

My thoughts exactly. A new `3.0_RC2` tag would be a good opportunity to add/modify things that we might otherwise have delayed for a future version. I reckon now is the time to create a `3.0_pre_RC2` _branch_, with a view to merging these changes along with the `mtlognorm` branch and some of the other pull requests into it, and we'll then push the lot out to `master` when ready and tag accordingly.

Lestropie:
> Good point re the tests - this is a pretty central command, it really does need to be protected with a few tests...

As an aside: I've been working on various aspects of the stats code lately (got multiple contrasts working!), so one thing I did just today was write a script that generates random dummy data to feed to `vectorstats`. That's the cleanest way to test the core GLM functionality. Testing the higher-order stats commands gets a bit trickier... but I figured it'd be easier to start from the bottom and work my way up.

PR #1034, opened 2017-07-04
mtlognorm

thijsdhollander:
Replaces `mtbin`, fixes all instabilities we encountered before, with bias fields becoming negative, and hence also infinitesimally close to zero in some voxels.  Key to this is working in the log-domain, but the whole algorithm has been redesigned from scratch as well.  It's now naturally/inherently incredibly robust.

Furthermore, "traditional" bias field correction as a preprocessing step has been reintroduced in the MT FBA pipeline.  We started to recently discover that the removal of this was the cause of frequent quality concerns with brain masks generated by `dwi2mask`.  I also noticed that `dwi2response tournier` is in cases more sensitive to this than one might naively expect.  You'll often still get a decent response, but the selected voxels can really end up in different places.  The reason why the response is often still quite excellent as to angular shape and contrast, is that there is a surprising abundance of good voxels in most healthy brains.

But well, at least the gross effects of certain bias fields should really be corrected before computing response functions and brain masks for certain scenarios.

This may all be quite relevant in context of hypothetical review papers on the topic of the FBA pipeline.

I'll try and merge as soon as possible, since there's several internal studies and external collaborators that depend on this urgently.


thijsdhollander:
For future reference: there were also convergence criterium issues with `mtbin`, as well as a problem with the basis functions.  Due to how global intensity normalisation used to be defined in `mtbin`, parts of its fit was also bouncing a bit back and forth between the 2 main blocks of the algorithm.  There's an important difference between the tissue rebalancing and the global intensity normalisation; they should not be mixed up, but well separated to be dealt with correctly.

bin/dwibiascorrect
```diff
@@ -118,7 +118,7 @@ elif app.args.ants:
   run.command('mrconvert mean_bzero.mif mean_bzero.nii -stride +1,+2,+3')
   run.command('mrconvert mask.mif mask.nii -stride +1,+2,+3')
   bias_path = 'bias.nii'
-  run.command('N4BiasFieldCorrection -d 3 -i mean_bzero.nii -w mask.nii -o [corrected.nii,' + bias_path + '] -s 2 -b [150] -c [200x200,0.0]')
+  run.command('N4BiasFieldCorrection -d 3 -i mean_bzero.nii -w mask.nii -o [corrected.nii,' + bias_path + '] -b [150,3] -c [1000x1000,0.0]')
```
thijsdhollander: Takes a tiny bit longer, yet still fast... but greatly improves suppression of non-brain bits in the lower frontal part of the head (nasal cavity and related stuff), which makes life much less difficult for `dwi2mask`.

cmd/mrmodelfield.cpp
```diff
@@ -1,131 +0,0 @@
-/* Copyright (c) 2008-2017 the MRtrix3 contributors.
```
thijsdhollander: Note that, if needed, this can now be done in practice with `mtlognorm` (yet with the new, correct log-domain computations), since `mtlognorm` now also accepts a single "tissue type" (or any other image to be modelled).

thijsdhollander:
We'll attack the old `mtnormalise` soon too, to be as equivalent in theory with `mtlognorm` as possible.  To avoid confusion and align with the new terminology, it'll end up being `mtbalance` or `mtrebalance` or something like that...

jdtournier:
OK, I've just reverted this merge, `master` now points to the commit previous to this merge. These changes are quite intrusive, they invalidate people's existing scripts and analysis pipelines, remove existing functionality, and change our recommendations for one our main pipelines - all on a release candidate. This is not something any developer should be doing unilaterally, this needs a frank and open discussion, and a consensus to be reached. I'm not sure why there was a rush to merge this, ongoing projects that need these changes should be able to use this branch without any problem (?).

To prevent this happening in the future, I've now added further restrictions on `master`: pull requests will now need to be reviewed and approved by other devs before a merge to `master`, on top of the usual TravisCI tests. This will make it more awkward to push out quick bug fixes unfortunately, but now that we are nearing a more formal release cycle, we have to be more careful.

So let's start a new pull request for this, and have a proper conversation. This new approach is no doubt a valuable improvement, but I for one would like to know more about how it works, and I'd like to discuss the rationale for a whole new command to do essentially the same thing (even if it does it better).

thijsdhollander:
Ok, I'm just going to document this once and for all, and I won't be spending much time and words on it.  All intentions are good; but I did want to give some context, because "unilaterally" is quite a strong label, that deserves a bit of nuance.

The above work was done over several weeks with 2 developers knowing every detail about every bit of code, algorithm and maths.  They both have PhD's, one in mathematics and one in engineering (let's call it computer science).  Every detail was discussed mutually between them to quite amazingly great length (whiteboards were filled with *layers* of diagrams and maths), and every exceptional case those 2 PhD's could think of has been considered.  A third person was involved too, and had quite some authority as well as insights to lead to the decision to try and get this fixed ASAP, and go to great lengths to avoid users doing flawed science as well as being confused by command naming; all with the external image of FBA as the main driver.  Nothing was "rushed", as it did take several weeks to do things proper.  So we balanced carefully between fixing a quite important part of FBA ASAP, as well as fixing it "once and for all", to avoid a further avalanche of patches to a bigger problem.  All 3 people in the story are listed as part of the team on mrtrix.org.  Furthermore, back and forth testing, communication and discussion between these and an RA, as well as an external collaborator has been part of the process too.  There were efforts to get other people on the job too, but those people weren't in the possibility to be on the job.  Which is totally fine, but we did do the effort to get them involved.  No secret rushes from our part involved.

So far for nuancing "unilaterally" and "rush". 

That said, I do take it this has caused some upset and discomfort to at least 1 other developer.  I happily apologise for this; it was definitely not the intention of all of the above people to cause this upset and discomfort.  On the contrary, as we were sacrificing a lot of time "for the greater good", rather than what would've been the greater personal benefits in terms of spending our time.

So well, sorry for the upset, from the bottom of our hearts, but let's move on now.  Thanks.

PR #1041, opened 2017-07-07
Silence compilation warnings on GCC7

Lestropie:
Trying to remove a bunch of new compiler warnings that myself and others have gotten since installing GCC7.

- Prevent use of multiplication when assigning to a boolean target, when a TWI target image is boolean.

- When performing oversampling using the reslice filter, compute the sum of image values using a `default_type` rather than the native datatype (which could be a boolean, or could experience integer overflow), then normalise appropriately depending on the native type (again, avoiding use of multiplication if the datatype is boolean).
***Not currently working***: Going gaga trying to figure out why my SFINAE is SNAFU.

- Prevent re-definition of `abs()` for unsigned integer types, which appears to have cropped up in a compiler header file. Some info [here](http://cplusplus.github.io/LWG/lwg-defects.html#2192) regarding the standard, but looks like `gcc` has slipped them in despite not being in the standard.
***Needs testing***: Need to know if `std_abs.h` (and its corresponding `#define _GLIBCXX_BITS_STD_ABS_H`) are appropriate for detecting the prior definition of these functions.

- Updated JSON for Modern C++ header. Though maybe we should consider instead creating a dummy *MRtrix3* GitHub user account in order to make commits that involve importing somebody else's code...?

### DIFF
--- a/core/adapter/reslice.h
+++ b/core/adapter/reslice.h
@@ -27,6 +27,37 @@ namespace MR
   namespace Adapter
   {
 
+
+
+    namespace
+    {
+      // Partial specialisation for boolean value_type in order to avoid compiler
+      //   warning regarding use of multiplication when assigning to a boolean
+      template <typename value_type>
+      typename std::enable_if<std::is_same<value_type, bool>::value, value_type>::type
+      inline normalise (const default_type sum, const default_type norm)
+      {
+        return ((sum*norm) >= 0.5) ? true : false;
+      }
+
+      // Partial specialisation to invoke round-to-nearest when taking an average of integers
+      template <typename value_type>
+      typename std::enable_if<!std::is_same<value_type, bool>::value && std::is_integral<value_type>::value, value_type>::type
+      inline normalise (const default_type sum, const default_type norm)
+      {
+        return value_type(std::round (sum*norm));
+      }
+
+      template <typename value_type>
+      typename std::enable_if<std::is_floating_point<value_type>::value, value_type>::type
+      inline normalise (const default_type sum, const default_type norm)
+      {
+        return (sum * norm);
+      }
+    }
+
+
+
     extern const transform_type NoTransform;
     extern const vector<int> AutoOverSample;
 
@@ -150,7 +181,7 @@ namespace MR
           using namespace Eigen;
           if (oversampling) {
             Vector3 d (x[0]+from[0], x[1]+from[1], x[2]+from[2]);
-            default_type result (0.0);
+            default_type sum (0.0);
             Vector3 s;
             for (int z = 0; z < OS[2]; ++z) {
               s[2] = d[2] + z*inc[2];
@@ -159,12 +190,11 @@ namespace MR
                 for (int x = 0; x < OS[0]; ++x) {
                   s[0] = d[0] + x*inc[0];
                   if (interp.voxel (direct_transform * s))
-                    result += interp.value();
+                    sum += interp.value();
                 }
               }
             }
-            normalise (result);
-            return result;
+            return normalise<value_type> (sum, norm);
           }
           interp.voxel (direct_transform * Vector3 (x[0], x[1], x[2]));
           return interp.value();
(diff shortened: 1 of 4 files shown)


Lestropie:
```
error: prototype for 'typename std::enable_if<std::is_same<typename MR::Adapter::Reslice<Interpolator, ImageType>::value_type, bool>::value, void>::type MR::Adapter::Reslice<Interpolator, ImageType>::normalise(MR::default_type&) const' does not match any in class 'MR::Adapter::Reslice<Interpolator, ImageType>'
error: candidate is: void MR::Adapter::Reslice<Interpolator, ImageType>::normalise(MR::default_type&) const
```

It's as though the `enable_if` isn't being resolved before it looks for function prototype matches...?

core/adapter/reslice.h
```diff
@@ -183,10 +186,37 @@ namespace MR
         default_type from[3], inc[3];
         default_type norm;
         const transform_type transform_, direct_transform;
+
+        void normalise (default_type&) const;
```
jdtournier: You could rewrite this as a small external helper function, templated on your `value_type`:
~~~C++
namespace { // in anonymous namespace to avoid any chance of external conflict
  template <typename X> inline void normalise (default_type&) const;
}
~~~

and invoke the explicit specialisation you're after at line 163:
~~~C++
normalise<value_type> (result);
~~~
That should simplify your SFINAE somewhat...? Also, you could have a default implementation, a specialisation for `bool`, and a SFINAE specialisation for all integer types - I think C++ would allow for that? The only catch would be to make sure that the SFINAE specialisation doesn't also catch `bool` - but you already check that anyway...

jdtournier:
Just a couple of comments, since I've seen these warnings come up too.

> Prevent re-definition of abs() for unsigned integer types. 

The issue you link to is slightly different, and relates to the fact that two headers include conflicting versions of the same `abs` call: one returns a `double`, the other returns an integer of the same type as the input. But I don't think that's the issue here. I think the warning you're trying to fix is this one:

    "./core/types.h:280:32: warning: optimization attribute on 'uint8_t std::abs(uint8_t)' follows definition but the attribute doesn't match [-Wattributes]"

Interestingly, I only get this on MSYS2 (GCC 7.1.0), but not the latest version of GCC on Arch (7.1.1)...

The reason these calls were added in the first place was because the standard doesn't provide them for unsigned types (because they can't be negative anyway). That's fine, but if you want a generic template class that accepts any type of value, and needs to invoke `abs()` somewhere, the compiler will try to use one of the existing _signed_ versions, with two consequences: one is it might pick a version that can't represent the full range of the input (so a large `uint32_t` might get cast to a `int32_t` and get corrupted because it's beyond the representable range), and the other is that the decision as to which version to use is ambiguous, so it just errors out. So I added these overloads so the compiler could just get on with it when the types passed happened to be unsigned. 

In this case, I think what's happening is that the compiler can see the equivalent definition for _signed_ integers, but they are defined with slightly different compiler attributes. For some reason, it thinks the overloads should have matching attributes. No idea what the issue is really, and there's zero information about this warning that I can find - I think it's a new addition. And it might be a bug in this particular version of GCC, since I don't get on Arch with GCC 7.1.1... 

The other warning I get on GCC 7.1.1 (Arch Linux) is this one:
~~~
./core/math/math.h: In function ‘MR::vector<MR::vector<ValueType> > MR::load_matrix_2D_vector(const string&)’:
./core/math/math.h:147:35: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
       while (getline (stream, sbuf)) {
~~~
I have absolutely no idea why it's getting upset here - there is no pointer dereferencing, only passing by reference. No scope for type-punning that I can see... Wondering whether it's a bug in the latest version of GCC...? Doesn't happen with GCC 7.1.0 on MSYS2... :confused:

core/adapter/reslice.h
Lestropie: Yeah, the anonymous namespace trick should do it. Probably why SFINAE has been done like that in the past.

Lestropie:
> The issue you link to is slightly different, and relates to the fact that two headers include conflicting versions of the same `abs` call

It starts that way, but the ongoing discussion leads into how `abs()` for unsigned integers should actually be handled conceptually. The standard clearly now says that a program invoking such is ill-formed; yet `gcc` appears to be providing functions to do it.

> I think the warning you're trying to fix is this one:

I need to double-check: I thought I was getting an error related to re-definition. Specifically conflicting with [this block](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/std_abs.h#L82-L97). Hence why `#ifndef`'ing got around it.

> `warning: dereferencing type-punned pointer will break strict-aliasing rules`

Argh, I've seen that one before with `getline()`, but now I can't remember how it was fixed ><

Lestropie:
> In this case, I think what's happening is that the compiler can see the equivalent definition for signed integers, but they are defined with slightly different compiler attributes. For some reason, it thinks the overloads should have matching attributes.

OK I think this is correct:

```
In file included from ./core/mrtrix.h:36:0,
                 from ...:
./core/types.h:284:34: warning: optimization attribute on 'uint32_t std::abs(uint32_t)' follows definition but the attribute doesn't match [-Wattributes]
   inline uint32_t abs (uint32_t x) { return x; }
                                  ^
In file included from c:\msys64\mingw64\include\c++\7.1.0\cstdlib:77:0,
                 from...:
c:\msys64\mingw64\include\c++\7.1.0\bits\std_abs.h:56:3: note: previous definition of 'long int std::abs(long int)' was here
   abs(long __i) { return __builtin_labs(__i); }
```

I saw the "previous definition" bit, but didn't realise it was referring to a signed definition. I'm tempted to call this a compiler bug too, since it's non-matching optimization attributes on two different functions that happen to have the same name, not a prototype / definition pair.

jdtournier:
I agree that both the 'previous definition' and 'dereferencing type-punned pointer will break strict-aliasing rules' warnings are probably compiler bugs that'll hopefully go away at some point. They don' happen with `clang++` anyway. Happy for you to merge these changes once the checks pass.

PR #1046, opened 2017-07-11
package_mrtrix: update script

jdtournier:


package_mrtrix
```diff
@@ -72,7 +64,7 @@ EOF
 
     QTCORE=$(for n in $LIBS; do echo $n; done | grep 'libQt.*Core')
     [[ $QTCORE ]] && ( # sort out Qt:
-      QTPLUGDIR=$(strings $QTCORE | sed -n 's/qt_plugpath=\(.*\)$/\1/p')
+      QTPLUGDIR=/usr/lib/qt/plugins/ #$(strings $QTCORE | sed -n 's/qt_plugpath=\(.*\)$/\1/p')
```
maxpietsch: I believe `/usr/lib/qt/plugins/` is `/usr/lib/x86_64-linux-gnu/qt5/plugins/` on ubuntu? QTCORE is /usr/lib/x86_64-linux-gnu/libQt5Core.so.5 and $(strings $QTCORE | sed -n 's/qt_plugpath=\(.*\)$/\1/p') yields `/usr/lib/x86_64-linux-gnu/qt5/plugins`.

jdtournier: Good point. I'd forgotten about this fudge. Problem is that there is no `qt_plugpath` string to be found within my version of the core Qt5 library - my guess they've changed things around, so that approach can't be relied on. This is why I'd hard-coded in the path. I've no idea how to make this robust, other than forcing the user to provide this path as an additional argument/option/environment variable - which isn't great...

maxpietsch: I'd make it an optional option and error out if it can't be detected automatically. The error message could suggest using `/usr/lib/qt/plugins/`. Given the target audience of the script that should suffice?

jdtournier: Sounds good. To be honest, I anticipate it'll only be used by us to prepare binary packages... Eventually I'm hoping neurodebian will pick up the bulk of our Linux packaging load, and this would be used purely to prepare binary releases for those on non-Debian based systems... So not hugely fussed either way.

PR #1048, opened 2017-07-12
Scripts: Image header import

Lestropie:
Proposing to initialise a `dev` branch, where any developments that may influence outcomes / interfaces (and therefore shouldn't go to `master` without an explicit tag) can live. I figure that most developers will probably leave this branch as their base installation so that things get tested a reasonable amount - just like `tag_0.3.16` was for a long period of time. Even though the branch won't be "locked", I would suggest that any reasonable-size changes should still be done with a pull request, but little inconsequential things can be committed straight to the `dev` branch without issue. Comment if you wish.

This particular pull request is the result of my frustration in dealing with importing image header information in Python scripts: having to keep track of what variables are stored as raw text vs. what has been converted to integers / floats, and having to make multiple calls to `mrinfo` for the same image just to get different header fields / key/value pairs.

Instead, the entire header contents can now be imported into a class instance using a single function call. Everything will be stored using appropriate types; e.g. the header transformation is automatically parsed and stored as a 4x4 floating-point array.

Few other bits and bobs, but that's the bulk of it.


Lestropie:
Closing as these changes are now encapsulated in #1098.

PR #1049, opened 2017-07-12
Stash purge

Lestropie:
Another one trying to kick off a dedicated `dev` branch. This is just a bunch of stuff that had filled up my Git stash because of fixing one thing while working on another, but I didn't want to lose when my hard drives finally earn promotions from their current "pre-fail" status.


jdtournier:
OK, let's start using this `dev` branch for upstream changes - although it might be better to name it `upstream`...? Not too worried as long as we all know what it's for.

I've just set up protection for this branch, so pull requests and successful tests are required to merge into it. No reviews required in this case though, in contrast to `master`. Everyone happy with that?

Lestropie:
> ... although it might be better to name it `upstream`...?

`dev` was just the first thing that came to mind. Since we're trying to be more "professional" / "standardised" about the whole thing, if "`upstream`" is more "default" that's fine by me.

> Everyone happy with that?

Fine by me. Not completely against having open access to such a branch (anyone who breaks tests on it is going to get a public lashing anyway), but if you'd rather lock it, it's all good. Just means a little more work to deal with nitty little things.

jdtournier:
@Lestropie: any intention to merge this into `tag_3.0_RC2`...?

Lestropie:
No, deliberately leaving it as `dev` target.

Lestropie:
This started as what I had in my git stash, and then sort of became my dump site of assorted changes and fixes. I've now dealt with the merge after #1098, and am now looking to get all of this into `dev`.

Hard to review since it contains a bit of everything, so I'll try to list potential points of contention / interest:

- `-shell` option in DWI commands changes to `-shells`. Always felt weird filling in "shell" with a comma-separated list.

- `mrinfo` now has `-shellvalues` and `-shellcounts` rather than `-shells` and `-shellcounts`. Partly to separate from the option from the bullet point above, but also because if there's multiple options with a common prefix, it's better to give them all a unique suffix in order to separate them.

- Moved the `tckgen` options specific to the iFOD2 algorithm into their own option group. There's a good chance that future algorithms will also need their own option groups...

- Moved some `tckgen` code from `.h` to `.cpp` files. Not sure if @jdtournier will like it; but many of the functions / constructors are called once only, so it's not like there's execution speed benefits in inlining them. Maybe some of the smaller ones would be better off inlined though...

- Since it looks like a lot of the `5ttgen fsl` failures that have been occurring have been due to people unwittingly down-sampling their T1s, the script now warns if the voxel size is greater than 1mm, and then re-samples the T1 to 1mm just for the FIRST step. Don't think this is the only issue going on with `5ttgen fsl`, but it seems to be a fairly common mistake due to using `flirt` with default options for the registration, so hopefully the bright warnings convince people to check their data.

Anyone is welcome to review / comment on above. Need to hunt down my MIA `labelconvert` test data that's causing the CI to fail before I can merge, so no hurry.

jdtournier:
Haven't looked through all (ahem, any...) of this in any detail, but regarding the changes that I'm explicitly mentioned in:

> - Moved some tckgen code from .h to .cpp files. Not sure if @jdtournier will like it; but many of the functions / constructors are called once only, so it's not like there's execution speed benefits in inlining them. Maybe some of the smaller ones would be better off inlined though...

Seems entirely fair enough to me. These are not template functions, so there's no _need_ to have them in headers. Most of the functions are also heavy enough that the overhead of invocation is negligible compared to the body of the function itself, so I don't anticipate any noticeable loss of performance there. However, some of the calls are more like convenience wrappers around other calls, and these might be better left in the headers - I came across [this one](https://github.com/MRtrix3/mrtrix3/blob/8ddee9b4117460c6b9e513d4cc50a37fe2bc785b/src/dwi/tractography/tracking/method.cpp#L92-L95) on a quick look through, but there may be other similar instances that are worth moving back to the headers. Not sure whether this would actually make any detectable difference to the runtime, but it would make me feel better...

Lestropie:
Yea, I think it was that same one function that came into my memory when I was writing that it might need to be re-checked. I've re-inlined that one, but I reckon it's the only function short enough to justify it.

-----

There's been an ongoing issue with signal handling in the Python code (#858)... Just to explain (even if only for historical purposes):

- In `master`, the signal handling isn't actually working: [This line](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/app.py#L99) fails because the `signal` module isn't imported, and this gets caught & passed by the broad `except` statement. In the vast majority of cases, a Ctrl-C is actually caught [here](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/run.py#L156-L158) by `run.command()`, and the handler invoked directly.

- In `dev`, I fixed the signal handler setup as part of #1098, but this highlighted a new issue: On a Ctrl-C, the handler would get called *twice* (once by Python invoking the signal handler, and again by explicitly calling the handler function from the `try .. except` block in `run.command()`). Removing the explicit handler call in `run.command()`, and relying on catching signals, removes this duplication, but then the handler function itself has all sorts of wacky issues (subprocess threads not terminating, fatal Python errors in the subprocess module if threads weren't shut down, I/O reentry errors, ...).

- In `stash_purge` (this PR), the best fix I could come up with is, inside `run.command()`:

  - Reinstate the default Python handler for `SIGINT`.

  - Catch any `KeyboardInterrupt` exceptions while command is executing, and explicitly call `app.handler()`.

  - Reconfigure the custom signal handler for `SIGINT` once command completes successfully.

It's race condition central, but this is the only way I could figure out that would appropriately handle Ctrl-C on both Python 2 and 3, both with and without `-info` / `-debug`. Fingers crossed it doesn't break anything on Windows...

Lestropie:
This now contains a bunch of fixes for things that have already been pushed to `dev`. So in the absence of any concerns raised / bugs arising overnight I'll hopefully merge this to `dev` tomorrow.

thijsdhollander:
Had a quick look, but couldn't check everything due to how varied all the changes are.  The one thing that I did spot which I'm slightly concerned about is the renaming of the `-shell` option to `-shells`.  To get another one already out of the way: I'm ok with the `-shellvalues` and `-shellcounts` pair to `mrinfo`, that's nice and clear indeed.

I understand the motivation behind changing `-shell` to `-shells`, but I'm not sure it's worth inconveniencing a lot of people (with scripts).  It's also not consistent with a few other options: we've still got `-lmax` (not `-lmaxes` or something like that), or `-stride` (not `-strides`) or a couple of other options that take essentially a vector of values as well, yet their name refers in singular to what *one* of those values would be.  So for that reason, I think a change from `-shell` to `-shells` is not worth it.

However, thinking about (changing) that option's name, there may be an opportunity to change it to something else.  I know this is going to be controversial, but let's give it a shot any way: what about `-bvalues`?  First things first: this doesn't mean the variable names in the back-end would have to change, happy to keep the "shell" terminology for ourselves there; we perfectly know what we're talking about in the context of MRtrix processing.  But for the user, it may definitely help a great deal.  I've seen many users time and time again make the mistake of overlooking the b=0 in context of this option, because they don't see it as a shell.  We can't really blame them for that, since about 99% of the literature uses the same definition for "shell" (b=0 not being one, that is).  I noticed this just as well when I started introducing my SS3T-CSD: I constantly need to (re-) stress that it needs 2 b-values, but that that means single-shell data.
In terms of MRtrix context: that's still compatible with the fact that we cluster b-values together (in "shells").  Even though people may have b-values that slightly vary (in a "shell"), they'd still not refer to them as having acquired distinct b-values (i.e. they'd still say "I've acquired 4 b-values").

So well, there's my attempt at a potentially controversial change then. :slightly_smiling_face:  I'd say, if we do consider changing `-shell` (and inconvenience people), it should be for a worthwhile change.  Just changing `-shell` to `-shells` will create more confusion than it's worth, in my personal opinion...

thijsdhollander:
Just to quickly add, here's for instance one of those users who run into this mistake: http://community.mrtrix.org/t/multi-tissue-multi-shell-fod-and-response-function/1091

jdtournier:
Just adding my 2 cents here:

- changing from `-shell` to `-shells` is unlikely to inconvenience anyone, as long as no other options begin with `-shell` - MRtrix3's option shortening feature will resolve `-shell` just fine. The other changes are likely to cause more trouble though - though not enough trouble to hold back IMO.

- I like @thijsdhollander's suggestion of changing to `-bvalues` though... I agree it is less prone to misinterpretation. If we do follow through, I'd also suggest the equivalent query options to `mrinfo` should also change accordingly.

Lestropie:
"`-strides`" rather than "`-stride`" would actually make sense too, as possibly would some other e.g. `mrinfo` options... would have to go through the list. "`-lmaxes`" not quite so much: It's not a word, so making it plural doesn't work as well. Then again, maybe using either "`-lmax`" or "`-lmaxes`" depending on the nature of the command, rather than trying to have consistency across commands, would actually help in this case? Not sure. Also with `-lmax` it was originally used only in the singular, and even now it's most commonly used in the singular sense, whereas `-shell(s)` was compatible with a list from the outset & is more frequently used in that way. Can create a new topic if we want to explore those possibilities.

Possible disadvantage with "`-bvalues`" is that it could be perceived as setting a *range* rather than a list of clustered values, or be somehow more or less restrictive in its use to select volumes than the current shell-clustering algorithm; the very nature of having "shell" in the option has that implicit message that the DWI volumes must be arranged in shells in order for the option to work, which may make complaints like [this one](http://community.mrtrix.org/t/error-in-dwipreproc-eddy/1164) even more frequent. Not entirely against it, but the justification would need to exceed the script-breaking capacity.

The nature of "single-shell" is always going to be tricky: you only need to look at the changes in `DWI::Shell::select_shells()` and `dwi2response` code to appreciate it's not a simple topic full-stop. So my instinct is that more verbosity in the command / option help is probably more likely to help than changing the option string. Inevitably with SS3T release there'll need to be some explicit documentation on precisely this.

thijsdhollander:
> "-strides" rather than "-stride" would actually make sense too, as possibly would some other e.g. mrinfo options... 

Definitely not against that.  The actual current description even starts with "specify the stride**s** of..."

> "-lmaxes" not quite so much: It's not a word, so making it plural doesn't work as well.

Yeah, no, I wouldn't actually change that; but it was just to point out that the consistency was certainly not there (yet) for a range of other options.  For `dwi2response`, for instance, `-shell` is (by some users) quite often used directly together with `-lmax`, and both lists even relate to each other in that use case.  But I fully agree "lmaxes" is very unwieldy and awkward; not a good thing for a UI.

> Possible disadvantage with "-bvalues" is that it could be perceived as setting a range rather than a list of clustered values,

I do disagree with that, to be honest.  If that were the case, those same users would struggle across the board for a lot of things that can be specified in a vector.  Take for example `mrconvert` with `-coord (axis coord)`.  There's a very clear distinction there between how we specify ranges/sequences (with `:`) versus individual values.  As far as I'm aware, we've got a consistent logic in MRtrix that commas separate individual values, and don't specify ranges.

> or be somehow more or less restrictive in its use to select volumes than the current shell-clustering algorithm; the very nature of having "shell" in the option has that implicit message that the DWI volumes must be arranged in shells in order for the option to work

The documentation can clarify that.  This could even be a good candidate for a documentation page/article of its own, since MRtrix in its entirety is structured around shelled data (with a few exceptions, mostly DTI related).  But more importantly: this shell-clustering happens for most commands (that e.g. have `-shell` as an option) _regardless of the user specifying that option_.  It'd be weird for the name of an option that is often not even specified to have the responsibility of communicating this fact, and even then it's debatable whether the word "shells" really drives that point home for those particular users who didn't realise in the first place that the command/algorithm they were running required shelled / b-value clustered data.  And even then beyond that, if they did feed the wrong data, or there was an issue with the clustering versus the selected b-value, there's often a very descriptive warning or even error message already in place (and we can tune those further to communicate the point in carefully chosen words, and numbers that come from the users' data or their specified options itself, and use that information to point out the mismatch in their case).

Note for instance even in the specific case (on the community forum) that you [link](http://community.mrtrix.org/t/error-in-dwipreproc-eddy/1164) to: the user didn't use or come across the `-shell` option, so there was no way it could've communicated anything to them.  What they *did* get, was this error message:

```
dwi2mask: [ERROR] DWI volumes could not be classified into b-value shells; gradient encoding may not represent a HARDI sequence
```

...which does communicate what's going on (but see how even then, the kind of user that does get in trouble, is quite often not able to figure things out even from such a lengthy message that even includes the word "shells").  Bringing up this particular error message though, it looks like we could improve it.  The wording "HARDI sequence" could be a bit confusing here (although I understand myself what it's trying to say).

> The nature of "single-shell" is always going to be tricky: you only need to look at the changes in DWI::Shell::select_shells()

The latest changes I made there now actually make dealing with that more easy for commands that need to clearly indicate what they require; and via `dwiextract` that functionality is being made directy available to scripts as well.  Single-shell can now be requested independently of (1) requiring, or (2) explicitly not wanting, or (3) not caring about the presence of b=0 images.  But that doesn't even really matter in this discussion, I would say, since it's about the back-end which most end-users never get to see.

> So my instinct is that more verbosity in the command / option help is probably more likely to help than changing the option string.

More verbosity is always a good thing (as long as it doesn't become too unwieldy of course).  But it doesn't stand in the way of making the option name better reflect the terminology that is used in the domain.  Or avoid that it's not reflecting that.  I experienced that first hand when initially naming my method "2S3T-CSD": I've literally had to explain to every single person I presented that stuff to that the "2 shells" in there actually meant single-shell + b=0 data.  Naming it that way initially was probably the biggest mistake I could've made (well, "did make" that is).  Since immediately changing it to "SS3T-CSD" thereafter, I've never had any issues any more.  Even when talking about the requirements nowadays, and I often add "single-shell data ......and b=0 images", I get responses like "oh, of course" right after adding/stressing that latter bit.

All that just to say that about everyone but us (in MRtrix) treats shells as "all your (clustered) b-values apart from the b=0 images".

So while typing that all up, I'm actually becoming more and more convinced that `-bvalues` is effectively the better choice.  Now that we're still just before release, we've got the opportunity to fix this once and for all, so it doesn't come back to bite us later on.

> I like @thijsdhollander's suggestion of changing to -bvalues though... I agree it is less prone to misinterpretation. If we do follow through, I'd also suggest the equivalent query options to mrinfo should also change accordingly.

Yes, that makes sense of course.  It's funny how `-shellvalues` would then become `-bvalues`, but now thinking of it, even that sounds like an improvement.  "shellvalues" sounds in and of itself quite strange... and "b-values" is, well, the word we actually use for that across the domain.

Lestropie:
Maybe "setting a range of b-values" had the wrong connotation. Think this is what I was trying to get at: If your data have some volumes with *b*=2995, some with *b*=3000 and some with *b*=3005, it may be unclear with an option "`-bvalues`" whether or not you need to specify all three values or just one; whereas with "`-shell(s)`" it might be more intuitively clear that those volumes all belong to the same "shell". It's certainly not an outright ambiguity; but I see `-shell(s)` as the less ambiguous of the two options.

> Bringing up this particular error message though, it looks like we could improve it. The wording "HARDI sequence" could be a bit confusing here (although I understand myself what it's trying to say).

Yeah, it's already been revised once I think, I try to revisit these things when users report their confusion despite the error message being technically accurate.

> The latest changes I made there now actually make dealing with that more easy for commands that need to clearly indicate what they require ...

Not arguing to the contrary. What I meant was that it's one thing for a command / script developer to encode what they want/need from input data, and another thing for users to appreciate the nuances of these decisions; for instance in "single-shell" multi-tissue CSD.

> It's funny how `-shellvalues` would then become `-bvalues`, but now thinking of it, even that sounds like an improvement.

Not necessarily: In the case of `mrinfo`, `-bvalues` (in the absence of any description) could be perceived as being "the *b*-values in the image data", which would be "the *b*-value of each individual volume", i.e. effectively a `bvals` file.

The current `-shellcounts` option would also be harder to adjust; it'd need to be e.g. `-volumes_per_bvalue` or the like. Not that I'm a massive fan of the naming of these two options, either as they are in `master` or in this branch, but I don't know that `-bvalues` is any better.

-----

Personally I'm in slight preference of `-shells`, perhaps with additional text in the option help (or a const string that can be added to the `DESCRIPTION` field of commands) describing that in *MRtrix3*, *b*=0 is considered a "shell" and can be added to this list. Happy to go with the consensus though.

Lestropie:
Want to get this merged, if only to reduce my own branch-hopping and allow me to go back to smaller merges with discrete purpose. Are people content with including the `-shell` -> `-shells` change here going to `dev`, but for this & related discussion to remain ongoing in #1147?

jdtournier:
I'd strongly urge you to merge - there's too many disparate fixes in there, it makes it very difficult to review... I'm quite happy with your `-shell` ⇨ `-shells` change, since this won't break any existing scripts (courtesy option shortening). But likewise, this is a discussion that we might want to have on a dedicated branch / PR... 

I think all other changes are OK (?).

In general, I'd encourage people to keep pull requests to a single issue, it gives others a much clearer picture into what's going on. I know this one has an unusual history, given all the other stuff that was going on with `3.0_RC2` at the time, the discussions about the git flow model, etc. - not trying to criticise anyone here, just saying I find it much easier to keep up with small, coherent, single-issue threads... :confused:

Lestropie:
Agree completely; this just kept growing because it started as an unrelated group of things, so when small fixes / improvements were made they inevitably ended up in here rather than creating a new PR each time. Have every intention of not doing the same again.

Issue #1024, opened 2017-06-22
Scripts: Re-print warnings on completion


Lestropie:
Came to me while writing #1023.

Warnings may be issued to the user during Python script execution; either using `app.warn()`, or from within an *MRtrix3* binary. However if the terminal printout is long, the user may not actually see them, despite their redness and boldness. It should be possible to buffer any warnings delivered, and re-print them at script completion to catch the user's attention.

Lestropie:
Minor addendum:

If `FailOnWarn` is set in the config file, additionally terminate the script if a constituent command throws a warning. Apart from consistency with the intent of that flag, this may also help during development to remove erroneous command warnings / make scripts more robust.

Lestropie:
Bumping.

Warnings generated by *MRtrix3* commands could be printed upon completion of that command in addition to repeating them upon completion of the script.

PR #1051, opened 2017-07-14
Dwipreproc alignment

### LINKED ISSUE #874: dwipreproc: First volume alignment issue
Resulting from discussion [here](http://community.mrtrix.org/t/dwipreproc-register-fieldmap-to-dwi-other-questions/621/4).

`eddy` assumes intrinsic alignment between the first of its _b_=0 input volumes, and the first volume that was input to `topup`. There is however a `dwipreproc` usage scenario where this isn't the case: The `-rpe_pair` option (or, more generally, using the `-se_epi` option in 0.3.16). Any mis-alignment between the first volume acquired for susceptibility distortion correction, and the first _b_=0 volume in the DWIs, will result in incorrect inhomogeneity field correction.

There's two ways I can think of that this could be dealt with (that don't require modifying FSL or superseding those commands), but I want to make sure I pick the right one:

### 1. Add first DWI _b_=0 volume to `topup` input

Take the first _b_=0 volume in the DWI series that is to be fed to `eddy`, and concatenate it to the start of the image volumes to be fed to `topup`. That way, when `topup` generates all of its results with respect to this volume, it's intrinsically aligned to the volume that `eddy` subsequently uses as its reference.

* May introduce a bias in `topup`, e.g. estimating the field based on 2 A>>P and 1 P>>A volumes. Could potentially look for this in the phase encoding table and scrub a volume accordingly.

* I _think_ this should intrinsically account for differences in header transform between the single _b_=0 and the other SE EPI volumes, if my assumptions about `topup` and `eddy` are correct, but I'll need to test.

* If FoV size is not identical, would require re-gridding the _b_=0 volume to match the SE EPIs (or the other way around).

* Assumes equivalent TE == equivalent contrast between _b_=0 volume and SE EPIs (which wouldn't be the case for our protocols if I had my way).

### 2. Explicit registration

Perform explicit rigid body registration step to align first _b_=0 volume with first SE EPI volume.

* Would require more exhaustive testing to make sure everything's behaving as it should. E.g. My suspicion is that the b-spline coefficients are derived in image space and then applied in image space with no regard for the header transformations, which means modifying that transformation may do nothing.

* Would be able to deal with differing contrasts between the _b_=0 and SE EPI volumes.

* Registration would be imperfect since the first _b_=0 volume will still be inhomogeneity distorted when the registration is performed, and it's not guaranteed that an SE EPI volume can be found with equivalent distortion. Solving this would require performing iterative registration between the `topup`-corrected brain volume and the `eddy`-corrected volume _during_ `eddy`.

I think Option 1 is winning out having written it down. But open to comments. Also open to help given I'm supposed to be grant writing :-1:

Lestropie:
Closes #874.

Decided I should give this a red hot go finally. Especially with a tag update on the way, which such a change really does need to be associated with... Have I mentioned that I resent `dwipreproc`?

Ideally would subject this sort of thing to a battery of tests. Trouble is, the stumbling points of `dwipreproc` will usually be weird use cases and combinations of code branches I hadn't thought of. Tests can take forever too.

### DIFF
--- a/bin/dwipreproc
+++ b/bin/dwipreproc
@@ -363,19 +363,11 @@ if app.args.se_epi:
       app.error('No phase-encoding information present in SE EPI image header');
 
   # If we are using the -se_epi option, and hence the input images to topup have not come from the DWIs themselves,
-  #   we need to insert the first b=0 DWI volume to the start of the topup input image
-  # Additionally:
-  # - If the input to topup is currently balanced in terms of phase encoding, and insertion of this volume is therefore
-  #   unbalancing the input, and we can find an existing volume in the SE_EPI image with the same phase encoding as the
-  #   first DWI, remove that volume from the input to topup
-  # - TODO If the DWI is defined on a different voxel grid to the SE-EPI image, it will be necessary to re-grid the SE-EPI
-  #   data to match the DWI
-  # - Ideally, it would also be preferable to somehow quantify the overall image contrast between the b=0 image, and the SE-EPI
-  #   images, and instead use a registration-based solution. However I think this can wait for now, given I don't know of
-  #   anybody using such an acquisition.
-  # - TODO The first b=0 volume in the DWIs should also be the first volume of the DWI series; is basically always the case, but
+  #   we need to insert the first b=0 DWI volume to the start of the topup input image. Otherwise, the field estimated
+  #   by topup will not be correctly aligned with the volumes as they are processed by eddy.
+  # TODO Yet to do:
+  # - The first b=0 volume in the DWIs should also be the first volume of the DWI series; is basically always the case, but
   #   it wouuld nevertheless be preferable to test for this
-
   bzero_threshold = 10.0
   if ['BZeroThreshold'] in app.config:
     bzero_threshold = float(app.config['BZeroThreshold'])
@@ -386,17 +378,38 @@ if app.args.se_epi:
     dwi_first_bzero_index += 1
   if dwi_first_bzero_index == len(grad):
     app.error('Unable to find b=0 volume in input DWIs to provide alignment between topup and eddy')
-  run.command('mrconvert dwi.mif dwi_first_bzero.mif -coord 3 ' + str(dwi_first_bzero_index))
+  run.command('mrconvert dwi.mif dwi_first_bzero.mif -coord 3 ' + str(dwi_first_bzero_index) + ' -axes 0,1,2')
   dwi_first_bzero_pe = dwi_manual_pe_scheme[dwi_first_bzero_index] if overwrite_dwi_pe_scheme else dwi_pe_scheme[dwi_first_bzero_index]
+
+  dwi_te = image.headerKeyValue('dwi.mif', 'EchoTime')
+  fmap_te = image.headerKeyValue(topup_in_path, 'EchoTime')
+  if dwi_te and fmap_te and not dwi_te == fmap_te:
+    app.warn('It appears that the spin-echo EPI images used for inhomogeneity field estimation have a different echo time to the DWIs being corrected. '
+             'This may cause issues in estimation of the field, as the first DWI b=0 volume must be additionally used as input to topup for alignment purposes. '
+             'SE-EPI images with different contrast to the DWI b=0 images is not yet supported by dwipreproc.')
+
+  # Since image.match() is very basic, and won't accept an image pair with different numbers of volumes,
+  #   need to get just the first volume of each in order to test
+  run.command('mrconvert ' + topup_in_path + ' first_fmap_volume.mif -coord 3 0 -axes 0,1,2')
+  fmap_needs_regrid = not image.match('dwi_first_bzero.mif', 'first_fmap_volume.mif')
+  file.delTempFile('first_fmap_volume.mif')
+  if fmap_needs_regrid:
+    app.warn('DWIs and SE-EPI images used for inhomogeneity field estimation are defined on different image grids; '
+             'The latter will be autmoatically re-gridded to match the former to facilitate spatial alignment '
+             'between topup and eddy stages')
+
   topup_pe_sum = [ 0, 0, 0 ]
   topup_volume_to_remove = len(topup_pe_scheme)
   for index, line in enumerate(topup_pe_scheme):
     topup_pe_sum = [ i + j for i, j in zip(topup_pe_sum, line[0:3]) ]
     if line[0:3] == dwi_first_bzero_pe[0:3]:
       topup_volume_to_remove = index
-  if (topup_pe_sum == [ 0, 0, 0 ]) and (topup_volume_to_remove < len(topup_pe_scheme):
-    app.console('Balanced phase-encoding scheme detected in series provided via -se_epi option; volume ' + str(topup_volume_to_remove) + ' will be removed and replaced with first b=0 from DWIs')
-    run.command('mrconvert ' + topup_in_path + ' - -coord 3 ' + ','.join([index for index in range(len(topup_pe_scheme)) if not index == topup_volume_to_remove]) + ' | mrcat dwi_first_bzero.mif - topup_in_dwibzero.mif -axis 3')
+  if (topup_pe_sum == [ 0, 0, 0 ]) and (topup_volume_to_remove < len(topup_pe_scheme)):
+    app.console('Balanced phase-encoding scheme detected in \'' + app.args.se_epi + '\'; volume ' + str(topup_volume_to_remove) + ' will be removed and replaced with first b=0 from DWIs')
+    if fmap_needs_regrid:
+      run.command('mrconvert ' + topup_in_path + ' - -coord 3 ' + ','.join([index for index in range(len(topup_pe_scheme)) if not index == topup_volume_to_remove]) + ' | mrtransform - - -interp sinc -template dwi_first_bzero.mif | mrcat dwi_first_bzero.mif - topup_in_dwibzero.mif -axis 3')
+    else:
+      run.command('mrconvert ' + topup_in_path + ' - -coord 3 ' + ','.join([index for index in range(len(topup_pe_scheme)) if not index == topup_volume_to_remove]) + ' | mrcat dwi_first_bzero.mif - topup_in_dwibzero.mif -axis 3')
     file.delTempFile(topup_in_path)
     # Also need to update the phase-encoding scheme appropriately if it's being set manually
     #   (if embedded within the image headers, should be updated through the command calls)
@@ -411,7 +424,10 @@ if app.args.se_epi:
       topup_manual_pe_scheme = new_topup_manual_pe_scheme
   else
     app.console('Unbalanced phase-encoding scheme detected in series provided via -se_epi option; first DWI b=0 volume will be inserted to start of series')
-    run.command('mrcat dwi_first_bzero.mif ' + topup_in_path + ' topup_in_dwibzero.mif -axis 3')
+    if fmap_needs_regrid:
+      run.command('mrtransform ' + topup_in_path + ' - -interp sinc -template dwi_first_bzero.mif | mrcat dwi_first_bzero.mif - topup_in_dwibzero.mif -axis 3')
+    else:
+      run.command('mrcat dwi_first_bzero.mif ' + topup_in_path + ' topup_in_dwibzero.mif -axis 3')
     file.delTempFile(topup_in_path)
     # Also need to update the phase-encoding scheme appropriately
     if topup_manual_pe_scheme:
--- a/bin/dwipreproc
+++ b/bin/dwipreproc
@@ -376,8 +376,7 @@ if app.args.se_epi:
       break
     dwi_first_bzero_index += 1
   if dwi_first_bzero_index == len(grad):
-    # TODO Need to test whether different grids between SE-EPI and DWI will crash eddy in this scenario
-    app.warn('Unable to find b=0 volume in input DWIs to provide alignment between topup and eddy; script will proceed on the assumption that motion between the SE-EPI and DWI series is negligible')
+    app.warn('Unable to find b=0 volume in input DWIs to provide alignment between topup and eddy; script will proceed on the assumption that motion between the SE-EPI and DWI series is negligible. Note however that eddy itself may crash in this scenario.')
   else:
     run.command('mrconvert dwi.mif dwi_first_bzero.mif -coord 3 ' + str(dwi_first_bzero_index) + ' -axes 0,1,2')
     dwi_first_bzero_pe = dwi_manual_pe_scheme[dwi_first_bzero_index] if overwrite_dwi_pe_scheme else dwi_pe_scheme[dwi_first_bzero_index]
--- a/bin/dwipreproc
+++ b/bin/dwipreproc
@@ -299,7 +299,7 @@ else:
 
 # This may be required by -rpe_all for extracting b=0 volumes while retaining phase-encoding information
 import_dwi_pe_table_option = ''
-if dwi_manual_pe_scheme:
+if overwrite_dwi_pe_scheme:
   with open('dwi_manual_pe_scheme.txt', 'w') as f:
     for line in dwi_manual_pe_scheme:
       f.write(' '.join( [ str(value) for value in line ] ) + '\n')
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Ok, so having reviewed the code here, as I understand it:

1. Problem is motion between RPE pair (or other separate SE EPI input) and DWI data to be corrected for EPI distortions.  We should be quite careful what we do/change with/about this scenario, as I reckon this is probably the most common use case nowadays (i.e. acquiring an RPE pair for this purpose).
2. Proposed implemented strategy is looking for the first b=0 image in the DWI data, and add this to the start of the topup input image.  As (I think) I understand it, topup also deals with motion itself, and expresses its output relative to the first input image.  So that should then make the estimated EPI distortion appropriate relative to that first b=0 volume that came from the DWI data.  That in and of itself sounds good to me.  Only uncertainty I've got here: does topup deal correctly with the different/changed balance between PE images?  What I mean is: say I've got a simple RPE pair, as an AP and PA image and the DWI dataset is AP.  The strategy will add an AP image from the diffusion data to the pair, so there's now two AP images and one PA image.  Is this a problem?  I suppose this is dealt with (correctly) somewhere?
3. Concerns/limitations:
   a. Scenarios that aren't eligible for correction (any more): different echo times between SE EPI images and DWI dataset, and DWI datasets without b=0 images.  More generally, the scenario where the SE EPI images don't have an image in the DWI dataset with corresponding contrast, because one of the latter has to be added to the former set.  Does this happen a lot in practice?  And if it does, shouldn't for instance the old strategy be applied, where the user reasons motion is negligible or something?  That's what is the current strategy any way, so at the moment this is already assumed (but it may not per se be good of course).
   b. The strategy may still not be perfect, because the first b=0 volume in the DWI dataset may not be the first volume of the DWI dataset, so not per se the best reference volume, as motion is dealt with in eddy relative to the first volume in the DWI dataset (right?).

So well, I'm not entirely sure about this one.  I'm not against the strategy; it does make sense to deal with it like this.  But my biggest worry would be how common those non-handled cases are (a question to which I don't really know the anwer...).  Also: how big is the impact of (a bit of) motion between the SE EPI images relative to the DWI data on the EPI distortion correction?  EPI distortions are spatially still "relatively" smooth, right?  Everything is relative of course, but I guess I'm trying to get a grip on how big the original problem really is versus the (potential) concerns that the proposed solution may bring.  This is not a negative assessment of the proposed solution: I'm just unsure about a few things that are key to making a good assessment of the proposed solution.

@jdtournier and @dchristiaens , I've added you as (proposed) reviewers.  Not to put time pressure on you guys (I understand if you're busy), but just to hear what are your ideas on this, and whether my semi-assessment makes sense (or to add nuance to it where needed).  If too many people are still undecided about this and we detect it may need more discussion (entirely justified if the case), then I would personally hold off packing it in RC2.  If you both are confident this solution is safe to proceed with already, then by all means, allow it to be merged in.

jdtournier:
I'm not sure I'm the best person to review this - I don't get to process data very often these days, and when I do, it's typically neonatal, so heavily non-standard...
dirgen 30 tmp.txt -power 16 -force && dirstat tmp.txt | grep 'nearest-neighbour angles:' | head -n 1 | awk '{ print $5$8 "," $10 }' | tr ',' ' ' > tmp_stats.txt && testing_diff_matrix tmp_stats.txt dirgen/stats_30_pow16.txt
That said, a few comments:

> does topup deal correctly with the different/changed balance between PE images? What I mean is: say I've got a simple RPE pair, as an AP and PA image and the DWI dataset is AP. The strategy will add an AP image from the diffusion data to the pair, so there's now two AP images and one PA image.

This has come up in discussion before, and I'm not sure what the answer is here. I'd expect it might introduce some bias since the cost function is a simple least-squares on the difference, but I'm not sure I have any particular insights into how much of an impact this might have. My best guess is: not a huge amount - but that's based on gut feeling rather than hard facts...

> Scenarios that aren't eligible for correction (any more): different echo times between SE EPI images and DWI dataset, and DWI datasets without b=0 images. 

Were these ever eligible for correction? I don't see that our scripts ever allowed these kinds of cases...

> The strategy may still not be perfect, because the first b=0 volume in the DWI dataset may not be the first volume of the DWI dataset, so not per se the best reference volume, as motion is dealt with in eddy relative to the first volume in the DWI dataset (right?).

I guess this is what [this comment](https://github.com/MRtrix3/mrtrix3/blob/79cbc104cc17c7c0e523dd01daadaa11aec8cbbb/bin/dwipreproc#L368-L370) is about... I agree it would make sense to at least verify that the first volume is the same b=0 image as was used in the TOPUP stage. I assume this is in the works (?).

Lestropie:
> Only uncertainty I've got here: does topup deal correctly with the different/changed balance between PE images?

*I* deal with this: code from [here](https://github.com/MRtrix3/mrtrix3/pull/1051/files#diff-9a2510352b82dd6f73530274c31a7e1fR401). If the SE-EPI data are "phase-encode balanced", i.e. the sum of the phase encode directions of all volumes is a vector `[0, 0, 0]`, then I *erase* a volume from the SE-EPI image as I insert the first *b*=0 volume, in order to preserve that "balanced-ness". If this is *not* the case in the input image, then I just [say "bugger it"](https://github.com/MRtrix3/mrtrix3/pull/1051/files#diff-9a2510352b82dd6f73530274c31a7e1fR429) and insert the DWI *b*=0 volume as-is.

Exactly how `topup` behaves in the latter scenario, I honestly don't know, will depend on the particulars of the implementation. Could grab all the *b*=0 images from a DWI series and a single reversed phase-encode volume and feed them in and see what happens. But I'm going off the premise that "balanced" = "good", and that failing to preserve such would probably attract criticism.

> ... different echo times between SE EPI images ...

Not aware of anybody doing this presently, but it's actually how I advocated that it should be done at the time, and wanted to do an ISMRM abstract on it, so would have liked to preserve compatibility with such. But nowadays there's other alternatives that would be more worthwhile pursuing.

> ... and DWI datasets without b=0 images.

Again, haven't encountered such myself. The proposed code will error out on this. One possibility would be to add a volume from the SE-EPI series to the start of the DWIs before running `eddy`, then erase it afterwards, on the assumption that `eddy` will appropriately deal with alignment between shells; but given that the method is called "*Post*-`eddy` alignment of shells", it's entirely possible that this *wouldn't* solve the problem, depending on if the realignment re-applies the inhomogeneity field. So I'm not sure I want to further complicate the script to support this, it's getting pretty hard to follow as it is.

> And if it does, shouldn't for instance the old strategy be applied, where the user reasons motion is negligible or something?

Maybe. At least with a warning. Though I worry that warnings are going unnoticed; should [fix that](https://github.com/MRtrix3/mrtrix3/issues/1024).

> The strategy may still not be perfect, because the first b=0 volume in the DWI dataset may not be the first volume of the DWI dataset, so not per se the best reference volume, as motion is dealt with in eddy relative to the first volume in the DWI dataset (right?).

> I guess this is what this comment is about... I agree it would make sense to at least verify that the first volume is the same b=0 image as was used in the TOPUP stage. I assume this is in the works (?).

My interpretation is that `topup` defines the estimated field with respect to the first volume in its input, and `eddy` assumes that this is "intrinsically" aligned to the first volume in *its* input. So yes, if there's motion between the first volume and the first *b*=0 volume, you're screwed; hence the floating `TODO`. Again not an acquisition I've ever seen...

Solution is similar to a previous comment: re-arrange the volumes as they are input to `eddy`, then re-arrange again afterwards. It's just, again, more complexity on a script that's already doing my head in.

> Also: how big is the impact of (a bit of) motion between the SE EPI images relative to the DWI data on the EPI distortion correction?

It was enough to cause a bit of concern on the FSL mailing list, and enough for me to get contacted and harassed about it. Certainly MCRI were concerned since kids have a lot of motion, and inter-protocol motion can often be bigger than within-protocol (i.e. "The scanners's stopped, I can get off this pressure point now"), even with correct instructions from radiographers. 

> EPI distortions are spatially still "relatively" smooth, right?

This is still unclear. With the method I worked on at the time, I was picking up some high-frequency stuff at fluid-tissue interfaces, but it's not clear whether or not it was over-fitting. Looks like `topup`'s resolution maxes out at 4mm, but there can be sharp transitions between those points.

Lestropie:
Scenario where first volume of input DWIs is not a *b*=0 is now handled & tested.

Also permitted processing where the input DWI contains no *b*=0 volumes to proceed with a warning rather than erroring out.

Lestropie:
I've also tested a case where the DWI and SE-EPI images were defined on a different voxel grid, and that completed without issue.

> ... and DWI datasets without b=0 images.

This crashes `eddy`.

thijsdhollander:
Quickly settling the previous concerns that I had about quite exotic scenarios: all good, we probably shouldn't worry too much about those.  But I'm starting to get cold feet about the actual main change of this pull request (replacing an SE EPI image by one from the DWI data):

Ok, I've done a few quick tests with `topup` to try and (de)mystify some things.  The result of this is some concern...  Here goes a quick overview:

Data from a collaborator, so probably can't share the raw data, but the results still make a point.  The data that I got came with a separate pair of AP and PA images, and the DWI data was AP.  The DWI data itself started with a b=0 image, and 3 others (so total of 4 b=0 images) were spread through the acquisition of a single b=2800 shell.  This is quite representative for what I think is the most common scenario nowadays.  Furthermore, there was only very little motion in this case; a pretty "safe" case so as to say.  I ran `topup` with:

1. Just the AP and PA pair, as one would "usually" do, both inherently via `dwipreproc` as on `master`, as well as via the example scenario in the FSL documentation
2. Replicated the AP image 5 times (i.e. exact copies), followed by the PA image.  Goal: test the balance question we had directly.  I.e., "balance due to numbers of volumes"
3. Multiplied the AP image intensities by a factor 5 (but just 1 copy as in the usual scenario), followed by the PA image.  Goal: test a difference balance question, i.e., "balance due to intensity (difference)" and just see how dependent topup is on normalised intensities
4. The AP image, followed by the 4 AP images from the DWI data, followed by the PA image.  That's the same balance (due to numbers of volumes) as experiment 2, but then with actual different b=0 images.  This experiment (and what it was supposed to check; i.e. just add some different images and a tiny bit of motion versus experiment 2) was flawed in its setup though, since there was another difference that I hadn't anticipated when setting it up.  See below.

So, results:

1. Field map looks normal (i.e. as expected), motion parameters file shows no motion for the first volume (which is just hard coded, the first volume isn't actually checked for motion relative to itself) and a tiny bit for the second (PA) volume.  All makes sense.
2. Produces very close to the same field map as experiment 1.  Motion parameters shows no motion for the first volume, but (very) little motion for all others, even though the next 4 volumes are exact copies of the first volume.  A bit funny, but not problematic.  This confirms that `topup` deals with "balance due to numbers of volumes" correctly.  I was first a bit more naive (in my head) about what `topup` does in practice, but reading a bit about it, this actually makes sense.  This could lead us to feel safe about bringing in all b=0 volumes from the DWI data, so we don't need to throw an SE EPI one out and we can use all data.  But we're not there yet...
3. Same thing: produces very close to the same field map as experiment 1.  I was a bit surprised here, because I (too) would've thought `topup` just used a least squares metric.  But so: it deals well with *gross* global intensity differences.  Motion parameters in line with experiment 1 too.
4. So given the previous experiments' findings, this one should've worked well too.  But it didn't: the field map was definitely obviously different.  Then I noticed that the b=0 images from the DWI data we're somewhere around a factor 1.8 ~ 2 different in intensity range compared to the SE EPI pair.  But even that shouldn't be an issue, if we combine our successful conclusions from experiments 2 and 3.  But *then* I noticed the contrast was also not exactly the same.  I don't have all the acquisition parameters (nor the expertise to judge this correctly), but: even though the contrast was different, the distortions definitely were *not* different: all those AP images from the DWI data looked very similar in shape (due to distortions) to the AP image from "the pair".  So the pair does, as far as I can tell, a very good job at estimating an appropriate field map for the DWI data, but mixing in b=0 image(s) from the DWI data to compute the field map becomes a disaster...

Just to at least give some visual insight and the confidence that the distortions were the same (and motion was very limited), yet the contrast and intensity level wasn't, here's those input images:

AP from "the pair":
screenshot0000

4 b=0 AP images from DWI data:
screenshot0001
screenshot0002
screenshot0003
screenshot0004

PA from "the pair":
screenshot0005

All the above are shown using the same intensity range / windowing, bottom of the range being 0.  Note the DWI b=0's have lower maximum intensity (e.g. in CSF), but actually slightly *higher* minimum intensities (e.g. in WM).  So the DWI b=0's show less contrast.  Note that (and I've checked this from many different angles) the distortions between the AP from "the pair" versus the APs from the DWI data are very (very, very) similar.

So this is where I'm getting cold feet... as feeding this into the proposed `dwipreproc` in this pull request would go quite wrong (I haven't tested it explicitly, but it follows almost directly from the above experiments now).  This opens up maybe a "bigger" question: how much (and what) should we aspire to automate in `dwipreproc`, and of course by doing so, take quite some responsibility for.  At the moment (`master`) the value of `dwipreproc` is mostly to make interfacing with FSL's tools much easier, and probably most importantly (for us) in line with MRtrix-style of interfacing.  We essentially wrap FSL's tools here to not break a user's MRtrix-style environment and experience (as well as concatenate a few of those tools, and offer more direct ways for the most common "normal" scenarios).  All of these things are of great value, and very (very) much justify the existence of `dwipreproc`.  But what this pull request is proposing does go fundamentally beyond: it automatically makes certain ("design") kinds of decisions for the user.  Of course we can document all that, but it would still happen *automatically*.  I'm starting to wonder if that's not (far) too dangerous...?

To turn this the other way around: if a user would want to do what this pull request implements and offers directly (e.g. let's consider the example of replacing the AP image of their pair by the first b=0 image from the DWI data), it's pretty simple for them to do this (consciously) by `mrconvert ... -coord 3 0` piping into `mrcat ...` with their PA image, right?  And at the moment for even just their AP-PA pair out of the box, they need to do that `mrcat` step already anyway... so concatenating something else and creating this kind of customised setup is pretty simple with the existing `dwipreproc`.  Wouldn't it be safer to leave this design decision with them?  What we're proposing here now goes even beyond the FSL manual's example scenarios (so isn't even explicitly "advised" there, even though I admit it can make good sense).

So well... I don't know, but I'm starting to feel that it might be a better idea to stick with a (very) convenient wrapper-style `dwipreproc` as it is now, and not automate too many things that could be (too) debatable and should maybe be left to the user to decide upon given their own scenario at hand.

thijsdhollander:
Just before I leave the office (where I'm starting to live almost... erm... :sos:), something that just came to mind: a difference that *I think* may have been there between the acquisition of the AP-PA pair versus the DWI data *may* have been multi-band.  The DWI data was _certainly_ acquired multiband, but I'm not sure if the AP-PA pair also was...  I'm far from an expert in how this may affect the image contrast, so I could be talking complete nonsense here in implicitly suggesting that this explains the contrast difference.  But it's certainly not an uncommon case nowadays, since many labs are starting to implement multiband for their DWI acquisition.

Lestropie:
I was going to say that there'd need to be something fundamentally different in the acquisition; ourselves we literally use the diffusion sequence with one b-value of 0, so it's even twice-refocused. But multi-band would indeed have a TR effect; single volume acquisitions would almost certainly *not* be multi-band. The distortions might be different due to differing bandwidth (sounds like the recommendation is to disable GRAPPA with multi-band); this should be captured in the header entries if things are being imported & working as they should, but certainly won't work if using the older `-rpe_*` options.

Could activate this feature only when a command-line option is provided; that'd mean it could be used by those for which it's applicable, but this particular solution may not remain in place down the track, so far from ideal.

We need to carefully consider `dwipreproc` more generally (again): there's all sorts of other things I could capture in header entries and use, and I could try the alternative solution to #874, but the "convenience" of being able to use the older `-rpe_*` options may be opening the door for bad pre-processing. And trying to handle more and more use cases may end up requiring more time and effort than I'm willing to put into wrapping somebody else's methods, but removing functionality that was once provided never goes down well.

Going to have to sleep on this one. Good catch...

Lestropie:
My instinct is to close this and omit from #1052. Any discussion regarding this code / solution specifically can happen here; discussion of alternative possible solution(s) / scope of `dwipreproc` in general to happen in #874.

thijsdhollander:
> My instinct is to close this and omit from #1052.

Agreed.

The core of the "issue" (motion between DWI series and e.g. RPE single volume acquisitions) has definitely been brought to the attention of the FSL developers as well... so while discussion can continue of course, I'd also sit back a bit and see what they come up with; both as part of their software as well as recommendations.  As to our own users: for those that face a scenario where this issue is relevant, I'd say it's still pretty easy for them to take the necessary steps (even via the current `dwipreproc`) to deal with it if they wish (and are themselves certain that their data allows for it correctly).

Lestropie:
Closing this off without merge to clean up the PR list. Not deleting the branch in case there's some parts of the implementation that may be useful when re-attempting #874.

PR #1036, opened 2017-07-05
mtlognorm

thijsdhollander:
I offer `mtlognorm` as a valuable input of a new method to overcome an important issue.  This is a new algorithm, that optimises an inherently different target function.  I don't give permission to put this new method into "mtbin" or under its name.

Beyond that, feel free to do what you want with my input based on several weeks of intensive scientific research and proper data.

I would personally encourage to do away with `mtbin`, as there are important issues with it.  I would personally encourage to update the docs accordingly, such as in the way offered by the work we did for this pull request.  I would personally encourage to reintroduce dwibiascorrect in the pipeline any way, because I observed important impact on `dwi2mask` and `dwi2response tournier` in realistic cases.  I would personally encourage to adopt the lower threshold for the template fixel masks that I have suggested as a docs change in this pull request, since 0.2 has been found to be consistently inappropriate in a lot of studies now, and this impact is very important on CFE, which affects other fixels even if they are in the mask at the higher threshold.  @draffelt , as a direct example: Remika used 0.2, as copied from the docs, to reanalyse her data.  Trying to reproduce the existing result we had, *several whole bundles disappeared from the result*, even results from fixels that were still in the fixel mask became not significant.  It may be worthwhile (in the future) adding documentation on the importance of a comprehensive fixel analysis mask, and that it also impacts the power of stats for the fixels that are already in the mask at a given threshold.

The reason that I retained `mtbin` before, but made it provide the warning/error, was to in the mean time alarm users that they should probably switch.  As mentioned before, I have good reasons to not want to label the new method `mtbin` in any case, but on top of that, we hoped to not let this significant change of method go unnoticed.  The challenge with this in practice is that, even if `mtbin` were deleted altogether, the binary will still sit there for users if they update via the "usual" means (`git pull` and `./build`: many users do this on a frequent basis, and they don't come and read our blog every time when they do).  Yes, this _will_ break their potentially automated script, but that was actually the intended point: to force them consciously to fix it up, as their current FBA pipeline is essentially broken, but happily still spits out a wrong/biased result (in many realistic cases, I can back this up with existing very different independent studies where I observed this happening) at the end without any signs of issues.  This is not very good for FBA, in my humble personal opinion.

Issue #1037, opened 2017-07-05
Anisotropic template resolutions


thijsdhollander:
...aren't fun or anything we'd like to have in our studies.  We saw them recently popping up everywhere.  The reason is `mraverageheader` and the changes made to it, that were (I believe) intended to deal with slice angulations larger than 45 degrees.  As I mentioned over email, I understand the issue and the how/why what has happened here, but in context of `population_template`, I don't think it still makes as much sense, especially since non-rigid alignment still happens... and actually even in the absence of that too.  The gist being that we don't know the resolution relative to the anatomy, and it's the latter that defines what can reasonably be encoded in a template without stretching the information content in the data.  I could write up a paper about it, but I don't have the time, and I believe everyone that's into this understands all arguments from all (historical) sides enough now.  Here's a few copy-pastes of my earlier email for added value:

> .... due to mraverageheader obviously.  A very quick inspection shows that it works with "projected resolutions".  Even with the "-resolution max" option, this will happen: it does choose the maximum, but still of projected resolutions, so in any realistic case with slice angulations, you'll get an anisotropic template resolution that is strictly lower than the isotropic (upsampled) resolution of 1.25mm^3 that we typically feed in. ....

....

[...]

PR #1052, opened 2017-07-14
Tag 3.0 RC2

Lestropie:
Constructing a discrete pull request for the new tag. Not just to create the button, but also to collate the changes coming in & going out. I think we'll all agree that time is a priority on this one, so only bits that really need to go into the tag should be listed; need to avoid scope creep.

Incoming:

- #1036 (`mtbin` replacement): Main reason for tag.

- #1030 (`fixelcfestats -mask`): Already merged in.

- #1047 (`dwipreproc` fixes for acquisitions that include repeating all DWIs): Basically done, but I might think about & tweak a couple of things regarding handling of the header entries. Realistically doesn't need to go with a tag since it's an outright bug fix and shouldn't influence outcomes for use cases that were otherwise working, but there's no harm in going along with.

- #1051 (`dwipreproc` handling alignment between `topup` output and `eddy`): Looks like I might have something working, but ideally needs a bit of testing. Would be reasonably valuable if it could be validated in time; distortion correction errors could be substantial if there's a fair bit of inter-protocol motion, so it's an important one to fix.

Other pull requests remain as merging into `dev`.


thijsdhollander:
I was just slightly confused by the way: I had reasoned this pull request would be created from the `dev` branch?  There's a sense of urgency and time priority indeed, but even then I'd reasoned these things with the highest priority would just end up on `dev` first, for them to be pull requested to `master`?  Well, no worries, this is fine too! :+1:

jdtournier:
Just echoing @thijsdhollander here: I was just thinking we'd go with a straight merge of `dev` to `master`. Part of the reason is that `dev` is protected (need pull requests to merge into it), which I think is a good idea as it provides guarantees regarding visibility of features that get merged into it - we will all be notified about features being pushed into `dev`, and I think that's important so nothing slips in that would otherwise elicit a bit of discussion. Given that it's destined to become the next `master`, I think it's a good idea to make sure that we only use `dev` for these kinds of major upstream merges.

More generally, I think it's time we had a discussion about the repo structure and how we manage a stable `master`, and an almost equally stable development branch. I reckon [this structure](http://nvie.com/posts/a-successful-git-branching-model/) would be a good starting point. What we have now (as of yesterday) is a slightly simpler version, although it does require changing how we think about this a touch:

- `master` should now be purely for tagged releases. If we need to release bug fixes fast, they can be pushed directly to it from their own branch (no need for an actual hotfix branch), and that will require an explicit pull request and review process. Whether that should trigger a new tag is up for discussion - there's a good argument to be made that if a bug fix is urgent enough to warrant immediate release on `master`, it's important enough to warrant a new (micro) tag too. Note I'm using the typical convention around version numbers: major.minor.micro.

- `dev` should now be used in the same way that we used `master` till now: bug fixes and new features that have reached some level of maturity, and are in line for inclusion in the next major/minor release. This is more in line with the `release` branch in the post I linked to above (maybe we should rename our `dev` to `release`...?). But I'm not sure there is that much of a distinction between their `develop` and `release` branches in our case - again, up for discussion. 

- otherwise, everything else can reside in feature branches, as is the case currently anyway. 

So that doesn't change a great deal, but I reckon it's probably a good time to discuss this...

Lestropie:
There's a bunch of stuff currently slated to merge to `dev` that are not at all necessary for `3.0_RC2` to go out, and not all are ready to go either. The ones that I've personally got listed going to `dev` I _don't_ want going out: they need to be tested for a longer period of time (and/or I need to get a Python linter set up as part of the testing). We don't want scope creep delaying this update like what happened with `0.3.16`. Long-term I agree with you completely; it's just this particular instance, because the concept of the `dev` branch hasn't settled, that separating the tagged update from `dev` makes sense.

jdtournier:
OK, I hadn't realised what you had in `dev` wasn't destined for release. I'd assumed it was. What is in `dev` that you don't pushing out now, just out of interest? I'd have thought any such features would remain on feature branches until ready...

thijsdhollander:
For now this is all good, I'd personally say.  But I do agree mostly with what @jdtournier has in mind here as to future work flow.  Especially this may in the future be a good guideline:

[about non-very-close-to-finalised stuff on `dev`]:
> I'd have thought any such features would remain on feature branches until ready...

This actually with in the mind the fact that sometimes there may be quite some urgency to releasing something to `master` (not just a bug fix, which can directly go there).  This with indeed the scenario of 0.3.16 in mind, which I hope we all agree wasn't something we'd aspire having again.  You could basically merge new and very finalised features to `dev`, and if at some point one comes in (to `dev`) that is for some reason quite urgent to end up on `master` with a proper minor (not micro) version increment, this can be done relatively efficiently together with other features that have been sitting ready in `dev`.  And if of course some new features start piling up in `dev`, this could naturally create the desire to merge to `master` and tag (minor).  But if very-far-from-finalised things end up on `dev`, these may hold off merges of `dev` to `master` for undesirable lengths of time (echo 0.3.16 experience).  Having very-far-from-finalised things on `dev` is also not really handy to discuss them, as you're getting an avalanche of discussions mixed on the `dev` pull request that may be getting prepared at the time (echo 0.3.16 again).  This is also how things go unnoticed (echo the anisotropic and changed voxel sizes via `population_template` here a bit: I would've loved to discuss that a bit better back at the time where it popped up, but it went under my radar because everything was one big 0.3.16 "mess").

Apart from that: I agree we probably have no need to distinguish between `dev` and `release` as separate branches. IMO, that's going to be overkill for what we are trying to manage here (and bring a lot of unnecessary overhead).  Whether the one branch is named `dev` or `release` then isn't really something I feel strongly about: whatever works.  `dev` is shorter to type; I do like that.  `release` risks sounding like it's the same as the "current" latest stable release (i.e. `master`).  So if I'd have to vote, I'd vote for `dev`.

But with all of this, let me repeat this one thing I do care a lot about again: I like this proposed work flow (it makes good sense), but it has to be flexible enough in practice.  If one of us ever wants to introduce a major feature with some level of urgency (to `master`, but via `dev` of course), this has to be possible.  This can definitely involve discussion, but it doesn't always have to mean everyone was involved from step 1 of the development of the feature (though if it's going to impact others directly, these others are better made aware early on of course).  ...so there's got to be some trust that a feature is taken care of by its developers.  This becomes unavoidable as the software grows, and maybe more people get involved, who can all have a different focus and expertise and (and ranges of kinds of) features they work on.  If introducing the level of indirection (that `dev` is), ends up becoming a practical roadblock that puts unreasonable restrictions on MRtrix' ability to be our vehicle to get methods offered to the users (which is primarily done via `master`) or to even have complete (and relatively fast/flexible) control over "our own" methods and their documentation... then I think I'd have a problem with that.  But if we take people's needs with respect to such things into account while working in a `master` - `dev` - [feature branches] organisational structure, I've got confidence in this working well and to our benefit.

Lestropie:
> No slacking off please...

You're more than welcome to contribute by performing code reviews and testing proposed changes.

thijsdhollander:
> slacking off please...

As in: not forgetting about moving on with the pull request, not per se adding things to it.  We're all pretty busy, so I'm happy with delaying other bits if people can't find the time to code review / test (including myself).  My point is about getting the things that are done to the users.

thijsdhollander:
> You're more than welcome to contribute by performing code reviews and testing proposed changes.

Ok, I've done all I could at the moment on that front now.

Let me clarify again that I'm not trying to push/force those 2 other pull requests through.  But if we detect those will either require (much) more discussion, or people just don't have the time now to do the work/discussion that has to be done on those things to get them 100% finalised, then I personally think the release of the other 2 features shouldn't be held back by it for now.  With "not slacking off", I was referring to "not losing focus" about getting those initial features in users' hands; not to rushing other independent features.

This is sort of the worry I formulated above about flexibility: it's good and entirely ok if whole groups of features are merged as bigger batches in `master` and tagged accordingly (not against that at all), but I'm hoping independent things don't become too artificially dependent on each other.  "Faster" and smaller releases of smaller groups of features are more flexible.  It even allows for a finer granularity of tags; so if a new batch of features breaks something crucially for a user, they can go back to the previous tag without losing the benefit of a lot of independent other (earlier introduced) features.  So by all means, I'm just *trying* to be constructive here.

jdtournier:
OK, just a couple of comments here, to address concerns raised by @thijsdhollander:

The git workflow we're talking about shouldn't have any impact of flexibility - if something needs to be pushed out fast, then it can. The only question is whether it warrants a new release (i.e. increasing the version number to label a clean break) - this should only happen when there is a change in behaviour or interface, as we're talking about here. Also, we've now yet starting using this workflow, so things are a little bit in flux. But if we had been using it, then `dev` should have been clean, usable, and ready for release, and the only addition to it would have been to merge the emergency feature and release. 

In the rare instance when a feature needs to be released ASAP without waiting for `dev` to be ready (something that should basically _never_ happen because no-one should be merging to `dev` if their feature isn't already ready and approved), then there's nothing stopping you from either creating a pull request for that particular feature directly to `master`, or cherry-picking the relevant range of commits from whatever other feature branch is relevant, and creating a fresh pull request to `master` from that - I've often cherry-picked a few commits off one branch to push to `master`, especially when I feel the commits fix a different issue than the main gist of the feature branch then happen to have been pushed to. 

On the other hand, I would welcome a little debate about how people think this should work. There are other models out there, and the one I put forward is just one possibility... But I think it should suit our workflow fairly well. What I'm not sure about is what would be OK to push to `master` directly: if we restrict it to urgent bug fixes only, it's great for stability of the code, but not so good for fast turnaround for more minor bug fixes and enhancements, of the kind we routinely push out in response to reports on the forum. So I think there is an argument for using `dev` to hold compatibility-breaking changes and/or major new features, and pushing everything else to `master` as we normally do. I reckon as people start using the tagged releases more often (by way of downloading pre-compiled packages), pushing changes to `master` will be something we probably won't need to worry about so much - although we will still want to ensure stability as much as possible.

thijsdhollander:
> The git workflow we're talking about shouldn't have any impact of flexibility - if something needs to be pushed out fast, then it can.

:+1: Sounds great.  I'm certainly hopeful it won't impact flexibility in practice, but I'm happy to give it a chance.  It should be possible in principle indeed, but it'll definitely be (at least initially) a challenge, I reckon.

> The only question is whether it warrants a new release (i.e. increasing the version number to label a clean break) - this should only happen when there is a change in behaviour or interface, as we're talking about here. 

Makes full sense.  The minor version increments I was referring to, would be for significantly new features; but I wanted to stress that these can also come on their own, or are at least not required to be part of a "big" batch of features.  I was basically arguing for fast turnaround.

> But if we had been using it, then dev should have been clean, usable, and ready for release, and the only addition to it would have been to merge the emergency feature and release.

Sounds perfect (to my ears).  If we can get something like that going, I'd be very much for it.  This also still allows for a few last minute quick(er) fixes of details that had slipped under the radar of the initial pull request of the feature.  A good example would be the 2 commits I just did, that updated the `dirgen` docs and fixed a couple of typos.  Nothing that would've broken `master` or would've upset anyone at all should they have ended up on `master` already, but I did accidentally pick them up anyway, so the opportunity was still there to get them sorted already.  So for "less urgent" yet in principle 100% finalised features, there's a place on `dev` and actual devs keeping an eye out for small hiccups related to those new features.

> In the rare instance when a feature needs to be released ASAP without waiting for dev to be ready (something that should basically never happen because no-one should be merging to dev if their feature isn't already ready and approved), then there's nothing stopping you from either creating a pull request for that particular feature directly to master, or cherry-picking the relevant range of commits from whatever other feature branch is relevant, and creating a fresh pull request to master from that - I've often cherry-picked a few commits off one branch to push to master, especially when I feel the commits fix a different issue than the main gist of the feature branch then happen to have been pushed to.

Good point, but I'd rather stress the bit between the brackets there.  Actual software bug fixes are an exception, I'd say.  Makes sense to directly create a pull request to `master` there (and after merge potentially tag, inspired by the importance of the fix).  But features, even when relatively urgent, should in principle still quite efficiently (however with appropriate review) flow from feature branch to `dev` and pick up whatever else was already sitting *ready* on `dev` and on to `master`.  Such a scenario would optimally also avoid annoying bigger merges of `master` into `dev` as much as possible: for such bigger (even urgent) new features, you'd work based on `dev` in your feature branch, rather than earlier features sitting ready in `dev` having to work around new features (that may have gone to `master` first) again.  This was a small frustration I personally had throughout the 0.3.16 saga: at times bigger changes did happen to `master`, sometimes quite important ones, yet some particular ones either piled up or were on their own already a pain to merge into 0.3.16 again.  Or 0.3.16 had already taken a different route to addressing (part of) the issue, whereas an independent feature branch went off on its own to master.  So well, this all as an encouragement to go through `dev` as much as possible (I reckon); but for this to be feasible even in scenarios where flexibility and efficiency is needed to get features flowing through to `master` relatively fast (if needed), `dev` definitely has to be in a "ready" state at most/all times.  And at the same time an encouragement to not pile up *too* many things on `dev`, as the risk of the aforementioned annoying scenarios happening increases when `dev` and `master` diverge more.

> There are other models out there, and the one I put forward is just one possibility... 

I'm happy with this one and keen on giving it a shot; but open to other suggestions of course.

> What I'm not sure about is what would be OK to push to master directly: if we restrict it to urgent bug fixes only, it's great for stability of the code, but not so good for fast turnaround for more minor bug fixes and enhancements, of the kind we routinely push out in response to reports on the forum. 

I think that's fine, and *if* we don't let `dev` pile up for too long (as well as keep in in a "ready" state at all times), we should still be able to have a fast turnaround, as we can have the turn(around) happen at about any moment we feel there's a need to.

> So I think there is an argument for using dev to hold compatibility-breaking changes and/or major new features, and pushing everything else to master as we normally do.

Here goes probably the concern I brought up above about `master` heading off on its own *too* much versus `dev`, which basically ends up "punishing" those that have used `dev`, because the onus will be on `dev` to merge `master` in again.  The risk of this would of course again be lower if `dev` doesn't sit there for too long without being merged to `master` at a reasonable frequency.  An actual incentive for a merge of `dev` into `master` could actually be any change that would otherwise let `master` stray away from `dev` too much.  If whatever change/feature has the risk of causing this, it should reasonably take the responsibility of building on `dev`, and taking `dev` along to `master`, I reckon.

> I reckon as people start using the tagged releases more often (by way of downloading pre-compiled packages), pushing changes to master will be something we probably won't need to worry about so much - although we will still want to ensure stability as much as possible.

True, although we're not there yet of course.

Lestropie:
#1047 merged in.

Regarding #1051, open to input but my instinct is to leave it out and pursue an alternative solution independently of this tag / PR (details to be discussed in #874). If the automated solution makes assumptions that don't hold for reasonably common data, and there's no trivial way to detect when that assumption doesn't hold, I don't think it's a process we can automate.

Given #1036 is the primary, @thijsdhollander do you want to draft a merge announcement?

Lestropie:
> The git workflow we're talking about shouldn't have any impact of flexibility - if something needs to be pushed out fast, then it can.

As long as what's being proposed is adhered to, then yes. The bit that's stumbled me at this particular point in time is things like #1048. Sometimes with larger changes to the Python scripts, where currently I don't have any mechanism for automated refactors, it's possible for bits of code here and there to not be appropriately updated to reflect changes. Automated testing of the scripts is currently very minimal, and there's no compilation process to catch outright errors. So I'm looking into whether a linter can possibly help (possibly within the CI). I'm also looking (both with this request in particular and more generally) to merge changes such as this into `dev` immediately *after* a tag release, so that if things do get overlooked there's a better chance of being caught by a developer within `dev` and getting a quick fix without affecting users.

> But if we had been using it, then `dev` should have been clean, usable, and ready for release, and the only addition to it would have been to merge the emergency feature and release.

:+1:

> So I think there is an argument for using `dev` to hold compatibility-breaking changes and/or major new features, and pushing everything else to master as we normally do.

Agreed. Anything that's an outright bug should go to `master` so that anyone who does a straight `git clone` has minimal issues, and we're less likely to get duplicate bug reports.

> The minor version increments I was referring to, would be for significantly new features; but I wanted to stress that these can also come on their own, or are at least not required to be part of a "big" batch of features. I was basically arguing for fast turnaround.

There may need to be at least some level of restraint on regular feature releases, depending on the overhead associated with each tag. If we need to configure documentation builds, generate pre-compiled binaries, notify third-party distribution platforms, make forum announcements etc. every time there's a tagged update, we may not want to be doing that too frequently (e.g. every time we add one new command-line option to one command, as the far extreme case). So we probably just need to feel out exactly where we end up on the spectrum between that extreme case and the opposite, i.e. no updates until *MRtrix4*.

> ... at times bigger changes did happen to master, sometimes quite important ones, yet some particular ones either piled up or were on their own already a pain to merge into 0.3.16 again.

With divergence like that, merging is going to be a pain regardless. Restricting `master` merges to bug fixes *only*, reviews before merging to `dev`, basing feature branches off `dev`, and shorter turnarounds, should hopefully all contribute in reducing this.

Basing feature development branches off `dev` is also related to my first point RE: Python library changes. This was quite annoying in 0.3.16 with the extent of changes that happened in there.

-----

One other point regarding this model in conjunction with tags: If a git-naive user were to use `git checkout 3.0_RC2`, it would be my preference for that user to receive all of the bug fixes that are pushed to `master` after the `3.0_RC2` tag, up until `dev` is merged and the next tag is created. I suppose that technically what I'm describing would be better handled with version *branches* rather than *tags*... but could the tags be "rolled" if there are critical bug fixes on `master`? Or should we just use this to justify faster turnarounds and encouraging users to update? Or are we optimistic that such crucial bugs will progressively become less frequent and therefore this will be less of an issue over time?

thijsdhollander:
>>  The minor version increments I was referring to, would be for significantly new features; but I wanted to stress that these can also come on their own, or are at least not required to be part of a "big" batch of features. I was basically arguing for fast turnaround.

> There may need to be at least some level of restraint on regular feature releases, depending on the overhead associated with each tag. If we need to configure documentation builds, generate pre-compiled binaries, notify third-party distribution platforms, make forum announcements etc. every time there's a tagged update, we may not want to be doing that too frequently (e.g. every time we add one new command-line option to one command, as the far extreme case). So we probably just need to feel out exactly where we end up on the spectrum between that extreme case and the opposite, i.e. no updates until MRtrix4.

Maybe adding some further nuance to my original aim/argument there: so the argument is probably rather for the *possibility* of a fast turnaround, if circumstances ask for it.  So definitely not the whole overhead for each new option to a command, but the possibility for it in a case where the developer of the new feature has a good incentive to make it available ASAP.  In the latter case, it shouldn't matter whether there were already some other new features queueing up, or none at all.  MRtrix should remain our vehicle for making our own new methods available, and as soon as we deem necessary.  As long as a new method doesn't change any existing functionality, but only adds new and independent things, it should be able to be introduced relatively easily, and the (independent) state of other developments shouldn't hinder that.  I essentially want to avoid we set some "predetermined" kind of bureaucracy in stone where e.g. "you can't release a feature now, because it's the only one and we have just released a batch of 5 new features last week"... or something like that.

> Given #1036 is the primary, @thijsdhollander do you want to draft a merge announcement?

 :+1:  Yep, I'll initiate something (the bit on `mtnormalise`) via the staff section on the forums; I'll keep it brief and simple.  Further drafting / adding to / tweaking the announcement can then be done over there.

jdtournier:
Just a few scattered responses:

> Automated testing of the scripts is currently very minimal, and there's no compilation process to catch outright errors. So I'm looking into whether a linter can possibly help (possibly within the CI). 

Sounds like a good plan, given how prominent scripts have become over the years. :+1:

> I'm also looking (both with this request in particular and more generally) to merge changes such as this into dev immediately after a tag release, so that if things do get overlooked there's a better chance of being caught by a developer within dev and getting a quick fix without affecting users.

I'm not sure what the best strategy is here - these are complex scripts that get used all kinds of different ways. I agree these should be merged to `dev` as early as possible to get some developer testing, but this is unlikely to exercise these scripts to the levels that the community manages. So I think we have to acknowledge that bugs will always creep in with new features, and this is why we should push out relatively frequent bug fix releases - maybe every 6 weeks or so, as the need arises. So if we were to release a new version where behaviour has changed significantly, that should herald a minor version increase (e.g. `3.0.3` → `3.1.0`), while a set of bug fixes that don't change behaviour should trigger a micro version increase (e.g. `3.0.3` → `3.0.4`). In practice, you'll find that most open-source projects are a little buggy in their initial `X.Y.0` release, and a lot better with later micro releases - it's unavoidable when you introduce significant new features. 

> There may need to be at least some level of restraint on regular feature releases, depending on the overhead associated with each tag. 

I agree we can't be releasing new versions every week... It'll annoy users, and increase the workload for us considerably. But there is a bit of a conflict here:

> MRtrix should remain our vehicle for making our own new methods available, and as soon as we deem necessary. As long as a new method doesn't change any existing functionality, but only adds new and independent things, it should be able to be introduced relatively easily, and the (independent) state of other developments shouldn't hinder that. I essentially want to avoid we set some "predetermined" kind of bureaucracy in stone where e.g. "you can't release a feature now, because it's the only one and we have just released a batch of 5 new features last week"... or something like that.

It depends on what the feature is obviously, but I would argue that something that adds functionality should really be held off for the next _minor_ release... There's nothing to stop interested parties from using it by building the `dev` branch directly, but I do think we need to enter a stable cycle of fixed releases, in line with how other open-source projects operate. If the added functionality is necessary to get around other issues that have been identified by users, then maybe there's an argument for pushing it out as a micro version - it's closer to a bug fix in this case. 

On the other hand, if the new feature is interesting and important, then I expect it would have taken longer than a week to materialise, so the developer would have had the opportunity to flag his intention of merging it into the next minor release in time for the rest of the team to discuss the merits of delaying the release to allow this new feature to be included. Again, it depends what we mean by 'feature' in this context - if it's just a new command-line option to e.g. provide additional diagnostic information to an existing command, in a way that has no other impact on existing usage, then I don't have a problem with pushing it out as a micro release (I think). It depends on how urgent we feel these changes are - bearing in mind that most people rarely update their installations, so a quick release in no way means everyone will be using it immediately... 

> One other point regarding this model in conjunction with tags: If a git-naive user were to use git checkout 3.0_RC2, it would be my preference for that user to receive all of the bug fixes that are pushed to master after the 3.0_RC2 tag, up until dev is merged and the next tag is created. I suppose that technically what I'm describing would be better handled with version branches rather than tags... but could the tags be "rolled" if there are critical bug fixes on master?

Yes, I've often wished it was possible to do something like this. But the reality is once we start producing binary releases, then we can't be changing its contents without labelling it differently - i.e. with a different version number... We can't allow for people to download what looks like the same package as before, but is actually not the same - and it would be very difficult for anyone to really know which exact version was actually running, given the way the versioning is currently handled: they'd all show up as the same version number (i.e. just the tag name - no SHA1 checksum is shown if the commit is also a tag). 

The way I envisage this working is that if a bug is fixed on `master` that we agree is sufficiently critical that it needs to be released immediately, then that's what we do - with a new micro version, even if it's a single commit. But most bugs aren't that critical, so can be released along with other fixes in due course.  

> Or should we just use this to justify faster turnarounds and encouraging users to update? 

Well, hopefully by the time we do release, it will have been picked up by NeuroDebian, so these updates will roll out automatically. Not sure how this would be handled on Windows or macOS yet, though. 

> Or are we optimistic that such crucial bugs will progressively become less frequent and therefore this will be less of an issue over time?

:rofl: I wish! Unlikely though...

maxpietsch:
I've a added the voxel size option to `population_template` as it does not change any behaviour. In 55ab618, I changed the FBA docs so that `population_template` is run with 1.25mm resolution. Is everyone ok with that?

thijsdhollander:
> I've a added the voxel size option to population_template to address #1037. It is an optional argument and does not change any current behaviour. In 55ab618, I changed the FBA docs so that population_template is run with 1.25mm resolution. Is everyone ok with that?

:+1: Has my support 100%.

At some point, when I find some time again in a couple of weeks or so, I'll comment a bit further in the issue, as I think there are very reasonable (and sampling-theory justifiable) arguments to change the default behaviour as well.  I may collect some opinions and evidence in the mean time to support it decently.  But so, no time now, and the discussion shouldn't hold this particular pull request back any more.  The FBA docs change addresses the issue sufficiently for the scenarios that I currently have to manage.

thijsdhollander:
What's still needed to release this?  I vote we add the volumes view for the light box mode (#694), and close it off right there.  The original scope of this pull request has already been settled a while a go; it's time to move on.

jdtournier:
Sorry, I was away last week - was hoping to find the time to tie this off, but a grant deadline got in the way... 

All that's missing is the release statement. I'll work on that this afternoon, hopefully we can release tonight or tomorrow.

thijsdhollander:
:+1: Sounds good.  Let's definitely still get #694 in, since it's a finished and complete feature.  Further extensions to it can be included in another future release.

thijsdhollander:
Ok, everything's ready and all tests check out.  The original scope and more is done.  I note a few thumbs up in the staff category discussion.  As mentioned there, unless anyone voices concerns, the new RC2 can become a reality tomorrow. :relieved:  I'll keep an eye on here and the staff category discussion, and unless concerns are voiced, make things happen somewhere AM AEST... That should give the other time zones some final (day) time to respond if needed.

maxpietsch:
@thijsdhollander In `mtnormalise`, is there a way to separate "bias" field from global scaling? Typically "bias" fields are scaled or shifted so that they average to one. I realise that you named the field as "norm" not "bias" but typically bias fields are scaled or shifted so that they average to one which makes them interpretable separately from a global intensity scaling. Conceptually, there should be no difference to `mtnormalise`'s "norm" field.

In the case of a zeroth order polynomial, the "norm" field equals the estimated scale factor which is also stored in the image header under the entry `lognorm_scale`. But I've noticed that the bias or "norm" field in `mtnormalise` no longer averages to 1 in the estimated mask as it did with `mtbin` - nor does it average to `lognorm_scale` for third order polynomials.

I might have missed it buried in the discussions here but what is the meaning of `lognorm_scale` in the case of a third order polynomial? Unless I got it wrong, I'd suggest changing the `lognorm_scale` factor in the image header to reflect the global intensity scaling.

jdtournier:
Happy for release - as long as @maxpietsch is happy, of course... :wink:

40°C over here in Greece - sweltering... If it all goes south, I'm off-grid. 😁

maxpietsch:
Don't let my comment deter you from merging!

thijsdhollander:
@maxpietsch : I'll merge somewhere this morning, but I'll get back to your comments above too.  They're indeed valid points, or rather "choices", that can be made when providing those outputs, and the overall rationale of the new `mtnormalise` on that front is indeed different compared to `mtbin` before.  There's good reasons though (beyond those that you'd probably identified as well).  Briefly though: you can always "normalise" (forgive the confusing terminology here) the output `-check_norm` yourself of course, using any mask or average that you deem appropriate.  The (or rather "an") easy way would be to just divide the `-check_norm` image by the `lognorm_scale` value.

thijsdhollander:
>  thijsdhollander dismissed jdtournier’s stale review via 9635243

I can overcome this now because I wasn't the one creating the pull request, but we may want to think about this setting/behaviour in the future.  After looking this up, it looks like this can be set: i.e. if new commits undo any reviews.  This makes sense in theory, but in practice may be annoying for our work flow if e.g. any last minute bug fixes are done (as is essentially the case here now).

Lestropie:
Something like `dwi2fod-shell-option` I'd normally expect to go straight to `master`, since it's an overt bug fix. It's possible that if you then need to merge changes in `master` back into a branch with a pull request, it won't invalidate a review *as long as* that merge is a simple push-button; any manual merge would indeed need to invalidate a review. But I wouldn't know without testing it.

jdtournier:
Good to see this is now released! Good effort everyone. 

Just one thing: I note the tag is not signed...? At least it's not showing up as verified on the releases page. I'd like us to stick to signed releases to give the software a clear seal of approval. @thijsdhollander, any chance you can rectify? Procedure is explained [here](https://github.com/blog/2144-gpg-signature-verification) and other places. I expect you'll need to delete the existing tag and create a new signed version, which you can do on your local repo, but you won't have the rights to force-push the updated tag to the main repo from your account. Simplest at this point would be for you to push the correct tag to your own fork of the repo, and let me know where to fetch it from. I can then fetch your repo and force-push it from my side. How does that sound? 

Note there's no great rush to do this, take your time...

jdtournier:
>> thijsdhollander dismissed jdtournier’s stale review via 9635243
>
> I can overcome this now because I wasn't the one creating the pull request, but we may want to think about this setting/behaviour in the future. After looking this up, it looks like this can be set: i.e. if new commits undo any reviews. This makes sense in theory, but in practice may be annoying for our work flow if e.g. any last minute bug fixes are done (as is essentially the case here now).

Well, the whole point of enforcing these reviews is to ensure _all_ changes that get pushed to `master` have the chance to be properly reviewed and discussed. In many ways, this kind of last minute addition is precisely what this procedure is supposed to prevent... It's no problem in this case since @bjeurissen's pull request was clearly itself reviewed and approved, but in general I wouldn't want to allow last-minute commits to be pushed to an already-approved pull request if that means changes that would otherwise warrant discussion might get pushed out to `master` prematurely - I assume that's the whole reason this setting exists in the first place...

As far as our workflow is concerned, I don't see that this should be a problem in any way. From this point on, we should _not_ be pushing _anything_ to `master` without thorough review. [As discussed earlier on in this thread](https://github.com/MRtrix3/mrtrix3/pull/1052#issuecomment-316342530), any urgent bug fixes can be pushed to `master` directly anyway in their own pull request (in which case it should trigger a new micro release), anything else should be merged to `dev` with a view to making it into the next release. 

In this case, all this would have done is delay the release a touch while someone else provides an up to date review - which you were able to do yourself here anyway. So I don't see the need to change anything - hope you all agree...?

Lestropie:
Ah; should probably have rebased #1083 to `master` given it's not urgent, sorry. Maybe not as critical right now with `RC` tags, but should get into the habit of playing by the more stringent rules.

Agree with keeping the stale review settings as-is. It's possible to create a PR but not yet request a review if there's a good chance of minor fixes before merge. Also, splitting development work in `dev`, which would trigger a minor release, with overt bug fixes to `master` that would trigger a macro release, should reduce the chance of repeating a case like this (i.e. last-minute small bug fix invalidating review of large changeset).

thijsdhollander:
> As far as our workflow is concerned,  ....... So I don't see the need to change anything - hope you all agree...?

Fair enough; happy with it.  Let's see how it plays out in practice.

> Just one thing: I note the tag is not signed...?

Aha, interesting.  I hadn't spotted this (functionality) yet, so was unaware of it all.  Ok, I've checked out a bunch of documentation on it and set up my GPG key to get it to work.  I've forked the repo here:

https://github.com/thijsdhollander/mrtrix3

...and pulled it locally, removed the original (unsigned) tag, created a new (signed) tag to the exact same commit (the merge commit), and force pushed again to the forked repo (apparently had to force push because a / the tag with the same name already existed).  This all looks ok now:

https://github.com/thijsdhollander/mrtrix3/releases/tag/3.0_RC2

@jdtournier, I suppose this gives you the means to fetch my fork and force push to the main one?  I can make it into a "release" then again as well.

Going through the process and learning about the steps involved, I did start to wonder how I should have done it it practice originally though: wouldn't I have run into the same problem, i.e. that I can't push directly to our `master` branch?  Or could I have tagged the last commit in the release branch before it was merged to master?  But then it wouldn't have been the same as tagging the merge commit of course... and that sounds like it could lead to unwanted effects if e.g. any other later (in time) commits were added to `master` in between.  The latter just to say that I'd definitely want to tag the merge commit, but then I would run into the former issue (I think).

thijsdhollander:
Ok, forget most of my post above: I've figured out the answer to my last question *and* I fixed the signed tag straight on the main MRtrix3 repo's `master` branch.  And I just learned a lot... :thinking: 

So I found out that branch protection doesn't interfere with pushing a tag.  I *did* think that *force* pushing a tag (deletion and addition with the same tag name) would almost certainly be stopped by branch protection... but guess what: it isn't.  After I'd done my tag business on my local copy of the *main* MRtrix3 repo's `master`, I could simply `git push --tags -f` without anything stopping me.  It all checks out too:

https://github.com/MRtrix3/mrtrix3/releases/tag/3.0_RC2

Weird stuff. :open_mouth:

jdtournier:
> So I found out that branch protection doesn't interfere with pushing a tag.

😲 That's unexpected... I guess it makes sense since a tag is not a branch, and there's not tag protection option, but since tags show up as releases, it's a little scary...

In any case, I guess at least that problem is now solved... :+1:

jdtournier:
> Ah; should probably have rebased #1083 to dev given it's not urgent, sorry. Maybe not as critical right now with RC tags, but should get into the habit of playing by the more stringent rules.

Yes, I didn't spot that either... Oh well, not to worry. I'm not sure how to play this to be honest. Maybe the simplest is to do as you did: push overt bug fixes (whether critical or not) to `master` directly, and if one of these happens to be critical, then we tag and release with a new _micro_ version. This means any micro version gets all bug fixes up to and including the critical one. Anyone building from `master` gets a broadly bug-free version (at least they get all benign fixes - i.e. those that don't modify behaviour). And as discussed before, any new features or behaviour-modifying changes go into `dev`, destined to be part of the next _minor_ release. 

I think we'll need to codify all this and document it somewhere at some point... Maybe via doxygen, so it can show up in the [online dev docs](http://www.mrtrix.org/developer-documentation/)...? With a link to it in the [README](https://github.com/MRtrix3/mrtrix3/blob/master/README.md)?

thijsdhollander:
>  That's unexpected... I guess it makes sense since a tag is not a branch, and there's not tag protection option, but since tags show up as releases, it's a little scary...

Sleeping a night over it, this makes full sense to me as well now.  Tags have essentially nothing to do with any particular branch.  You can just tag any commit, or even any object (doesn't even have to be a commit apparently, but GitHub will only show tags to commits on its website).  But I agree it's a bit scary (though we can't really do anything about that).

Note furthermore that I didn't have to re-create the "release" thing online ("releases" are basically further annotation and possibly attached binaries etc... that only live/exist on GitHub, not in the actual repository; and are linked to a tag)!  You can see this by looking at the date/time of the release versus the tag: the tag (since I had to re-do it) has a time stamp that is more recent than the release (which retains its original time stamp).  This behaviour is what I find the most scary in a way: you could make a tag and release, then add compiled binaries or other files to it via GitHub, yet subsequently someone (of us) could change the tag with the same name to point to anything else, and that would inherently change where that "release" points to (even though the binaries or other attachments are completely independent of course).  So you can in principle end up with an inconsistent state.  Note even furthermore that I could've changed the tag even more easily without deleting the former, by using the `-f` option to `git tag`: that allows to directly create a tag using an existing tag's name (so essentially delete and create).

So well, lesson here is probably that we'll simply have to be careful.  There's no GitHub facilities that I'm aware of that offer any further protection on this front...

Finally, @jdtournier @Lestropie @maxpietsch @rtabbara (since you all had open pull requests, or branches that were rebased to `dev`, or were already based on `dev`): just a heads up that I just merged master into `dev`.  So `dev` is now up to date and we can in principle start working using the git flow model.  For most branches that were rebased, this should now have minimal consequences, since `master` (and now `dev`) is essentially the same as the previous RC2 branch was, apart from 2 extra commits that only touched a tiny bit of code related to `dwi2mask`.  Things that were originally based on (and branched from) `dev` will of course need to take into account that `dev` has now jumped quite a bit ahead.  We won't have this kind of situation (where your feature is being "surpassed" by a massive batch of different features related to a release *all at once*) any more *if we stick quite strictly to the git flow model*.  The latter will be the challenge. :wink:

PR #1055, opened 2017-07-17
Image diff relax tolerance on spacing

jdtournier:
Minor issue - but just noticed it caused some testing failures...


jdtournier:
Just realised this also contains a commit from #1056 - duh... I'll clean this up and try again.

Issue #1012, opened 2017-06-06
Catching erroneous Siemens mosaic import


Lestropie:
Following from [this thread](http://community.mrtrix.org/t/dwipreproc-eddy-error/913).

Ideally any operation that is not valid for a 3D image should throw a meaningful warning if provided with such data. `dwipreproc` in particular should catch this as early as possible so that `eddy` doesn't segfault. 

What confuses me though is:

> If you strip out the Siemens CSA info, MRtrix3 can't reformat these into 3 volumes - they remain as 2D slices with all the slices in a single large montage.

What are the dimensions reported by `mrinfo`? Because even if the user has provided the gradient table manually, I would expect `dwi2mask` to fail if the number of rows in the gradient table doesn't match the number of volumes.

jdtournier:
There'll be something like 1024×1024×1×64 - i.e. the z dimension will be one. The DICOM import code will still treat individual volumes as individual volumes, since they'll be labelled as such in the DICOM headers. They will genuinely be 4D.

On a different note, I did have a crack at supporting these files a while back even if the CSA info is missing - see [this branch](https://github.com/MRtrix3/mrtrix3/tree/handle_siemens_mosaic_even_if_csa_information_missing). The problem there is that if the CSA info is missing, so will the diffusion encoding info, so these data will likely be impossible to process anyway - not without obtaining that information from a different source, with all the pitfalls entailed in the process. Other issue is that you end up with blank slices at the top if the number of slices isn't an exact square of an integer (i.e. doesn't fill the mosaic exactly). 

Probably the best option would be to detect these cases and reporting them visibly to the user with a big warning, and then try our best to import the data regardless. The only remaining issue from that point of view is detecting these additional slices so that the imported data is cleaner. This could easily be done by looking for the last non-zero slice in the mosaic of the first volume, and assuming that is the number of slices. A bit messy, but doable...

Lestropie:
> There'll be something like 1024×1024×1×64

In that case it shouldn't be too hard to simply check that all three spatial axes have size > 1 in any algorithm that will break / not be applicable if it's not the case. Should just be a simple function in `image_helpers.h`. Can define something similar in the Python library so that it can be caught as early as possible in scripts. Even if the DICOM import is improved in the future, these would still be good sanity checks.

jdtournier:
:+1:

PR #1061, opened 2017-07-19
DICOM: try to handle Siemens mosaic even if CSA fields missing

jdtournier:
see related discussion in #1012. 

Is this worth considering including? Slight problem in that it requires a bit of work to finalise in order to throw out redundant slices, but should still be do-able relatively quickly...

### DIFF
--- a/lib/file/dicom/image.cpp
+++ b/lib/file/dicom/image.cpp
@@ -1,16 +1,16 @@
 /*
  * Copyright (c) 2008-2016 the MRtrix3 contributors
- * 
+ *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
- * 
+ *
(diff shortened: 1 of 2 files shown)


Lestropie:
@jdtournier: Did my best to resolve this with `dev`, but it will need your look-over / a run with your DICOM test suite / a test with scrubbed CSAs.

thijsdhollander:
Is this still a goal for RC3?

jdtournier:
> Is this still a goal for RC3?

To be honest, I don't think it should be. It certainly shouldn't delay the release, and even if I do get around to implementing the remaining changes, it'll require testing to make sure nothing breaks. I'll give it a crack if I have a minute, but personally I think we should release ASAP regardless.

Lestropie:
If it requires any more than just running your DICOM import test script, then let's leave it out.

jdtournier:
OK, let's leave it out. 

For completeness, the testing I'd do is just to make sure it doesn't break existing well-formed data. I'd need additional testing to check that it can read malformed data correctly, but there's no reference data for that to test against. So yes, it's a bit of work...

PR #1066, opened 2017-07-21
5ttgen fsl: Fix bug when provided mask does not match T1

Lestropie:
Silly errors in handling the case where the user provides a pre-calculated brain mask using the `-mask` option, but that mask is not on the same voxel grid as the input T1 image.


Lestropie:
Closing: Fix was duplicated in 2e30340 and merged to `dev` in #1049. No review received here, so fix can instead wait for imminent tag update.

Issue #1062, opened 2017-07-20
Advanced ROI handling (MR spectroscopy VOI support) - feature request


ansk:
Dear experts,

I am working with MR spectroscopy datasets (single-voxel MR spectroscopy). The MR spectroscopy voxel is large cuboid VOI with arbitrary size and orientation.

1.
I would need to compute some metrics from the VOI using the 3D anatomical image acquired in the same session, i.e. GM/WM/CSF content of voxel (when providing partial-volume fractions 3D images from `FSL-FAST` or `SPM`) and in case of providing segmentation image (such as `aparc.a2009s+aseg.mgz` from `FreeSurfer`) to provide volumes of particular anatomical compartments in the VOI. 

So my inquiry would be to have a functionality within mrtrix to be able to create cuboid VOI with arbitrary size and orientation and compute summary statistics using this VOI and arbitrary 3D dataset. The option maybe would be to create .mif image defining this VOI (1x1x1 image with large anisotropic voxel size?) to be able to use existing mrstats tools. However, I am not sure how mrstats handles cases when ROI is of differing orientation and voxel size with respect to 3D dataset (i.e. how the voxels lying only partially in the VOI are handled). Ideal would be to account partual-volume voxels properly.

2.
Also, displaying this VOI in `mrview` using ROI tool as intersection with every slice plane would be very nice and handy. 

3.
Another handy tool would be to be able to produce „partial-volume fraction“ image using 3D dataset and this VOI in the same geometry as the provided 3D dataset (this will definitelly belong to mredit framework), where each voxel value would encode whether the image is lying entirely in VOI (value 1), entirely outside VOI (value 0) and for voxels lying only partially in VOI to compute their exact partial volume fraction. This could provide useful partial-volume information for 3D datasets with larger VOIs, such as DWI or fMRI and enable to generate various ROIs for tractography (for example to generate ROI of insula voxels inside spectroscopic VOI using intersection of `aparc.a2009s+aseg.mgz` and this ROI)

The MR spectroscopy dataset is in DICOM provided as class `CSA Non Image`, therefore it can be   peculiar to convert it to `NIfTI` and direct reading of position and orientation from that. 
I tried mrconvert (`mrtrix`), `dcm2nii`, `mri_convert` (`FreeSurfer`) and neither of them was able to read it.

[...]

PR #1067, opened 2017-07-21
mrtransform: add -oversample option

jdtournier:
As discussed on #1062.


maxpietsch:
I'd add a warning or throw an exception if the oversample option is provided but no template is provided. 

Also, do we want to offer the oversample option for [warps](https://github.com/MRtrix3/mrtrix3/blob/master/core/filter/warp.h#L79) as well? Not sure if it makes much sense to change the oversampling factor of the reslice filter that is applied to the warp.

Issue #708, opened 2016-07-06
mrview Tractography tool: Display geometry option


Lestropie:
Currently the tractography tool is 'locked' to displaying track data as pseudo-streamtubes. Similar to other `mrview` tools, it should be possible to re-compile the shaders and branch in the main `draw()` call to display different geometries, selectable using a combobox. E.g.:
- Pseudo-streamtubes (i.e. current)
- Lines (i.e. pre-geometry-shader), for when you absolutely must be able to see all tracks in a large dataset, and don't care about line thickness
- Points (I get asked all the time about the ACT paper figures displaying dots for the streamline endpoints, currently the only solution is MRtrix 0.2's `mrview` with depth blend enabled; would also be handy for assessing tracking algorithms / track resampling)

Supersedes #14.

draffelt:
Points could be handy. I needed this recently for debugging something

Lestropie:
Closed by #1073.

PR #1073, opened 2017-07-27
mrview: Tractography: Adding line and point geometry

rtabbara:
As requested in #708. A few things:

* I've set lines as the default (rather than tubes). Don't think stream-tubes were very well received and for performance reasons I think having lines (which skip the geometry shader entirely) are the way to go for the default

* Thickness and lighting are disabled when rendering lines and points.

* Geometry type is associated to a specific tractogram so the tool does support rendering different types concurrently. Not sure how much use there is for this, but it does seem to be a very Donald thing to do.

Anyway, please check it out.


Lestropie:
With regards to the point geometry, the two applications I have for this are examining individual streamlines very closely for assessing tracking / resampling / image sampling, and visualising streamlines terminations (e.g. as described [here](http://mrtrix.readthedocs.io/en/latest/troubleshooting/FAQ.html#visualising-streamlines-terminations)). So in both these cases each point really needs to be larger than one pixel. I think I played with antialiased (round vs. square) points when writing the connectome toolbar, but ended up not using them, so I don't have a code example. I also don't know what the maximal radius might be, but I'd think it'd be large enough for any reasonable use case.

rtabbara:
@Lestropie Thanks for taking a look at this and clarifying the use-case for point geometry. Even though this PR is based to RC2, happy for this to be pushed down later given that it's not a show-stopper. That'll give me plenty of time to fix it up.

jdtournier:
Just had a quick look at this - looks good. One minor suggestion: maybe we could default to having the lighting on the first time you switch to pseudotube rendering? Otherwise it might look like nothing happened, and after all, lighting is the thing you're most likely after if switching to that rendering mode...

> Don't think stream-tubes were very well received

Really...? Where did that come from? It's a thing of beauty!

thijsdhollander:
> I've set lines as the default (rather than tubes). Don't think stream-tubes were very well received and for performance reasons I think having lines (which skip the geometry shader entirely) are the way to go for the default

Not sure I'd agree with that! :open_mouth:  Especially as a default, streamtubes are incredibly user-friendly and give you (via lighting) very useful hints on spatial depth of structures, that is otherwise much harder to see with lines.  Having lines as an option is good though, for those scenarios that do require exceptional performance... but I'd say these scenarios are relatively exceptional indeed (so not the default, I'd say).

Let's do this after / independent of RC2 though.  That one's scope is starting to extend too much...

Lestropie:
Yeah the pseudotubes are awesome, and way faster than other packages that actually triangulate cylinders. The issue is that having *only* the pseudotubes resulted in a loss of functionality, since files that could have been opened previously could no longer be opened. These changes absolutely address that. The remaining question is whether the basic line geometry should be the default (so that the program doesn't crash or lag out trying to render a tractogram with pseudotubes that would otherwise be displayed OK as lines), and if so, whether that would be "hiding" the pseudotubes functionality too much from the user (and whether we could do something about that).

Not urgent, so I support re-basing to `dev` and further discussion.

rtabbara:
> Not urgent, so I support re-basing to dev and further discussion.

@Lestropie  Yep, I agree. Will do.

> Not sure I'd agree with that! :open_mouth: Especially as a default, streamtubes are incredibly user-friendly and give you (via lighting) very useful hints on spatial depth of structures

@thijsdhollander Nice to know they're appreciated :smile: Happy to discuss down the line.

jdtournier:
> The issue is that having only the pseudotubes resulted in a loss of functionality, since files that could have been opened previously could no longer be opened.

I'm intrigued by that. Is the memory use any different? Why would the pseudotubes implementation crash? I can see it taking too long to render, but it shouldn't crash - or if it does, the regular lines version ought to also crash. If this is genuinely the case, I'd like to investigate this...

Lestropie:
Well, maybe exaggerated with "could no longer", more a question of having an interactive frame rate vs. not. And suppose I was leading the question RE: which should be the default. If lines are default users may be underwhelmed with the render quality, whereas if pseudotubes stay as default they may get hit with a low frame rate if opening a large file. Or could have a config file entry that auto-switches from pseudotubes to lines when loading a file if the number of tracks exceeds some number.

rtabbara:
I've now updated this PR to add support to render tract points with a size > 1px. The thickness slider also allows a user to dynamically adjust the point size. A few points (no pun intended):

* After discussing with @Lestropie, similar to pseudotubes, I've set it so that the rendered point-size is relative to the FoV.

* I have set the default geometry types to psuedotubes and as @jdtournier suggested, enabled lighting by default.

thijsdhollander:
I would simply have the pseudo-tubes as the default (i.e. exactly as how it currently behaves; I've had *a lot* of *very positive* feedback since the introduction of this, increases usability a lot and well, just looks fancy and bumps up the perception of our software), yet include a config-file entry for making line geometry the default.  My reasoning is always the same: the "default defaults" should cater for the most mainstream users, who are looking for usability in the most common scenarios and in this case, want high quality visualisations for their papers and presentations.  People who want to go beyond, and have a good use case for showing so many streamlines (and not only in a slab, but a whole brain volume) and at the same time don't have the fancy hardware to support where they are pushing things... well, they are probably determined enough to at least find out about the possibility of a config-file entry.  Even without it, initially, they can still open the file (without a crash), and switch to line geometry.  They'll just naturally discover the limits of their own hardware and decide that their particular work flow requires line geometry all the time, and set the config-file entry accordingly.

I wouldn't per se do this based on a number of streamlines.  There's many other factors that influence performance, going from length of those streamlines (and number of triangles essentially rendered), to whether they view them all at once (or only a slab), to what they are showing at the same time (a couple of high-res overlays and volume rendering also slows things down eventually; even without streamlines), to their own personal hardware and even other software they might be running at the time.  I also doubt there's going to be many people that want a mix of renders based on what the streamline count for each of them happens to be...  I reason the most common scenario is still pseudo-tubes, followed by people who want line geometry *consistently* for a particular scenario.

In general for many MRtrix features, where applicable of course, I'd rather shy away from too much (hidden) automatic heuristics.  There's exceptions of course, but in general I find these make the software over-complicated and too much of a mystery black box for most users, which in turn burdens us with more support work, because many of these confused users end up on the forum with us having to explain our intricate hidden logic.  Again, there's surely exceptions, but in general keeping things simple and clear where possible should help software maintenance and user support, and will also frustrate users less because they may otherwise feel they're not in control of simple things... just my 2 cents of course. :slightly_smiling_face:

thijsdhollander:
Looks like we were typing things at the same time; but I agree with both points @rtabbara!  So the only suggestion left may be to have a config-file entry for people who would want lines as default; if there is demand for it of course.

jdtournier:
Just to clear up a slight misunderstanding:

> - I have set the default geometry types to psuedotubes and as @jdtournier suggested, enabled lighting by default.

What I meant was that _if_ we default to lines, then the _switch_ to pseudotubes should have lighting enabled by default - otherwise it might look like not much has happened, I expect most users will be after a '3D' look when switching to streamtubes, which implies lighting. 

But if we stick to pseudotubes by default, I'm not sure whether I'd like to have lighting enabled by default... I guess it's not going to hurt, performance isn't that much worse with lighting enabled, and maybe it's best to stick with the highest 'quality' render by default. A config-file entry to modify the default would be good though, I'd probably stick to the current default (pseudotubes, no lighting) given the choice...

Lestropie:
If the option were available, I'd personally use a number of streamlines threshold to downgrade from pseudotubes to lines, since my machine isn't tied to one specific workflow. But that approach makes less sense if there's also a config file option for default lighting. Adding command-line options for altering the geometry of files opened at the command-line would help deal with reducing lag when opening very large tractograms, so I'd be content with that.

Couple of other thoughts, not sure if they are requisite for this merge or whether they can be listed separately:

- Currently "thickness" is a global parameter. Particularly when drawing both tracks and endpoints at the same time, it would be nice to be able to scale the size of the endpoint dots independently of the tracks.

- Lighting currently does not influence the point geometry. In an ideal world, these would be "pseudo-spheres" when lighting is enabled. Personally I don't care particularly, but I can imagine someone posting about this on the forum. @rtabbara Maybe you have some sense of how much effort this would take, and hence whether or not it's worth bothering?

rtabbara:
> Lighting currently does not influence the point geometry. In an ideal world, these would be "pseudo-spheres" when lighting is enabled.

Enjoy @Lestropie 

myimage

I prefer the name stream-pearls over pseudo-spheres, but that's just me.

thijsdhollander:
screenshot0000

Whole-Brain Stream-Pearl Imaging (WBSPI)

jdtournier:
👏 very nice, @rtabbara! Hope you had fun working out the maths for the lighting... 😜

thijsdhollander:
Looks like it's ready to merge to `dev`?  If so, don't hesitate to push the button, @rtabbara. :wink:

rtabbara:
@thijsdhollander Just pushed a few small things, but other than that I think it should be good to go. 

Might want to quickly check whether people are happy with how the dynamic stride updates when rendering points (i.e. under/over-aggressive when skipping points depending on point-size).

thijsdhollander:
> the dynamic stride updates when rendering points

Hmm, good that you mention this: we probably need to document this a bit more explicit now that we also render points.  In case of lines, this didn't matter so much (the documentation aspect, that is), but now that we render points explicitly, people may always "naturally" expect them to be all the points stored in the file; i.e. they may overlook the fact that points may be skipped, because when they do look at the points... their interest is probably in all those specific points.  What do you reckon, @Lestropie @jdtournier?

I suppose a common use case for the point geometry will be looking at track endpoints.  In case of skipping points, would this affect that when zooming out?

rtabbara:
>  In case of skipping points, would this affect that when zooming out?

Nope, because the point-size is invariant relative to the FoV, we only recompute the stride when a user explicitly adjusts the thickness slider.

thijsdhollander:
> Nope, because the point-size is invariant relative to the FoV, we only recompute the stride when a user explicitly adjusts the thickness slider.

Ok, that makes it much safer than I imagined it to be earlier.  In that case, I'd say it's safe to merge.  It's still in `dev` anyway, so documentation can always be added before the next release cycle.

Lestropie:
> Might want to quickly check whether people are happy with how the dynamic stride updates when rendering points

Feels fine I think - for tracks at least. It's not as sensible a thing to do when rendering points compared to rendering lines (points being entirely absent vs. slight errors in paths of lines), and it only kicks in if the points are really huge which isn't really a useful situation anyway, so the precise behaviour isn't all that important.

The nastier problem is what happens when using the point geometry to render streamlines endpoints (i.e. a track file generated from `tckresample -endpoints`, which has strictly 2 points per streamline). When the thickness is increased, first half the endpoints disappear, and then *all* points disappear. When downsampling in the [C++ tractography code](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/resampling/downsampler.cpp), I ensure that the streamline endpoints are always preserved regardless of the severity of downsampling; it looks like you've achieved similar by padding data with duplicates of the two endpoints? In which case, I don't quite know why it's glitching out in this scenario.

Maybe just disabling strides when displaying as points would be the easy fix.

rtabbara:
@Lestropie thanks for the PR tweaks.

>  ...it looks like you've achieved similar by padding data with duplicates of the two endpoints

Yep, that was the intention behind the padding in theory, but I agree it's not behaving as expected.

> Maybe just disabling strides when displaying as points would be the easy fix.

I'll take a closer look at this before yielding.  Regardless, I think disabling strides when rendering points isn't a particularly satisfying solution. Even if users are solely interested in visualising endpoints as indeed points, it's going to be pretty jarring to see entire tractograms appearing or disappearing based on geometry type.

rtabbara:
OK, had a more careful look at this and think I figured it out. There was a slight indexing issue when specifying track starts, but the main problem was that I incorrectly assumed that the amount of pre/post padding between each track would correspond to the max downsampling factor, when in fact max downsampling factor = padding_amount / 2.

The reason for this is that every call to the vertex shader contains the prev, current and next vertices for a given index in the vertex buffer. Track starts and ends are special cases however, where we have {track_start, track_start, next_ver} and {second_last_v, track_end, track_end} respectively. In short, we have to guarantee that even after downsampling there are at least two copies of both track start and track end.

The padding amount that I've used remains at 6 which means we have a max downsampling factor of 3 (please check arithmetic :wink:). In practice I think this is fine (?) &mdash; you would want to have some pretty chunky streamtubes for it to warrant a downsampling > 3, but please let me know otherwise...

jdtournier:
Looks good to me. It's fun rendering huge points - makes the brain look like bubble gum. :nerd:

Just a few minor comments though:

- on my system (nVIDIA GTX 780) , rendering a ~1.7GB tractogram with lines gives me a framerate of ~5 FPS, while rendering with pseudotubes gives ~3FPS (even with lighting). Pretty far from a deal-breaker. I'm not convinced of the need to have standard line rendering in there... Might be different on other systems though, I'll admit. But if that difference in performance is similar to that on other systems, I don't think it warrants introducing yet more complexity (and maintenance overhead)... 

- I'm not sold on the argument for downsampling when rendering points. The only real usage for this is to show the exact location of each vertex, so it strikes as counter-productive (or even potentially misleading) to decimate... It's also not detrimental to display quality (whereas it is a serious problem for pseudotubes when the width of the tube exceeds the step size), so I don't see a real argument for it. Just my 2 cents...

rtabbara:
> rendering a ~1.7GB tractogram with lines gives me a framerate of ~5 FPS, while rendering with pseudotubes gives ~3FPS (even with lighting).

Well, just as a point of comparison, I'm on a nVIDIA Quadro 600 and definitely see substantial differences between geometry types when rendering tractograms around the 100-200k mark. (e.g. 25fps lines vs 6fps tubes) but admittedly this is a pretty dated graphics card, so not sure if it's a fair comparison. Regardless, I don't really have a dog in this race, so if the consensus is to remove lines, I'm happy to do that.

> I'm not sold on the argument for downsampling when rendering points. 

Yep, that's fair enough. I was originally doing it for consistency, however I agree that downsampling points could lead to confusion. If anyone else thinks otherwise, please let me know.

Lestropie:
- Given the utility of displaying points, hence requiring branching for displaying different geometries, I don't know that removing basic line geometry would make a big difference to complexity. Plus it helps the frame rate (more so on older cards apparently). So I'd vote to keep it.

- I tend to agree with disabling downsampling of point geometry. Probably keep it enabled for line geometry though: Even if it doesn't have the `width > step size` issue, it might help with frame rate. Might be worth double-checking your math on the track data padding vs. downsampling ratio: If downsampling with point geometry is disabled, do the requirements become less stringent?

thijsdhollander:
Yeah, happy to keep lines too.  I agree with @Lestropie that we've already got "geometry" facilities in place anyway, both in the back-end and the GUI, so apart from streamtubes and spheres/points, there shouldn't be any harm in having a third option for lines.  It also sounds like the one with the least risk for maintenance overhead, since it's the least complex one.

I agree with disabling downsampling for point geometries too, for the same reasons stated by everyone before.

jdtournier:
Sorry, forgot to respond to @rtabbara: that difference in frame rate is indeed more than sufficient to justify hanging on to the basic line rendering - no argument there. 

Interestingly, this is likely to be due to the artificial differences introduced by nVIDIA to differentiate their Quadro range from the GeForce (gaming) range - as described in [this post](https://stackoverflow.com/a/29170393/3244166) (and many others if you look around), the Quadro drivers are very heavily optimised for wireframe rendering (i.e. rendering GL_LINES), since this is what CAD applications tend to rely on. Conversely, that functionality is probably not well supported by the GeForce drivers (despite the hardware being essentially identical). But on the GeForce cards, you get the same level of performance by using polygon rendering for line drawing (which is essentially what I observe on my system). 

Regardless, we need to allow for potentially very large differences in performance between different OpenGL environments, so I agree we should keep it.

rtabbara:
Ok, downsampling for points has been disabled. I'm going to leave this PR to cook over the weekend, but if there are no objections I'll be merging this into <del>master</del> dev next week.

Lestropie:
You merged as I was writing my review! ><

:+1:

PR #1076, opened 2017-08-02
Docs: Update advanced debugging page for 3.0_RC1

Lestropie:
Triggered by [this post](http://community.mrtrix.org/t/error-segmentation-fault-invalid-memory-access/1071).

Is nice to have an up-to-date documentation page on debugging so that a link can be provided rather than having to repeat the details each time. However "Advanced debugging" documentation page was not updated to reflect changes to binary location and multi-build-configuration with `3.0_RC1`.

### DIFF
--- a/docs/troubleshooting/advanced_debugging.rst
+++ b/docs/troubleshooting/advanced_debugging.rst
@@ -21,24 +21,26 @@ resulting information. The instructions for doing so are below.
 3. Configure and compile *MRtrix3* in debug mode:
 
    ::
-   
-       ./configure -debug -assert debug; ./build debug
-       
-   Note that this compilation will reside *alongside* your existing *MRtrix3*
-   installation, but will not interfere with it in any way. Commands
-   that are compiled in debug mode will reside in the ``debug/bin``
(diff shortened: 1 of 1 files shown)


Lestropie:
Are the instructions otherwise precisely equivalent? Because if not, it needs additional details, and I've never used it.

jdtournier:
Good question. I think they are, but maybe someone with actual macOS experience can confirm...?

chunhungyeh:
Oops... sorry, I thought that the commit comment would appeared over here... Well, anyway... just want to let you know that I added some instructions about using ``lldb`` on macOS, but I am not sure whether ``valgrind`` (which can be installed through Homebrew or Macports) would function similarly.

Lestropie:
@chunhungyeh: The commit comment can still be accessed on this page: Click on the "..." button next to the first line of the commit message. All good for additional details to go into commit messages, that way they're stored in `git` itself. Thanks for the addition. If `valgrind` is also available on Mac, I'd expect the interface to be exactly the same, so there shouldn't be any need for modification there.

PR #1077, opened 2017-08-02
dwipreproc: Fix volume matching with zero-length gradient vectors

Lestropie:
As raised [here](http://community.mrtrix.org/t/error-unable-to-determine-matching-dwi-volume-pairs-for-reversed-phase-encode-combination/1070).

**Edit**: Will wait overnight to make sure the test with the provided data passes.

### DIFF
--- a/bin/dwipreproc
+++ b/bin/dwipreproc
@@ -184,8 +184,17 @@ do_topup = (not PE_design == 'None')
 
 def grads_match(one, two):
   # Dot product between gradient directions
-  dot_product = one[0]*two[0] + one[1]*two[1] + one[2]*two[2]
-  if abs(dot_product) < 0.999:
+  # First, need to check for zero-norm vectors:
+  # - If both are zero, skip this check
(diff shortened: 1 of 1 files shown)


Lestropie:
While my test failed, it was for a different reason (similar issue previously encountered & solved [here](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/_5ttgen/fsl.py#L195-L200)):
```
$ echo $FSLOUTPUTTYPE
NIFTI
$ fsl5.0-applytopup --imain=dwi_pe_1.nii --datain=applytopup_config.txt --inindex=1 --topup=field --out=dwi_pe_1_topup.nii --method=jac
$ mrconvert dwi_pe_1_topup.nii dwi_pe_1_topup.mif -json_import dwi_pe_1.json
mrconvert: [ERROR] cannot stat file "dwi_pe_1_topup.nii": No such file or directory
$ ls
dwi_pe_1_topup.nii.gz
```
🤕

PR #1078, opened 2017-08-02
RNG: use std::random_device on all platforms

jdtournier:
as discussed here:
http://community.mrtrix.org/t/tck2connectome-edge-statistic-sift2-questions/1059/5?u=jdtournier

### DIFF
--- a/core/math/rng.h
+++ b/core/math/rng.h
@@ -62,15 +62,8 @@ namespace MR
           if (from_env) 
             return to<std::mt19937::result_type> (from_env);
 
-#ifdef MRTRIX_WINDOWS
-          struct timeval tv;
-          gettimeofday (&tv, nullptr);
-          return tv.tv_sec ^ tv.tv_usec;
-#else 
-          // TODO check whether this does in fact work on Windows...
           std::random_device rd;
           return rd();
-#endif
         }
 
 


jdtournier:
Weird - works fine on my Win10 VirtualBox, and on my Win10 laptop - both using up to date MSYS2, GCC 7.1.0... Looks like it's been fixed in more recent releases of GCC. But if we can't rely on users having the right version of GCC installed, I agree we can't merge this as it stands.

But having read through that blog post you pointed to, I think we might be better off replacing our RNG classes with the author's [randutils.hpp](https://gist.github.com/imneme/540829265469e673d045) header. He describes the interfaces in a [subsequent post](http://www.pcg-random.org/posts/ease-of-use-without-loss-of-power.html), and it seems pretty solid. My only concern with it is what happens on copy-construction - our threading backed relies on the copy-constructor giving each thread a different seed, so we'd need to make sure that requirement is also satisfied. But that'll probably have to be left for some other time...

jdtournier:
> Weird - works fine on my Win10 VirtualBox, and on my Win10 laptop - both using up to date MSYS2, GCC 7.1.0... Looks like it's been fixed in more recent releases of GCC.

Apologies, you're completely right - it's totally deterministic on Windows. I was fooled by the randomness introduced by multi-threading. I get identical output if I disable multi-threading... Good catch :+1:

PR #1080, opened 2017-08-03
Python fsl.findimage() function

Lestropie:
Another script fix necessary to get through my test of #1077.

### DIFF
--- a/bin/dwibiascorrect
+++ b/bin/dwibiascorrect
@@ -51,10 +51,6 @@ if app.args.fsl:
       app.error('Could not find FSL program fast; please verify FSL install')
     
   fsl_suffix = fsl.suffix()
-  if fast_cmd == 'fast':
-    fast_suffix = fsl_suffix
-  else:
-    fast_suffix = '.nii.gz'
 
 elif app.args.ants:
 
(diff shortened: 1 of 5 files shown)


Lestropie:
I've successfully executed all scripts affected by these changes.

I also made a bad stnank in #1047 [here](https://github.com/MRtrix3/mrtrix3/pull/1047/files#diff-9a2510352b82dd6f73530274c31a7e1fR483): Because `applytopup` is now run separately for each phase-encoding direction, the list of its outputs may only contain one entry, and therefore `mrcat` can't be run... So this very much needs to go with `3.0_RC2`: `-rpe_none` and `-rpe_pair` are currently broken. Have I mentioned that I resent `dwipreproc`? Perhaps it's time to make a matrix of tests (not necessarily in the CI) to make sure `dwipreproc` handles each possible case as it should; the different possibilities are so intertwined that it's really easy to fix one use case but break something else.

Open to review... Any volunteers?

jdtournier:
Not sure I'm the best person to be reviewing this... I'm happy for it to be merged if you're 'happy' with it. As you say, this is the kind to things that requires extensive community testing to really identify all the edge cases. It would be good to have tests for all of these, but that will be a lot of work, and definitely won't make it into `tag_3.0_RC2`... So I'd merge now and add tests later if/when you get the chance...

Issue #1054, opened 2017-07-17
CI: Check for TODO / FIXME


Lestropie:
Came to me writing [this comment in #1051](https://github.com/MRtrix3/mrtrix3/pull/1051#issuecomment-315756258), probably spurred on by recently seeing a similar message pop in in a Python linter.

I often write `TODO`s / `FIXME`s in my code as I work, and end up looking for them through my `git diff`s to make sure I haven't missed anything I intended to get to. But there's also a few floating around in the code base here and there, which can be worrying when you find them, especially when there's no further information in the comment.

So we could have a script similar to `check_memalign` (or an expansion of its functionality) that looks for such symbols and fails the CI test.

jdtournier:
Not sure I'd fail the CI tests for something like this. Often these are wish-list level items, or only a concern in rare edge cases. One option though is to use Doxygen's [todo command](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdtodo) - you then get a page in the online docs with these listed, but to be honest I'm not sure it would be much use, I don't think many of us consult these pages routinely. Personally, I think anything that genuinely warrants a dedicated TODO or FIXME line probably deserves an issue on GitHub. Otherwise it's just unfinished. That said, if that's how people currently use TODOs (i.e. as a placeholder for them not to forget to deal with something before merging), then yes, ensuring it fails CI would be a useful safety net.

maxpietsch:
I tend to use TODOs to highlight future extensions and I've seen a few FIXME in warning, error and assert failure messages. I guess we could change non-critical TODOs to something like FUTURE but I doubt it's worth writing a dedicated test for it.

thijsdhollander:
Also, some "FIXME"s at the moment are more aimed towards developers using a certain part of the API, right?  As in: watch out, developer, you're violating some *precondition* of this functionality.

But I agree with @jdtournier , that "future" TODOs/FIXMEs deserve a GitHub issue to highlight them.  It's the most direct way of bringing them to the attention of other developers as well.  Things like "TODOs before this feature is actually *ready* to be merged somewhere" could also be highlighted in a pull request as a note somewhere.  A comment in the code itself can also still (additionally) be useful of course, to easily track down where exactly the TODO is still supposed to take place.

Lestropie:
Also: Could similarly detect `TRACE` and `VAR()` calls.

Lestropie:
[...]

PR #1092, opened 2017-08-14
mrview Tractography tool: Fix to downsampling

Lestropie:
Setting the stride for downsampling should be based on the output step size (if present) rather than the tracking step size. If no step size information is available at all from the track file header, then the stride should be fixed to 1.

Should have been included in #1069.

Also note relation to #1054.


jdtournier:
Yep, sounds good to me.

PR #1098, opened 2017-08-18
Pylint tests and script image header import

### LINKED ISSUE #1074: Test Python scripts using linter
As first mentioned in #1052 & #1054.

Will most likely use `pylint`. Requires creation of a runtime configuration file to set up our desired code "style", and to set up what severity of error is necessary in order to fail CI testing.

Lestropie:
Closes #1074. Supersedes #1048.

As the library & provided scripts have increased in complexity, there's been too much grief with modifications breaking things, Python 2/3 compatibility, and a lack of testing in CI (since most of them can't be tested on a small FoV image sample, or have external dependencies e.g. FSL). So I've bitten the bullet and integrated a linter into the TravisCI tests. Along the way I've re-formatted various bits of code, and found all sorts of little bugs.

I'm going to close #1048 since I used that branch as the base of this one, and there were so many issues found along the way that it didn't make sense to try to break the two branches apart for the sake of independent pull requests. So included in here is everything in #1048, the major component of which is the `image.Header` class, which pulls all header information from an image file into a read-only class instance. This is *much* neater than wrapping individual `mrinfo` calls to extract individual header properties as was done previously, particularly as each header element is intrinsically imported as the appropriate type; so you don't have to e.g. manually convert the output of `mrinfo -output size` from a string to a list of integers, it's all done at the import step.

@jdtournier has expressed discomfort on multiple occasions about reviewing changes to the script library since he basically hasn't touched it. So I'm flagging @maxpietsch, but am putting a call out more generally to find out who around here feels competent to review Python changes.


Lestropie:
Fixed merge conflicts with `dev`. Is anybody likely to be able to check over this? I would like to have the `pylint` tests integrated into `dev` before I deal with merging #1049, since there'll be a bit more conflict there.

Lestropie:
Mostly variable names were overriding built-ins / reserved terms. For one of them it complained about a variable name being re-used as a function input argument. For `input` -> `inputs`, I don't recall whether or not `pylint` complained, but it struck me as very odd during editing that this one variable name was singular yet every other variable containing a list with one entry per input file had a plural name.

`pylint` catches attempts at reading variable names before they are written, which helps a lot for these kinds of refactors. If it's looking like a problem let me know and I can help merge.

Cheers for the look-over.

PR #1099, opened 2017-08-18
File::MMap: Use delayed write-back on ZFS

Lestropie:
Even though the file system is local rather than network-based, exceptionally slow write speeds were observed (~ 1 day vs. < 10 seconds).

Has only been tested on a ZFS mirror setup; have no idea what performance is like on other ZFS setups, nor whether the reported filesystem type varies for different ZFS setups. Cannot find any documentation; the magic number appears only a handful of times on the entire interwebs.

### DIFF
--- a/core/file/mmap.cpp
+++ b/core/file/mmap.cpp
@@ -20,7 +20,7 @@
 # include <sys/param.h>
 # include <sys/mount.h>
 #elif !defined(MRTRIX_WINDOWS)
-# include <sys/vfs.h> 
+# include <sys/vfs.h>
 #endif
 
 #ifdef MRTRIX_WINDOWS
@@ -53,9 +53,9 @@ namespace MR
       mtime = sbuf.st_mtime;
 
 
-      if (msize < 0) 
+      if (msize < 0)
         msize = sbuf.st_size - start;
-      else if (start + msize > sbuf.st_size) 
+      else if (start + msize > sbuf.st_size)
         throw Exception ("file \"" + Entry::name + "\" is smaller than expected");
 
       bool delayed_writeback = false;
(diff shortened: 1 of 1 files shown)


bjeurissen:
I can test this on my ZFS setup. Do you have a scenario that I can use to reproduce?

Lestropie:
Was just a `mrconvert` from `.nii.gz` to `.mif` of HCP DWI data. Smaller images seemed to be OK.

jdtournier:
I wouldn't be surprised to find this isn't related to the type of filesystem as such, but rather to the kernel's handling of dirty pages - as [outlined in the documentation](http://mrtrix.readthedocs.io/en/latest/troubleshooting/FAQ.html#linux-very-slow-performance-when-writing-large-images). What evidence do you have that this is specific to ZFS? Could it be related to the specific system's virtual memory management settings? 

@maxpietsch & I spent quite a bit of time trying to figure this one out, and it's a very tricky one to deal with. There's a long history of issues with writing to memory-mapped regions. In the past, we used memory-mapping read-write as a matter of course, but this led to very obvious issues on network filesystems - so this motivated c7ec2dd8, which implemented delayed write-back for _any_ read-write memory-mapped file (though technically the file is no longer memory-mapped as such). But then this [led to some users running out of RAM](http://community.mrtrix.org/t/hcp-connectome-tutorial/144/6?u=jdtournier) for operations that would work fine when using memory-mapping. This is when we implemented 45673184: detection of specific filesystems so we could use delayed write-back only for files where direct memory-mapping would be problematic. 

The problem we were addressing was the huge amount of traffic generated by the filesystem trying to maintain consistency of the file contents when the process is concurrently writing at potentially very distant offsets - this can lead to massive amounts of network traffic as any 4KB memory page marked as dirty gets dispatched to update the server, only to be potentially written to again shortly thereafter as the process come around to update a nearby value. 

But it turns out that's only half the story:  we were (or at least _I_ was) relying on the kernel caching the file contents in RAM until the file was closed, and then writing-back the file in its own time. But as we now know, on Linux there are limits to how many dirty pages the kernel will hold in RAM before forcing them to be committed to disk. So if you have a process that loop over the data in such a way that it touches most 4KB memory pages once before looping back again (i.e. badly matched strides for the output image), then we end up in a situation where the system forces a write-back at some point early on, then allows processing to continue, only to find it needs to force a write-back again pretty soon after that, and so on. This really impacts on performance, in some cases quite badly. It's also a problem for piped images, by the way (these are also accessed using memory-mapping). But when the strides are a decent match, there's no real problem: the pages being written back are unlikely to be written to again, so each page only gets committed to disk once - no real performance hit. But when the strides are a long way from contiguous, this kind of situation can occur. 

As to a solution to the problem: the obvious one is to force delayed write-back every time - i.e. revert 45673184, and accept the fact that this will mean increased minimum RAM requirements. To be honest, in such low RAM situations, even though the system can deal with the demands of the process, it will entail a lot of swapping, and performance under these circumstances is likely to be bad anyway... So it may simplest to do this. Otherwise, we could try to detect when the strides and size of the image are likely to break the system's limits, but that will be very fiddly to get right - detecting the system's limits requires peering into the system's internals, and is likely not very portable to boot. I can't think of any other reasonable way to deal with this...

If we do revert to enforced delayed write-back, one aspect that will be moderately upsetting is the slight loss in performance for piped operations, due to the additional `write()` - but I think there are ways of dealing with this, particularly using [POSIX shared memory](http://man7.org/linux/man-pages/man7/shm_overview.7.html) - which essentially behaves like memory-mapping, but with files guaranteed to be RAM-only. But I'll save that for later...

Lestropie:
Ah, forgot about the dirty pages settings. I'll try that when I get the chance (not my system).

Lestropie:
Yep, it was the dirty pages settings.

Just thinking about the documentation:

- Currently the "Hanging on network file system when writing images" section is duplicated across the FAQ and "Hanging or crashing" pages. Anyone recall whether or not this was deliberate? (They have two different authors)

- The section on dirty pages is currently in the FAQ's. Should this perhaps go into the "Hanging or crashing" page, and rename that page to e.g. "Performance & crashing issues" (since in this case it's not an outright hang)?

- Should the end of each installation page include a link to the aforementioned page, so that people are more likely to check that if they encounter an issue immediately after installation?

maxpietsch:
"Hanging on network file system when writing images" used to be in "display issues" from which I have moved it to the FAQs (e4f2b573) not realising that it also was in the "Hanging or crashing" page.
I am in favour of a "Performance & crashing issues" page which then also should contain the dirty pages setting.

PR #1100, opened 2017-08-21
mrview: add Rami Tabbara in AUTHOR entry

jdtournier:
Just noticed this burning injustice, and had to do something about it...


thijsdhollander:
Well done Rami, we are so proud of you! :man_dancing:  :tada:

PR #1101, opened 2017-08-22
Documentation: Various fixes and updates

Lestropie:
Couple of changes to the documentation suggested by a colleague based on their introductory experience. Mind you, I still dislike some of the explanations; @thijsdhollander you might be able to improve bits.

The fixels / dixels page could probably do with a couple of images; but then again, so could the vast majority of the documentation. Reckon we might need to do a mass screenshot-addition session at some point.


Lestropie:
Might also perform the change discussed in #1099 before merging this one due to the overlap.

thijsdhollander:
While we're at it, I'm adding some other changes based on community input, both online and offline.  I fully agree this will also require a more thorough session at some point (with respect to both images and text).

thijsdhollander:
@bjeurissen, probably worth checking: I've added the `dhollander` response algorithm to the multi-tissue CSD docs page, which should make things much easier for users. I'm getting these questions all the time, sometimes by users who seem to blame the MSMT-CSD technique for a suboptimal outcome, while the input response function obtained from `dwi2response msmt_5tt` were to blame.  I'm hoping this will lead to more positive initial experiences with multi-tissue CSD, both in terms of convenience as well as quality of the output.

thijsdhollander:
I'll be doing some further quick edits soon (this week), to resolve some of the most pressing issues I'm always finding myself re-explaining to users on/off-line.  If others find the time, this may be a good opportunity to do some quick doc updates here and there, potentially based on user (feedback) experience?  The community has been a very good resource for these kinds of things lately; we should put it to good use!

thijsdhollander:
While I'm looking at the docs anyway: I've noticed that the versions of the docs that are automatically built online, are starting to get a little outdated.  Currently we have these versions built:

```
    latest
    stable
    3.0_rc1
    0.3.16
    0.3.15
    0.3.14
    master
    doctest
    better_reporting_for_unsupported_dicom_transfer_syntax
```

A few quick things I can think on here:
* We should add `3.0_rc2`
* Maybe `0.3.14` and `0.3.15` are starting to get a little too outdated to appear in this list?
* I've probably missed this (so there's probably a good reason), but what about `better_reporting_for_unsupported_dicom_transfer_syntax`?

@rtabbara, is it typically you that edits this list if needed; or do we each do it ourselves?  (I don't recall myself doing this... not sure whether this was e.g. something I should've taken care of when `3.0_rc2` happened)

jdtournier:
> I've noticed that the versions of the docs that are automatically built online, are starting to get a little outdated.

Yes, I added these myself - didn't really discuss with anyone, sorry... I think our policy should be to maintain the docs for all releases, including historical ones. I don't think we should be removing older versions unless there is a clear need to. The default will be `master` anyway, and users will need to figure out which version they're after if they need to. 

I'll add `3.0_RC2` and clean up that `better_reporting_for_unsupported_dicom_transfer_syntax` - that was definitely supposed to be transient... :blush:

jdtournier:
By the way, [this is the particular post](http://community.mrtrix.org/t/link-to-earlier-version-of-mrtrix3-fixel-based-pipeline-workflow/878?u=jdtournier) that prompted me to add the documentation for each release explicitly.

thijsdhollander:
> By the way, this is the particular post that prompted me to add the documentation for each release explicitly.

That's a very good motivation indeed; fully agreed to keep each version for that purpose.

> I'll add 3.0_RC2 and clean up that better_reporting_for_unsupported_dicom_transfer_syntax - that was definitely supposed to be transient...

Looks much better now. :relieved: 

While still in the mood to keep looking to clean up as much as possible, and avoid confusing users: now that we're adopting a style relatively close to git flow: aren't `latest`, `stable` and `master` *thematically* becoming one and the same?  `master` and `latest` are already that, but given the great care we take with `master`, the latest version of that branch should be the most "stable" by definition (anything else sitting either in feature branches, or in `dev`).  As always, I'm trying to look at this from a (simply) user point of view: to a lot of them, the word "master" isn't very meaningful.  "latest" though, does make intuitive sense.  And the tagged versions do of course as well, especially with the aforementioned (on the forum) scenario in mind.  So what I guess I'm wondering about: could we just have `latest`, followed by the version tags, followed by `doctest` at the very end for our own purposes, and get rid of the explicit `stable` and `master` in that list?

To add to that: we do have to (very slightly) worry about search engines of course spidering all of the docs, so users potentially clicking their way into an older version of the docs (i.e. one of the version tagged ones).  At the moment, the hints for them looking at a page in such a non-latest version are rather subtle (the URL, which most will overlook, and 2 *very* subtle hints in the menu).  Would it be easy to have something (probably at the top of the menu on the left) that says in some bold red shouting font (yet stylish, if feasible) "you're not looking at the latest version of the docs"?  Apart from the scenario on the forum (where users are very aware that they're on purpose looking at a non-latest version of the docs), most users should probably find themselves looking at the latest version, right?

jdtournier:
> So what I guess I'm wondering about: could we just have latest, followed by the version tags, followed by doctest at the very end for our own purposes, and get rid of the explicit stable and master in that list?

Done. Although technically it would be nice to have an explicit `stable` version too, it doesn't quite work as intended in practice for us. The difference between `latest` and `stable` is that `stable` should point to an actual release, while `lastest` points to `master`, which is an untagged (non-release) version. But the explicit versions are available anyway, and in practice `stable` doesn't quite point where we'd like (although it might in future): it will [only consider tags whose names 'look like' a version number](http://docs.readthedocs.io/en/latest/versions.html#how-we-envision-versions-working), and unfortunately our `3.0_RC1` and `3.0_RC2` tags aren't recognised as such... So I've disabled the `stable` version, along with the `master` version (since as you say `latest` is build from `master` anyway...). 

> Would it be easy to have something (probably at the top of the menu on the left) that says in some bold red shouting font (yet stylish, if feasible) "you're not looking at the latest version of the docs"? 

I'd be very surprised if that could be done... I can't find any relevant settings for this. And we can't modify our sources to do this, since that would involve a commit, and modifying the tag (which is assumed immutable). Not sure it's worth the effort...

jdtournier:
I've just also added a page about software deployment. Just got a question about it, seemed like a good idea to put it out there now...

thijsdhollander:
> Done.

:+1: 

> Although technically it would be nice to have an explicit stable version too, it doesn't quite work as intended in practice for us. The difference between latest and stable is that stable should point to an actual release, while lastest points to master, which is an untagged (non-release) version. But the explicit versions are available anyway, ...

Well, mostly due to that last point, combined with the fact that that last "stable" tag/version appears immediately next to the `latest` version, we should be good.  Users can then also decide for themselves what definition of "stable" they apply.  If we're careful about merging to master, then `latest` should in principle always be the most stable, since it will include the latest (both major and minor) bug fixes that we deem important to make `master` more stable.

Overall, I like how clean the list now is.  The only thing in there that could still realistically confuse end-users is the `doctest` one; but I'm not so worried about it, since it sits at the very end of the list... so that's fine. (and of course, we ourselves still need relatively convenient access to that one)

> I'd be very surprised if that could be done...

No worries, I was just checking if anyone had something easy in mind.  If not, it's not a major concern.

> And we can't modify our sources to do this, since that would involve a commit, and modifying the tag (which is assumed immutable).

Oh, of course, I definitely wouldn't want to do this in the actual documentation itself.  I was more wondering about the existence of a potentially easy way to just highlight the (non-latest) version if that's what's currently being browsed by any user/reader, potentially with some warning kind of message, or colour, or otherwise.  So it was more about a readthedocs feature (if such feature existed) that allows to change the style in function of the version that's being browsed.  But as mentioned, this isn't overly important nor urgent at all.

> I've just also added a page about software deployment. Just got a question about it, seemed like a good idea to put it out there now...

:+1:  I suppose we can treat doc changes like these always like something that can go directly to `master`?  As long as such doc changes/updates aren't with respect to changes or features that are still sitting in `dev`, bringing the docs up to date or otherwise improving them is something that doesn't interfere with the software itself, nor with scripts or workflows that users may have set up, etc...

In any case, if people can think of other updates to the docs and have the time to do so, I'd still encourage to get some in as part of this pull request.

thijsdhollander:
> (one of) the edits I'll be working on is the response function estimation page

Ok, ran out of time today.  I'll do this some time next week.

thijsdhollander:
(update: still far from done obviously; working on it whenever I find a few minutes/seconds)

There's of course still a lot missing from all the documentation in general, but while staring at things, one that I spotted is something on the `mrdegibbs` in the "DWI pre-processing" category, since there's already denoising and distortion correction in there.  @bjeurissen @jdtournier, if you find time of course, it may be good to put something small/simple in there (akin to the denoising, shouldn't be very long I reckon).  The main value of something like that actually being there would probably more be that people are aware that `mrdegibbs` is available (and very valuable!).  So not urgent, I reckon, but still a valuable todo. :slightly_smiling_face:

Lestropie:
Changed PR title since there's a lot of other documentation changes in here now.

thijsdhollander:
Ok, slight warning, and *please don't panic* (as no harm is intended): I'm going to merge this one into the renaming/etc.. branch that has a pull request into `dev`.  As we'll have this RC release soon anyway, I figured it'll be much easier to manage all documentation changes that still may need to happen (e.g. with respect to the SH basis fix, etc..) in `dev`, rather than in bits and pieces in `dev`, and `master`, and then having to merge this all down the line.  So I'll just put this there now, and all doc changes deemed important to be included with the upcoming RC, can all be done nicely over there.

thijsdhollander:
Alright, done.  I'll close this pull request, since all commits have been fast-tracked to `interface_name_changes`, which will hopefully end up soon in `dev`, which will hopefully end up soon in `master`.  :pray: GitFlow

Lestropie:
> readthedocs doesn't update the doctest version of the docs anymore.

It's not updating *any* version any more: All builds are failing as of about 3 weeks ago. Hopefully [this](https://github.com/rtfd/readthedocs.org/issues/3296) fixes it.

jdtournier:
Good catch - looks like it's building fine now. :+1:

PR #1103, opened 2017-08-23
TravisCI: Incorporate documentation testing

Lestropie:
Running a little experiment that requires creating a PR to see if it works.

Within the TravisCI tests:

- Run `sphinx-build` on the `docs/` directory, and ensure that the documentation compiles without any warnings or errors.

- Run `generate_user_docs.sh`, and fail testing if the generated documentation files are not identical (i.e. the developer has forgotten to run this script, and hence a mismatch remains between the command-based help pages and the documentation reference pages).


Lestropie:
:champagne: :rofl: That was fun. 

@jdtournier: You'll probably like [this line](https://github.com/MRtrix3/mrtrix3/pull/1103/files#diff-354f30a63fb0907d4ad57269548329e3R29)... :nerd_face:

.travis.yml
```diff
@@ -9,22 +9,30 @@ addons:
...
-  - export PATH=`pwd`/release/bin:`pwd`/scripts:${PATH}
+  - export PATH=`pwd`/bin:${PATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
   - (cd ..; hg clone https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
+  - pip install recommonmark sphinx sphinx-rtd-theme
+  - pip3 install recommonmark sphinx sphinx-rtd-theme
 script:
+  - ${py} -m sphinx -n -N -W -w sphinx.log -c docs/ docs/ tmp/
   - ./check_memalign && $py ./configure -assert && $py ./build -nowarnings && ./run_tests
+  - ./docs/generate_user_docs.sh && ! ( git diff -- . ':(exclude)docs/reference/commands/mrview.rst' ':(exclude)docs/reference/commands/shview.rst' | tee gitdiff.log | grep '^-\|^+' | grep -q -v 'commands_list.rst\|mrview\|shview' )
```
jdtournier: :nerd_face:  nice use of `tee` - a rarely used little gem...

jdtournier: By the way, I reckon it would be cleaner to use the `--stat` or `--numstat` option to `git diff` to check for differences - much cleaner to parse (just need to check for non-empty output)... Something like:
```
./docs/generate_user_docs.sh && ( 
  gitdiff=$(git diff --numstat -- docs ':!docs/reference/commands/??view.rst' ); 
  [[ $gitdiff ]] && ( 
    git diff > gitdiff.log; 
    echo "documentation is not up to date - please run docs/generate_user_docs.sh and commit again"; 
    echo $gitdiff 
  ) 
)

```

Does this achieve the same thing...?

jdtournier: Actually, your use of `--exit-code` is even cleaner - but `--quiet` is probably better still (implies `--exit-code`):
```
./docs/generate_user_docs.sh && 
git diff --quiet -- docs ':!docs/reference/commands/??view.rst' || ( 
  echo "documentation is not up to date - please run docs/generate_user_docs.sh and commit again"; 
  git diff --stat -- docs 
  git diff -- docs > gitdiff.log; 
)
```

jdtournier: Not sure I follow... But in any event, I'm trying to get the GUI commands to work as expected with these options even without a display, which would obviate the need for this.

Lestropie: Spoil sport... :-P

jdtournier:
Guess that did it then...? Slightly concerned about whether the changes I made may impact on some other aspects though, since it slightly changes the order of execution for the various initialisations (config file loading, running the `usage()` function, argument parsing, etc.). I had a quick go a checking that argument parsing still worked for `mrview`, and clearly it works OK for command-line apps, but there may be some interactions with config-file options that may be affected (issues like [this one](https://github.com/MRtrix3/mrtrix3/pull/526#issuecomment-208088706), for example). Might be worth double-checking before merging this...

Lestropie:
Only one I could find was [here](https://github.com/MRtrix3/mrtrix3/blob/master/core/progressbar.cpp#L154): Calling `DEBUG()` before `parse_standard_options()` would never have an effect.

I also moved the `AUTHOR` and `SYNOPSIS` checks to their own function `verify_usage()`: Although that check should occur after `usage()` and before `::MR::GUI::App()`, they don't really belong inside something called "`parse_special_options()`". So they get their own little home.

jdtournier:
Seems good to me - are you ready to merge otherwise? Go for it if it's good to go from your point of view.

PR #1104, opened 2017-08-24
system-wide config file location can now be overridden by MRTRIX_CONFIGFILE environment variable

jdtournier:
This is prompted by a HPC sysadmin who prefers to keep all files related to a given package in a single location (likely to be a very common scenario). This is fine for everything apart from the config file, which is hard-coded to `/etc/mrtrix.conf`. It's difficult to see how the binaries could be modified to auto-detect the location of their config file, so it looks like an environment variable might be the only way to allow this. It's a trivial change, seems to work on my system.

Still need to fix up the scripts to match though...

### DIFF
--- a/core/file/config.cpp
+++ b/core/file/config.cpp
@@ -32,17 +32,21 @@ namespace MR
 
     void Config::init ()
     {
-      if (Path::is_file (MRTRIX_SYS_CONFIG_FILE)) {
-        INFO ("reading config file \"" MRTRIX_SYS_CONFIG_FILE "\"...");
+      const char* sysconf_location = getenv ("MRTRIX_CONFIGFILE");
+      if (!sysconf_location) 
+        sysconf_location = MRTRIX_SYS_CONFIG_FILE;
+
+      if (Path::is_file (sysconf_location)) {
+        INFO (std::string("reading config file \"") + sysconf_location + "\"...");
         try {
-          KeyValue kv (MRTRIX_SYS_CONFIG_FILE);
+          KeyValue kv (sysconf_location);
           while (kv.next()) {
             config[kv.key()] = kv.value();
           }
         }
         catch (...) { }
       } else {
-        DEBUG ("No config file found at \"" MRTRIX_SYS_CONFIG_FILE "\"");
+        DEBUG (std::string ("No config file found at \"") + sysconf_location + "\"");
       }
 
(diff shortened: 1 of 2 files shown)


jdtournier:
@Lestropie, could you confirm whether 6290ae6 is sufficient for the scripts...?

jdtournier:
On a different note: this is technically a new feature, not a bug fix as such, so really we should be merging this to `dev` with a view to inclusion in the next release. But this is currently holding up a HPC sysadmin, so could be argued to be a fix of some sort... And it's a relatively innocuous change, so if everyone is OK with this, I reckon it's OK to merge direct to `master`. Thoughts...?

jdtournier:
> I've got the feeling that applying this git flow thing strictly isn't going to be our strongest point.

Yes, I think it'll take some getting used to... Part of the problem is that it's not clear yet how we should be using this. I reckon we really need to document this somewhere - I think I'd mentioned one option might be as a page in the [developer documentation](http://www.mrtrix.org/developer-documentation/). I've mentioned different (although very similar) models in #1052, but I think the one we'll end up using is probably what I suggest in [this particular post](https://github.com/MRtrix3/mrtrix3/pull/1052#issuecomment-315769793) (towards the end of it in particular). Essentially, that would entail using `master` more or less as we do now, but push compatibility-breaking changes or major new features to `dev` - basically anything that requires an announcement and a tag (I reckon a major new feature needs a tag, just so we can say "feature X is available as of version X.Y.Z" or something - #1013 is a case in point). I'll try to add a page to the dev docs about this at some point, we can discuss the details in the corresponding pull request. 

> So long story short: happy with whatever, as long as it applies equally for everyone and everyone's features/contributions. 

I'd hate for anyone to feel that way - which is why I raised the issue for discussion in my previous post. But if the workflow turns out to create these kinds of issues, then we'll need to fix it. There will always be some degree of subjectivity here, but as long as we're all engaged in the discussion, I reckon it'll work out OK. 🤞

bjeurissen:
I noticed that if I set MRTRIX_CONFIGFILE, that particular config file is indeed read, but then the default one in my home-directory  is subsequently read, overriding my custom config file:

```
mrinfo: [INFO] reading config file "/Users/ben/test_project/mrtrix.conf"...
mrinfo: [INFO] reading config file "/Users/ben/.mrtrix.conf"...
```

which seems to beat the purpose of being able to set a custom config file.

PR #1106, opened 2017-08-25
Doxyfile: update input folders to list 'core' rather than 'lib'

jdtournier:
The [developer documentation](http://www.mrtrix.org/developer-documentation/) is currently missing heaps of stuff since `lib/` has been moved to `core/`...


jdtournier:
BTW, this change only modifies `Doxyfile` - it has no impact on the code whatsoever, perfectly safe to merge to `master`.

PR #1108, opened 2017-08-29
Changes to enforce requested thread counts

### LINKED ISSUE #1105: computational requests not matched by commands
I'm running into issues with the amount of cores allocated by mrtrix commands, does not actually match those specified by `nthreads` option.

In particular, when I run _tcksift2_ with:

`tcksift2 25M.tck wm.mif sift_weightfactor.txt -act r5TT.nii -fd_scale_gm -force -nthreads 8
`

I've set a watcher (see attached) which records the amount cpu's being used, as now indicated by the first column:

`911 tcksift2 25M.tck wm.mif sift_weightfactor.txt -act r5TT.nii -fd_scale_gm -force -nthreads 8
`

As you can see it's over the specified limit, and has been an issue I've encountered across multiple commands. This is a snapshot at each second, and so I've calculated all the instances where the cpu's are over-specified:

> [root@hpcnode040 ~]# awk '$3>=900' watcher.20170823.second | wc -l
> 2398 (about 40 minutes)

This issue could be data-dependent as the segmentations/co-registrations weren't entirely accurate, but nevertheless, this is sub-optimal when running on an institute-wide cluster. 

Just a few technical details:

>  CentOS 7 running kernel 3.10.0-327.3.1.el7.x86_64
>  File system is an nfs mount of an xfs file system
> 
> [root@hpcnode040 ~]# ldd $(which tcksift)
> linux-vdso.so.1 =>  (0x00007ffe8ff82000)
> libmrtrix-3.0_RC2.so => /software/mrtrix3/mrtrix3-20170815/bin/../lib/libmrtrix-3.0_RC2.so (0x00007f36b9141000)

Best,

Alistair
[watcher.20170823.second.txt.zip](https://github.com/MRtrix3/mrtrix3/files/1250810/watcher.20170823.second.txt.zip)

Lestropie:
As discussed in #1105.

- At the track mapping stage of `tcksift` / `tcksift2`, multiple threads were being invoked for both the streamline-voxel mapping and the streamline-fixel visitation encoding, resulting in twice as many threads being spawned as what was requested. These two tasks are now combined into a single functor such that the requested number of threads is honoured.

- In `tckgen -seed_dynamic`, multi-threading both the tracking and streamline-voxel mapping functors could result in the use of more resources than had been requested. In these changes, the back-end keeps track of whether the number of threads was explicitly requested by the user, or whether it was set automatically based on the detected hardware concurrency: If an explicit request was made, then each of these stages will run with half of the requested number of threads, such that the total requested resources are not (substantially) exceeded; whereas if no explicit request is made, then the maximum number of threads will be spawned for both stages, with the back-end then being responsible for sleeping & waking the different threads.

- Another small little gotcha I came across in here was that the value of `__number_of_threads` was being initialised to 0 such that it would be determined appropriately on the first call to `number_of_threads()`, yet a value of 0 for this variable is totally legitimate. In this use case, the checks in `number_of_threads()` would be repeated on each call, rather than using the pre-determined value value. So here I instead explicitly flag this property as being uninitialised, and its appropriate value is determined once only.

@AlistairPerry, @CanWood: Let us know if you have found any more instances of this type of behaviour. This fix won't go out straight away, but it'll be a part of the next release (whether another release candidate is required, or the software release proper).

Closes #1105.


CanWood:
Thanks for this!  We fetched this branch and Alistair has been running a test run since yesterday.  His `tckgen -seed_dynamic` stage has been running for a bit with 8 threads requsted but 4 in use.  I presume he's only stressing one of either the tracking or streamline.  We'll let you know how we go when he hits the tcksift stage.

Lestropie:
Yes, the dynamic seeding is a tricky one. The alternative would be to assign N threads to the tracking, and just one thread to the mapping; a more efficient tracking algorithm that spits out more tracks per unit time would however need more mapping threads, hence why that stage currently multi-threads in `master`. But maybe that would be more appropriate, rather than half and half, given the computational expense of the default tracking algorithm.

Did you previously observe a peak usage substantially over the requested amount specifically for `tckgen -seed_dynamic`, or did I jump the gun on altering that one?

CanWood:
Sorry for the late reply.  Been off for a long weekend.  

There wasn't much of a peak over 8 with `tckgen -seed_dynamic` before this fix.  It sat nicely in its 8+abit before your change.  With this branch it sat at 4+abit.  That said, I'm not familiar with the data (nor the program to be honest as I'm just the sys-amin trying to wrangle resources) so I can't say if your fix is more correct for the general case with Alistair's data using half resources being an exception.  

All of that said, Alistair's pipe took longer than before (as expected if he was only using 4 threads for `tckgen -seed_dynamic`) but was much better with the thread counts.  In a 126997 second job, he went over 9 cores for six seconds and over 10 cores for only three.  That, I can live with.

```
10212 46432 1010           mrmath mesh2pve_L_Accu.mif mesh2pve_R_Accu.mif mesh2pve_L_Caud.mif mesh2pve_R_Caud.mif mesh2pve_L_Pall.mif mesh2pve_R_Pall.mif mesh2pve_L_Puta.mif mesh2pve_R_Puta.mif mesh2pve_L_Thal.mif mesh2pve_R_Thal.mif mesh2pve_L_Amyg.mif mesh2pve_R_Amyg.mif mesh2pve_L_Hipp.mif mesh2pve_R_Hipp.mif sum - -nthreads 8 -force
10212 46432  913           mrmath mesh2pve_L_Accu.mif mesh2pve_R_Accu.mif mesh2pve_L_Caud.mif mesh2pve_R_Caud.mif mesh2pve_L_Pall.mif mesh2pve_R_Pall.mif mesh2pve_L_Puta.mif mesh2pve_R_Puta.mif mesh2pve_L_Thal.mif mesh2pve_R_Thal.mif mesh2pve_L_Amyg.mif mesh2pve_R_Amyg.mif mesh2pve_L_Hipp.mif mesh2pve_R_Hipp.mif sum - -nthreads 8 -force
10492 46432 1063           mrcat cgm.mif sgm.mif wm.mif csf.mif path.mif - -axis 3 -nthreads 8
10492 46432  932           mrcat cgm.mif sgm.mif wm.mif csf.mif path.mif - -axis 3 -nthreads 8
10647 42470  942         5tt2gmwmi 5TT.nii 5TTgmwmi.nii -nthreads 8 -force
23735 23589 1176           /software/mrtrix3/mrtrix3-nthreads_fix/bin/mrtransform 5tt.mif 5tt_regrid.mif -template fa.mif -interp linear -nthreads 8
```
As always, thanks for responsiveness and attention.

Lestropie:
OK, thanks for the feedback. In that case, I'm reverting the dynamic seeding to still use the maximum permissible number of threads, in order to run as quickly as possible. Hopefully either the small excess usage is not a problem, or the user can simply use e.g. `-nthreads 7` for that command instead.

@jdtournier I've left the changes to the threading back-end in case it's useful in the future, and it also prevents repeated initialisation in the case of `-nthreads 0`.

core/thread.h
```diff
@@ -233,26 +233,36 @@ namespace MR
...
     /*! the number of cores to use for multi-threading, as specified in the
      * variable NumberOfThreads in the MRtrix configuration file, or set using
      * the -nthreads command-line option */
     size_t number_of_threads ();
 
+    /*! provides information regarding whether the number of threads has been
+     * initialised, set explicitly, or determined implicitly. This may affect
+     * how particular algorithms choose to launch threads depending on the
+     * presence of a user request. It also prevents repeated testing of
+     * config file / environment variables, since 0 was previously used both
```
jdtournier: I don't think documenting the fact that the code was previously buggy really belongs in the docs...? I totally agree that was a bug, but it's already clearly described in the git history, which is where it should be, IMO...

Lestropie: True enough. Probably trying to prove to myself at the time that I wasn't overlooking anything...

jdtournier: :+1:

PR #1111, opened 2017-09-02
mrview: roi save QMessageBox: overwrite standard buttons for qt4 compatibility

### LINKED ISSUE #1109: mrview: popup window with query to save ROI but with only OK button
When user creates a ROI in `mrview` and attempts to quit `mrview` without saving ROI, the popup window with notification that ROI has been modified but not saved and question whether user wants to save ROI appears, but there is only one option for user - OK button. When user clicks on OK, ROI is actually NOT saved to file and therefore it is lost.

maxpietsch:
This fixes #1109. Not merging as I am not sure whether this should go to master instead.

### DIFF
--- a/src/gui/mrview/tool/roi_editor/roi.cpp
+++ b/src/gui/mrview/tool/roi_editor/roi.cpp
@@ -32,7 +32,7 @@ namespace MR
       namespace Tool
       {
 
-            
+
 
 
         ROI::ROI (Dock* parent) :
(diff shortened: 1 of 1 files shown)


jdtournier:
Agree that should probably go to `master` - it's innocuous enough, can't possibly affect anyone's analysis, and fixes an overt bug.

jdtournier:
OK, changing base to `master` on the UI is a bad idea when the pull request is itself based on `dev` - means everything currently on `dev` gets pulled in too. This commit will need to be cherry-picked onto a new branch from  `master`, and a new pull request created for it... Assuming of course everyone is happy for this fix to go straight to `master`...?

Lestropie:
I'd go to `master`. We'll need to re-revise our branch & merging conventions after release, but under RC conditions I'm happy having overt fixes sent straight to `master`.

maxpietsch:
Created a PR to master in #1113.

Issue #1107, opened 2017-08-26
Error building test suite


justbennet:
I just built 3.0_RC2 from source.  It build cleanly, but it errors when I try to `./run_tests`.  I've pasted the `testing.log` below.

I can run `mrconvert mrconvert/in.mif -datatype float32 tmp.nii.gz` from the `testing/data` folder, and it produces `tmp.nii.gz`, which seems to be a valid NIfTI file when I read it with `nibabel`.

Have I done something egregiously wrong?

```
$ cat !$
cat testing.log
-------------------------------------------
  Testing MRtrix3 installation
-------------------------------------------


-------------------------------------------

## building test commands... 


ERROR: ( 8/22) [CC] tmp/cmd/testing_diff_image.o

g++ -c -std=c++11 -pthread -fPIC -DMRTRIX_WORD64 -DMRTRIX_TIFF_SUPPORT -DEIGEN_FFTW_DEFAULT -I /sw/arcts/centos7/fftw/3.3.4-gcc-4.9.3/include -L /sw/arcts/centos7/fftw/3.3.4-gcc-4.9.3/lib -l fftw3 -lm -fopenmp -Wall -O3 -DNDEBUG -Isrc -I../src -I../core -Icmd -isystem /usr/include/eigen3 cmd/testing_diff_image.cpp -o tmp/cmd/testing_diff_image.o

failed with output

In file included from src/diff_images.h:23:0,
                 from cmd/testing_diff_image.cpp:20:
../core/image_diff.h: In instantiation of struct MR::check_images_abs(ImageType1&, ImageType2&, double) [with ImageType1 = MR::Image<std::complex<double> >; ImageType2 = MR::Image<std::complex<double> >]::<lambda>:
../core/image_diff.h:50:7:   required from void MR::check_images_abs(ImageType1&, ImageType2&, double) [with ImageType1 = MR::Image<std::complex<double> >; ImageType2 = MR::Image<std::complex<double> >]
src/diff_images.h:56:50:   required from void MR::Testing::diff_images(ImageType1&, ImageType2&) [with ImageType1 = MR::Image<std::complex<double> >; ImageType2 = MR::Image<std::complex<double> >]
cmd/testing_diff_image.cpp:46:33:   required from here
../core/image_diff.h:51:67: error: const qualifiers cannot be applied to MR::Image<std::complex<double> >&
       .run ([&tol] (const decltype(in1)& a, const decltype(in2)& b) {
                                                                   ^
[...]

PR #1114, opened 2017-09-04
fix const-qualifier issue in image_diff.h

jdtournier:
As discussed in #1107.

### DIFF
--- a/core/image_diff.h
+++ b/core/image_diff.h
@@ -48,7 +48,7 @@ namespace MR
     {
       check_headers (in1, in2);
       ThreadedLoop (in1)
-      .run ([&tol] (const decltype(in1)& a, const decltype(in2)& b) {
+      .run ([&tol] (const ImageType1& a, const ImageType2& b) {
           if (std::abs (cdouble (a.value()) - cdouble (b.value())) > tol)
             throw Exception ("images \"" + a.name() + "\" and \"" + b.name() + "\" do not match within absolute precision of " + str(tol)
                  + " (" + str(cdouble (a.value())) + " vs " + str(cdouble (b.value())) + ")");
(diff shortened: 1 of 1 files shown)


jdtournier:
OK, cherry-picked commit and changed base to `master`. I agree it might as well go there, it is completely harmless for users, only affects testing...

jdtournier:
Seems there is an issue with the changes in #1103:
```
$ ./docs/generate_user_docs.sh && git diff --exit-code docs/ > gitdiff.log || echo "Documentation is not up to date; run docs/generate_user_docs.sh and commit changes"
./docs/generate_user_docs.sh: line 24: reference/commands_list.rst: No such file or directory
find: `../cmd/': No such file or directory
find: `../bin/': No such file or directory
./docs/generate_user_docs.sh: line 101: ./format_config_options: No such file or directory
Documentation is not up to date; run docs/generate_user_docs.sh and commit changes
```
I reckon we should revert that command to run within the `docs/` folder...

Lestropie:
The changes in #1103 are to make it possible to run that script from *any* directory. If that's not working, it's something I need to fix in `dev`; mind you, I've no idea how it could have passed testing in the first place if that line weren't working...?

Lestropie:
You might need to check exactly what you're running there: The current version of `generate_user_docs.sh` in `dev` [clearly](https://github.com/MRtrix3/mrtrix3/blob/dev/docs/generate_user_docs.sh) has the requisite path determinations, and wouldn't produce the errors you're getting. So you're running the TravisCI test from #1103 without the corresponding changes that came with #1103 to make it possible. I think.

jdtournier:
Well this is straight off the current `master`, so not sure what's going on... You'll not there's only my commit in the diff here.

Lestropie:
My guess is it's an issue with the PR re-basing then. The docs test is clearly [not present](https://github.com/MRtrix3/mrtrix3/blob/master/.travis.yml) in `master`, and as you say there's nothing listed in the diff of this PR outside of your one commit. So maybe create a fresh PR?

jdtournier:
Yes, strange... I'll look into it later today when I have a minute.

jdtournier:
OK, looks like TravisCI is getting confused by the forced push, and is trying to build the old commit (or something else altogether - there is no commit b0d6574, should be 92e57e7). Seems [others have also encountered issues with this](https://github.com/travis-ci/travis-ci/issues/6991). I'll close this pull request and create a fresh one, see if TravisCI picks it up properly this time...

Issue #761, opened 2016-08-22
Sticky move in mrview


maedoc:
I just got a shiny mac, and for some two touch combination on the touchpad, mrview gets stuck moving the image around in the window and no amount of clicking or esc seems to stop it; I usually have to quit and relaunch.

I compiled recently from Git (I can pull, recompile and test when plugged into AC again) against Qt5 from brew, in case that helps. Maybe it's a glitch in Qt or Mac, not mrview. Any tips would be helpful.

maxpietsch:
I can confirm this issue on my mac. I also had to close mrview. It happened often when drawing ROIs via touch pad which might be coincidental but makes this bug very annoying... From memory, it might have happened after something like a pan gesture via [tap and hold](http://doc.qt.io/qt-5/qtapandholdgesture.html) but today it proves hard to reproduce. I have Qt 5.6.1.

jdtournier:
I reckon @maxpietsch is right that it might be related to the handling of the pan gesture in Qt5. As far as I can tell, this is activated via a tap and hold (i.e. place your finger on the touchpad and leave it there without clicking or moving for a second or so), and then drag around to pan. It should revert back to normal behaviour if you take your finger off the touchpad for a second or so. 

That said, on the few occasions where I've actually had to use MRView on a Mac, I've found that this feature gets in the way more than it helps. In terms of getting stuck in some unusable state, I'm guessing the most likely explanation is that mouse press events that would have been propagated to the appropriate handlers are being intercepted by the gesture handling backend, which can cause trouble if for example our code registered a mouse click event that starts pan mode, but fails to receive the corresponding release event due to it being usurped by the gesture handler. That would prevent MRView from coming out of pan mode, for instance. 

So two options: get rid of the pan gesture (no great loss IMO), and/or make sure the gesture handler resets any structures related to mouse click events (probably wise to do this even if we get rid of the pan gesture, actually, just to avoid similar interactions with the pinch to zoom and rotate gestures...).

maedoc:
Thanks for describing the mechanics of the gesture: at least now that I can do it intentionally, it doesn't seem to get stuck.. pebcak?

jdtournier:
Maybe... :Wink:

But it does sound like there might be an issue with its interaction with the ROI tool. Definitely worth taking a look. And I'm open to feedback on whether that pan gesture is actually useful in any way. I don't have a Mac, so I don't have much experience with actually using it...

maxpietsch:
[...]

PR #1115, opened 2017-09-05
mrview: removed updateGL() after pan_event. 

maxpietsch:
proposed fix for #761

The updateGL()  in the pan_event causes pan gestures on slow systems to remain in state Qt::GestureUpdated, never reaching Qt::GestureFinished on macOS qt 5.9.1 

After and during panning, the crosshair is often displaced from the mouse cursor. If I had to guess the reason for the gesture not finishing, I'd say it has to do with the window not reaching the cursor position if the mouse is moved too quickly.

Removing the updateGL seems to have no negative side effects but would be good to double check that.


Lestropie:
Could not reproduce any faults with the modified code, but I also couldn't reproduce the reported issue with the current `dev` branch on any system. So happy for this to be merged in if it yields a benefit at your end.

src/gui/mrview/mode/base.cpp
```diff
@@ -192,7 +192,7 @@ namespace MR
           const Projection* proj = get_current_projection();
           if (!proj) return;
           set_target (target() - proj->screen_to_model_direction (window().mouse_displacement(), target()));
-          updateGL();
+          // updateGL(); # updateGL() causes pan gestures to remain in state Qt::GestureUpdated, never reaching Qt::GestureFinished on macOS
```
jdtournier: Happy to remove if no side effects - but I think you'd be better off deleting the line altogether, and placing the comment you've added in the git history. I'd rather keep the code free of unnecessary legacy, and document the rationale for changes in the history - that's what it's for... :wink:

PR #1116, opened 2017-09-05
fix const-qualifier issue in image_diff.h

jdtournier:
As discussed in #1107.

This is take 2 - see original in #1114.


jdtournier:
OK, looks like this one worked OK. Just need someone to approve, and it should be good to go.

PR #1117, opened 2017-09-05
Use tmp prefix in testing for population template

jdtournier:
This had been bugging me for a while now, so I finally did something about it...

Issue was that `population_template` would invariably fail a test on my system. The issue was an incompatibility with Python3, due to writing a `str` to a binary file, without converting to `bytes` via `encode()` (fix in this pull request). What this did was to cause `population_template` to leave its temporary folder behind, and since that folder was not being deleted by `run_tests` either, this left untracked content in the testing repo, which would show up in the `git status` of the main repo. So I'd be doing `./run_tests` on a clean repo, and a subsequent `./build` would change the version string to add `-dirty` to it, simply because `population_template` had failed a test...

These changes rely on the fact that `run_tests` will clean up any files or folders that start with `tmp` - a feature that most tests rely on. Also, any files prefixed with `tmp` are ignored (or at least they should have been - just [fixed that too on the testing repo](https://github.com/MRtrix3/test_data/commit/5e0f3ccb63ef2bdea941092ccd017b4af94dcc5d)). So the output file and the (now explicitly named) temporary folder both now start with `tmp`, and there is no need to manually delete the output any more.

### DIFF
--- a/bin/population_template
+++ b/bin/population_template
@@ -59,7 +59,7 @@ except ImportError:
       fmt = [fmt, ] * ncol
       format = delimiter.join(fmt)
       for row in X:
-        fh.write((format % tuple(row) + '\n'))
+        fh.write((format % tuple(row) + '\n').encode (errors='ignore'))
     finally:
         fh.close()
 
--- a/testing/tests/population_template
+++ b/testing/tests/population_template
@@ -1,2 +1,2 @@
 population_template population_template/in tmp-population_template-out.mif.gz -type rigid_nonlinear -rigid_scale 0.2,0.3 -rigid_niter 20,20 -nl_scale 0.2,0.3 -nl_niter 1,1 -tempdir tmp-population_template -force 
-population_template population_template/in tmp-population_template-out.mif.gz -type nonlinear -nl_scale 0.2,0.3 -nl_niter 1,1 -tempdir tmp-population_template -force 
+population_template population_template/in tmp-population_template-out.mif.gz -type nonlinear -nl_scale 0.2,0.3 -nl_niter 1,1 -tempdir tmp-population_template2 -force 
--- a/testing/tests/population_template
+++ b/testing/tests/population_template
@@ -1,2 +1,2 @@
-population_template population_template/in population_template/out.mif.gz -type rigid_nonlinear -rigid_scale 0.2,0.3 -rigid_niter 20,20 -nl_scale 0.2,0.3 -nl_niter 1,1 -force && rm population_template/out.mif.gz  
-population_template population_template/in population_template/out.mif.gz -type nonlinear -nl_scale 0.2,0.3 -nl_niter 1,1 -force && rm population_template/out.mif.gz  
+population_template population_template/in tmp-population_template-out.mif.gz -type rigid_nonlinear -rigid_scale 0.2,0.3 -rigid_niter 20,20 -nl_scale 0.2,0.3 -nl_niter 1,1 -tempdir tmp-population_template -force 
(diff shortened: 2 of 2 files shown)


jdtournier:
By the way, I have a sneaky suspicion that there will be quite a few other Python3 compatibility issues in there - I've only dealt with the one that was causing issues for me... 

The other slight concern was that the output is not clean when redirected to file. Here's what it looks like in `testing.log`:
```
## running "population_template"...

# command: population_template population_template/in tmp-population_template-out.mif.gz -type rigid_nonlinear -rigid_scale 0.2,0.3 -rigid_niter 20,20 -nl_scale 0.2,0.3 -nl_niter 1,1 -tempdir tmp-population_template -force [ ok ]
population_template: numpy not found. using replacement functions
population_template: ESC[03;32mGenerating a population-average template from 2 input imagesESC[0m
population_template: ESC[00;31m[WARNING] no masks input. Use input masks to reduce computation time and improve robustnessESC[0m
population_template: ESC[03;32minitial alignment of images: massESC[0m
population_template: ESC[03;32mlinear registration stages:ESC[0m
population_template: ESC[03;32m(00) rigid     scale: 0.2000, niter: 20, no reorientationESC[0m
population_template: ESC[03;32m(01) rigid     scale: 0.3000, niter: 20, no reorientationESC[0m
population_template: ESC[03;32mnonlinear registration stages:ESC[0m
population_template: ESC[03;32m(00) nonlinear scale: 0.2000, niter: 1, no reorientationESC[0m
population_template: ESC[03;32m(01) nonlinear scale: 0.3000, niter: 1, no reorientationESC[0m
population_template: ESC[03;32mGenerated temporary directory: /home/jdt13/mrtrix3/testing/data/tmp-population_templateESC[0m
population_template: ESC[03;32mChanging to temporary directory (/home/jdt13/mrtrix3/testing/data/tmp-population_template)ESC[0m
population_template: ESC[03;32mGenerating initial templateESC[0m
ESC[03;36mCommand:ESC[0m  mraverageheader /home/jdt13/mrtrix3/testing/data/population_template/in/s1.mif.gz /home/jdt13/mrtrix3/testing/data/population_template/in/s2.mif.gz average_header.mif -fill
ESC[03;36mCommand:ESC[0m  mrregister /home/jdt13/mrtrix3/testing/data/population_template/in/s1.mif.gz average_header.mif -rigid_scale 1  -rigid_niter 0  -type rigid  -noreorientation  -rigid_init_translation mass  -datatype float32 -rigid linear_transforms_initial/s1.txt
...
```
Not sure what's handling this output (`app.console()`?), but it might be worth thinking about the kinds of gymnastics we had to implement in #686 to deal with this in the main code... Could detect redirection using [`sys.stderr.isatty()`](https://docs.python.org/2/library/stdtypes.html#file.isatty) - [as used in the build script](https://github.com/MRtrix3/mrtrix3/blob/master/build#L178), for example?

Lestropie:
Terminal output comes from a variety of places. Disabling control codes based on `sys.stderr.isatty()` is fairly easy; the trickier part will be having `class progressBar` behave appropriately (it's currently unused in this repo I think, but only because I haven't integrated it yet).

jdtournier:
I didn't realise there was a progressbar class for the scripts - but if it's unused, and dealing with the current `app.console()` calls is easy, might be worth including in this pull request? Otherwise, we might as well merge, to be fair the changes here were originally nothing to do with this...

Lestropie:
I didn't quite manage to finish the progress bar fixes / integration yesterday; some inconvenient individual called a meeting :-P  Will fix it over at #1049 along with all my other junk. Might as well keep this PR on-topic.

Lestropie:
Issue with output from scripts going to text file that was mentioned here is now fixed in #1049.

PR #1121, opened 2017-09-12
Update dev from master

jdtournier:
This will need to be done from time to time - see discussion on #1120...

PR #1123, opened 2017-09-13
TravisCI: Try explicitly requesting "Trusty"

Lestropie:
Examining testing issue raised in #1121.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ install:
   - export PYTHONPATH=`pwd`/lib:${PYTHONPATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
   - (cd ..; hg clone https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
+  - pip install urllib3[secure]
   - pip install pylint recommonmark sphinx sphinx-rtd-theme
   - pip3 install pylint recommonmark sphinx sphinx-rtd-theme
 script:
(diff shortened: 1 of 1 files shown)


Lestropie:
From [here](https://docs.travis-ci.com/user/reference/overview/): Suspect the issue has arisen due to TravisCI changing its default from Precise (12.04) to Trusty (14.04) in the last month.

Lestropie:
Solution implemented in #1121.

Lestropie:
For the record, there was two separate issues reported in #1121:

```
OSError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/isort'
```
Fixed by adding `--user` option to the `pip install` calls to change the install location. It's possible to enable `sudo` in TravisCI, but that [results](https://docs.travis-ci.com/user/reference/overview/) in a VM-based environment rather than a container, and our tests correspondingly failed due to timeout.

```
/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
```
Seems to be fixed by the line `pip install urllib3[secure]`, which surprised me slightly since [the documentation](https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2) says i t needs to be explicitly enabled.

PR #1125, opened 2017-09-13
fixelcfestats: fix import of file list

jdtournier:
Use the MRtrix3 provided strip() function to remove trailing non-printing characters, and ignore empty lines.

This was prompted by [this forum post](http://community.mrtrix.org/t/fixelcfestats-error-with-dicom-images/1140/6). Simple fix, worth merging to `master` directly?


jdtournier:
I note there are no tests for `fixelcfestats`... I haven't tested these changes myself, it would be good if someone with data ready to go could verify that nothing falls over - or better, add a few tests... :grin:

Lestropie:
Fixed. Testing for an empty string works best *before* concatenation :+1:

Auto-testing `fixelcfestats` is going to be non-trivial. As mentioned in #1030 I'm starting with automatic testing of `vectorstats` to verify the GLM basic operation, particularly now with support for multiple rows in the contrast matrix; this will hopefully expand over time as I add that functionality to the other statistical inference commands, to verify that they're working "correctly". But `fixelcfestats` is at the end of that list...

PR #1129, opened 2017-09-20
Updating dev

thijsdhollander:
I've reused the latest hot fix branch (since we were starting to lag again a bit), which automatically brings along the backlog of hot fixes.

I thought I'd be clever and also try to reuse the pull request of that one, but clicking the edit button in that pull request, I discovered I can't change the base branch from `master` to `dev`...  I suppose this makes sense since the pull request had already resulted in a successful merge to `master`.  It'd probably be weird for a single pull request to document merges to 2 different base branches.  It's a pity though, it would've been nice to circumvent having to create a different pull request; and would also document the git flow way of things better.  Well, nothing we can do about that, I suppose...


thijsdhollander:
So order of things as I did them:

1. restored the `fixelcfestats_fix_file_list_import` (latest hot fix) branch; this would in a normal (git) flow not be necessary if we don't delete it in the first place of course :slightly_smiling_face: 
2. merge `dev` into `fixelcfestats_fix_file_list_import`
3. create this pull request for `fixelcfestats_fix_file_list_import`, but be careful to (upon creation of the pull request on github) immediately select `dev` as the base (by default `master` was selected, I suppose because that's where `fixelcfestats_fix_file_list_import` originally branched off)
4. as I'm typing this, wait for the tests to run; they should be the `dev` version of the tests, since `dev` is already merged in
5. upon hopefully successful completion of the tests, hit the button and `dev` should be up to date

thijsdhollander:
So we should in principle do this for each hot fix, but the above of course implies it's not a disaster if we forget one; nothing special needs to be done to bring it along with any future hot fix.  This does of course increase the risk of compatibility problems with `dev`, and in a way it's not being nice to future developers of hot fixes, since they may be forced to deal with your "mess" (should you have created a mess; let's don't assume we create messes by default :smile: ).

thijsdhollander:
Ok, that seems to check out.  I'll hit the button, so things don't linger around in this intermediate stage.

Issue #810, opened 2016-11-02
tcknormalise: Naming


Lestropie:
I'm currently re-jigging some streamlines mapping / manipulation code. One command that currently strikes me as odd naming, probably because it existed before _MRtrix3_ had image registration, is `tcknormalise`. This is probably the only reference to 'normalisation' in a spatial sense... We tend to talk about 'warps' and 'transforms', with 'normalisation' instead referring to intensity modulation to a common scale.

Would people be happy with renaming to `tckwarp`? Only trouble with that is that I'd probably want to make it compatible with affine transformations also, and it'd seem wasteful to me to also have a `tcktransform` as a separate command, whereas buried within `tckwarp` it might not get spotted. The functionality definitely _shouldn't_ be in `tckedit`, as that introduces an order dependence problem.

draffelt:
Can we just have one `tcktransform` command that does both (like `mrtransform`)?

thijsdhollander:
That seems to make the most sense to me too!

jdtournier:
Happy with `tcktransform` too, go for it.

Lestropie:
`tcktransform` it is. (Won't change on `master` for a while yet though)

Issue #502, opened 2016-03-31
Tractography::Writer: Preserve order


Lestropie:
There are now a couple of places where streamlines data are processed in a single-threaded fashion, as the order of output tracks must match the order of input tracks, and that order cannot be guaranteed if multiple processing threads are used.

A possible solution would be to have contained in this class a `std::set<size_t, Streamline<>>`, that stores data for any streamlines that are received by the Writer class out-of-order. Whenever the Writer functor is called and data is written (and therefore the streamline index is incremented), this structure is also checked to see if there is further data that can be pushed to the output file / buffer.

There's a couple of situations where streamlines data order doesn't need to / can't be guaranteed, e.g. `tckgen` / some `tckedit` usages. So I'm thinking this functionality could just go in the primary `Tractography::Writer` class, and be activated / deactivated via a flag.

Lestropie:
Closed by #1939.

PR #1131, opened 2017-09-20
tcknormalise: fix extraneous NaN triplets in output

jdtournier:
This occurs when streamlines venture outside the deformation field, and causes additional empty streamlines to show up in output. As discussed in http://community.mrtrix.org/t/bug-in-tcknormalise-when-using-warpcorrect/1175

Not sold on merging to `master` on this one - happy to cherry-pick and create a fresh pull request to `dev` if this isn't felt sufficiently critical, it does modify behaviour slightly....

### DIFF
--- a/cmd/tcknormalise.cpp
+++ b/cmd/tcknormalise.cpp
@@ -65,9 +65,19 @@ class Warper { MEMALIGN(Warper)
       interp (warp) { }
 
     bool operator () (const TrackType& in, TrackType& out) {
-      out.resize (in.size());
-      for (size_t n = 0; n < in.size(); ++n) 
-        out[n] = pos(in[n]);
+      out.clear();
+      for (size_t n = 0; n < in.size(); ++n) {
+        auto vertex = pos(in[n]);
+        if (!vertex.allFinite()) {
+          if (out.empty())
+            continue;
+          if (!out.back().allFinite())
+            continue;
+        }
+        out.push_back (vertex);
+      }
+      if (!out.back().allFinite())
+        out.resize (out.size()-1);
       return true;
     }
 


Lestropie:
This reminds me of `tckedit`, where applying a mask can result in multiple output tracks for one input track. There it's dealt with explicitly: the pipe functor stitches those multiple segments into a single `Streamline<>` to send down the queue, and the sink functor disentangles them and calls the `Tractography::Writer` functor once for each.

In this case the potential for one-to-many mapping is probably less of an issue, but nevertheless even with your fix there's a chance that a user could still report the number of input and output streamlines being different. Would it be better to instead just keep only the longest connected segment of the warped track?

thijsdhollander:
Hmm, I hadn't appreciated the challenge of this yet, but it's something worth "getting right" indeed.  Whatever "right is...

> This reminds me of tckedit, where applying a mask can result in multiple output tracks for one input track. There it's dealt with explicitly: the pipe functor stitches those multiple segments into a single Streamline<> to send down the queue, and the sink functor disentangles them and calls the Tractography::Writer functor once for each.

So essentially, there's already a precedent here in a way.  Warping streamlines using a deformation field that is only defined for a certain field of view is only generating an output for that field of view, so it's masking the streamlines just as well.  I reckon whatever behaviour we prefer should be applied equally for `tckedit` with a mask and `tcknormalise`.

> Would it be better to instead just keep only the longest connected segment of the warped track?

Hmm, not sure about that one.  That could lead to all sorts of strange behaviours.  Whereas with splitting users may indeed report mismatches in streamline count, with only keeping a certain part of the original streamline in the "field of view" users could for instance report mismatch in track density.  So it's always going to change certain properties in both scenarios, and confuse users who happen to focus on said property...

Something else that comes to mind is connectivity: tracks that get split in different part change their "connectomic" (whoever reads this: please don't adopt this word... :neutral_face:) behaviour.  They even change their "anatomically constrained" behaviour by potentially introducing end-points (at the edge of the field of view) that may violate anatomical constraints.

The only solution that I can think of which deals with all of the above at the same time would be to effectively introduce a value for streamline vertices who's position is not defined (so essentially "NaN", but it's annoying/challenging that that one is already used for end-points of course).  You end up:

1. not changing streamline count: streamlines that are broken up are still one streamline, just with defined and undefined bits
2. not losing track density: at least in the field of view, because nothing is defined beyond (but that's the same for images that get masked or warped)
3. not losing connectivity: if both original end-points are in the field of view, they're still connected by the same streamline
4. ...with a strict definition of a single step size, you in principle don't even lose the (original) length

So the question(s) is/are:

1. Is this overkill?  It's subjective, but yeah, it's not unreasonable to call this overkill. :slightly_smiling_face: 
2. If not overkill, but desirable: what value to use then, since "NaN" is already taken (and I wouldn't change "NaN" for end points at this stage, that's going to cause very widespread severe headaches)...

But whatever the solution, I'd like to stress my initial point again: it probably makes sense to go for the same strategy in `tckedit` with a mask and `tcknormalise` with a limited field of view.

thijsdhollander:
To add to my overkill solution:

5. doesn't lose anatomically constrained-ness
6. up to a certain extent, still retains similar behaviour for track-weighted imaging; even with a distance window or function along the streamlines

Lestropie:
> > Would it be better to instead just keep only the longest connected segment of the warped track?

> Hmm, not sure about that one. That could lead to all sorts of strange behaviours. 

I tend to agree; though genuinely having a single streamline get broken into multiple pieces from having travelled outside the warp field, and each of those pieces actually being meaningful / useful, is probably unlikely. Nevertheless it might not be worth the added complexity. Just wanted to raise the possibility.

> This should probably go to `dev`.

:+1:

> While thinking about this anyway: we should really reconsider the name of `tcknormalise`. 

I've already changed this to `tcktransform` in my Fixel TWI development branch. Was discussed briefly in #810. Not against it being changed in `master` before that point; was just going on the premise that it's just easier to justify to users changing a command's name when it comes with a change in capability / scope as well.

thijsdhollander:
>> While thinking about this anyway: we should really reconsider the name of tcknormalise.

> I've already changed this to tcktransform in my Fixel TWI development branch. Was discussed briefly in #810. Not against it being changed in master before that point; was just going on the premise that it's just easier to justify to users changing a command's name when it comes with a change in capability / scope as well.

Ah, I'd missed that.  Yeah, that sounds certainly ok.  There's no urgency as to the rename, so if it comes along at any stage, that should be fine.  If it happens still just in `dev` any way, it's not going to be in `master` immediately.  I would make sure that all significant renames of commands and options that we've still got in mind happen before release though (or at least as much as possible).  It may as such even not be a bad idea for us to go over all names thoroughly before release...

Lestropie:
> Ah, I'd missed that.

You [commented](https://github.com/MRtrix3/mrtrix3/issues/810#issuecomment-257783149) on it! :grimacing: 

> It may as such even not be a bad idea for us to go over all names thoroughly before release...

Fair call; already have one other I've been wanting to change (`mesh2pve` -> `mesh2voxel`).

jdtournier:
OK, all fair points being raised here. Thinking about this some more, I reckon the best approach _might_ be to simply skip those points that map to NaN. It's _almost_ the same as what @thijsdhollander was suggesting, only difference being that these vertices would be literally non-existent rather than undefined. Unfortunately, NaN can't be used to signal an undefined vertex (it's the streamline delimiter) and neither can Inf (it's the end-of-file marker), so I can't see an easy way to label vertices as undefined (certainly not without modifying a lot of back-end code to support that, and most likely breaking compatibility with older versions). Skipping the vertices altogether achieves all but point 4 (and potentially 6) from [thijs' list](https://github.com/MRtrix3/mrtrix3/pull/1131#issuecomment-331025911) (constant step size). It's a trivial change, what do you reckon?

Also agree that it should go to `dev` - I'll create a fresh pull request for that once we've decided how to handle it. 

One final concern: [the process is currently multi-threaded](https://github.com/MRtrix3/mrtrix3/blob/672901b5a3474ec3bddefe5c348c68b82a0a257b/cmd/tcknormalise.cpp#L132-L138), and hence runs the risk of modifying the order of the streamlines - this may give rise to inconsistencies with streamlines weights, etc. Is this handled by the back-end? If not, should we do something about it, and what would be the best way to do this? Maybe I can modify the multi-threaded queue backend to support that transparently..?

thijsdhollander:
>>    Ah, I'd missed that.

> You commented on it! :grimacing:

Ok, that's probably how it even ended up in my brain to begin with... well, it's clearly all good then! :+1: 

> I reckon the best approach might be to simply skip those points that map to NaN. It's almost the same as what @thijsdhollander was suggesting, only difference being that these vertices would be literally non-existent rather than undefined.

That's actually not too bad indeed.  The only minor "annoyance" that this also leaves is visualisation: depending on where the original streamline exited and re-entered the "field of view" of the warp or mask, you're left with a long streamline segment that will be rendered as well in `mrview`.  Really worst case (but this may be rather rare in practice) is a streamline thet exist the FOV and re-enters on some completely other side of it: in that case the long segment will go *through* the field of view (rather than sitting "along" the edge of it).  But this is quite a minor thing, and if really a problem for someone's scenario, they should consider defining their field of view differently, I reckon.

> Unfortunately, NaN can't be used to signal an undefined vertex (it's the streamline delimiter) and neither can Inf (it's the end-of-file marker), so I can't see an easy way to label vertices as undefined (certainly not without modifying a lot of back-end code to support that, and most likely breaking compatibility with older versions).

Now thinking of it, doesn't negative infinity exist as a separate value as well?  While that would then be the "easy" way to label vertices as undefined, I do agree you're still left with making about every application that deals with streamline tractograms compatible with that.  And I can see how that may in at least a couple of cases mean a way of dealing with "undefined vertices" that is specifically tailored to individual applications... possible resulting in a lot of headaches and potentially long discussions.  It probably not worth the hassle indeed.  Even if at some point we think it is (but probably not), the solution of simply skipping the vertices doesn't hinder that in the future.

So with that, I think I can see no remaining significant issues with the skipping vertices solution of @jdtournier above. :+1: 

> One final concern: the process is currently multi-threaded, and hence runs the risk of modifying the order of the streamlines - this may give rise to inconsistencies with streamlines weights, etc. Is this handled by the back-end? If not, should we do something about it, and what would be the best way to do this? Maybe I can modify the multi-threaded queue backend to support that transparently..?

Not entirely sure what's all in place currently, but `tckedit` has a `-tck_weights_in` and `-tck_weights_out` option, which I suppose plays well with multi-threading?

`tcknormalise` doesn't have such an option yet, so would need it for this scenario.

Note that it's not (only) about the order of the streamlines (so just making things single-threaded for instance in `tcknormalise` is not the solution either): entire streamlines may disappear if completely outside of the new FOV.  Now thinking about that further (you can tell I'm only thinking this all through while I'm typing...), in my initial "undefined vertices" solution, that would've introduced a choice we still had to settle on as well: would such streamlines have just disappeared, or would they still exists, yet will all nodes now undefined.  The latter would've had the "advantage" (depends on how you look at it) that the streamline count was still preserved.  Not sure if that would matter for any application though...  But now trying to go with the "skipping vertices" solution, I suppose there's no other (reasonable) way than to indeed have streamlines disappear if they fall entirely outside the FOV. (which is probably ok)
It does show we'll have to be careful with treating these cases right: there's streamlines going out and in the FOV, there's some going out but maybe not going in again, and there's some being entirely out(, and there's some having only 1 point still in, not an entire segment; but that's not per se a separate case).

Well, all of that to probably also say that `tcknormalise` will need `-tck_weights_in` and `-tck_weights_out` options; there's no way around that I think?  This may be strange to users though.  With `tckedit` it's more obvious you'd want to bring your weights along through the track editing (so you'll be looking for such options), but with `tcknormalise` it may be less obvious for users to realise this "naturally".

While I'm thinking while typing anyway, I final thought: we could still retain some empty entry for streamlines that were removed...?  In that case, we *can* actually just try to retain the order of streamlines in the file, and it would still match with that in any corresponding weights files.

Lestropie:
Looks like `tcknormalise` is indeed multi-threaded; I thought it was not, for this reason. Should probably write a generic `constexpr std::string` that can be added to the `DESCRIPTION` of multiple commands, describing that if retaining order is required then `-nthreads 0` should be used, at least until #502 is done.

There is however an outstanding issue of streamlines existing entirely outside the FoV. In this case, the track data will be empty, and the tractography file writer will *skip* the track, rather than writing an *empty* track. I reckon this needs to be  dealt with more fundamentally. Sending an empty track down the queue as a "failed" streamline in `tckgen` was done a long time ago, but it gets in the way in applications outside of `tckgen` that output track data. So the writer class might need to be modified so that e.g. the `GeneratedTrack` functor uses empty track data as such a flag, but when receiving a `Streamline<>` the data are written regardless of length.

> Not entirely sure what's all in place currently, but `tckedit` has a `-tck_weights_in` and `-tck_weights_out` option, which I suppose plays well with multi-threading?

This plays with multi-threading, in that the streamline order may be permuted, but the track weight is "attached" to the vertex data within the `Streamline<>` class, so the output weights file reflects the new streamline order. A more generic solution is however required for more generic processes, which would be #502.

jdtournier:
OK, a few thoughts on this before I had a go at implementing the changes we're talking about:

- I don't think we need to worry about streamlines that are fully outside the FoV, and hence show up empty. By the looks of things, they'll still be included in the output as empty streamlines (i.e. you'll get two NaN triplets back to back). So the count will still be correct, etc. 

- This may be relevant for streamline piping (currently a sub-issue of #411). I was wondering how to store the per-streamline attributes in the same file, so that we could pipe the data without requiring a whole new format. As things stand, I think all implementations will check for delimiters by only checking whether the _x_ coordinate is non-finite (both in the [C++ code](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/tractography/file.h#L104) and in the [Matlab reader](https://github.com/MRtrix3/mrtrix3/blob/master/matlab/read_mrtrix_tracks.m#L73)). At the very least, a compliant application _should_ really check whether _any_ of the vertices are non-finite (but then our own code doesn't...). But nonetheless, the point is that the other 2 floats in the delimiter are currently ignored. We could easily recruit say the _y_ coordinate to store the preceding streamline's weight - and we could do this by default when the weight is available (in addition to explicitly storing them in a standalone file). This would do away with the need to store the information separately (we could also modify the `tck` file in-place to inject that information so users don't need to duplicate their entire tractogram?), and immediately allow the current file format to be used for streaming through pipes. It would also be completely backwards compatible with existing apps. What do you think...?

jdtournier:
closed with #1154

Lestropie:
> By the looks of things, they'll still be included in the output as empty streamlines (i.e. you'll get two NaN triplets back to back)

No, I don't think so: The `Writer` class will hit [this line](https://github.com/MRtrix3/mrtrix3/blob/dev/src/dwi/tractography/file.h#L395), and no delimiter will be written; therefore the streamline will be *absent* from the output track file rather than being *empty*. Hence my earlier [spiel](https://github.com/MRtrix3/mrtrix3/pull/1131#issuecomment-333431976).

> We could easily recruit say the y coordinate to store the preceding streamline's weight ...

Not a fan of hacking the existing format I don't think. For one, it's entirely reasonable to run `tcksift2` multiple times on the one track file with different parameters, resulting in different weights. Injecting this data into the input track file also blurs the line between command input and output, which we've tried to avoid in other contexts. It's also entirely reasonable to generate e.g. a TDI both with and without taking the streamline weights into account: if the weights were embedded in the track file, there would need to be an explicit e.g. `-ignore_weights` option to calculate the latter, as opposed to currently using `-tck_weights_in` to do the former; so it doesn't really simplify the issue.

If some other format is capable of storing per-streamline weights, and a `tckconvert -tck_weights_in` imported the estimated weights and embedded them into the output file, that'd be fine; indeed there's a discussion kicking off from the DWI derivatives extension to BIDS related to definition of a standardised track file format, and this would be a useful inclusion in such a format.

jdtournier:
>> By the looks of things, they'll still be included in the output as empty streamlines (i.e. you'll get two NaN triplets back to back)
>
> No, I don't think so: The Writer class will hit this line, and no delimiter will be written; therefore the streamline will be absent from the output track file rather than being empty. Hence my earlier spiel.

I stand corrected. Yes, I agree it would be good to modify to allow outputting empty streamlines. Might be an idea to do this as part of #1154 - I'll look into it if I have the time...

>> We could easily recruit say the y coordinate to store the preceding streamline's weight ...
>
> Not a fan of hacking the existing format I don't think. 

I have similar reservations. But these files are so large, and the use case sufficiently compelling (I think) to warrant at least discussing it... 

> For one, it's entirely reasonable to run tcksift2 multiple times on the one track file with different parameters, resulting in different weights. 

Fully agree. But this would not be expected to be a common occurrence. You might do this relatively often as part of debugging, but the average user would simply run the command and leave it at that. Besides, we'll have to support standalone weights files for the foreseeable future anyway, so we can make sure that we preserve support for writing these weights to file rather than updating the main tracks file.

> Injecting this data into the input track file also blurs the line between command input and output, which we've tried to avoid in other contexts. 

Yes, I agree. My general unease with this though is more about ensuring that the file contents remain consistent even if the command crashes. Having clear read-only inputs and read-write outputs provides strong guarantees in that regard. In this case though, I think we can code this up in a relatively safe way, to ensure that if the command crashes halfway through, at worst what we end up with is inconsistent _weights_, leaving the vertex data itself unaffected. 

The reason I'd contemplate doing this is because the files we generate are so large that it would really be worth avoiding duplicating the data - if we did want to go with a format capable of storing weights, that is. Obviously the current approach handles this fine, _apart_ from when anything happens to the streamlines that might affect their order or might remove streamlines - which we've been discussing here. While we do have strategies to deal with this, they are far from foolproof. No worse than handling bvecs/bvals, mind you... But I do think there might be a merit in extending the format to handle weights - and if so, it's worth thinking about in-place injection of the weights for an existing file... 

> It's also entirely reasonable to generate e.g. a TDI both with and without taking the streamline weights into account: if the weights were embedded in the track file, there would need to be an explicit e.g. -ignore_weights option to calculate the latter, as opposed to currently using -tck_weights_in to do the former; so it doesn't really simplify the issue.

OK, it doesn't necessarily simplify _all_ operations, but it probably makes the _default_ operation simpler - in this case, you'd presumably want to generate a weighted TDI if you've bothered to compute them, so the simple default (just the streamlines file) would actually be the safest default, and simpler to manage. I reckon a lot of other operations would likewise be simplified. But the simplification of invocation is only really a minor issue, more important is the better guarantee of consistency that it provides (no need to remember to provide the weights file, or indeed which weights file, or realising that you'd forgotten to provide the weights file in your `tckedit` command, etc.). 

But really, the main reason I thought about this in the first place was to allow streaming the data through pipelines - for this to work properly, the format needs to multiplex the weights in there somehow. And the format as currently defined would allow this to be done with no modification to previous versions of the code. It seems a waste not to take advantage of that... But even if we decide to support this only for piping, I guess that would already be useful. 

Anyway, food for thought...

Lestropie:
The weights are also not the only parameters that one may want to ascribe to each streamline (e.g. TWI factors); or, indeed, to each streamline vertex (*a la* `.tsf`), and have those data correspondingly modified whenever some operation is applied to the streamlines, or passed with the streamlines down a pipe. That's why I'm expecting that the track pipe handler will *not* be `.tck`-based, but will be something like [`.trk`](http://www.trackvis.org/docs/?subsect=fileformat) with a MRtrix-style variable-size key-value-based header. So the above suggestion would be a hackjob, that's only applicable to one of many use cases, and it's the one case for which there's already a mechanism in place for dealing with modifications to track data... personally I'd rather put the effort into solving the fundamental issue.

PR #1134, opened 2017-09-22
mrtransform: Bug fix for warp field regridding

Lestropie:
If non-linear warp field and template image are not defined on the same voxel grid, the warp field must be resampled onto the template grid before warping can take place. This needs to be done using an identity transform rather than an uninitialised transform.

### DIFF
--- a/core/filter/warp.h
+++ b/core/filter/warp.h
@@ -15,10 +15,11 @@
 #ifndef __filter_warp_h__
 #define __filter_warp_h__
 
+#include "datatype.h"
+#include "adapter/reslice.h"
 #include "adapter/warp.h"
 #include "algo/threaded_copy.h"
 #include "algo/threaded_loop.h"
-#include "datatype.h"
 #include "interp/cubic.h"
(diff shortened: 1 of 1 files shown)


Lestropie:
```
transform_type uninitialised_transform;
std::cerr << uninitialised_transform.matrix() << "\n";
```

```
5.72622e-321  6.9428e-310 4.67708e-310 2.26486e-306
 6.9428e-310 3.95253e-323 4.67708e-310 1.18576e-322
7.90505e-323 6.95309e-310 4.67688e-310 3.18642e-318
```

:-/ Maybe the behaviour changed with Eigen 3.3?

rtabbara:
@Lestropie Whelp, I get something similar on Eigen 3.2. Think I got it confused with https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrregister.cpp#L805 which does identity-initialise the underlying matrices.

Anyway, great catch, although I still don't get how previously multiplying a junk matrix when reslicing would have gone unnoticed :grimacing:

Issue #1140, opened 2017-09-26
mrview: lightbox: fixel plot: inconsistent line thickness


rtabbara:
Fixel line thickness is inconsistent across all slices. Specifically, appears thickness changes are triggered by slices moving in and out of focus.

thijsdhollander:
Just to add to clarify for others: this didn't only happen with fixels, but also with tractograms.  I was first starting to think it may have to do with slab thickness of tracks versus slice distance (increment) of the lightbox slices versus potentially the exact location of the focus (which isn't always the centre of the slice shown, so things may get complex with respect to (track) slab location versus slice location versus focus position).  Then again, that clearly didn't appear to be (at least the fundamental) the problem, since e.g. switching crop to slab completely off still showed things weirdly changing when clicking other lightbox panels, even though all panels *should* be showing the exact same full tractogram...  So well, let's see what #1142 gives. :slightly_smiling_face:

rtabbara:
This is fixed in #1142

PR #1142, opened 2017-09-26
mrview: lightbox: Reinitialise GL state when rendering each pane

rtabbara:
* Fix for #1140
* However, this affects any rendered tool. Issue was that we need to ensure GL state is re-initialised because draw calls sequence is slice1, tools1, slice2, tools2 etc.
* Therefore, can't guarantee GL state remains unchanged between slices
* Ideally, want to ensure any tools restore original GL context at the end of their draw call, but regardless probably safer to do this

### DIFF
--- a/src/gui/mrview/mode/lightbox.cpp
+++ b/src/gui/mrview/mode/lightbox.cpp
@@ -177,21 +177,21 @@ namespace MR
         void LightBox::draw_plane_primitive (int axis, Displayable::Shader& shader_program, Projection& with_projection)
         {
           ASSERT_GL_MRVIEW_CONTEXT_IS_CURRENT;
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Did some more testing, and don't seem to run into any immediately obvious issues.  I've updated with the latest changes from `master`, so the tests can run again.  If nobody spots any further issues, I'm probably looking to hit the button tomorrow.  The bug that got fixed was quite significant and relevant for anyone visualising their FBA results via lightbox mode.  They could be showing wrong stuff (and putting the images in their papers) without fully realising.

thijsdhollander:
:+1: Perfect, I'll hit the button!

Issue #1072, opened 2017-07-26
Change configure debug optimization flag


rtabbara:
The configure script is using the optimizer flag `-O1` for a debug build, but this results in optimized-out variables which hinders the ability to step through the code. The flag should be set to `-O0`.

Here's the part in question: https://github.com/MRtrix3/mrtrix3/blob/master/configure#L1232

Lestropie:
Was set in 05b9ad19.

Need to make sure that both `gcc` and `clang` are happy with `-O0`.

jdtournier:
Yes, the default used to be `-Og`, which is the recommended optimisation level for debugging with `gcc`. But `clang` doesn't understand it. The commit @Lestropie pointed out contains a link to a post suggesting `-O1` as a good compromise between optimsation and debugging that works across both `gcc` and `clang` - but I'm happy to switch to `-O0` if both compilers support it...

PR #1143, opened 2017-09-26
Configure tweak

jdtournier:
Minor change, arose out of my attempts at figuring out a simple way to release binaries for _MRtrix3_. This simply turns on some optimisation (`-O1`), allows slightly tighter packing of common symbols (if I understand `--sort-common` correctly), and does not include references to libraries not used in the executable (`--as-needed`). Should hopefully work on all platforms as-is, but probably needs testing... 

No rush on this one, BTW - it's just a minor tweak.


Lestropie:
- Should `-O1` be omitted in debug mode? Sounds like it's really only a compile time effect & won't affect debug-ability, but nevertheless might be preferable to disable for debugging.

- `ld` has a `--strip-debug` option for omitting debugging symbols from the shared library; might be worth including in release mode.

- `--as-needed` only applies to entries that appear *after* the `--as-needed` flag. So currently it won't be applied to *MRtrix3* objects, only system objects. Don't know if it's deliberate or not, just checking.

jdtournier:
> - Should `-O1` be omitted in debug mode? Sounds like it's really only a compile time effect & won't affect debug-ability, but nevertheless might be preferable to disable for debugging.

Yes, that's probably fair enough. To be honest, it's really hard to find any information about what it does exactly... The only reason I'd included it in the first place is because Qt5 does (via its default `qmake` config), and I don't think they'd do this if it wasn't safe. But we can definitely disable for debug builds.

> - `ld` has a `--strip-debug` option for omitting debugging symbols from the shared library; might be worth including in release mode.

Well, according to [this post](https://stackoverflow.com/a/1349244), `--strip-debug` removes only those symbols introduced by `gcc -g` (bearing in mind that `--strip-debug` is the same as `-g`, at least according to the [strip man page](https://sourceware.org/binutils/docs/binutils/strip.html) - not `ld`, I know, but probably does the same thing...). Since that flag is only included for debug builds, that would probably be redundant...?

> - `--as-needed` only applies to entries that appear after the `--as-needed` flag. So currently it won't be applied to MRtrix3 objects, only system objects. Don't know if it's deliberate or not, just checking.

Not quite. From the [ld man page](https://linux.die.net/man/1/ld): 

> This option affects ELF DT_NEEDED tags for dynamic libraries mentioned on the command line after the --as-needed option. 

So it's only relevant for shared libraries, and only affects whether the final executable includes a reference to it (to inform the runtime linker that it'll need this library). That's the main thing I was trying to achieve here. 

For objects, the build script only includes objects if they were listed as dependencies in the first place, and it determines this based on whether any of the corresponding headers were included in the code required to build the executable (it's a bit complicated, but [the rules are outlined here](http://www.mrtrix.org/developer-documentation/build_page.html#build_section). So I think it's all fine...

Lestropie:
> (bearing in mind that `--strip-debug` is the same as `-g`, at least according to the `strip` man page - not `ld`, I know, but probably does the same thing...).

From `ld` [man page](https://linux.die.net/man/1/ld):
```
-g
Ignored. Provided for compatibility with other tools.
```
:no_good_man: :rofl: 

> Since that flag is only included for debug builds, that would probably be redundant...?

:+1:

-----

So in the process of having `ld` use `-O0` in debug mode and `-O1` in release mode, I noticed that #1072 was never resolved, and `g++` / `clang++` are currently using `-O1` in debug mode. I've had a bit of frustration lately with inability to use `gdb` due to optimisations / inlining. So I've made that change here as well, given it requires comparable testing (I've successfully compiled these changes on both compilers, in both debug and release modes).

(The other option would be to branch: When in debug mode, if `cpp` contains "`g++`", use `-Og`, otherwise use `-O0`)

PR #1145, opened 2017-09-26
Fix mrcat for complex-valued images.

dchristiaens:
Concatenating complex-valued images currently produces an image with correct header datatype, but incorrect image values. only the real-valued part of the input data got copied over to the output.

This commit fixes this bug without affecting behaviour for real-valued data.


Lestropie:
While this'll fix the behaviour, longer-term it'd be cool to have many more generic data manipulation commands templated by datatype just like `mrconvert` 😈

jdtournier:
> While this'll fix the behaviour, longer-term it'd be cool to have many more generic data manipulation commands templated by datatype just like `mrconvert` 😈

We had precisely this conversation yesterday... Are you volunteering...?

jdtournier:
Just pushed minor changes to do with signed/unsigned comparison warning. @dchristiaens, any chance you could give this a spin and merge if all OK?

dchristiaens:
Tested and approved.

PR #1146, opened 2017-09-26
mrview: document FontSize config file option

jdtournier:


thijsdhollander:
Tests have completed; this should be a completely safe merge, since it doesn't affect any behaviour at all, just documentation.  I'll hit the button, so we can close it off.

Issue #1096, opened 2017-08-15
average_response: De-mean before averaging


Lestropie:
Currently, `average_response` does a direct mean of response function ZSH coefficients across subjects. This means that if the overall amplitude of the response function is larger in one subject than in others, the shape of the response function for that subject will have a greater influence on the shape of the group average response function. If equalising the intensity of WM *b*=0 signal across subjects using `dwiintensitynorm` before this is done, this could be perceived as a desirable effect: subjects with a smaller RF after `dwiintensitynorm` would have less influence on the shape of the mean response. However in the current [multi-tissue FBA documentation](http://mrtrix.readthedocs.io/en/latest/fixel_based_analysis/mt_fibre_density_cross-section.html), the `dwiintensitynorm` step is no longer performed. Hence the scanner reconstruction gain factor could influence which subjects have greater influence on the shape of the group average response.

In this case, it could be preferable to instead normalise each subject's RF to a fixed *l*=0, average, then re-scale the resulting average response to the mean *l*=0 across subjects.

What's not written in stone is how this normalisation should occur in a multi-shell context, i.e. whether to normalise independently for each shell or take e.g. a geometric mean of scaling factors across shells for that subject.

Tagged `question` for a reason. 👂

Lestropie:
Implemented as part of #1449. Demonstration below for comment.

Two inputs:

- `[9 0 0 0 0]` (isotropic, large);
- `[1 -0.5913 0.3126 -0.127 0.0337]` (just an exemplar single-subject WM response, small).

![1](https://user-images.githubusercontent.com/5637955/52095466-1500fb80-2617-11e9-93ee-da2816264da4.png)

![2](https://user-images.githubusercontent.com/5637955/52095470-19c5af80-2617-11e9-90ae-c836437a68bf.png)

`responsemean tmp1.txt tmp2.txt average_response.txt -legacy` (identical to `average_response`): `[5.00000 -0.29565 0.15630 -0.06350 0.01685]`

![average_response](https://user-images.githubusercontent.com/5637955/52095472-1cc0a000-2617-11e9-80d1-9cf152f4a315.png)

`responsemean tmp1.txt tmp2.txt responsemean.txt`: [1.80000 -0.53217 0.28134 -0.11430 0.03033]`

![responsemean](https://user-images.githubusercontent.com/5637955/52095474-1f22fa00-2617-11e9-9595-21e620555e4c.png)

-----

[...]

Issue #944, opened 2017-03-30
dwibiascorrect -ants problem with sub-millimeter voxel size


bjeurissen:
The -b parameter of N4BiasFieldCorrection is interpreted in mm.

When dealing with images with sub-millimeter voxel sizes, N4BiasFieldCorrection will allocate an image with dimensions (150/vdim x 150/vdim x 150/vdim), which quickly results excessive memory consumption even on HPC machines.

What was the rationale behind the 150mm parameter?

Lestropie:
My guess is manual testing & tuning. That's @draffelt's domain though.

If said tuning was based on a 2.5mm voxel size, that's 60 voxels per axis. One could I suppose flip the equation and calculate the necessary value for `-b` based on the input image size to get that many field voxels?

thijsdhollander:
Yeah, that's all just ad hoc tuning... parameters are really just designed for human heads.  I'm not too fussed about this; this is really a step to be looked into manually, if users need it at all at this stage.  I like having it included in the FBA pipeline for improved masks (when it works :grin:) and a bit less "bias" during `dwi2response tournier`.  But in this day and age of multi-tissue processing, `mtnormalise` should pick up where `dwibiascorrect` left off, or even do the job entirely on its own.

Note that in #1184, my tuning has led me to a 100mm distance between spline control points indeed, so this may help a bit with the issue described here.  But there's data and projects where I've instructed RA's / students / collaborators to use a particular set of N4 parameters that is not the scripted one as well.  The scripted one is just a "relatively generally applicable" one.  We could eventually allow users to set the N4 parameters via the script as well, but as I mentioned, I don't see this as an urgent issue, or an issue at all... Very happy with `mtnormalise` performance since we've started using it over here, and since I've gotten some feedback from studies where it was used in the mean time.

PR #1148, opened 2017-09-27
dwibiascorrect ANTS N4BiasFieldCorrection parameters re-tuned

thijsdhollander:
It looks like biasfield correction is becoming my personal nightmare of sorts... :fearful: :scream: 

So when I retuned the N4 parameters last time, it seemed to improve things for `dwi2mask` for all data that I had access to... until now.  @Lestropie ran into / spotted a case where the opposite was true: the corrected data got slightly too large intensities towards the edge of the FOV (basically extrapolation from the bias field estimated in the brain mask).  This then becomes an issue for `dwi2mask`, where it automatically thresholds the b=3000-ish data (b=0 is not a problem).  The role of that bit of data though, is to essentially give a rough brain parenchyma mask (b=0 gives a rough CSF mask).  If that one fails, and there are no other shells to play that role decently, you only end up with a CSF + few bits of edge-of-FOV crap.  That's already a fail at that stage, but the crap-cleaning filter makes things even worse thereafter (but again: at that stage all was lost already, so it doesn't matter any more).

The answer is to re-tune the N4 parameters once more.  The trick is to keep them smooth enough; there's really only so much you can estimate from an average b=0 (T2w) contrast...  N4 was really only designed to perform well on a T1w image sadly.  So we just keep it smooth enough, so we still gain from correcting the gross effects of a bias field (if a gross effect is present).  That'll help `dwi2mask` as well as `dwi2response tournier` (or any dwi2response algo that runs tournier at some stage).  In a decent multi-tissue FBA (or other analysis), you can at some stage perform `mtnormalise` any way, which takes care of things much more decently.

Alright, lots of words for a simple pull request: this comes from extensive tuning and should work decently for what can decently be done at that stage in a pipeline.  Since this does change behaviour, it's gotta go `dev` sadly... but it's quite important again for people doing FBA... so it may not be a bad idea to think towards a tagged release in the near future.

I think it'd be worthwhile to at least have another tagged release (candidate) before the "release release", so we can get a bunch of features on `master` again.  It'd be good if the actual true "release" doesn't have to come with too many other features other than the benefits of a release (i.e. some kind of binaries distribution (mechanism).  What do people reckon?

Feel free to test the dwibiascorrect changes if you've got the time.  I'm quite convinced they're another change for the better, after careful tuning.  I'll be looking to at least merge them into `dev` relatively soon if nobody comes across important examples that would argue against the changes.  Speak up if you feel strongly about this, or remain silent forever (or until the next issue arises). :smile:


Lestropie:
Do you feel as though this change is compromising the quality of the bias field correction? Because if so, we might be able to retain the better performance of the current settings within the brain, and instead more directly address what happens outside the brain.

As an example: With `dwibiascorrect -fsl`, it applies a multiplicative factor to each voxel in the input mask, which is somehow scaled so that the values are distributed around 1.0, and voxels outside the mask remain unchanged. So that can lead to weird effects at the edge of the mask, where a voxel in the mask can be multiplied by e.g. 1.3 but the adjacent voxel is unaffected. So even just performing this would solve the reported issue, but obviously comes with a different non-ideal behaviour, especially if the masking process is imperfect.

So a couple of thoughts. Firstly, the factors coming out of N4 are nowhere near centred around 1.0, which makes applying a "default" scaling outside the mask difficult (and may be consequential for #1096 as well, if that overall factor varies a lot between subjects and `dwiintensitynorm` isn't used). But we could, for example, optimise the bias field within a brain mask, apply the estimated field within a *dilated* brain mask, and then null values outside of the dilated brain mask. That would hopefully allow reasonable correction at the outer extent of the brain; and limiting the extent to which that field is applied outside the brain would prevent the effect reported here (i.e. brightness in the FoV corners) from appearing.

Alternatively, rather than putting a hard threshold on how far outside the brain mask the field can be applied vs. nulling the image data, we could do something like what Max tried in #1016, i.e. smoothly constrain the bias field toward 1.0 the further away you get from the brain mask. This would require first re-scaling the N4 bias field so that it is in fact centred around 1.0 within the brain mask; though that's probably not a bad idea to add to the script anyways.

Just riffing...

thijsdhollander:
> Do you feel as though this change is compromising the quality of the bias field correction?

Nope.  This new parameter set is a bit less aggressive in (over)fitting.  The quality, as in "accuracy" is hard to assess.  As I mentioned, the problem is mostly that what can be done realistically of a T2w contrast at a low spatial resolution is very limited.  Aggressive fitting is riskier, because if the algorithm fails to separate white matter and grey matter as intensity classes / peaks in the histogram (which it often does), it'll end up being an over-fit, with more wacky extrapolation beyond the brain mask.  The new choices are simply "safer", and may as such actually be "better" for lower quality datasets as well.

> Because if so, we might be able to retain the better performance of the current settings within the brain, and instead more directly address what happens outside the brain.

So this shouldn't be needed.

I wouldn't add heuristic mechanisms to N4, because there's not a lot that can be safely relied upon.  Intensity normalisation, as you mention, is indeed something that isn't there inherently.  Also, the main goal of actually adding `dwibiascorrect` again to the pipeline, is to produce better brain masks, so relying too much on the brain mask estimated before N4 (in the `dwibiascorrect` script) is also not a safe idea.

The normalisation issue with respect to response function averaging, by the way, is also not too big of a deal.  Data will typically not be entire orders of magnitude different between subjects, and response functions have already been shown to be varying very, very little between subjects of a similar species.  If anything, that response function averaging step is just to provide a certain feeling of elegance.  In most (e.g. FBA) studies, you won't notice any impact of significance depending on whether you got your (set of) responses from all subjects, a particular group, or even a single subject.

More important though, is what a certain response function represents, i.e., which area(s) of the brain it typically gets estimated from.  That's where the bias field versus the tournier algorithm comes into play, since the latter one is (by conscious design) biased towards higher amplitude responses.  So it's not entirely unimportant to try and correct gross bias field effects before response function estimation.  This is where these new proposed N4 parameters should come into play, hopefully making this again a little bit more robust.

thijsdhollander:
Just a reminder:

> Feel free to test the dwibiascorrect changes if you've got the time. I'm quite convinced they're another change for the better, after careful tuning. I'll be looking to at least merge them into dev relatively soon if nobody comes across important examples that would argue against the changes. Speak up if you feel strongly about this, or remain silent forever (or until the next issue arises).

I'll be looking to merge (to `dev` of course) sometime this week.  ...with the hope it gets some extra inherent testing on `dev` dev of course.

Lestropie:
I don't really have a large amount of data from different sites / acquisitions, so can't really contribute to the assessment here over and above the problematic data I sent you.

Lestropie:
Will this change be enough to sort out #944 @bjeurissen?

thijsdhollander:
I'll merge this for now to `dev`, since it resolved the issue, and I've seen it work decently on a few other cases.  Maybe in the future, we could indeed allow the parameters to be set via the script, so people are aware that defaults are involved and they can be overridden in case it makes sense for particular data... but I'm not going to lose too much sleep over it.  Over all the changes we've made to this internally at some point, and via other mechanisms to correct for bias fields or normalise, I've barely ever seen an impact on our (FBA) studies' results.  This is never going to be perfect (the data doesn't even have the specificity to resolve the solution to this problem uniquely).  Even if spatially biased, as long as it's somewhat consistent across subjects in spatial behaviour, this isn't a big issue for sufficiently powered fixel- or voxel-wise studies.

PR #1151, opened 2017-09-29
fix dwigradcheck

dmordom:
fixed error in the mrconvert call generating data.mif (was not including gradient data)

### DIFF
--- a/bin/dwigradcheck
+++ b/bin/dwigradcheck
@@ -43,7 +43,7 @@ num_volumes = image_dimensions[3]
 app.makeTempDir()
 
 # Make sure the image data can be memory-mapped
-run.command('mrconvert ' + app.args.input + ' ' + path.toTemp('data.mif', True) + ' -stride 0,0,0,1 -datatype float32')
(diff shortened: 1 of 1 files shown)


Lestropie:
That won't work unfortunately: If the user does not provide the `-grad` command-line option, that `run.command()` call will fail since `app.args.grad` won't be defined.

I suspect what you're looking for is [this change](https://github.com/MRtrix3/mrtrix3/commit/9bdba675c4ccf7f1303d1f5606c22b8abb89cf0a#diff-5b8aa0c09296f88fa5d4a6abcc2e6c4c), which I've made in a development branch but didn't think to propagate across to `master` at the time. Essentially there's no need for `dwi.mif` to have an embedded gradient table, since a gradient table is always explicitly imported in every `tckgen` call. What does get stuck however is the `dwi2mask` call, which fails if there isn't an embedded gradient table. So the solution is to just explicitly import the gradient table for the `dwi2mask` call.

Would you like to make this change & get credit for it? Otherwise I can set it up myself, no biggie.

Lestropie:
Closed in favour of #1168.

PR #1152, opened 2017-09-30
population_template and dwiintensitynorm: ignore hidden files in input directories

maxpietsch:
to avoid problems with hidden .DS_Store and similar files as reported [here](http://community.mrtrix.org/t/dwiintensitynorm-valueerror-empty-searator/1200/2)


maxpietsch:
Untested on Windows so would be good if someone could run the test for `population_template` on windows.

Lestropie:
Solution didn't work for me: `win32api` requires explicit installation via `pip install pypiwin32`. I implemented an alternative based on [this post](https://stackoverflow.com/a/6365265), which I've tested with a hidden file in place.

Note also that the two images in `testing/data/population_template/in/` are softlinks to existing images in `testing/data/`, but the softlink property of these two files isn't propagated to Windows machines, which results in `run_tests` failing on Windows.

maxpietsch:
Thanks, Rob!

lib/mrtrix3/path.py
```diff
@@ -26,7 +26,7 @@ def allindir(directory, dir_path=True, ignore_hidden_files=True): #pylint: disab
   def is_hidden(directory, filename):
     if app.isWindows():
       try:
-        attrs = ctypes.windll.kernel32.GetFileAttributesW(unicode(os.path.join(directory, filename)))
+        attrs = ctypes.windll.kernel32.GetFileAttributesW(u"%s" % str(os.path.join(directory, filename)))
```
Lestropie: http://python-future.org/compatible_idioms.html#unicode

Lestropie:
Wondering if `path.allindir.is_hidden()` should test the hidden file attribute on Windows, but then *also* check for a name with a leading "." character. If running under e.g. MSYS2, it's entirely reasonable that the latter may indeed exist despite being reported as a Windows machine.

Lestropie:
Whoops, apologies for the [false announcement](http://community.mrtrix.org/t/dwiintensitynorm-valueerror-empty-searator/1200/8). Though given it's caught out [multiple people](http://community.mrtrix.org/t/dwiintensitynorm-error-empty-separator/1224) in quick succession, maybe it should be merged to `master`?

thijsdhollander:
...or maybe time for a another tagged release candidate, after the ISMRM deadline chaos? We've got some significant new features piling up on `dev` again already...

PR #1154, opened 2017-10-02
tcknormalise: skip vertices if outside warp

jdtournier:
This occurs when streamlines venture outside the deformation field, and
causes additional empty streamlines to show up in output. As discussed in
http://community.mrtrix.org/t/bug-in-tcknormalise-when-using-warpcorrect/1175
and #1131.


Lestropie:
More changes to come here as discussed in #1131, specifically [this comment](https://github.com/MRtrix3/mrtrix3/pull/1131#issuecomment-333979607). I reckon the combination of skipping vertices outside the warp, and being able to write empty streamlines to file if they exist entirely outside the warp, should be satisfactory and not introduce too much additional complexity. I will however need to do a bit of a code search: I believe I may have used sending empty `Streamline<>`'s to the track file writer in order to influence the `total_count` field, whereas the change described in the linked comment would instead result in empty streamlines being written.

PR #1070, opened 2017-07-25
5ttgen fsl: Fixes

Lestropie:
- Disable boundary correction in FSL FIRST, since it not only does not influence the estimated surface meshes and hence the output of `5ttgen fsl`, but may also erroneously alter the name of the combined output image that is used to detect success or failure of the segmentation.
- Add missing "`from mrtrix3 import path`" call.
- In `mesh2pve`, ensure that the output image is floating-point.

Closes #1068.

PR #1155, opened 2017-10-03
fsl.checkFirst() function

### LINKED ISSUE #1124: 5ttgen fsl: More changes to FIRST error detection
As discussed in #1070 after the relevant merge.

A local user managed to produce this same fault in a one-off, but it did not reproduce when we tried again... will find out soon enough if repeatedly running the script overnight can reproduce the same fault, but this time with more information for me to extract after the fact.

It appears that in some instances, the file `first_all_none_firstseg.nii.gz` is not produced, despite `.vtk` files being successfully produced for all structures of interest. In this particular case, one of the jobs generated a vector bounds checking segfault, such that the corresponding mask image was not generated, but it did successfully create the mesh file before that point. Therefore, in the post-execution script, when these mask images are combined together into a single volume, this one image was missing and therefore the `fslmerge` command failed.

I think the solution is going to have to be:

- If not using SGE, check for the presence of all `.vtk` files, and generate an appropriate error if any one file is missing.

- If using SGE, a modified `file.waitFor()` function will need to wait based on a *list* of expected files, preferably updating the terminal text as new files appear. This way if a single segmentation fails, the user will still be left with an error along the lines of "waiting for files (9 of 10 detected)".

Should also move this functionality to `lib/mrtrix3/fsl.py` as much as possible, so that `5ttgen fsl` and `labelsgmfix` can use the same code.

Lestropie:
Closes #1124. 

Related to discussion in #1070.

Discussed in [this](http://community.mrtrix.org/t/5ttgen-waiitng-for-creation-of-new-file/1044) forum thread (and others...).

The latest issue in this ongoing saga was that I tried to use the presence of one of FIRST's summary output images as a proxy for detecting overall FIRST success (& hence possibility for the script to continue). However in some instances, creation of this image would fail, and/or its path would not be consistent with the `FSLOUTPUTTYPE` environment variable, yet all of the requisite `.vtk` files were present and hence the `5ttgen fsl` script could theoretically progress.

Therefore, in this alternative solution I expand the `file.waitFor()` function to be able to wait for a *list* of files, rather than just one; and use it to look for just the `.vtk` files that will be subsequently read by `5ttgen fsl`. If not all files are immediately present in an SGE-enabled environment, the user will get a progress bar, with the progress percentage showing what fraction of files have been created. In the absence of SGE, the error message will report how many of those requisite files were created. The `FSLOUTPUTTYPE` non-conformity issue is bypassed, since only the `.vtk` files are tested, not any derived images.

This segment has always been difficult to test, so don't really expect others to test it thoroughly. Instead I'll demonstrate what I did:
```
  run.command(first_cmd + ' -s ' + ','.join(structure_map.keys()) + ' -i T1.nii' + first_input_is_brain_extracted + ' -o first')
+ os.environ['SGE_ROOT'] = 'NULL'
+ shutil.move('first-L_Accu_first.vtk', 'first-L_Accu_first_moved.vtk')
  fsl.checkFirst('first', structure_map.keys())
```
*added lines prepended with "+"*

This would execute `run_first_all` as normal, but then engage the SGE branch of `fsl.checkFirst()`. As expected, I get a progress bar saying 90% of files are present; if I then (in another terminal) move that file back to its intended location, the script eventually detects its presence and proceeds. If the `os.environ` line is erased, the script appropriately errors out, reporting that only 9 of 10 structures were segmented successfully.

Would like to merge this to `master`, since it's an ongoing stumbling block in processing for many people but doesn't influence analysis results. I'm confident that this solution is better than the existing code, but do appreciate that it's a difficult one to assess / review / test.


jdtournier:
While this is fresh in my mind: can we add an option to disable SGE altogether in the script? Having looked at the problem reported [here ](http://community.mrtrix.org/t/5ttgen-waiitng-for-creation-of-new-file/1044/15?u=jdtournier) with Vasiliki, it's clear that trying to run `5ttgen` (and `labelsgmfix` for that matter) over a cohort via an SGE script causes all kinds of issues, since `run_first_all` then decides to also submit its own jobs via SGE - so you get an explosion in the number of jobs, or the cluster's policies catch this and kill your command (in our case, it just killed `run_first_all`)... 

It's a simply fix to add `unset SGE_ROOT` within the SGE job script to prevent any commands within the script from using SGE (I'm surprised SGE doesn't do this automatically, to be honest), but it would be easier still if this was handled transparently. [SGE-capable environments](https://docs.oracle.com/cd/E19957-01/820-0699/chp4-21/index.html) have the `SGE_ROOT` variable set, so we can use that both to detect when problems are likely to arise, and to prevent sub-tasks from running under SGE by unsetting that variable. Also, jobs running under SGE will have the `JOB_ID` variable set, so we could at script invocation immediately detect whether our top-level script (i.e. `5ttgen`, etc) was invoked directly on the head node (in which case it's OK to allow `run_first_all` to use SGE), or indirectly as part of an SGE task (in which case we _don't_ want any further SGE calls). 

So that might look like:
```python
if os.environ.get("SGE_ROOT") and os.environ.get("JOB_ID"):
  del os.environ["SGE_ROOT"]
```
I _think_ this last line unsets the variable for child processes too, which is what we'd be after...

Note this won't solve the problem of detecting when `run_first_all` has completed following a direct invocation in an SGE-capable environment - but it might alleviate a few related headaches...

Lestropie:
> ... or the cluster's policies catch this and kill your command (in our case, it just killed `run_first_all`)...

OK, I hadn't fully decoded Vasiliki's comment to understand that one. If that's the case, then yes absolutely, disabling nested SGE calls makes sense. If the change you propose doesn't propagate to child processes, it shouldn't be too hard to make the requisite change at the `run` module initialization instead, and pass that environment to `subprocess`. I didn't want to disable SGE globally since I'd like for the script libraries to be more widely applicable; I did disable it at one point in `5ttgen fsl` / `labelsgmfix` but then people complain that it's not running as fast as it should :-/

Lestropie:
[bump](http://community.mrtrix.org/t/problem-using-labelsgmfix-command/1373)

Lestropie:
[bump again](http://community.mrtrix.org/t/5ttgen-fsl-error/1379)

Issue #1138, opened 2017-09-23
dwipreproc with half sphere sampling


neurolabusc:
I tried dwipreproc with historical data where only half-sphere sampling was used and there was no reverse phase encoding. I was surprised that the script selected eddy instead of the older eddy_correct, as this is discouraged
  https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/eddy
I think it would be great if the dwipreproc script included logic that if the user selects -rpe_none it either generates a warning or uses eddy if sampling is on the half sphere. Off the top of my head, I think you could detect a half sphere by computing the mean of bvec - if this is zero it is a whole sphere, if not normalize the length and compute find the bvec with the largest cross product. Alternatively, the [documentation](http://mrtrix.readthedocs.io/en/latest/reference/scripts/dwipreproc.html) could be enhanced to describe warn users to avoid this situation.

Lestropie:
It's discouraged to use half-sphere sampling at the acquisition stage, but personally I wouldn't interpret from that documentation that `eddy_correct` is recommended over `eddy` in that instance. Indeed I suspect it wouldn't be the case, and I doubt FMRIB would recommend as such; this is particularly true for higher *b*-values, where `eddy_correct` can cause more problems than it solves (since it doesn't predict the signal for a particular diffusion direction before performing the registration, so the image contrast between target and source can vary markedly).

It might be possible to automate the inclusion of `--slm=linear` in the case of a half-sphere acquisition as suggested in that linked page; but that can already be provided by the user, by utilising the `-eddy_options` command-line option in `dwipreproc`.

neurolabusc:
OK, automatically detecting this or changing the documentation to suggest `-eddy_options " --slm=linear "` sound like reasonable solutions.

jdtournier:
I just contacted Jesper Andersson about this to get his opinion, here's what he has to say:

> Yes, it sounds like Chris is a little overcautious. It is a good idea to sample on the whole sphere, and we encourage new protocols like that. That does not mean that eddy doesn’t work when sampled on the half sphere. But there are some gotchas to be aware of.
>
[...]

PR #1157, opened 2017-10-04
dirstat enhancements

jdtournier:
- Add report on symmetry of sampling, as discussed in #1138. This is the norm of the mean direction vector. Ideal value is zero, 0.5 is half-spherical sampling. Potentially useful for determining resilience to eddy-current induced distortions after correction with EDDY, and whether `--slm=linear` option to EDDY is warranted. 

- Add options for more fine-grained queries. Can select which section to report, by default everything is shown. As soon as a section is requested, all others are suppressed (unless also explicitly requested). Combine with `-quiet` for script-friendly output. Examples below.

- Add ability to read directions directly from an image, in addition to existing functionality (text file with 2 columns: [ az el ] pairs; 3 columns: [ X Y Z ] Cartesian vectors; 4+ columns: full gradient file with per-shell breakdown). 

Here's the full output on some old data I have:
```ShellSession
$ dirstat.exe ~/test_data/dwi.mif
dirstat: [WARNING] The following image volumes were not successfully assigned to a b-value shell:
dirstat: [WARNING] 15 (150.90495101627999), 36 (2357.8614957466802), 43 (234.15662571675)
C:/msys64/home/donald/test_data/dwi.mif (b=0.3149276342462895) [ 19 directions ]

    nearest-neighbour angles: mean = 87.2037, range [ 63.435 - 90 ]
    energy: total = inf, mean = inf, range [ 35.5388 - inf ]

    nearest-neighbour angles: mean = 87.2037, range [ 63.435 - 90 ]
    energy: total = inf, mean = inf, range [ 17.9511 - inf ]

  Spherical Harmonic fit:
    condition numbers for lmax = 2 -> 4: nan nan

  Symmetry of sampling:
    norm of mean direction vector = 0.0895422

C:/msys64/home/donald/test_data/dwi.mif (b=2983.5983574174184) [ 145 directions ]

    nearest-neighbour angles: mean = 12.0546, range [ 11.6942 - 12.3965 ]
    energy: total = 19649.5, mean = 271.027, range [ 264.074 - 274.567 ]

    nearest-neighbour angles: mean = 12.0555, range [ 11.6942 - 12.3965 ]
    energy: total = 12215.9, mean = 168.495, range [ 136.479 - 190.181 ]

  Spherical Harmonic fit:
    condition numbers for lmax = 2 -> 14: 1.05562 1.11509 1.19363 1.32383 1.5707 2.01918 3.10022

  Symmetry of sampling:
    norm of mean direction vector = 0.516308
    WARNING: sampling is strongly asymmetric - this may affect resiliance to eddy-current distortions
```
Note that it picks up on a couple of corrupt direction vectors (an old bug in the Siemens DICOM export, long since fixed), and provides a per-shell breakdown (not so useful for the b=0 shell, I'll admit). It also points out straight away that this is half-sphere sampling (I'd forgotten about that...). 

Here's a more script-friendly output, requesting just the mean nearest-neighbour angle assuming a bipolar electrostatic repulsion model, and the symmetry index for the b=3000 shell:
```ShellSession
$ dirstat.exe ~/test_data/dwi.mif -quiet -shell 3000 -bipolar -mean -nearest -sym
12.0546
0.516308
```

Hopefully that'll prove useful...?


thijsdhollander:
```
thijs@bri-pub-59:~/mrtrix3$ dirstat ~/Desktop/schemes/Skyra/dirs.txt 
/home/thijs/Desktop/schemes/Skyra/dirs.txt [ 64 directions ]

    nearest-neighbour angles: mean = 18.0604, range [ 17.5494 - 18.8115 ]
    energy: total = 3680.75, mean = 115.023, range [ 114.836 - 115.416 ]

    nearest-neighbour angles: mean = 19.1285, range [ 17.5614 - 22.0865 ]
    energy: total = 1784.69, mean = 55.7716, range [ 54.3991 - 57.2621 ]

  Spherical Harmonic fit:
    condition numbers for lmax = 2 -> 8: 1.00912 1.022 1.03874 1.09274 

  Symmetry of sampling:
    norm of mean direction vector = 0.0046772
```

Good to know it was worth the effort to re-design our standard acquisition scheme over here.  :wink:  Looks like I did a decent job. :blush: 

Note: when I ran this, it was only on a file that had the directions for the non-bzero shell.  Running it on the final scheme with the interspersed bzeros also gives this complete output for the bzero shell, which is indeed completely useless, and a bit weird/confusing as well (due to `nan` values for condition numbers, and `inf` values for the energy range upper value, etc...).  Maybe we should just not output all those stats for a b=0 "shell".  The output could still mention its presence and number of "directions" (images/entries) though.

Lestropie:
- Given the effort of naming commands based on their input / output data & operations, seems unusual to have one command with the `dir*` prefix able to receive an image as output, but not all of them. Moreover, the diffusion gradient table isn't the only potential source of a direction scheme from an image: Consider for example `sh2amp` [writing](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/sh2amp.cpp#L109) a "directions" key-value field to the output header. Would not have thought a `mrinfo -dwgrad` or `-export_grad_mrtrix` before `dirstat` was expecting too much from a user :-/

- Output for a *b*=0 shell definitely needs to be switched off, or reduced to just a count.

- The command-line options are inconsistent with other "script-friendly" capabilities e.g. `mrinfo`. There you either don't request a specific field, and get a pretty-printed output to `stderr`, or you do request a specific field, and get just the value(s) of that field only on `stdout`.
With these changes, using one or more (but not all) of the command-line options to get only a subset of fields, but then still pretty-printing those results, isn't really providing any capability over and above the default usage; it's just masking certain fields from the pretty printout. Having to combine those options with the `-quiet` option to access just the values for reading via a script, as well as receiving those values on `stderr` rather than `stdout`, feels unusual to me.

jdtournier:
OK, coming back to this after a slightly manic couple of weeks...

> - Given the effort of naming commands based on their input / output data & operations, seems unusual to have one command with the dir* prefix able to receive an image as output, but not all of them. Moreover, the diffusion gradient table isn't the only potential source of a direction scheme from an image: Consider for example sh2amp writing a "directions" key-value field to the output header. Would not have thought a mrinfo -dwgrad or -export_grad_mrtrix before dirstat was expecting too much from a user :-/

I see your point, but one of the the usefulness of `dirstat` in practice is to inspect the DW gradient in random datasets - I suspect that this will prove to be the single most common use case for general users. While it is perfectly possible to export the table to file and run the command on that, it is simply more convenient to allow it to run directly on the image. And I don't think it makes the command any less ambiguous - sure, it'll only look for the `dw_scheme`, but it's still perfectly possible to extract the other potential candidates from the header in those rare cases where that might be needed. I'm not massively fussed either way, but I do this often enough that I personally find the added convenience worthwhile. 

> - Output for a b=0 shell definitely needs to be switched off, or reduced to just a count.

OK, I have no issues with reducing the amount of output to the bare essentials for b=0 (i.e. only the stuff that makes sense, which is basically the number of volumes). But I'm not sure how to handle small _b_ shells (e.g. in my previous output, b=0.3149276342462895) - I think I'd rather keep providing the full information there, even if it would in most cases be pretty meaningless. The issue there being that sometimes that information may be desired by the user. If it can be computed, then I think it should... 

> - The command-line options are inconsistent with other "script-friendly" capabilities e.g. mrinfo. There you either don't request a specific field, and get a pretty-printed output to stderr, or you do request a specific field, and get just the value(s) of that field only on stdout.
> With these changes, using one or more (but not all) of the command-line options to get only a subset of fields, but then still pretty-printing those results, isn't really providing any capability over and above the default usage; it's just masking certain fields from the pretty printout. Having to combine those options with the -quiet option to access just the values for reading via a script, as well as receiving those values on stderr rather than stdout, feels unusual to me.

Sure, happy to re-think the interface. Only slight issue is that requesting e.g. the mean nearest-neighbour angle and its range, and the energy per shell for both uni and bipolar models on a multi-shell scheme will dump a lot of information out, and it's not obvious what's what. With the current approach, you can check what you're getting before suppressing the verbose output. Maybe we could add a switch to turn on the verbose selective output instead...?

thijsdhollander:
> I see your point, but one of the usefulness of dirstat in practice is to inspect the DW gradient in random datasets - I suspect that this will prove to be the single most common use case for general users. While it is perfectly possible to export the table to file and run the command on that, it is simply more convenient to allow it to run directly on the image.

I also vote for convenience.  The command could output a warning that mentions it only uses the DW gradient table in this scenario, and no other kinds of direction sets that may be present in other header entries; but I doubt that even that is needed.  The intuition of the most common/sensible scenario speaks for itself here, I reckon.

> OK, I have no issues with reducing the amount of output to the bare essentials for b=0 (i.e. only the stuff that makes sense, which is basically the number of volumes). But I'm not sure how to handle small b shells (e.g. in my previous output, b=0.3149276342462895) - I think I'd rather keep providing the full information there, even if it would in most cases be pretty meaningless. The issue there being that sometimes that information may be desired by the user. If it can be computed, then I think it should...

Hmm, not sure.  We do use certain definition(s) of what is a b=0 "shell" here and there.  `dwiextract -bzero` uses a default of <10 equals bzero, or the config file entry "BZeroThreshold" to override it in case the software is set up to deal with some very unusual data.  That has worked pretty well for all scenarios that need the b=0 separately for "special" purposes.  Why not use the same criterium/mechanism for `dirstat`?  For reference: https://github.com/MRtrix3/mrtrix3/blob/master/cmd/dwiextract.cpp#L77 

...and if really needed (e.g. a scenario that deviates from what is, I reckon, a sensible default), the user can still extract the scheme and just feed the b=0 part of the directions separately to `dirstat`.  This may sound less convenient again, but that's ok if we think of it as a less default scenario.  I fully value the convenience argument (see my previous point as well), but only for a sensible "default" scenario.  By default, a b-value less than 10 is _reasonable_ to regard as essentially a b=0, and then the output shouldn't be cluttered by confusing, well, essentially nonsense (with all the artificial NaNs, etc...).  I think we're otherwise going to get a range of confused users on the forum that `dirstat`'ed their single-shell dataset, with half of the output being that confusing b=0 bit.

Another compromise could be a "`-full`" or something-kind-of option to `dirstat` that doesn't compress the b=0 bit of the output to a minimal one-liner?  That would still allow for convenience in that scenario, while not intervening with the (what I reckon is the more sensible) default scenario.

thijsdhollander:
> Hmm, not sure. We do use certain definition(s) of what is a b=0 "shell" here and there.

A further argument to my ramblings on how we treat the b=0 data exceptionally in other places, one that I just stumbled upon / got reminded of again: opening DWI data via the "ODF display" tool's functionality to show amplitude ("dixel") data, one can select the shell in the tool's interface, but the b=0 data is not an option in the "shell" drop down menu.  That's fine, by the way; this again makes full sense to me, given that the b=0 data isn't expected to be "angular" in nature.

Lestropie:
> opening DWI data via the "ODF display" tool ... the b=0 data is not an option in the "shell" drop down menu

That's probably laziness as much as anything else: The input data couldn't be used as-is, the tool would need to generate a sphere and then duplicate the mean *b*=0 data for all vertices. So I didn't bother. :man_shrugging:

thijsdhollander:
> That's probably laziness as much as anything else

Good feature though. :wink: 

> the tool would need to generate a sphere and then duplicate the mean b=0 data for all vertices

That would in a way already be a fit in principle, rather than directly displaying the data (which is what the "dixel" visualisation's main goal is, I would reason).  Just to conclude that the b=0 would again need different/special treatment to make it even sensible in the context of this tool.  But that the more sensible thing to do is probably not offer an option to select it here at all, since it would probably only confuse people.

jdtournier:
OK, come to think of it, if the user _really_ wants stats for the _b_=0 shell, we could allow them to request them via the `-shell 0` option. So that would mean we don't need to display the _b_=0 shell by default, and use the default interpretation of shells. I'll work on that when I find the time. Hopefully that would be OK with everyone...?

thijsdhollander:
> OK, come to think of it, if the user really wants stats for the b=0 shell, we could allow them to request them via the -shell 0 option. So that would mean we don't need to display the b=0 shell by default, and use the default interpretation of shells.

That's even better than a "special" `-full`, `-complete` or something option.  Sounds perfect! :+1:

Lestropie:
I'm happy for `dirstat` to take images as input now; the practicality beats out the naming nomenclature. I would however like the reporting of specific fields to behave more like `mrinfo` / `mrstats`. I'll also integrate this test into `dwipreproc` to check acquisition schemes (ran into precisely such data today) as part of this PR, but the interface needs to be finalised before I can do that. Using `-shells 0[,...]` to override hiding the *b*=0 stats sounds good too.

jdtournier:
OK, I think these latest changes will probably be sufficient now. I reckon the `-output` option I've added probably behaves much more in line with expectations now... Also added quite a bit of documentation while I was at it.

Lestropie:
@jdtournier: Make sure you're happy with 9f883e8. Otherwise your changes are spot-on.

jdtournier:
More than happy with your suggestions. Just fixed the outstanding memalign issue, I'll merge as soon as testing works out.

Issue #1071, opened 2017-07-25
warpconvert: Invert field


Lestropie:
As required for #874. Need the option to calculate the inverse of a displacement / deformation field.

Alternatively, would this be considered something other than a "conversion" and hence require a different command? E.g. `warpcalc` (since `transformcalc` has a `-inverse` option)?

Would also suggest that the `-type` option in `warpconvert` should in fact be a compulsory argument, since the default behaviour when this option is not provided is not entirely obvious.

Issue #1084, opened 2017-08-10
dwipreproc: Rework odd axis dimension solution


Lestropie:
Currently, if the SE-EPI image series provided to `topup` has one or more spatial axis with odd dimension, these images are cropped by `dwipreproc`; if this is not performed, `topup` crashes  when it tries to downsample the images. However the resulting field is currently passed to `eddy` as-is (and this worked for my testing in the past). This situation now however results in `eddy` crashing with:
`"Image Exception : #3 :: Attempted to add images/ROIs of different sizes"`

So, a few things then:

- Could simply perform a corresponding crop of the DWI input series as well, and this would probably work in most situations.

- It would however be better to pad the images by duplicating the outermost slice (for each problematic axis), then crop those padded slices out after `eddy` has completed.

- More fundamentally though, this would suggest that at some point, the output field from `topup` and the input DWIs to `eddy` are being concatenated and hence assumed to lie on the same grid. Therefore, if the two main inputs to `dwipreproc` (i.e. the DWI and the `-se_epi` option) are not on the same image grid, cropping / padding axes only solves one of a number of problems.
The most general solution would be to use `topup` to estimate the field, register the input to `topup` to the input to `eddy`, *re-grid* the estimated field to that of the input DWIs, then run `eddy`. This is basically the second option described for #874. It is however made more difficult by the fact that the primary output from `topup` (i.e. the one that `eddy` actually uses) is parameterised as spline coefficients, and therefore resampling these spline coefficients based on a new voxel grid would be quite tricky.

What I thought may have been the best approach is:

- If the DWIs have one or more axes with odd dimension, pad appropriately.

- If provided separately, resample the SE-EPI images to the same voxel size & dimensions as the DWIs.

- Run `topup`.

- Perform rigid-body registration such that the corrected output from `topup`, when simulated to have the same phase-encoding as the first volume in the `eddy` input, is aligned with that first volume in the `eddy` input.

- Run `eddy`.

This however would only work if `eddy` were to honour the header transformation in the `topup` output field image (the one containing the spline coefficients). If it *doesn't* (as I suspect is the case, given that it appears to be concatenating this field with the DWIs), then this approach doesn't work.

So, thinking aloud, here's an alternative:

- Pad DWIs for odd dimensions as before

- Run `topup` on the native SE-EPI data (padding odd dimensions if necessary)

[...]

PR #1159, opened 2017-10-05
Dwipreproc: -align_seepi option

Lestropie:
OK, here we go again... :worried: 

Related to #874, #1051, #1084.

An explicit option `-align_seepi` is added to the `dwipreproc` script, which instructs the script that the user wishes to enforce alignment between the inhomogeneity field as estimated by `topup`, and the estimations & corrections performed by `eddy`. It does this by ensuring that the first volume in the series input to `topup` is precisely the same volume as the first volume in the series input to `eddy`.

More specifics & addendums:

- If the volumes within the `-se_epi` input do not contain any phase-encoding contrast, previously the script would immediately abort, as it was impossible to estimate the inhomogeneity field. Now, if this occurs *and* the `-align_seepi` option is provided, the script will concatenate the DWI *b*=0 volumes with the `-se_epi` input, and re-check the requirement of phase-encoding contrast. This is to account for e.g. labs that acquire a single P>>A *b*=0 and then their A>>P DWIs; but also, if one is not going to acquire *all* DWIs with reversed phase-encoding, rather than e.g. acquiring one A>>P *b*=0, one P>>A *b*=0, then A>>P DWIs, for the same total acquisition time one could instead acquire ~ 5 P>>A *b*=0 images then A>>P DWIs, giving more data to `topup`. This is especially true for multi-band acquisitions due to the extra calibration time.

- If the first volume in the DWIs is not a *b*=0 (unlikely, but nevertheless possible), the DWI volumes are permuted such that the first *b*=0 volume becomes volume index 0, and then permuted back to their original order after `eddy`. This is necessary due to the `topup` / `eddy` silent mutual agreement regarding the reference space.

- If the `-se_epi` input image does not match the DWI voxel grid, it is immediately re-gridded to the DWI voxels. Even if the `-align_seepi` option is *not* used, it's still necessary for the `topup` estimated field and the DWIs input to `eddy` to be defined on the same grid: `eddy` appears to perform some internal concatenation step between these data, and so errors out if they don't match. Reports [here](http://community.mrtrix.org/t/dwipreproc-on-multiband-hpc-like-dsi-failed/1185) and [here](http://community.mrtrix.org/t/errors-with-dwidenoise-and-dwipreproc/1174).

- If `-align_seepi` is used, check header fields for information relating to TE, TR and flip angle, since these influence image contrast and may therefore invalidate the assumptions required for this option to work. Also capture these fields from DICOM import.

- `topup` cannot operate on images where one of the three spatial dimensions has an odd number of voxels; likely due to downsampling by a factor of 2 and experiencing a buffer overflow. Previously this was dealt with by cropping the image provided to `topup` by one voxel along relevant axes. However this removes image information, and stopped working when `eddy` started performing an internal concatenation step (since the DWIs were not also correspondingly cropped). This has been altered to not only manipulate both the DWI and SE-EPI images, but also to *pad* problematic axes with duplicate image information as opposed to cropping, and then crop the resulting series after `eddy` has completed. Therefore the image output by the script should always have the same size in the three spatial dimensions as the input image.

- `warpinvert`: New command, but not actually used (was going to be necessary for an alternative `dwipreproc` alignment approach).

There's a bit of gymnastics going on, and a lot of possible code paths, so I'm not expecting that anyone's going to be able to spot issues from the code alone; but you never know.

I've tested it on a reasonable range of possible input data configurations, including presence & absence of the above confounds, and different `-rpe_*` options. But ultimately it probably needs field testing...


Lestropie:
Closes #1071.

Lestropie:
Took me a lot of attempts, but this finally passed testing...

Merging as I've got multiple changes to `dwipreproc` stacking up, and this is unlikely to get more testing than I've already thrown at it by just sitting here. So I'm hoping to get all of the changes into `dev`, and then have our RAs giving all of the changes a good testing.

PR #1160, opened 2017-10-06
configure: handle path translation when importing settings from pkg-config

jdtournier:
Proposed solution to the issue raised [here](http://community.mrtrix.org/t/configure-and-build-fails-on-windows-install/323/4?u=jdtournier) and [here](https://github.com/MRtrix3/mrtrix3/pull/1090#issuecomment-334740664). Turns out there's a `cygpath` executable provided with MSYS2 (actually part of cygwin) to translate paths, so all this does is try to use that to translate include paths... Seems to allow _MRtrix3_ to build properly using the MinGW version of python. 

Note this only translates header paths (i.e. starting with `-I`), not library search paths (which start with `-L`). Still, it builds fine with the default packages, so I've not done anything about it at this point. Worth bearing in mind for the future though...

### DIFF
--- a/configure
+++ b/configure
@@ -485,6 +485,16 @@ def compare_version (needed, observed):
 
 
 
+
+
+
+def MSYS2_translate_path (path):
+  try:
+    return execute ([ 'cygpath.exe', '-w', path ], RuntimeError)[1]
+  except:
+    return path
+
+
 def get_flags (default=None, env=None, pkg_config_flags=None):
   """Return a list of the flags required for a given packagei
   
@@ -501,7 +511,7 @@ def get_flags (default=None, env=None, pkg_config_flags=None):
       flags = []
       for entry in shlex.split (execute ([ 'pkg-config' ] + pkg_config_flags.split(), RuntimeError)[1]):
         if entry.startswith ('-I'):
-          flags += [ '-isystem', entry[2:] ]
+          flags += [ '-isystem', MSYS2_translate_path (entry[2:]) ]
         else:
           flags += [ entry ]
       return flags
@@ -602,7 +612,7 @@ exe_suffix = ''
 lib_prefix = 'lib'
 
 system = platform.system().lower()
-if system.startswith('mingw') or system.startswith('msys'):
+if system.startswith('mingw') or system.startswith('msys') or system.startswith('windows'):
   system = 'windows'
 report ('Detecting OS: ' + system + '\n')
 if system == 'linux':


jdtournier:
There's been a handful of reports of issues that seem related to this issue on the forum over the last couple of days - see [this post](http://community.mrtrix.org/t/error-relating-to-c-compiler-when-running-configure-on-windows/454/16?u=jdtournier) and links therein. There is some urgency with this - can anyone review this and double-check that it does what it's supposed to...?

Note that testing this properly will probably need to be performed on an up to date installation of MSYS2 - you'll need to [update that first](https://github.com/msys2/msys2/wiki/MSYS2-installation#iii-updating-packages) to verify that there is indeed a problem on `master`, then checkout this branch and verify that this fix does what it's supposed to...

Lestropie:
On `master`:
- MinGW python:
  - `    ./configure` works
  - `    EIGEN_CFLAGS="-I/mingw64/include/eigen3" ./configure` works
  - `    EIGEN_CFLAGS="-isystem /mingw64/include/eigen3" ./configure` **does not work**
- MSYS python:
  - `    ./configure` works
  - `    EIGEN_CFLAGS="-I/mingw64/include/eigen3" ./configure` works
  - `    EIGEN_CFLAGS="-isystem /mingw64/include/eigen3" ./configure` works

On `fix_mingw_python_path_handling`:
- MinGW python:
  - `    ./configure` works
  - `    EIGEN_CFLAGS="-I/mingw64/include/eigen3" ./configure` works
  - `    EIGEN_CFLAGS="-isystem /mingw64/include/eigen3" ./configure` **does not work**
- MSYS python:
  - `    ./configure` works
  - `    EIGEN_CFLAGS="-I/mingw64/include/eigen3" ./configure` works
  - `    EIGEN_CFLAGS="-isystem /mingw64/include/eigen3" ./configure` works

So the only consistent fault I get is explicitly using the "`-isystem `" prefix on MinGW python.

I can't confirm whether or not your changes fix the issue. What I reported back [here](https://github.com/MRtrix3/mrtrix3/pull/1090#issuecomment-333152802) was using a manual `eigen3` install rather than through `pacman`; and as such, I couldn't verify whether `pkg-config` was working, let alone `cygpath`. Now that I've deleted that and installed `eigen3` through `pacman`, `configure` works out of the box for me on both branches, for both versions of `python`.

Also note the formation of `PATH` and `eigen_cflags` in `config`:
On `master`:
- MinGW python:
  - `    PATH = r'C;\msys64\mingw64\bin;C;\msys64\home\Rob\src\mrtrix3\bin;C;\msys64\ ...`
  - `    eigen_cflags = [ '-isystem', 'C:/msys64/mingw64/include/eigen3' ]`
- MSYS python:
  - `    PATH = r'/home/Rob/src/mrtrix3/bin:/home/Rob/src/robs_0.3/bin:/home/Rob/ ...`
  - `    eigen_cflags = [ '-isystem', 'C:/msys64/mingw64/include/eigen3' ]`

On `fix_mingw_python_path_handling`:
- MinGW python:
  - `    PATH = r'C;\msys64\mingw64\bin;C;\msys64\home\Rob\src\mrtrix3\bin;C;\msys64\ ...`
  - `    eigen_cflags = [ '-isystem', 'C:\msys64\mingw64\include\eigen3' ]`
- MSYS python:
  - `    PATH = r'/home/Rob/src/mrtrix3/bin:/home/Rob/src/robs_0.3/bin:/home/Rob/ ...`
  - `    eigen_cflags = [ '-isystem', 'C:\msys64\mingw64\include\eigen3` ]`

So for me it looks like the two different `python` versions have different "native" representations of paths, as evident by the reported `PATH`, but all `cygpath` is doing in `fix_mingw_python_path_handling` (at least for me) is swapping forward slashes to backslashes. 

Also: `EIGEN_CFLAGS="-isystem C:/msys64/mingw64/include/eigen3" ./configure` (i.e. manual Windows path rather than manual Unix path) works on either version of `python`.

So I have no idea how others are encountering the issue, assuming they have installed `eigen3` via `pacman`.

jdtournier:
superseded by #1176 - closing.

PR #1170, opened 2017-10-17
mrview: overlay load

maxpietsch:
`-overlay.load` selects the last overlay image so that other command line argument parameters (opacity, interpolation) apply only to the last overlay image loaded. 

As requested [here](http://community.mrtrix.org/t/mrview-multiple-overlays-colourmaps/1244/1).


Lestropie:
While this solves the issue as reported, I'll take the chance to question whether the behaviour RE: selection v.s. clear & selection should be altered for cases over and above loading from the command-line.

Example: You have one overlay already loaded, and it is selected. You then open a second overlay via the GUI. Currently, this will result in *both* overlays being selected, and hence any toolbar elements subsequently used will be applied to both overlays unless these selections are first altered. I wonder if it would instead be preferable to *always* clear the current set of selected overlays when loading & selecting new overlays, regardless of whether that loading occurs using the `-overlay.load` command-line option or the GUI load button.

I'm in preference of the latter alternative, but content for it to remain unchanged.

jdtournier:
>  I wonder if it would instead be preferable to always clear the current set of selected overlays when loading & selecting new overlays, regardless of whether that loading occurs using the -overlay.load command-line option or the GUI load button.

I think that would be my preference too...

PR #1172, opened 2017-10-19
Fix set_path to look for mrinfo.exe on Windows

jdtournier:
As reported in [this post](http://community.mrtrix.org/t/error-mrtrix3-executables-not-found-in-expected-location-in-windows-10-64-bit-installation/1253?u=jdtournier).

### DIFF
--- a/set_path
+++ b/set_path
@@ -17,13 +17,25 @@
 # particular). This will not work for the C shell and derivatives. 
 
 
-import sys, os, platform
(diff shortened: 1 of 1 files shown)


dchristiaens:
Jana ran into what seems to be the same issue today, despite already using the latest master. Could it possibly not yet be resolved?

jdtournier:
See #1176...

Issue #1173, opened 2017-10-20
Store slice axes in header for subsequent mrdegibbs


bjeurissen:
I was thinking it would be useful to store the slice axes in the mif header when importing from DICOM. This information can then be used to make sure `mrdegibbs` is using the correct slice axes, e.g. when processing data that was acquired with sagittal instead of axial slices. It could also be useful for other future commands that operate on acquisition slices.

Lestropie:
Yes, I should ideally grab this for #1169 as well. I didn't bother at the time since `eddy` assumes `k` is the slice axis, but since it's applicable here too it's probably worth the effort.

bjeurissen:
Definitely!

Lestropie:
Initial attempt at capturing the data in eb447bc. No corresponding modification to `mrdegibbs` yet, and `dwipreproc` currently only does an error check (ideally it would permute axes if required so that the slice encoding direction is `k`, since `eddy` assumes that to be the case and doesn't offer a command-line input to change this).

Could also do with confirmation from @jdtournier that the approach / interpretation is correct. My understanding of DICOM is still somewhat limited.

Lestropie:
Closed by #1175; will go out to `master` on next tag.

PR #1175, opened 2017-10-24
Slice timing

Lestropie:
Second of three PRs targeted particularly at `dwipreproc` (wanted to merge #1159 before this one). Some verbose commit messages in there that contain a decent amount of information; but the crux is:

- Capture slice encoding direction and slice timing from DICOM, and store them in BIDS-compatible header fields.

- In `dwipreproc`, make sure that if the slice encoding axis is known, it's the third axis. `eddy` doesn't have a command-line option to set this; ideally `dwipreproc` should permute axes in order to support this, but that could have weird side-effects, e.g. when `eddy`'s text file outputs are interpreted (which is the third of the three PRs). If `--mporder` is provided to `eddy`, either grab the text file that the user provides via `--slspec=`, or construct such a file based on the slice timing information present in the image header (which I expect to be the most common usage).

- In `mrdegibbs`, if the slice encoding axis is known, compare it to the default / user-defined within-slice axes selected, and report / modify as appropriate (as requested in #1173).


Lestropie:
Okay, I think I've done everything I want to with this branch. I had a brief look at permuting image axes if the slice encoding direction is not *k* or *k-*, but this was going to be quite difficult, since axis permutation may also affect phase encoding direction if one is not careful, and order of axis permutation relative to other functions may cause issues. So I'm giving up on that for now; if someone has a non-axial DWI acquisition, they can deal with the data manually (they'll at least get a meaningful error message from `dwipreproc`).

Unlikely to attract much critique here, so would prefer to merge and have it more thoroughly tested in `dev`.

One possible point of contention would be the creation of `core/axes.h` and `core/axes.cpp`, for dealing with encoding of axis & direction as `[ i, i-, j, j-, k, k- ]`; previously this only applied to phase encoding, so lived in `/core/phase_encoding.*`, but is now also used for slice encoding; and although the nomenclature has origins from NIfTI (apparently), the functionality doesn't only apply to NIfTI. So it's hard to find a place for.

Other thing is `core/file/dicom/mapper.cpp`: I *think* that certain lines should refer to frame properties rather than image properties (changed in 3c474c6 and 335038b), but when I did a manual merge in #1159 I accidentally reverted. This PR should restore the earlier fixes; @jdtournier might want to double-check.

PR #1177, opened 2017-10-24
Persistent build mode

jdtournier:
This adds an option to `build` to keep re-trying a build upon failure, until no new jobs succeed. This is to try to speed up the build on TravisCI whilst making it resilient to out of memory errors. Please treat this as an experiment at this stage...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   - py=python2
   - py=python3
 install:
-  - export NUMBER_OF_PROCESSORS=4
+  - export NUMBER_OF_PROCESSORS=2
   - export PATH=`pwd`/bin:${PATH}
   - export PYTHONPATH=`pwd`/lib:${PYTHONPATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
@@ -28,7 +28,7 @@ install:
   - pip3 install --user pylint recommonmark sphinx sphinx-rtd-theme
 script:
   - ${py} -m sphinx -n -N -W -w sphinx.log -c docs/ docs/ tmp/
-  - ./check_memalign && $py ./configure -assert && time $py ./build -nowarnings -persistent && ./run_tests
+  - ./check_memalign && $py ./configure -assert -nooptim && time $py ./build -nowarnings && ./run_tests
   - PYTHON=$py ./run_pylint
   - ./docs/generate_user_docs.sh && git diff --exit-code docs/ > gitdiff.log || echo "Documentation is not up to date; run docs/generate_user_docs.sh and commit changes"
 after_failure:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   - py=python2
   - py=python3
 install:
-  - export NUMBER_OF_PROCESSORS=4
+  - export NUMBER_OF_PROCESSORS=8
   - export PATH=`pwd`/bin:${PATH}
   - export PYTHONPATH=`pwd`/lib:${PYTHONPATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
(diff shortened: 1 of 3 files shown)


jdtournier:
While we wait to see whether what I've just tried is going to make any difference: according to the [TravisCI docs](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments), we're limited to 2 cores anyway... There's probably little point in increasing the number of concurrent compile jobs - in fact, we might be better off actually _reducing_ the number of cores. 

Not sure how we're going to get around TravisCI's 50 minute time limit...

jdtournier:
Using 2 processes, the `./build` command itself takes:
```
real	39m52.781s
user	77m43.446s
sys	1m44.528s
```
Using 8 processes went over time. Let's try 4 again, and time that...

Lestropie:
```
real	39m51.347s
user	77m44.159s
sys	1m46.409s
```
and
```
real	41m42.433s
user	80m56.102s
sys	2m10.097s
```

jdtournier:
Right. Well I guess this tells us we might as well reduce the number of processes down to 2... We can drop this PR, there's nothing of use here.

No idea how to move forward in future. The paid TravisCI plans are pretty steep. I'll see whether there's any funds and appetite to cover this from the department... I'll also contact TravisCI directly to ask what concessions they might make for open source projects.

thijsdhollander:
Ok, just coming in very naively here, but: would there be a way to split the build/test of certain groups of commands in different (even predefined) batches?  I'm not sure what TravisCI allows (or doesn't), but we do have separate "tests" ("build jobs") for the Python2 and Python3 parts of the tests.  Isn't there a possibility to split things in other/extra ways?

Then again, thinking about how--if there was such a way--this would essentially circumvent limitations that TravisCI may want to impose, I'm probably imagining impossible things... :thinking:

Lestropie:
The `pylint` tests could theoretically be run separately with some build matrix magic. That would also reduce install time of the larger jobs since they would no longer need to install `pip` etc..

Unfortunately the documentation test requires that the binaries be built; so you could have one job that compiles and runs the tests, and another that compiles and builds the documentation, but that's a lot of wasted CPU cycles.

Also need to look at whether we can reduce CI setup time using [caching](https://docs.travis-ci.com/user/caching/).

jdtournier:
Well, the primary issue really is just the build time... It takes 40 minutes just for the `build` command to complete... When I monitored the build yesterday, it took 5 minutes before `build` was actually invoked, most of it pulling dependencies. We could shave off a couple of minutes by moving the `pylint` tests to a different job, but realistically that'll only buy us a tiny bit of headroom, and we'll no doubt hit the limit again later on. 

One option we haven't tried yet is building non-optimised binaries. They're currently built with asserts, which still implies `-O3`. We don't necessarily want to build with `-debug`, since that adds full debugging symbols (`-g`), which might slow things down again. I'll add a `-nooptim` flag to `configure` to specifically disable optimisation, see if that improves build times...

jdtournier:
OK, build times on my desktop (Core i7, 6-core with HT):

- using the default `-assert`:
   ```
   real    7m51.320s
   user    80m20.417s
   sys     1m5.686s
   ```

- using the new `-assert -nooptim`:
   ```
   real    4m8.749s 
   user    44m7.981s
   sys     1m4.229s 
   ```

So almost half the build time... 😜 

Let's see if that works out on TravisCI...

jdtournier:
OK, maybe I spoke too soon... The tests themselves now run excruciatingly slowly - on my desktop again:

- using the default `-assert`:
   ```
   real    1m59.910s
   user    6m54.332s
   sys     4m43.916s
   ```

- using `-assert -nooptim`:
   ```
   real    17m55.693s
   user    107m17.911s
   sys     2m27.117s
   ```

😵  that completely kills any speedup from the compile process itself... Back to the drawing board.

But interesting to see how much of a difference optimisation can make, hey?

jdtournier:
OK, before we spend any more time on this: I have the go-ahead from the boss to spend as much as required to support MRtrix. I'll get a quote from TravisCI, and hopefully we can just keep going without having to restructure everything... I'll keep you posted.

PR #1180, opened 2017-10-25
Remove -failonwarn option

Lestropie:
This one has given me the irrits for a while, so thought I'd make the change and see what people think.

Ages ago I added the "`FailOnWarn`" config file entry, with the intention that if one was running large amounts of processing on computing hardware, it may be preferable to abort execution if a warning is issued, rather than having warnings displayed but lost in the noise. However I've tried to use this myself in the past and failed because of the frequency of warnings issued; for instance when doing something to an FOD image that still had `dw_scheme` lying around. Still in the process of cleaning some of these up.

At the same time I also added a `-failonwarn` option to every command, that would engage this behaviour as a one-off. I suspect now that I added this simply because I could; now I'm having difficulty envisaging anybody actually using this option ever. Indeed I intentionally didn't add this to the list of standard options in the Python scripts from the outset.

So I'm proposing removing the `-failonwarn` option, just to reduce the bloat in command help pages by just a little bit. Open to thoughts on the matter.

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -48,9 +48,8 @@ namespace MR
       + Option ("debug", "display debugging messages.")
       + Option ("force", "force overwrite of output files. "
           "Caution: Using the same file as input and output might cause unexpected behaviour.")
-      + Option ("nthreads", "use this number of threads in multi-threaded applications (set to 0 to disable multi-threading)")
+      + Option ("nthreads", "use this number of threads in multi-threaded applications (set to 0 to disable multi-threading).")
         + Argument ("number").type_integer (0)
-      + Option ("failonwarn", "terminate program if a warning is produced")
       + Option ("help", "display this information page and exit.")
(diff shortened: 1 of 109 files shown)


Lestropie:
That needs to be there since it's defined as `extern` in the corresponding `.h` file. Other similar app settings are loitering around in the same place.

jdtournier:
Sorry, did you remove the `fail_on_warn` variable? If it's also in the header, let's get rid of it there too...

Lestropie:
The variable's still there because it's still possible to set `FailOnWarn` in the config file. I still think that may be useful, e.g. when running on HPC, but I need to set it on my own systems and start revising what feedback does or does not need to be warning level, and what can potentially be modified to avoid such warnings altogether.

jdtournier:
Oh yes, of course, makes sense.

PR #1181, opened 2017-10-25
Experimenting with TravisCI changes

Lestropie:
Try to construct a build matrix where each job only runs a subset of the required tests.

No idea if this'll work, and AFAIK no way to know without triggering a CI test. So let's just see what happens...


Lestropie:
Despite the [comment](https://github.com/MRtrix3/mrtrix3/pull/1177#issuecomment-339341553) in #1177, I think this is still worth using: It gives much more immediate & clear feedback as to the aspect of testing that has failed, & makes it possible to do a quick abort & re-submission if one of the faster aspects of testing fails. E.g.:

```
 2051.1  Compiler: clang C++  py=python2 test=sphinx  2 min 30 sec
 2051.2  Compiler: clang C++  py=python3 test=sphinx  3 min 1 sec
 2051.3  Compiler: clang C++  py=null test=memalign   2 min 49 sec
 2051.4  Compiler: clang C++  py=python2 test=pylint  2 min 39 sec
 2051.5  Compiler: clang C++  py=python3 test=pylint  2 min 50 sec
 2051.6  Compiler: clang C++  py=python2 test=build   46 min 27 sec
 2051.7  Compiler: clang C++  py=python3 test=build   48 min 32 sec
```

Haven't found any clear answers with respect to caching of install dependencies. There's a chance that they could potentially all start up faster, but it'd take a lot of playing around.

I also think we should be testing an up-to-date `gcc` in addition to `clang`. There's been a few compiler warnings that have crept in that `clang` presumably hasn't picked up.

jdtournier:
OK, couple of comments:

- no issues with breaking it up into distinct jobs - it does indeed make it a bit clearer what has failed. 

- our current plan includes 5 concurrent jobs. But given that two of them will take up a lot more time than the others, I don't think that'll influence the overall run time in the slightest. 

- you're still using 4 processes, which provides no advantages given that we only get 2 processes - might help keep the GCC tests within RAM limits too. 

- is there any point in performing the sphynx and pylint tests in both python 2 & 3? I can just about see the rationale for pylint, but no sure I see the point for sphynx...

- I've a mind to simplify one of the python tests to get it to use a bogus 'compiler' that just returns creates an empty output file and returns success every time - would allow it to verify correct operation of the python code, without taking up time running the exact same commands as the other python test is going to run anyway...

Lestropie:
> But given that two of them will take up a lot more time than the others ...

Well, assuming (almost) everything runs in parallel it won't reduce the time to completion. But they may not always run fully in parallel. I've tried putting the longest job to the top, so other smaller jobs can run while that's going.

> ... might help keep the GCC tests within RAM limits too.

Maybe. I'm also trying the `-nooptim` flag you added in #1177: Might reduce RAM usage, and will also reduce overall time if e.g. we want to test different versions of different compilers.

> Is there any point in performing the sphynx and pylint tests in both python 2 & 3?

I have some vague recollection of a build warning in Sphinx that only appeared in one of the two versions.

For pylint it will almost certainly be the case, given the amount of trickery required to get scripts working in both 2 and 3. First one that comes to mind (not wanting to pick on it, just first I thought of) was [use](https://github.com/MRtrix3/mrtrix3/blob/master/bin/population_template#L142) of `raw_input()` in `population_template`, which was renamed and hence no longer defined in Python 3; that would definitely succeed with one version of pylint but not the other.

> I've a mind to simplify one of the python tests to get it to use a bogus 'compiler' ...

That could work. At the moment I'm just using one Python version for the job that actually executes the tests, and another for those that compile without optimisation but don't actually run the tests. So both get tested; but a dummy compiler would make the jobs more specific.

-----------

While I've added lines for caching, my suspicion is that the first caches will be created upon a successful merge, and therefore I won't be able to verify if it's working using this PR alone.

Lestropie:
Quick follow-up on this:

- Including GCC7 pushes out the minimum job time to > 10m due to downloading and installing dependencies. Unfortunately it looks as though caching is working for pip but not for packages. @jdtournier Have you been in direct contact with anyone from TravisCI that you could ask about this? I've seen other forum threads where caching kicked in further down the track, but this concerns me that it appears to have kicked in, it's just not caching what I want it to. Or we could merge to `dev`, see if it changes for the next PR, and modify if necessary. Otherwise, we might want to consider just using Clang 3.9 alone in order to reduce wasted cycles.

- Compiling GCC7 with `-nooptim` required a slight tweak (549e495).

Lestropie:
@jdtournier Would you like to try these changes on `dev`? If so, I'll re-implement these changes as a single commit so as not to bloat the git history with experimentation. I'm hoping that the build time of each job will decrease once the target branch is cached.

jdtournier:
OK, had a look through the changes. A few thoughts:

- I'd rather get rid of the `-nooptim` option, it really is no use in practice. I don't think it's worth including just for the sake of slightly faster testing - it might not make any difference in practice anyway if it runs in parallel with the clang test. Also might be fast enough using `-debug` - I'll investigate. 

- I also think we can remove the `-persistent` option. It's not used in testing anyway.

- might clean up some of the scripts using bash `case` statements - just because...

jdtournier:
Also: I wouldn't reduce this to a single commit - it's good to document all the experimentation you did, I'd keep the commit history as-is, even if it is a bit messy...

jdtournier:
Right, `-debug` does not produce the kind of speedup in the build that `-nooptim` does. It's a shame, since that `#ifdef` in the middle of `dirgen.cpp` is a bit ugly... :cry: 

I'll leave `-nooptim` in, otherwise let's wait to see whether the couple of other changes will test out OK. I notice TravisCI is really struggling at the moment, not sure what's going on...

jdtournier:
Also, I do wonder about whether it's worth ditching the GCC7 tests. It requires tweaking to the actual code (due to `-nooptim`), fetching (and caching) additional packages, and also takes up a lengthy job, which might slow things down when we have multiple pull requests being tested concurrently (as is the case right now) - we're only allowed 5 concurrent jobs, officially...

jdtournier:
OK, on the last test:

- GCC7 build test failed (with `-nooptim`), presumably due to out of RAM error:
  ```ERROR: (270/506) [CC] tmp/cmd/mrtransform.o
  g++-7 -c -std=c++11 -pthread -fPIC -DMRTRIX_WORD64 -DMRTRIX_TIFF_SUPPORT -isystem 
  /usr/include/x86_64-linux-gnu -Wall -O0 -DNDEBUG -Isrc -I./core -Icmd -I/home/travis/build/MRtrix3/mrtrix3/../eigen -DEIGEN_DONT_PARALLELIZE cmd/mrtransform.cpp -o tmp/cmd/mrtransform.o

  failed with output

  g++-7: internal compiler error: Killed (program cc1plus)
  Please submit a full bug report,
  with preprocessed source if appropriate.
  See <file:///usr/share/doc/gcc-7/README.Bugs> for instructions.
  /tmp/tmpIiaLuy (END)
```

- clang build & run tests failed, for no reason that I can see... All it says at the end is:
  ```
  $ case "${test}" in
        "sphinx")
          cat sphinx.log
          ;;
        "memalign")
          cat memalign.log
          ;
        "pylint")
          cat pylint.log
          ;;
        "build")
          cat configure.log
          cat build.log
          ;;
        *)
          cat configure.log
          cat build.log
          cat testing.log
          cat gitdiff.log
      esac
    
  /home/travis/.travis/job_stages: eval: line 63: syntax error near unexpected token `;'
  /home/travis/.travis/job_stages: eval: line 63: `        ;'

  $ sleep 10

  Done. Your build exited with 1.
  ```
  Might be due to a missing `;;` at the end of that last case in the `after_failure` script...? But presumably that would only be called when a failure has already occurred right...? Which is borne out by the earlier statement:
  ```
  The command "./travis.sh" exited with 1.
  ```
  But there's no hint of a failure... :confused:

Lestropie:
> GCC7 build test failed (with `-nooptim`), presumably due to out of RAM error:

Yeah that'd be RAM. Would prefer to test the most up-to-date version possible for the sake of capturing new compiler warnings; but whether or not to downgrade it depends on how frequently this issue occurs (it succeeded first time for me).

> Might be due to a missing `;;` at the end of that last case in the `after_failure` script...?

Nope; after the `memalign` case.

> But presumably that would only be called when a failure has already occurred right...?

Being a syntax error, I don't think it'd need to experience a failure, it'd crash at the initial parsing phase.

Lestropie:
Following #1246, I'm thinking we should try to include OSX testing as part of CI. This is now supported by TravisCI as tabulated [here](https://docs.travis-ci.com/user/reference/overview/).  They also give OSX-specific instructions [here](https://docs.travis-ci.com/user/reference/osx/), and instructions for multi-OS testing [here](https://docs.travis-ci.com/user/multi-os/).

Lestropie:
I'd like to merge this to `dev` so that hopefully the VM caching kicks in for other branches with PRs against `dev`. #1251, and OSX testing in TravisCI, can be pursued separately. So shout out if you want this held back for any reason, otherwise I'll merge tomorrow. GCC7 test succeeded first time again, so tempted to leave it there and can address later if it comes up more often (e.g. reduce `NUMBER_OF_PROCESSORS` to 1 just for the GCC test).

Lestropie:
> Would prefer to test the most up-to-date version possible for the sake of capturing new compiler warnings; but whether or not to downgrade it depends on how frequently this issue occurs (it succeeded first time for me).

`g++` failures are currently a bit too frequent for my liking. @jdtournier Was it 90 minutes per job that we got bumped to? The `g++` build is typically 35-40 minutes, so could single-thread it, though not with huge head room. Or could use an earlier version of `g++`, but then may miss newer compiler warnings and the like.

jdtournier:
Tricky one. If you single-thread it, all it would take a little bit of extra load on the TravisCI server for the build to exceed its time limit. So I'm not sure it would necessarily solve the problem. If it's RAM, maybe we can stick to 2 cores, but use the persistent build idea that I'd proposed a while back in #1177...?

PR #1182, opened 2017-10-26
allow quiet operation from environment variable MRTRIX_QUIET

jdtournier:
As [requested on the forum](http://community.mrtrix.org/t/mrtrix-redirect-output-and-error-to-a-file/314/7?u=jdtournier)...

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -1087,6 +1087,10 @@ namespace MR
 #endif
 
       srand (time (nullptr));
+      
+
+      if (getenv ("MRTRIX_QUIET"))
+        log_level = 0;
     }
 
 


jdtournier:
Another one of those I'm not sure can justifiably be added to `master` directly, given that it's not a bug fix... But it is such a trivial change, and I can't see that it can possibly influence existing users in any way - unless they happen to have the `MRTRIX_QUIET` variable set in their environment for some reason...

jdtournier:
OK, I'll go with your suggestion - it's cleaner and ensure `log_level` is set as early as possible. I wasn't sure whether there'd be any issue with it being global, but I can't see why not... I'll check anyway and commit if it works OK. Means you'll need to re-approve...

Lestropie:
Maybe needs to be global so that `LogLevelLatch` can get to it? Either way there's similar app-wide properties stored in the same place.

jdtournier:
> Maybe needs to be global so that LogLevelLatch can get to it? Either way there's similar app-wide properties stored in the same place.

Sorry, didn't make myself clear: I know it needs to be global - I wrote it that way. The issue is that initialisation of global variables can be unpredictable, but I _think_ this is only because the _order_ of initialisation of globals is undefined - means initialising a global dependent on the value of another is fragile. Not the case here though - was just me thinking aloud about previous bad experiences on that front...

PR #1186, opened 2017-11-02
delay signal handling initialisation

jdtournier:
Someone just brought to my attention the fact that pressing Ctrl-C while a help page is displayed in the terminal renders the display unusable. I'm not sure why, but the terminal is left in a strange state, with typed commands not showing... You can see that the interrupt is handled by the invoking MRtrix command though, not the page display command (i.e. not `less`): our signal handling message pops up. 

So I figured the simplest fix was to delay setting our custom signal handlers until _after_ the command-line parsing has been done, particularly the handling of the help page. This seems to more or less fix it. 

The cleaner solution would be to have the invoking MRtrix command actually exit as soon as it's done piping its help page contents to the display command, but that turns out to be basically impossible using `popen()`, or even using an explicit `pipe()` & `fork()` - the process writing to the pipe needs to wait for the child process to finish. This seems to differ from explicitly invoking the pipe at the command-line: in this case, it seems the feeding process exits when it's done - or at least the interrupt does not get passed to the feeding process, but to the last one, not sure. For example:
```
mrinfo > help.txt
cat help.txt | less -X
```
and pressing Ctrl-C does not result in `less` terminating - far from it, it seems to handles the interrupt itself. So there's a difference between the `popen()` call and what the shell does to set up pipes... More trouble than it's worth though, so I think I'll stick with this simple solution... 

Thoughts welcome.

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -20,6 +20,7 @@
 #include "progressbar.h"
 #include "file/path.h"
 #include "file/config.h"
+#include "signal_handler.h"
 
 #define MRTRIX_HELP_COMMAND "less -X"
 
@@ -79,8 +80,6 @@ namespace MR
 
     const char* project_version = nullptr;
     const char* build_date = __DATE__;
-
-    SignalHandler signal_handler;
 
     int argc = 0;
     const char* const* argv = nullptr;
@@ -1062,6 +1061,7 @@ namespace MR
         }
       }
 
+      SignalHandler::init();
     }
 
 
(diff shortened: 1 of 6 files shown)


Lestropie:
Given this is the only place in the code where `popen()` is used, I reckon you're right that this is safest & easiest. If its use were more common, we'd want a more general solution for shutting down the process group in the signal handler.

Does as advertised on Linux; will approve once I've tested on Windows.

PR #1187, opened 2017-11-03
DICOM: Enhanced Siemens CSA capability

Lestropie:
- Expand list of element codes that are used to identify Siemens CSA data.
- Support CSA1 format.

Collaborator's DICOMs had CSA data stored at 0x0029U, 0x1110U, which *MRtrix3* was ignoring. It seems to load without issue simply by including that code in the list. So I've expanded the list of permissible codes for CSA import based on [this page](https://scion.duhs.duke.edu/vespa/project/raw-attachment/wiki/SiemensCsaHeaderParsing/csa.html).

While I've written CSA1 support based on [this page](http://nipy.org/nibabel/dicom/siemens_csa.html), I don't actually have any data to test it (I hypothesized that that was why my collaborator's data wasn't importing CSA fields, but it turned out to be malformed data masquerading as CSA). I've left the code there though; @jdtournier Got any CSA1 hiding in your DICOM private test data?


jdtournier:
Just looking into this since it might overlap with changes I'm looking into making for #1193: 

- no problem with the changes in `dict.cpp`, `image.cpp` and `quick_scan.cpp` - all sensible, and I can't see how they can cause any trouble.

- as far as I can tell, the rest of the changes (apart from removal of trailing whitespaces) boils down to interpreting the first 4 bytes as an item count if not "SV10"...? The rest of the changes provide debugging info when the unused fields don't match with expectation? Again, I don't see that hurting particularly, but since I haven't come across CSA1 data, I don't think this is something we're going to need to worry about a great deal... I'll run this on my data repo and get it to report if it does ever encounter a CSA1 entry, just to make sure. But if I don't come across one, I think I'd rather not introduce support for something we can't verify - we can look into doing this as & when the need arises. What do you reckon?

Lestropie:
> ... the rest of the changes (apart from removal of trailing whitespaces) boils down to interpreting the first 4 bytes as an item count if not "SV10"...?

That's more-or-less CSA1 in a nutshell.

Unfortunately the only case I've come across was a genuinely malformed entry. So if you don't have any examples, I'm happy for it to go; though the behaviour if such data are ever encountered should probably be to ignore the entry or at the very least issue a warning, rather than the current debug level message.

jdtournier:
OK, I've come across quite a few 'CSA1' headers in various Siemens images. But not one of the came up valid - your code immediately ignored them due to `cnum` being (much) greater than 128. So the only thing I take from this is that either these tags are _not_ supposed to contain CSA data, or your implementation is incorrect...

Note that I had loads of hits from some Skyra data, all in tag 0029 1120. But I also have hits in some Trio data, this time in tag 0029 1110. Not a single tag came up valid... My guess is if they don't start with `SV10`, they're not supposed to be interpreted as CSA - they're Siemens private, but probably stored using some other convention. I'd rather remove any attempt at loading CSA1...

I am surprised though that you'd say your collaborator had CSA data in 0029 1110 - that's one of those that shows up as _not_ CSA2 or valid CSA1. But then, these are Siemens private tags, and there's every chance that their interpretation might change without notice... Not sure what to do about it, other than allow interpreting these as CSA as long as they start with `SV10`. What do you reckon?

Lestropie:
> So the only thing I take from this is that either these tags are not supposed to contain CSA data, or your implementation is incorrect...

The only reference I have to go off is the second page linked to in the original PR post. So if none of your data that didn't contain the "SV10" flag obeyed that format, we probably shouldn't trust it. 

> ... they're Siemens private, but probably stored using some other convention. I'd rather remove any attempt at loading CSA1...

:+1:

> I am surprised though that you'd say your collaborator had CSA data in 0029 1110 - that's one of those that shows up as not CSA2 or valid CSA1.

Based on what reference / information? Just what you've seen in your own data? I'm only going off the first page linked in the original PR post, and the fact that my collaborator had valid CSA2 data there (which resulted in `master` not importing phase encoding information on `master`, whereas this branch did). Is there any documentation evidence that any of the new added "CSA" codes in this PR should explicitly *not* be read as CSA? Otherwise I agree, just:

> ... allow interpreting these as CSA as long as they start with `SV10`

jdtournier:
>> I am surprised though that you'd say your collaborator had CSA data in 0029 1110 - that's one of those that shows up as not CSA2 or valid CSA1.
>
> Based on what reference / information? 

Based on this line from your original post:

> Collaborator's DICOMs had CSA data stored at 0x0029U, 0x1110U, which MRtrix3 was ignoring.

?

jdtournier:
>> I am surprised though that you'd say your collaborator had CSA data in 0029 1110 - that's one of those that shows up as not CSA2 or valid CSA1.
>
> Based on what reference / information? Just what you've seen in your own data? 

Yep, that's all I've got to go on... Basically, the only occurrences of that tag in the data I have access to (from a Trio) don't show up as CSA2 (don't start with `SV10`), so your code tries to read them as CSA1, and the number of items (`cnum`) invariably turns out to be much larger than the 128 maximum - so they get ignored as invalid. 

> Otherwise I agree, just:
> 
>> ... allow interpreting these as CSA as long as they start with SV10

Good - I reckon that's the only sensible option given the above...

Lestropie:
OK, I reckon that should be good now. Worth running through your private test again just in case?

jdtournier:
OK, merged with updated `dev`, tested, and good to go.

PR #1193, opened 2017-11-17
DICOM: better handling of empty fields

jdtournier:
To fix [this issue](http://community.mrtrix.org/t/segmentation-fault/1321?u=jdtournier), mentioned on the forum.

### DIFF
--- a/core/file/dicom/element.cpp
+++ b/core/file/dicom/element.cpp
@@ -354,7 +354,23 @@ namespace MR {
       }
 
 
+      void Element::error_in_get (size_t idx) const
+      {
+        const std::string& name (tag_name());
+        DEBUG ("value not found for DICOM tag " + printf("%04X %04X ", group, element) + ( name.size() ? name.substr(2) : "unknown" ) + " (at index " + str(idx) + ")");
+      }
 
+      void Element::error_in_check_size (size_t min_size, size_t actual_size) const
+      {
+        const std::string& name (tag_name());
+        throw Exception ("not enough items in for DICOM tag " + printf("%04X %04X ", group, element) + ( name.size() ? name.substr(2) : "unknown" ) + " (expected " + str(min_size) + ", got " + str(actual_size) + ")");
+      }
+
+      void Element::report_unknown_tag_with_implicit_syntax () const
+      {
+        DEBUG (MR::printf ("attempt to read data of unknown value representation "
+              "in DICOM implicit syntax for tag (%02X %02X) - ignored", group, element));
+      }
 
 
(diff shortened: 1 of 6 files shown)


core/file/dicom/element.cpp
```diff
@@ -354,7 +354,23 @@ namespace MR {
...
+      void Element::error_in_check_size (size_t min_size, size_t actual_size) const
+      {
+        const std::string& name (tag_name());
+        throw Exception ("not enough items in for DICOM tag " + printf("%04X %04X ", group, element) + ( name.size() ? name.substr(2) : "unknown" ) + " (expected " + str(min_size) + ", got " + str(actual_size) + ")");
+      }
+
+      void Element::report_unknown_tag_with_implicit_syntax () const
+      {
+        DEBUG (MR::printf ("attempt to read data of unknown value representation "
+              "in DICOM implicit syntax for tag (%02X %02X) - ignored", group, element));
```
Lestropie: I can see it's a copy-and-paste of code from `core/file/dicom/element.h:138`; but should these not be `%04X`?

jdtournier: Yes, good point. Doesn't make much difference in practice (it won't truncate to 2 characters, just zero-pad to at least 2 characters), but best to make all this consistent. I'll amend now.

jdtournier:
One more thing I noticed while running my test battery on this: it seems our handling of Siemens DEC maps (colFA) was not great - these errored out in testing. Looking at the cause, this is because these images are seriously non-DICOM compliant... These are stored as mosaics, which is already a bit of a hack, adding RGB storage on top of their existing grayscale mosaic format, which was itself a hack. But they also don't provide the DICOM-standard `ImagePositionPatient` and `ImageOrientationPatient` entries, which meant the orientation _and_ the sorting of the slices (which relies on slice positions along the slice normal) was all messed up. 

However, it turns out that they store that same information within their own private CSA entry. So I've gone and added support for pulling that information out if it exists, and that seems to fix the issue... The Siemens colFA imported with the updated code match the corresponding FA maps, and all other images check out as expected. Required a couple of changes to the CSA handling code, but nothing major. 

Also, it turns out that the original bug here also affected some of the existing data in my tests - I struggle to understand how they didn't segfault before. Probably related to slight changes in the memory layout due to e.g. different compiler versions, meaning the offending reads now extend over the page boundary or something... Anyway, all fixed now :grin:

jdtournier:
OK, ran the tests again, all checks out - merging now.
I'll merge to `dev` in a minute.

PR #1196, opened 2017-11-21
DICOM: fix bug introduced in previous commits

jdtournier:
Just noticed a minor screw up - better fix it ASAP. All checks out on testing.


jdtournier:
:joy: you and me both...

Issue #1198, opened 2017-11-21
mrcalc issues warning "header transformations of input images do not match" when the input images (albeit identical) are not in the same file format


ansk:
Example:
```
mrconvert test.mif test.nii.gz
mrconvert: [100%] copying from "test.mif" to "test.nii.gz"
mrconvert: [100%] compressing image "test.nii.gz"
mrcalc test.mif test.nii.gz -add test3.mif
mrcalc: [100%] uncompressing image "test.nii.gz"
mrcalc: [WARNING] header transformations of input images do not match
mrcalc: [100%] computing: (test.mif + test.nii.gz)
```

jdtournier:
OK, that's unexpected... 
What does this report:
```
mrinfo test.mif test.nii.gz
```
?

jdtournier:
OK, never mind, it's most likely a rounding / tolerance issue. I hadn't realised that the [test for whether the transforms match](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrcalc.cpp#L587) uses the [default tolerance of zero](https://github.com/MRtrix3/mrtrix3/blob/master/core/image_helpers.h#L420)... I'll amend and put in a more sensible value like 1e-4.

jdtournier:
PR in #1199 - I'll close this for now. 

Please do checkout the `mrcalc_transform_match_tolerance` branch and see whether that fixes the issue. Note that it'll be based off the `dev` branch, not `master`, so that may introduce other differences as well...

PR #1199, opened 2017-11-21
mrcalc: increase tolerance on transform match test

jdtournier:
As discussed in #1198

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - export PATH=`pwd`/bin:${PATH}
   - export PYTHONPATH=`pwd`/lib:${PYTHONPATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
-  - (cd ..; hg clone https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
+  - (cd ..; hg clone --insecure https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
   - pip install urllib3[secure]
   - pip install --user pylint recommonmark sphinx sphinx-rtd-theme
   - pip3 install --user pylint recommonmark sphinx sphinx-rtd-theme
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
       - python3
       - python-pip
       - python3-pip
+      - ca-certificates
 compiler:
   - clang
 env:
--- a/cmd/mrcalc.cpp
+++ b/cmd/mrcalc.cpp
@@ -585,7 +585,7 @@ void get_header (const StackEntry& entry, Header& header)
   for (size_t n = 0; n < std::min<size_t> (header.ndim(), entry.image->ndim()); ++n) {
     if (header.size(n) > 1 && entry.image->size(n) > 1 && header.size(n) != entry.image->size(n))
       throw Exception ("dimensions of input images do not match - aborting");
-    if (!transforms_match (header, *(entry.image)) && !transform_mis_match_reported) {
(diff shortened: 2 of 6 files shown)


jdtournier:
One slight concern with this pull request: currently [transform_check()](https://github.com/MRtrix3/mrtrix3/blob/dev/core/image_helpers.h#L407-L416) tests for equivalence within tolerance in an absolute sense. This is fine for the elements of the rotation part of the transform, since _MRtrix3_ keep this bit unitary.  But I think the translation components should really be checked within _relative_ tolerance, since the relevant units may differ for e.g. small animal studies, or in different contexts. It's no big deal since we're unlikely to encounter data where the spatial scale varies by several orders of magnitude compared to what we typically come across, but nonetheless... I'm considering making the translation part of the tests relative to the FoV - thoughts?

maxpietsch:
You could apply both transformations to the [tetrahedron](https://github.com/MRtrix3/mrtrix3/blob/57e44453cf96a01e5d75c8edb577821daa5d3da9/src/registration/metric/params.h#L103 ) or the image [bounding box](https://github.com/MRtrix3/mrtrix3/blob/57e44453cf96a01e5d75c8edb577821daa5d3da9/core/math/average_space.cpp#L77) coordinates and check the deviations between both sets in units of voxel sizes.

jdtournier:
@maxpietsch: good call. I've just pushed a change (76c57f7) that checks for the difference in real space positions of 4 opposing corners of the bounding box. See what you think...

maxpietsch:
I've changed the test using the same coordinates but different tolerance criteria. Full FOV coordinates are compared to smallest voxel size. I added a test for voxel size as well. The latter is violated by `mrconvert` for conversion from mif to nii if I use a tolerance of 1e-6, fine with 1e-4.

jdtournier:
OK, after a quick chat with @maxpietsch, I reckon a rethink is in order. At it stands, given the way our transforms are defined (i.e. image → scanner space), the test is not as practically useful as it seems. The main reason to invoke this check is to verify that it is safe to perform voxel-wise operations on two images, which implies that we are interested in verifying that voxels in the first image have a one-to-one correspondence with the voxels in the other. But this is not actually what this test does - or what its name suggests it does. 

So I reckon we should rename the test to something like `voxel_grids_match_in_scanner_space()` (a bit of a mouthful, I'll admit - but hopefully expressive :wink:), and in that test we'd check that:

- the dimensions for the first 3 (spatial) axes match;
- the projection of the bounding box corners into scanner space match within the specified tolerance, _relative_ to the smallest voxel dimension (note that this inherently ensures that the voxel sizes match). 

Wadayall reckon?

maxpietsch:
I went ahead and made the changes @jdtournier suggested above. Feel free to revert or amend! I noticed that we check the image spacings without tolerance so I changed `spacings_match` to compare up to a specified tolerance in units of voxel dimensions but left the tolerance at zero for back compatibility. I'd suggest relaxing the voxel spacing test in `core/image_diff.h` to a tolerance of 1e-6.

maxpietsch:
The last two commits change the nifti voxel size handling (reading) in that the header voxel spacing is taken from the pixdim header entry and not overwritten by the s-form-derived spacing.

This causes mrcalc to issue a warning about header transform mismatch if nifti files written before and after this change are compared or when a mismatching version of mrcalc is used. The difference in the read transformations is in the 7th digit.

(Travis fails due to a certificate error with bitbucket but tests complete sucessfully on my machine.)

jdtournier:
I'm happy with these changes. Any other reactions before we merge?

jdtournier:
No complaints - I'll merge now...

PR #1200, opened 2017-11-22
update docs lagging behind

thijsdhollander:
Docs generation was forgotten somewhere somehow.  Now docs were lagging behind... can't have that, it drives me nuts :grin: .  I suppose merging this to `dev` isn't too controversial, so I'll do that now, before every other branch off `dev` has to do this.


thijsdhollander:
Ok, that took quite a while.  Had to restart each test job once at some point. :unamused:

PR #1201, opened 2017-11-29
population_template: python 3 & no numpy fix

maxpietsch:
I can't remember why I had the python version dependent write mode in there...
Fixes http://community.mrtrix.org/t/error-when-run-the-dwiintensitynorm/1340

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - export NUMBER_OF_PROCESSORS=4
   - export PATH=`pwd`/release/bin:`pwd`/scripts:${PATH}
   - export EIGEN_CFLAGS=-I`pwd`/../eigen
-  - (cd ..; hg clone https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
+  - (cd ..; hg clone --insecure https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3)
 script:
   - ./check_memalign && $py ./configure -assert && $py ./build -nowarnings && ./run_tests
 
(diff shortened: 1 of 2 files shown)


jdtournier:
So I'm not too sure what the issue was exactly, but I would generally encourage using binary mode for any file IO, since otherwise the line-endings will differ depending the platform used, which can lead to other errors when trying to load the same information on a different system. We had similar issues in the past with the locale settings using commas as the decimal point on German systems for example. 

In your particular case, it seems to me that the solution would be consistently use a binary more and the string encode()/decode() methods (as used in the [build script](https://github.com/MRtrix3/mrtrix3/blob/master/build#L192), for example). Note that these work fine in both Python 2 & 3, so there'd be no need for that explicit check you had...

maxpietsch:
Good point! 

The line endings are no issue within python though as specified in the [docs](https://docs.python.org/3.3/tutorial/inputoutput.html#reading-and-writing-files):

> In text mode, the default when reading is to convert platform-specific line endings (\n on Unix, \r\n on Windows) to just \n. When writing in text mode, the default is to convert occurrences of \n back to platform-specific line endings. 

Our code seems to rely on '\n' as a line separator (function `split_lines`) so, yes, I'll change it.

jdtournier:
OK, I notice you reverted your changes on the write front. I think you should always read _and_ write in binary mode, and use the decode() & encode() methods respectively. As it stands, the read is in binary mode, but not the write - I think your output should be predictable and stable, regardless of platform or locale. Personally, I suspect DOS line endings are going to die eventually... (not so sure about locale variations though)

maxpietsch:
Now read and write are in binary, regardless of python version. I've also fixed a small bug regarding too few input images.

(Travis failed due to the certificate error.)

jdtournier:
Cool, thanks. 

Fix for TravisCI in #1199 - just cherry-pick e31d0a0...

Issue #1204, opened 2017-11-30
fod2fixel -peaks overestimates certain peak amplitudes


thijsdhollander:
First discovered on `master`, but also confirmed on `dev`.  All the below is using `dev`, since a few added features to `fod2fixel` in `dev` make this easier to illustrate.  **I marked this as "critical" since it directly impacts the behaviour of `dwi2response tournier`.**  I actually discovered this while playing around with some things related to `dwi2response tournier`.  I'll try to keep the description of the bug as short as possible, but I invite everyone to confirm that I'm not getting things wrong here.

Ok, so, as stated, I simplified things here; but this kind of problem happens in any possible setting (with(out) different options compared to this example) when using the `-peaks` option to `fod2fixel`.

So, simplified goal of this example here: just get the amplitude of the largest peak of all FODs.  First up, doing this with `sh2peaks -num 1` and `peaks2amp` works without issues.  These are the FODs I worked on, combined with those largest peaks using the fixel visualisation tool:

![screen0101](https://user-images.githubusercontent.com/11952694/33417378-5999c92a-d5f5-11e7-825f-13b8d2bce145.png)

Scaling is set the same for both, which causes the fixels/peaks to appear just slightly larger than the FOD peaks (not a problem with the actual data of either FODs or peaks, this is just what the viewer does with those scales in practice).  The hot colour scale for the peaks shows (in addition to their visual length) also encodes the peak amplitude.  I'll get to the grey scale voxels further on.

Next, trying to do exactly the same with `fod2fixel`.  I used `fod2fixel` with the `-peak` option output, as well as the `-fmls_no_thresholds`, `-maxnum 1` and `-dirpeak` options; but the essential problem/bug happens just as well without those latter 3 options.  Showing the fixels/peaks again combined with the same FODs as before (length as well as hot colour encoding the `-peak` output fixel datafile):

![screen0102](https://user-images.githubusercontent.com/11952694/33417563-22995962-d5f6-11e7-85f7-2ab947f7ec0f.png)

[...]

PR #1205, opened 2017-11-30
SH peak finder: fix behaviour at pole for z-aligned delta

jdtournier:
Briefly mentioned in #1204 - seems `sh2peaks` fails for peaks that are perfectly aligned with the z-axis. Seems to be a simple divide-by-zero issue - the check introduced here seems to fix the problem.

In practice, this is unlikely to be a problem, given how low the chances are of ever encountering a peak perfectly aligned with the z-axis. There is no bug report for it, but it does fix a bug, so I'm wondering whether this should go on `dev`, rather than directly on `master`... Happy either way, opinions welcome.

### DIFF
--- a/core/math/SH.h
+++ b/core/math/SH.h
@@ -19,6 +19,7 @@
 # warning using non-orthonormal SH basis
 #endif
 
+#include "debug.h"
 #include "math/legendre.h"
 #include "math/least_squares.h"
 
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Yeah, this is so minor and fixes a genuine technical software *bug* (not just behaviour of a method or something): these kinds of things make perfect sense to be fixed directly on `master`, IMHO.

jdtournier:
Ah yes, the certificate failure... I've just merged #1201 which includes a fix for that. I'll update, hopefully the tests will work fine then.

PR #1209, opened 2017-12-04
remove support for the non-orthonormal SH basis

jdtournier:
As discussed in #1206 - opinions welcome.


dchristiaens:
Looks complete to me. OK if it passes the tests.

jdtournier:
Ok, seems everyone is happy with this - merging now.

Issue #1147, opened 2017-09-27
Interface changes before release


thijsdhollander:
As has come up in a few other issues and/or pull requests: we'd probably better check that we're as happy as can be with all command and option names by release.  Not to rule out _any_ changes thereafter of course, but it would obviously be better if we spot obvious candidates for name changes before release.  I've created this issue to list a few, and would encourage anyone to add to the list if you come across or think of one otherwise.  This includes commands which are deprecated, and may be considered to be removed before release.  It's all about getting our interface cleaned up essentially, which is also a good thing with subsequent educational activities (e.g. workshops) in mind.

So, here we go:

- `tcknormalise` :arrow_forward: `tcktransform` : is inherently happening as we speak (#810)
- `mesh2pve` :arrow_forward: `mesh2voxel` : was brought up in #1131, [this post](https://github.com/MRtrix3/mrtrix3/pull/1131#issuecomment-331074380)
- `-shell` :arrow_forward: `-bvalues` : some ongoing discussion in #1049, [this post and beyond](https://github.com/MRtrix3/mrtrix3/pull/1049#issuecomment-330456279).  Includes impact on `-shells` and `-shellcounts` of `mrinfo`.
- `-stride` :arrow_forward: `-strides` : brought up in context of the previous point.
- old `mtbin` is now deprecated and already replaced by `mtnormalise`.  We should remove `mtbin` at some stage.
- old `dwi2noise` is now deprecated and already replaced by `dwidenoise` with the `-noise` option.  We should remove `dwi2noise` at some stage.

While fiddling around with #1146, I also noticed that [the config file options](https://github.com/MRtrix3/mrtrix3/blob/master/docs/reference/config_file_options.rst) are quite inconsistent in their "formatting": most just use MixedCaseFormatting but some also use a Category.DotFormat and all registration related ones are lower_case_with_underscores.  This clearly drives any user suffering from mild (or worse) OCD completely crazy :dizzy_face:, so maybe we should go for consistency there too. :relieved: 

That's all I could think of at the moment, but don't hesitate to add to this.

Lestropie:
Also question regarding naming of `dwipreproc`; comment [here](https://github.com/MRtrix3/mrtrix3/pull/1047#issuecomment-315278398) in #1047 and re-raised in #1161.

thijsdhollander:
A lot of the above was done in #1211.  What remains or is not done as above is:

[...]

PR #1211, opened 2017-12-05
Interface name changes

thijsdhollander:
Starting to implement some name changes on the TODO list (#1147).  I started with the least controversial one (`-stride` --> `-strides`).  It would be good to get some of this done if we prefer to do a `dev` to `master` release soon, for the sake of getting #1206 out.

I marked this as "help / test wanted" because some of this requires widespread scattered changes, in code as well as scripts and documentation.  It's very easy to overlook some bits here... so if anyone finds time: always welcome to help out, and even just check that nothing gets forgotten.


thijsdhollander:
Implemented the `mesh2pve` to `mesh2voxel` name change as well.  Part of the #1147 list, and mentioned [here](https://github.com/MRtrix3/mrtrix3/pull/1131#issuecomment-331074380).

thijsdhollander:
Asking while we're at it here: are we ready to lose (deprecated commands) `mtbin` and `dwi2noise`?  Since we're in a spirit of retiring stuff (e.g. non-orthonormal SH basis), I thought I could just as well bring it up. :slightly_smiling_face:

thijsdhollander:
Looking up the discussion about `-shell(s)` --> `-bvalues` again ([this post](https://github.com/MRtrix3/mrtrix3/pull/1049#issuecomment-330456279) and beyond), I note/hope we were comfortable enough with it to proceed...

> I like @thijsdhollander's suggestion of changing to -bvalues though... I agree it is less prone to misinterpretation. If we do follow through, I'd also suggest the equivalent query options to mrinfo should also change accordingly.

> Happy to go with the consensus though.

Most important arguments / points to reflect on were covered in some back-and-forth discussion of those points.  Remaining question here would be what the corresponding `mrinfo` options (currently on `dev`: `-shellvalues` and `-shellcounts`) would become in that scenario.

`-shellvalues` could trivially be changed to `-bvalues` as well.  Even though the same option name as the option to select the b-values for commands that support it, this shouldn't be an issue in practice (and actually makes it more intuitive, since the kind of information this option sets or gets is the same).  Just having looked into `-strides`, I note we've got the same scenario going on there: sets the strides for commands that support it, gets the strides via `mrinfo`.

`-shellcounts` could be changed to `-bvaluecounts`.  This sounds very awkward, but then again, if you forget for a moment how we ever got used to `-shellcounts`, the existing `-shellcounts` is in principle just as awkward.  Notably, there's a risk that it sounds like "the count of what comes before" (shells or b-values), while it actually refers to the count of images / gradient directions for each b-value.  So we can either embrace the awkwardness of `-...counts`, or think of something better.  Trying to think of something better, everything that comes to mind is excessively long and becomes essentially the existing description of this option with underscores to separate the words... :grimacing:  Open to ideas...

thijsdhollander:
And while we're trying to move this all forward anyway, a sensitive one: the renaming of `dwipreproc`, also reminded of again in this post of #1147, with a few references to the moments in our glorious history where this came up.

I recall @Lestropie and I being mentally ready to switch to something along the lines of `dwiundistort` (referring to all forms of distortion, including spatio-temporal ones such as motion).  However, I also recall @jdtournier not being so keen on such a change. :grin:  I'd personally still argue the "preproc" suffix is already, and may even only become more, inappropriate due to it not covering all preprocessing (far from even).  It's also not the *first* step anymore (denoising and unringing proceeding it).

So well, how are we feeling about this one these days...? :grimacing:

jdtournier:
OK, a few thoughts on this. These changes included so far are innocuous and entirely backwards-compatible, so I've no objections to those. This includes the `-stride` ⇨ `-strides` change, renaming `mesh2pve` ⇨ `mesh2voxel` (since it's rarely invoked by users directly), and deprecating `mtbin` and `dwi2noise` (although I still reckon there might be value in that last one, in spite of the superior properties of `dwidenoise`, since `dwi2noise` reports on all sources of noise, Gaussian or otherwise, but given that nothing currently relies on it, I'm happy to resurrect it later if need be). 

But the ones I'm not convinced by are the suggested `-shell` ⇨ `-bvalues` switch and related changes to the relevant options in `mrinfo`, and the `dwipreproc` renaming proposal. We've already discussed this a fair bit, but I'll just reiterate my opinion on the matter here.

My main concern is minimising disruption for users. I would prefer to keep such non-essential interface-changing modifications for non-maintenance releases, if not major releases. Here, I'm not convinced the case for the changes is all that compelling, and certainly not enough to warrant inconveniencing users. So I would rather stick to the status quo (or at least limit this to minor non-disruptive changes)... More specifically:

- `-shell` ⇨ `-bvalues`: I appreciate the confusion that the concept of a b=0 'shell' might induce, but I don't think it's anything that can't be explained and documented away. Given the way the shell selection actually works, which revolves entirely around the concept of selecting a 'shell', I think there's a good argument for sticking with the current option name (although I'd be happy to rename it to `-shells`, since that's a non-breaking change). I think some of the confusion can be minimised by changing the documentation for that option to show `-shells bvalues`, rather than `-shell list` as it is currently. I'm not convinced that `-bvalue` option is necessarily as clear in terms of its actual intent, which is to select a _shell_ around that b-value (it doesn't really convey any the clustering that happens behinds the scenes to make that happen seamlessly). Personally I reckon I prefer `-shells` for that purpose... (I admit I may have been brainwashed by years of having used that very option, but still...). 

- renaming `dwipreproc`: again, I appreciate the concern that `preproc` might suggest that this is all the preprocessing that needs to happen. One option we'd discussed is to expand that script to also perform denoising and Gibbs unringing, but there is some reluctance from Rob to do this (understandably given the amount of effort that script requires to maintain...). Besides, I think there is a good case to be made for keeping individual steps separate, and leaving it up to users to get to grips with how things work - it'll be to their benefit down the track to understand what actually happened. There's nothing stopping us or others from wrapping all of these different pieces into distributable pipelines (and maybe we could encourage this), but I reckon we shouldn't get into the business of producing 'black boxes' - they invariably require a lot of maintenance to ensure they work seamlessly (as Rob well knows). But back to `dwipreproc`: the other issue is that I'm not overly enthused about any of the alternative names proposed to far. `dwiundistort` really sells it short, especially since I would say that its primary purpose is motion correction, with optional distortion correction (primarily susceptibility-induced distortions, since eddy-current distortions are actually fairly minimal on modern systems, as far as I can tell?). As noted by @thijsdhollander, any name would need to be ridiculously verbose to do it justice... So from that point of view, `dwipreproc` doesn't seem such a bad choice. As long as its purpose is clearly documented in the synopsis and description, I reckon that would be enough for users to understand what it does. And we can add suggestions in the description encouraging the use of `dwidenoise` and `mrdegibbs`, if we haven't already?

So that's my 2 cents... Happy to discuss, obviously.

thijsdhollander:
> but I'll just reiterate my opinion on the matter here

:confused: Slightly confused here.  Last @jdtournier certified opinion I remember on the `-bvalues` proposal was:

> I like @thijsdhollander's suggestion of changing to -bvalues though... I agree it is less prone to misinterpretation. If we do follow through, I'd also suggest the equivalent query options to mrinfo should also change accordingly.

...right here: https://github.com/MRtrix3/mrtrix3/pull/1049#issuecomment-330465102

I wouldn't have kept discussing it to the length I did in the posts below that one, if it wasn't a positive one... :confounded: 

When to make such a change, which is indeed more major, is another matter indeed.  We did quite drastically change the `tckgen` interface with the last release candidate.  I reckon that one was quite a bit more high profile that this would be.  Many people don't need/use the `-shell(s)` option, and those who do, are typically already paying a bit more attention than others.

> renaming dwipreproc: again, I appreciate the concern that preproc might suggest that this is all the preprocessing that needs to happen. One option we'd discussed is to expand that script to also perform denoising and Gibbs unringing, but there is some reluctance from Rob to do this (understandably given the amount of effort that script requires to maintain...).

No, I'm against that too.  So that makes all of us.  Easy.  I simply find the `dwipreproc` name misleading, but we shouldn't change the contents to suit the packaging.

> dwiundistort really sells it short, especially since I would say that its primary purpose is motion correction, with optional distortion correction (primarily susceptibility-induced distortions, since eddy-current distortions are actually fairly minimal on modern systems, as far as I can tell?).

Fair point indeed.  However, in that respect, `dwipreproc` is over- or mis-selling it too.  It's indeed challenging to come up with a proper descriptive name for this one.  Hence I don't feel overly strong about the name change... but it's still worth considering.  We're planning on releasing soon (right?), so if we do think of more drastic changes like this, I'd argue now is the time to go for it.

> As noted by @thijsdhollander, any name would need to be ridiculously verbose to do it justice... 

You really shouldn't be reading github ramblings this late at night... :stuck_out_tongue_winking_eye:  That comment was about the name of another `mrinfo` option, not about `dwipreproc` alternatives...

Lestropie:
> `-shellcounts` could be changed to `-bvaluecounts`

If we were to go that way, I think it'd need be something like `-volumes_per_bvalue`. But I think I remain in favour of `-shells`, with some modification of the help description of that option to indicate that *b*=0 may be considered as a 'shell' when using that option. Some extra text could additionally be added to the [diffusion gradient scheme handling](http://mrtrix.readthedocs.io/en/latest/concepts/dw_scheme.html) documentation page?

> One option we'd discussed is to expand that script to also perform denoising and Gibbs unringing, but there is some reluctance from Rob to do this (understandably given the amount of effort that script requires to maintain...).

Reluctance is not just from a complexity standpoint. In #1190 I added options to extract `eddy`'s outputs regarding outlier detection / motion. One thing we're looking into is identifying volumes that are too severely corrupted by motion / dropout for even `eddy`'s slice-to-volume correction to fix, removing them, then re-running `eddy` so those corrupt volumes don't contribute to the alignment / outlier detection of other volumes / slices. But rather than adding this complexity to `dwipreproc` (and me becoming responsible for the heuristics employed), `dwipreproc` can instead be called within a wrapper script; in which case you don't want / need denoising and unringing to be part of that loop. Sure you could have command-line options to skip those steps, but it just becomes unnecessary encapsulation.

Also if `dwipreproc` included those steps, ideally there would also be heuristics available to detect from the data when those steps should *not* be used, since they're not applicable to all data. Having them exposed makes the user responsible for that decision.

thijsdhollander:
> and deprecating mtbin and dwi2noise (although I still reckon there might be value in that last one, in spite of the superior properties of dwidenoise, since dwi2noise reports on all sources of noise, Gaussian or otherwise, but given that nothing currently relies on it, I'm happy to resurrect it later if need be).

So this is done now as well.  I've done each in a separate commit, for verbosity and to facilitate potential future resurrection of dwi2noise.

jdtournier:
> :confused: Slightly confused here. Last @jdtournier certified opinion I remember on the -bvalues proposal was:
>
>> I like @thijsdhollander's suggestion of changing to -bvalues though... I agree it is less prone to misinterpretation. If we do follow through, I'd also suggest the equivalent query options to mrinfo should also change accordingly.
>
> ...right here: #1049 (comment)

Darn... Seems I changed my mind. And all on the public record too - no plausible deniability... :flushed: 

So, yes, I'm in two minds about this. Like I said, I can see the rationale to avoid confusion regarding the _b_=0 shell, but having let the idea sink in for a while, I'm not sure either way. I think `-shells` also has its merits, despite the 'oddness' of a _b_=0 shell. Given my ambivalence, the overriding concern for me is to avoid needless disruption, hence the fickleness of my opinion... 

> When to make such a change, which is indeed more major, is another matter indeed. We did quite drastically change the tckgen interface with the last release candidate. I reckon that one was quite a bit more high profile that this would be. Many people don't need/use the -shell(s) option, and those who do, are typically already paying a bit more attention than others.

Fair point - this would be a bigger concern following actual release. You're right that the main issue at this stage should be to make sure that the interface is as unambiguous as we can make it, even if that might cause some disruption (admittedly not a huge amount, you're right). So let's ignore that consideration for now, and focus on whether `-bvalues` is clearer than `-shells`. Like I said, my opinion on this is not particularly strong one way or the other :man_shrugging:, so on that basis I'd opt to minimise disruption. And like I said, having thought about this some more, I reckon I actually find `-shells` a closer match to what the option actually does... 

> You really shouldn't be reading github ramblings this late at night... ￼ :stuck_out_tongue_winking_eye: 

Too true. See above. However:

> That comment was about the name of another `mrinfo` option, not about `dwipreproc` alternatives...

I reckon it applies equally to `dwipreproc` alternatives... On that note:

> However, in that respect, `dwipreproc` is over- or mis-selling it too.

Yes, I appreciate that. But it remains sufficiently vague that it's still an accurate reflection of its purpose. The only issue is that it gives the impression that it does _all_ the necessary preprocessing. If someone comes up with a decent acronym, I'd be more than willing to consider it. One option is to actually reflect more closely what's under the hood, and rename it to something like `dwieddy` or something like that... Main issue here is that this would obviously tie the command's purpose very directly to its implementation - but then as things stand, that is exactly what it does... In any case, open to ideas on that front. 

>> `-shellcounts` could be changed to `-bvaluecounts`
>
> If we were to go that way, I think it'd need be something like `-volumes_per_bvalue`. But I think I remain in favour of `-shells`, with some modification of the help description of that option to indicate that b=0 may be considered as a 'shell' when using that option. Some extra text could additionally be added to the diffusion gradient scheme handling documentation page?

I was also thinking along similar lines (things like `-ndw_dirs` or something). But how about `-shellsizes`? I think that would make sense: a shell at _b_ = 3000 s/mm² of size 60 is pretty unambiguous, I reckon?

> Reluctance is not just from a complexity standpoint.

Understood. It's not just that it's already complex - you're actually trying to make it even more complicated... :wink:

thijsdhollander:
> even if that might cause some disruption

> so let's _ignore_ that consideration for now

> so on that basis I'd opt to minimise disruption

:confounded: 

Ok, I'm not getting anywhere with this discussion, it seems. :roll_eyes:   I'll leave it at that, but my opinion remains strongly in favour of aligning with the terminology that's already widely adopted in the dMRI, where a shell (and the number of such shells present in data or required for certain models) is defined strictly differently.  Using the term to (try to) reveal the underlying _implementation_ that happens to cluster b-values is in many ways a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction).

For my own applications, I've just come up with a solution that will then simply use the existing `-shells` option differently (than some of the other apps) so as to align with the more common definition of shells. :man_shrugging:   I'm really aiming to only develop apps where the user interface works for the application and the user, rather than the other way around.

> The only issue is that it gives the impression that it does all the necessary preprocessing.

Well, that, and it doesn't even mention that motion and/or distortion correction is included.

> But it remains sufficiently vague that it's still an accurate reflection of its purpose.

That's not how accuracy is defined in my textbook. :wink:  It doesn't convey anything on what is either guaranteed to be included, or excluded.  So allows for both false positive as well as false negative interpretation of the set of included preprocessing steps.  Both of those punish accuracy, by definition of accuracy. :slightly_smiling_face: 

> One option is to actually reflect more closely what's under the hood,

> dwiundistort really sells it short, especially since I would say that its primary purpose is motion correction,

If we were to agree that its primary purpose is motion correction, then `dwicorrectmotion` could work.  Before anyone argues that's too long: "`fixelcorrespondence`". :wink: 

> But how about -shellsizes? 

Definitely beats `-shellcounts`.  And `-shellvalues` could at least become `-shellbvalues`.  Because, to _any_ dMRI scientist, what even is a "shell-value"..?

thijsdhollander:
`tcknormalise` --> `tcktransform` implemented, in line with the consensus opinion at #810.  I know this was eventually also going to come along via (a number of) different branch(es), but figured it's not a bad idea to already have it in here; just in case those other features aren't ready for inclusion in a release that may now happen sooner than anticipated.  And since there will be a few scattered name changes already... well, you get the picture.  There's also no other code/scripts that depend on this one, so pretty safe.

thijsdhollander:
While we're powering through some of this stuff anyway, last discussion from #1147 to open up: [the config file option names](https://github.com/MRtrix3/mrtrix3/blob/master/docs/reference/config_file_options.rst).

> I also noticed that the config file options are quite inconsistent in their "formatting": most just use MixedCaseFormatting but some also use a Category.DotFormat and all registration related ones are lower_case_with_underscores.

I reckon fixing this up to obtain consistency would then mostly impact the registration related ones.  Any opinions? @maxpietsch, this may concern you in particular...?

thijsdhollander:
Ok, I updated the state of things in the #1147 issue, for future reference.  I'm ok with this pull request for now, so we can move on with getting this into `dev`, wrapping `dev` up, and getting the SH fix in there to be included in a release (candidate).

This pull request has been sitting here for a while now, with opportunities to comment... I suppose people are fine with where it went, or don't feel too strongly about it.  I'll merge this to `dev` during the course of this week at some point.  Speak up if you feel you need to!

Final thing here: the development in this pull request required modification of some tests (all just naming stuff).  I nicely made a separate branch for that purpose in the test data repository.  However, this is the first time I'm facing this together with gitflow on top of that; since this will first go to `dev`, not `master`.  Do we need a `dev` branch on the test data repository too to keep things clean?

maxpietsch:
> I reckon fixing this up to obtain consistency would then mostly impact the registration related ones. Any opinions? @maxpietsch, this may concern you in particular...?
Not too concerned about renaming them.

jdtournier:
OK, happy with everything in there, just a few issues outstanding:

#### config option renaming

What convention should be opt for? I've been going with a CapitaliseEachWord convention, and the overwhelming majority of options follow that convention. There's 4 `NIfTI.OptionX` ones with a dot in there, we can easily take out that dot and make them consistent. There's also a few other options added by @maxpietsch that he's happy to have renamed - @maxpietsch, do you want to handle that? Otherwise, I'm happy to do it. 

#### `dwipreproc` rename

I'd argued that `preproc` is not misleading as it encompasses what it does, but yes, it is still ambiguous as it does not state precisely what it is that it does do... My main motivation was that in the absence of a good enough alternative, the case to change it is not overwhelming. There is the recent suggestion of `dwicorrectmotion` (or `dwimotioncorrect`?), which I like better than `dwiundistort`, but it's still not perfect. I'm happy enough with this change if there is a consensus on this - it is admittedly clearer than `dwipreproc`, and the distortion correction is admittedly optional (at least the more dominant EPI distortion correction). Please cast your votes...

#### `-shell` option rename

This somehow feels like it's become an emotionally charged issue, judging by the emoji count... Lots of comments have been made, some of which I feel compelled to respond to. So here goes:

First off, yes, I had originally expressed support for the idea, and have since changed my mind. I'd like to think that this was at least tolerated, if not actively encouraged - we wouldn't be in research if we didn't allow our thinking to evolve. I appreciate this means that sometimes our efforts will be in vain (I've got countless examples of such). All I can do is apologise for the inconvenience - but the fact remains that I've changed my mind.

>> even if that might cause some disruption
>
>> so let's ignore that consideration for now
>
>> so on that basis I'd opt to minimise disruption
>
> 😖

Selective quoting is something the British tabloids engage in liberally, invariably with the aim of discrediting the source. It _really_ doesn't help the conversation. 

What I meant is that if we need to make disruptive changes to improve the API, by all means let's do that. But if the improvement is not all that convincing, I'd opt to minimise disruption and stick with the status quo. Is that such an odd position to take?

> my opinion remains strongly in favour of aligning with the terminology that's already widely adopted in the dMRI, where a shell (and the number of such shells present in data or required for certain models) is defined strictly differently. 

I'm not sure I understand where our definition of a shell differs from the rest of the field. Are we referring purely to the b=0 shell concept? If so, yes it can be seen to differ, but really I find that just a more expedient way to talk about those volumes that happen to have b=0 in the same sentence as those that don't. I think we can all agree that it arise naturally as _b_ tends to zero. If we don't refer to that as a shell, then what is the correct term? The _b_=0 set? The _b_=0 volumes? From a different point of view, if we were to use `-bvalue` as an option, what would the documentation state? Something like:

- **`-bvalue list`**
  select the shells centred on the _b_-values listed in the comma-separated list provided. In the case of _b_=0, select the _b_=0 volumes. 

I could probably phrase this better, but the issue remains: we'd need to make an exception in the documentation for the _b_=0 volumes if we don't allow them to come under the label of a shell - and at this point, we do need to refer to this being a shell selection process. So either way there's a disconnect. I personally find it more elegant and natural to refer to the set of _b_=0 volumes as a shell like any other, and clarify this in the docs and help page. 

Note that this issue is not restricted to our software - I struggle with this in papers too. How do you refer to a multi-shell sequence? For example: we have a 3-shell sequence with 4 b-values at 0, 400, 1000 & 2600 s/mm², with 20, 40, 60 & 100 directions per shell respectively. Note already the ambiguity / slight of hand in the number of b=0 volumes, on top of the disconnect between the number of shells and the number of b-values. An alternate way to write this is: we have a 3-shell sequence with b-values at  400, 1000 & 2600 s/mm², with 40, 60 & 100 directions per shell respectively, along with 20 b=0 volumes. So again, we have to introduce a distinction between _b_=0 volumes and the remaining shell, which I would argue is historical and entirely artificial, and it leads to awkwardness when talking about these things. I think we should lead from the front here, and allow the b=0 volumes to reclaim their rightful and long-lost shell status.

Note also that I think the discussion around this option is distinct from (although strongly related to) how different techniques might be described. Your 2S3T-CSD / SS3T-CSD method is a case in point. For instance, technically you can't do DTI with a single b-value, but you can do it with single-shell data (even though that term doens't make it clear that _b_=0 data are also required). Oddly enough, you can do DTI with 2+ shell data and no _b_=0 data... So here is an instance where actually, I'd argue that the field has been using the wrong terminology: DTI strictly requires at least 2 shell data (as discussed by [Derek Jones back in 1999](https://www.ncbi.nlm.nih.gov/pubmed/10467296)), even if one of those shells can have a b-value of zero, and normally does since this is the most optimal setting SNR-wise. The original CSD method is probably one of the only truly single-shell techniques (along with QBI). So to cut a long story short: I think there's a lot of ambiguity in the terminology used in the field anyway, but I don't think that would necessarily muddy the waters as far as the `-shell` option is concerned. The only type of circumstance where I might see confusion arising is in cases where people select a specific shell to use in a 'single-shell' technique that actually requires a b=0 shell, e.g.:
```
dwiextract dwi.mif -shell 1000 - | dwi2tensor - ...
```
when really they need to do:
```
dwiextract dwi.mif -shell 0,1000 - | dwi2tensor - ...
```
But I don't see the confusion being particularly long-lasting here. 

> Using the term to (try to) reveal the underlying implementation that happens to cluster b-values is in many ways a leaky abstraction.

OK, you've kind of lost me. If the selection process inherently requires some form of tolerance around the stated b-value, this is not some detail of the implementation, it's very much part of the upfront requirements. There's many ways to implement that, but that functionality would be demonstrably useless without some process to group similar-valued volumes together. I personally like the fact that the option name currently suggests that it selects whole shells rather than matching the exact b-value stated. 

> For my own applications, I've just come up with a solution that will then simply use the existing -shells option differently (than some of the other apps) so as to align with the more common definition of shells. 

Can you clarify what you mean here? If your definition of a shell differs from that implied in other apps, this is going to cause yet more confusion... You mentioned the issue around the naming of your original 2S3T-CSD algorithm because users were confused as to whether their single-shell data would be suitable given its stated requirement for 2 shells, and I understand that. But you fixed that by relabelling the method as SS3T-CSD, which no longer states that it requires 2 shell data. Why would you then need to change the handling of the `-shells` option? Under what circumstances does a problem arise? Can you provide an example of a command-line where this is a genuine source for confusion?

> I'm really aiming to only develop apps where the user interface works for the application and the user, rather than the other way around.

Funny that. Up until that point, that's what I thought I was aiming for too.

dchristiaens:
My two pennies worth:

**-shell option rename:** I personally don't see a need for renaming this, and actually prefer the current terminology. If people feel that there is some confusion currently, I think @jdtournier's suggestion of updating the docs to `-shell bvalues` is a brilliant compromise.

**dwipreproc rename:** Again, I don't feel the need, but I can see the logic for this one. However, if we decide to change the name, I would suggest/prefer the name to reflect the FSL tools being called (`dwieddyproc` for example), since it would be the most accurate description of what the script currently does. I am also hesitant about `dwimotioncorrect` since that's the name I'm currently using in my development code, and which can hopefully be released in the not too distant future. Whilst that command can of course also be renamed to something else if need be, I do anticipate a need for separate commands (rather than algorithm options within a single script) since the interface will be somewhat different.

Lestropie:
> config option renaming

Happy with PascalCase.

> dwipreproc rename

`dwi_motion_correct` is also the name of Dave's old script for doing motion correction only based on per-volume DWI masks, as shown in the 2012 AFD paper.

I don't see this as any better or worse than `dwipreproc` given it omits eddy current distortion correction, in addition to the (optional) EPI distortion correction. I don't think we can justify changing this unless there's a clearly superior alternative. Can maybe adjust the synopsis of the script a little?

> -shell option rename

Vote for sticking with `-shells`, with added clarification in the option help that *b*=0 counts as a 'shell' in *MRtrix3*.

Happy with `mrinfo`'s change to `-shell_bvalues` and `-shell_sizes`. We could also theoretically have a `-shell_volumes` that would provide the indices of those volumes assigned to each shell.

> For my own applications, I've just come up with a solution that will then simply use the existing `-shells` option differently (than some of the other apps) so as to align with the more common definition of shells.

This might be worth elaborating on if it potentially alters the outcome of the discussion. I'd have thought your changes to `select_shells()` & `dwiextract` would have sufficed, but it sounds like you may have encountered further logical gymnastics?

thijsdhollander:
>>I reckon fixing this up to obtain consistency would then mostly impact the registration related ones. Any opinions? @maxpietsch, this may concern you in particular...?

>Not too concerned about renaming them!

> What convention should be opt for? I've been going with a CapitaliseEachWord convention, and the overwhelming majority of options follow that convention. There's 4 NIfTI.OptionX ones with a dot in there, we can easily take out that dot and make them consistent. There's also a few other options added by @maxpietsch that he's happy to have renamed - @maxpietsch, do you want to handle that? Otherwise, I'm happy to do it.

Too late, already done!

thijsdhollander:
Updated the copyrights for 2018 while at it.  Makes things easier to get this through at the same time.  I suppose the upcoming RC will be around the new year's start anyway, right?  Just trying to be efficient, because I've still got lots else on my plate before the holidays.

FYI, I'll be gone next week Friday, and not-working for 2 weeks.

thijsdhollander:
> dwipreproc rename

> it omits eddy current distortion correction, in addition to the (optional) EPI distortion correction.

Final proposal: `dwimotiondistortion`.  Still not longer than `fixelcorrespondence` by the way.

I'm not spending more time on this though (happy to even "vote" to remain with `dwipreproc` just to move on already); my main argument remains that a command's name should reflect the task it performs, on a high level.  For specific algorithm choices we have the algorithm selection sort-of logic that we've been implementing in several scripts as well as (binary) commands.

> I am also hesitant about `dwimotioncorrect` since that's the name I'm currently using in my development code

That must then be specifically motion correction, and motion correction *only*, I suppose?

> I do anticipate a need for separate commands (rather than algorithm options within a single script) since the interface will be somewhat different.

It really would depend on the details of that, I reckon.  But not knowing those, I'm not judging at all of course.  However, the interface being somewhat different doesn't hinder combination in a joint command with an algorithm selector: we have other algorithms bundled in commands as well that still differ significantly in interface and options.  An important reason to bundle things anyway, is again clarity of the interface: the command name guides users based on the task they want to perform, the algorithm selector lets them choose a strategy for that task, and beyond that sits the interface of the algorithm they've selected, which can also be fully elaborated upon in the manual that appears uniquely for that algorithm when "command algorithm" is entered at the command line.  No problems there.

However, at some point, it would be interesting to elaborate on why the interface is (so) different.  There's no harm in researching whether the interfaces of both can be united up to a certain extent.  I'm sure @Lestropie would be happy to know what, and mostly why, is different for such a command interface wise.  `dwipreproc` has a lively interface history as well. :smile:

Lestropie:
> Final proposal: `dwimotiondistortion`. Still not longer than `fixelcorrespondence` by the way.

Only if you count characters rather than syllables.

> I do anticipate a need for separate commands (rather than algorithm options within a single script) since the interface will be somewhat different

We can visit this when you're closer to release. If two blocks of code perform the same task but using different approaches, then ideally they'd appear as different algorithms within a semi-unified interface. The Python algorithms approach permits both command-line arguments and options to be both common among algorithms and specific to particular algorithms. However if `dwipreproc` is in Python, and your approach is in C++, then I'd agree that the naming of `dwipreproc` may need to be adjusted to reflect the nature of the FSL command wrapping, since there wouldn't really be a neat way to unify the two, and therefore the command names themselves would need to reflect their differences.

thijsdhollander:
Ok, all recent doc updates have been merged inhere as well.

I'm voting to not change `dwipreproc`, because we're not going to be able to make up our minds.  Let's wait until @dchristiaens has his development code ready, and reopen the discussion on `dwipreproc`'s interface, or that other one's interface, or their potential joint interface, or any other interface, at that point.

Ok, so I'm done here.  Every commit in this pull request until before the doc merge in here had already been approved and/or not protested against somewhere in all of the above.  The doc changes aren't anything controversial, and I would encourage anyway to thoroughly go over *all* documentation if we intend to pack this SH fix with `dev` into an RC soon.  So, conclusion being: this is ready to go into `dev`.  If no dramas happen before, let's say, Monday late afternoon or something, I'll merge this into `dev` so we can all move on.

thijsdhollander:
Bonus: via this, the latest `master` changes will also end up on `dev`. :tada:

bjeurissen:
@dchristiaens: I also have ``dwimotioncorrect`` in my development code. I really hope this does not mean we have been working on the same stuff again... We should talk!

thijsdhollander:
@Lestropie , I suppose you may understand this better than I do; see these tests failing: https://travis-ci.org/MRtrix3/mrtrix3/jobs/317859241

From my end, it seems more (pylint) tests started failing since I've merged in `dev` just today (which included the commits from 2 pull requests you've merged into `dev` a couple of days ago).

thijsdhollander:
FYI, in preparation for this merging to `dev`, I've created a `dev` branch on the test data repo as well.  We'll need this, now and potentially in the future, in case test data gets changed/removed (anything but just incrementally added) in context of developments on `dev` or feature branches.

thijsdhollander:
Alright, there we go.  Just to stress the above again: there's now also a `dev` branch on the test data repo, which at the moment includes the changes that have been made here (due to renaming of a few commands).  Since there's 2 other outstanding branches on the test repo, they'll need to have `dev` merged in them for their respective features to eventually go in `dev` on the main repo.

thijsdhollander:
For anyone who wishes to continue discussing anything interface (naming) related, please resort to #1147.  The issue is left open for that purpose.  Should it lead to anything, happily create another branch and pull request.

Lestropie:
Late to the party, but:

Should `mrmesh` be `voxel2mesh`? It strictly only works for 3D data.

jdtournier:
> Should mrmesh be voxel2mesh? It strictly only works for 3D data.

Up to you. I think the `mr` prefix makes it clear that it expects an image as input - even if that would effectively only be a 3D image, but I don't really see that as a problem. The only other command that starts with `voxel` is `voxel2fixel`, which is warranted in this case to highlight the conversion from voxel-wise to fixel-wise data. I guess a similar argument could be made for converting between voxel-wise to surface mesh data? But I don't feel particularly strongly either way...

Issue #1210, opened 2017-12-05
NIfTI qform / sform handling


Lestropie:
Thinking aloud about recent updates to [this forum thread](http://community.mrtrix.org/t/dwipreproc-error-inconsistent-left-right-order-stored-in-sform-and-qform/179/33).

Firstly:

> I note that the NIfTI standard states relatively clearly that the qform should be the transformation to scanner coordinates, while the sform should be the transformation to some template space. 

This to me suggests that we should be reading the `qform` by default if present, and only using `sform` if `qform` is invalid. It might be worth trying to backtrack and confirm why *MRtrix3* was switched to using `sform` by default: It might have been due to some other issue that's since been rectified.

Secondly:

At NIfTI write, we're writing the same transformation to both `qform` and `sform`, since the `Header` class only stores one transformation. However we could use a similar trick to #970:

- At image read, additional header fields within the image format are simply written as key-value parameters in text-readable format.

- At image write, presence of these key-value entries is detected and used to fill the additional header fields within the image format.

This already works for additional 4x4 transformation matrices stored within the "header" (footer) in the MGZ format; so using the same technique for e.g. NIfTI `qform` would probably be not too difficult.

jdtournier:
OK, in short: no, I don't think our handling needs to be modified particularly. I'll expand:

- regarding whether the `sform` or `qform` should be given priority, I can see how it might _seem_ that way. If the `qform` only is set, there is no ambiguity (the idea is that this is what would be read directly from the scanner). If the `sform` only is set, this would be unexpected, given that the intention is for the `qform` to store the transform to scanner coordinates, so that kinda should always be set, but at least there's no ambiguity either. It's when both are set that you have to pick one. While the `qform` gives you the transform to scanner, if the `sform` is also present, it was presumably added subsequently, and for a good reason - so if you think about it, the chances are this is the transform that the users expects to be used, but the NIfTI standard doesn't provide a mechanism to determine for sure which is the right one to use under what circumstances. So our reading the `sform` preferentially is I think in keeping with the _spirit_ of the standard, trying to make the best of an ambiguous situation. 

[...]

PR #1212, opened 2017-12-07
NIfTI: check consistency of sform and qform

jdtournier:
As discussed in #1210

Could still do with added FAQ or similar in documentation...


Lestropie:
@jdtournier: Check that you're happy with that description in the docs.

Lestropie:
I'm happy to go with that. Users will still get a warning in the case of any ambiguity once this is merged. Docs would need to be updated in addition to the code.

jdtournier:
I have to say, I'm not that sure. I distinctly remember someone having issues due to us using the`qform` preferentially (I think there was a time when that was the default behaviour), when interacting with some other package... I think maybe we should add a config file option to allow users to decide for themselves? That way we could adopt what seems to be the spirit of the standard, but allow users to maintain the current behaviour if it causes problems for them. What do you think?

Lestropie:
I thought of a config file entry on the way home but forgot to post it... Definitely let's do that.

I agree using `qform` is more consistent between the standard and *MRtrix3*'s own convention. So if a config file entry would catch that potential problematic case that was a long time ago and the details can't be remembered... I reckon go for that.

jdtournier:
Good to hear, let's do that.

About that issue I mentioned just earlier, thinking about it some more, I think the problem was that at the time, we were only writing out the `sform`, and that other package relied exclusively on the `qform`. So it's a moot point anyway...

Lestropie:
```
# command: tckgen SIFT_phantom/dwi.mif -algo seedtest -seed_random_per_voxel SIFT_phantom/mask.mif 27 tmp.tck -force && tckmap tmp.tck -template SIFT_phantom/dwi.mif - | testing_diff_image - tckgen/SIFT_phantom_seeds.mif [ ERROR ]
tckgen: [WARNING] existing output files will be overwritten
tckgen: preloading data for "SIFT_phantom/dwi.mif"...  [==================================================]
tckgen: [  0%]        1 seeds,        1 streamlines,        1 selected
tckgen: [  1%]       38 seeds,       38 streamlines,       38 selected
tckgen: [  1%]       39 seeds,       39 streamlines,       39 selected
tckgen: [  2%]       79 seeds,       79 streamlines,       79 selected
tckgen: [  4%]      157 seeds,      157 streamlines,      157 selected
tckgen: [  8%]      313 seeds,      313 streamlines,      313 selected
tckgen: [ 16%]      624 seeds,      624 streamlines,      624 selected
tckgen: [ 32%]     1246 seeds,     1246 streamlines,     1246 selected
tckgen: [ 64%]     2490 seeds,     2490 streamlines,     2490 selected
tckgen: [100%]     3888 seeds,     3888 streamlines,     3888 selected
tckmap: mapping tracks to image...  [==================================================]
testing_diff_image: [ERROR] images "/tmp/mrtrix-tmp-iQigO4.mif" and "tckgen/SIFT_phantom_seeds.mif" do not match within absolute precision of 0 (28 vs 27)
testing_diff_image: [ERROR] images "/tmp/mrtrix-tmp-iQigO4.mif" and "tckgen/SIFT_phantom_seeds.mif" do not match within absolute precision of 0 (26 vs 27)
testing_diff_image: [ERROR] exception thrown from one or more threads "loop threads"
```

o.0

Lestropie:
Just discovered another confound related to this and other issues.

SPM's "Coregister: Estimate" functionality, which is commonly used for inter-modality intra-subject rigid-body registration, directly modifies the `sform` of the input NIfTI, but does *not* modify the `qform`. As such, if *MRtrix3* loads the `qform`, while a command-line warning is issued about this fact, such warnings are often ignored, and the image will appear in `mrview` as though the registration has not yet been performed.

maxpietsch:
ANTs' (and ITK's) NIfTI handling differs again: 
`Therefore it can only use qform when reading a NIFTI-1 image. When writing NIFTI output, only the qform is written; the sform is replaced with zeros.`
https://github.com/ANTsX/ANTs/wiki/How-does-ANTs-handle-qform-and-sform-in-NIFTI-1-images%3F

Lestropie:
> When writing NIFTI output, only the qform is written; the sform is replaced with zeros.

That's not so much of a concern; if anything *MRtrix3* is quite happy if one is set and the other is not. The problem with SPM is specifically because it modifies the sform in place but leaves the qform untouched. So if the qform is filled with valid data before that operation takes place, then after that operation, both will be flagged as being valid and both will contain valid data, yet they will be different from one another; hence the ambiguity.

PR #1214, opened 2017-12-14
fixel2tsf NaN fix

Lestropie:
Details in commit messages.

Would merge straight away if it were just the bug fix, but it does raise a couple of questions:

- Should fixels outside the processing mask in `fixelcfestats` be zero-filled rather than having NaNs? Personally I prefer the latter, but I'm wondering whether or not there's anywhere else in the code where the possibility of non-finite fixel values may not have been considered and may therefore lead to erroneous behaviour.

- I didn't end up really needing the new `tsfvalidate` command here since the `fixel2tsf` issue turned out to be simpler than I thought, but it might be a useful command to add to the repo nonetheless. Even if not for the thorough validation check, it might make more sense to many users to use this to cross-reference `.tck` and `.tsf` files rather than checking header timestamps.


thijsdhollander:
> Should fixels outside the processing mask in fixelcfestats be zero-filled rather than having NaNs? 

Zero-filling is probably safer / more "robust", I would argue.

> I didn't end up really needing the new tsfvalidate command here since the fixel2tsf issue turned out to be simpler than I thought, but it might be a useful command to add to the repo nonetheless. Even if not for the thorough validation check, it might make more sense to many users to use this to cross-reference .tck and .tsf files rather than checking header timestamps.

It's a solid command with a clearly defined purpose / task.  I see no harm in including it, and even before we may use it, others might (want to) use it if it comes in handy.  So it has my full support to be included. :+1:

PR #1219, opened 2017-12-19
5tt2gmwmi: Fix -mask option

Lestropie:
Error was made in porting to new syntax as part of 0.3.14, within commit ec49edcaf.

(Pretty sure this would have been one of the very first commands I ported...)


thijsdhollander:
(I've just restarted the python 2 job, 2166.1, because it had failed due to timeout)

PR #1221, opened 2017-12-20
Python: Fix pylint consider-using-enumerate

Lestropie:
Rather than disabling this Pylint warning, instead use more Pythonic approaches where warnings were issued.

Reverts b54c55a7 committed to `dev` in #1211.

@jdtournier: I tested the `convert_bruker` change in interactive Python, but it might be worth running it on data just to be sure.

Also, short aside: I decided recently that when Python variables correspond to option strings to be used in a `run.command()` call, explicitly naming them as such really helps to separate those variables from other pieces of data flying around. :grin:


thijsdhollander:
Yeah, that's fair enough. :+1:   That was originally not much more than a temporary workaround just to get #1211 done and over with.  I did/do wonder though how that suddenly popped up there (in #1211), since it didn't in any other pull requests that had just been merged to `dev` only days before, whereas e.g. convert_bruker and others where also already looking the way they did in #1211... or I must be overlooking something else somehow.

Lestropie:
My guess was that TravisCI started pulling newer versions of `pylint`. But they appear to have not changed. ¯\(°_o)/¯

thijsdhollander:
> My guess was that TravisCI started pulling newer versions of `pylint`.

Yes, that was the only explanation I could imagine at the time as well.  If this would somehow have been the case, or if this can indeed happen in the future as well, we could (eventually) consider enabling a list of pylint checks, rather than disabling a list.  I mean, it's in a way a bit weird that our TravisCI checks would change behaviour "all by themselves" over time.

Issue #1043, opened 2017-07-11
Over-writing output fixel directories


Lestropie:
http://community.mrtrix.org/t/fixel-analysis-pipeline-index-mif-files-are-smaller-than-expected/1010

To me, suggesting that users explicitly make a copy of their fixel template before adding subject data to it (purely in order to retain a "clean" template), then having `fixelcorrespondence` / `warp2metric` receive a single fixel image as both input fixel template and output directory target, made more sense logically, and would be robust against this issue.

Lestropie:
http://community.mrtrix.org/t/tck2fixel-segmentation-fault/2497

Lestropie:
Integrating into the discussion on this issue the fact that the `-force` command-line option should work sensibly for fixel directories.

I think what may be needed here is a comprehensive list of commands that output fixel data (whether a complete directory, or a single fixel data file), and consider what the appropriate behaviour should be with equivalent input & output directories / with & without `-force`. I'll probably also propose how I would prefer the fixel command interfaces to be set up, which I think would resolve both the issues originally listed here and those mentioned in #2167.

jdtournier:
There's a lot of moving parts here... Agree we need to list at least a few scenarios and try to figure out whether there's a simple strategy that would make sense - at least in principle. 

I'm not sure I'm completely on top of the issue, but as far as I can tell there are at least two conflicting situations where `-force` could legitimately be invoked:

#### A command that produces a fresh fixel directory is asked to overwrite an existing fixel directory (e.g. `fod2fixel`)

In this case, there's no reason to expect the new fixel index & directions files to match any existing data file (at least not _a priori_). We should consider that such an operation should invalidate all existing data files, and so the logical thing to do is to wipe the entire folder clean. 

This is quite a drastic operation, and very likely to lead to unexpected loss of data for users - they might not have realised that the command would remove all files, not just the ones that the command would ordinarily produce. Solutions might include (one or more of):

- perform a full comparison of the new and existing index and directions files, retain the originals if they match. 
- provide a big warning and request confirmation before wiping any additional data files (if any are found). Could also be done with an additional option to be supplied on top of `-force` (e.g. `-confirm`, `-clean` , `-wipe`, ...), since requesting confirmation in an interactive manner would get in the way of scripts, etc. 
[...]

PR #1222, opened 2017-12-22
App: New argument types for input/output directories

Lestropie:
Was throwing together a command to serve a niche purpose involving fixel data, and the lack of this functionality was giving me the irrits.

Though the fact that many fixel commands have an output directory that may also be an input directory, and therefore can't actually use this functionality properly, reminded me of #1043...

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -1076,11 +1076,6 @@ namespace MR
           throw Exception ("input file \"" + text + "\" is not a valid track file");
         if (i.arg->type == TracksOut && !Path::has_suffix (text, ".tck"))
           throw Exception ("output track file \"" + text + "\" must use the .tck suffix");
-        if (i.arg->type == ArgDirectoryOut) {
-          const std::string base_name = Path::basename (text);
-          if (base_name.find (".") != std::string::npos)
-            throw Exception ("output path \"" + text + "\" is not a valid directory path (contains a \'.\')");
-        }
       }
       for (const auto& i : option) {
         for (size_t j = 0; j != i.opt->size(); ++j) {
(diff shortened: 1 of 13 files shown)


thijsdhollander:
I've got nothing against this, *if* someone tests that it works properly and doesn't break anything of course. :+1:

core/app.cpp
```diff
@@ -1042,27 +1052,68 @@ namespace MR
...
+        if (i.arg->type == ArgDirectoryOut)
+          check_overwrite (text);
+        if (i.arg->type == TracksIn && !Path::has_suffix (text, ".tck"))
+          throw Exception ("input file \"" + text + "\" is not a valid track file");
+        if (i.arg->type == TracksOut && !Path::has_suffix (text, ".tck"))
+          throw Exception ("output track file \"" + text + "\" must use the .tck suffix");
+        if (i.arg->type == ArgDirectoryOut) {
+          const std::string base_name = Path::basename (text);
+          if (base_name.find (".") != std::string::npos)
+            throw Exception ("output path \"" + text + "\" is not a valid directory path (contains a \'.\')");
```
jdtournier: Why explicitly disallow dots in the directory name? There's no particular reason Unix-wise to prevent this... (?)

Lestropie: Yeah, I umm-ed and arr-ed about that for a while. Even though it's legal, it's not something I would generally recommend, and could see it leading to confusion in cases where the user does not fully understand that the output is intended to be a directory rather than a file (e.g. ending up with fixel directories with ".mif" postfixes). Happy for that test to be removed if you think it's too un-Unix-ey.

jdtournier: I know what you mean. But while it would be odd to have directory with a `.mif` suffix, it wouldn't actually cause any problems for the code in any way - everything would work as expected. And preventing it would limit what users can legitimatelly do, e.g. they couldn't label their fixel folder `FBA_03.01.2018/`, which strikes me a perfectly sensible thing to call it. I'd rather take than out, I think that's trying too hard to guide novice users, when the worst case consequence of them mislabeling their folder is minor confusion...

PR #1227, opened 2018-01-02
Revert "fix SH precomputer"

thijsdhollander:
Reverts MRtrix3/mrtrix3#1206

Happy holidays all.  Just quickly chipping in to mention we had not agreed on merging this: there are several outstanding issues in this one, raised a couple of times by me, that have to be addressed first.  I've said I'll look into those thresholds when I'm back, next week that is.  We've delayed pull requests for much longer, and for much lesser reasons in the past too; I reckon this can wait until addressed properly.  I personally insist on it.  Thanks.

### DIFF
--- a/cmd/sh2peaks.cpp
+++ b/cmd/sh2peaks.cpp
@@ -70,10 +70,7 @@ void usage ()
 
   + Option ("mask",
             "only perform computation within the specified binary brain mask image.")
-  + Argument ("image").type_image_in()
-
-  + Option ("fast",
-            "use lookup table to compute Associated Legendre Polynomials (faster, but approximate).");
+  + Argument ("image").type_image_in();
 }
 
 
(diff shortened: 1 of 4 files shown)


thijsdhollander:
FYI: https://github.com/MRtrix3/mrtrix3/pull/1206#issuecomment-352948378 

First dot point is propably ok, the rest is not (yet).

jdtournier:
:confused: ?!? We're merging into `dev` here, not `master` - these other issues can and will be addressed, but there's no particularly reason to delay the merge at this point, is there? It's a self-contained fix, although it does have repercussions on defaults in other apps - but I don't see that we need to hold off on merging the fix itself...

In terms of the remaining points to be addressed (is that the complete list?):

> - check default tckgen FOD cutoff; as indicated above, I think this should consequently be lowered

Maybe, but right now the fix actually reverts the cutoff to exactly what it was before the bug was introduced. So while I agree this threshold should be revisited, I personally view this as a separate conversation. 

> - just thought of now: probably same thing for the default -fmls_peak_value in fod2fixel

That is potentially more problematic, since that value has (probably?) always been subject to the bug. But again, it sounds like there's been some incentive to change this default anyway as it wasn't appropriate, regardless of this bug. So I would again consider that to be separate conversation.

> - in case the previous bullet is indeed a thing, lower the "default" recommendations for this parameter in both (single tissue and multi tissue) FBA pipelines in the userdocs

This would follow from the previous point, so would need to be discussed as part of it.

> - potentially similar changes to such values in other userdocs pages

As above. 

So I'm not sure which of these issues you consider to be severe enough to hold off from merging? Bearing in mind this is going into `dev`, not `master`...

PR #1235, opened 2018-01-18
Tracking code: Proper use of polymorphism

### LINKED ISSUE #1234: method.truncate_track not overridden and lacks ACT check
In advance: please excuse me if this is a wild goose chase – I don’t have any ACT-ready data to test this on. These also aren't issues that affect me, but they might affect others now, or be a headache for you in the future.

I’ve just been playing with the MRtrix source code, particularly with relevance to `method.truncate_track(tck, max_size_at_backtrack, revert_step);`

I needed this method for a mod I’m working on and found what I can only assume (?) are two bugs when combined with iFOD2:

1)	iFOD2 has its own version of this method but this lacks the override keyword. Thus, when you call `method.truncate_track` using `iFOD2`, it calls the version in `MethodBase`
Whether this currently impacts ACT, I’m not sure. Easy enough to test (Add `WARN("not a problem!");` in `iFOD2::method.truncate_track` and run ACT with backtracking) but – as said previously, I don’t have any 5TT etc files ready to investigate.
The solution to this is obviously to implement `virtual…override` in the relevant classes.
2)	Within the `iFOD.truncate_track` method, the final line:
`act().sgm_depth = (act().sgm_depth > points_to_remove) ? act().sgm_depth - points_to_remove : 0;`
needs an additional check as to whether ACT is actually in progress. Currently if this is called when ACT is not in progress, the method throws an exception 
```
if(S.is_act()){
act().sgm_depth = (act().sgm_depth > points_to_remove) ? act().sgm_depth - points_to_remove : 0;
}
```

Lee

Lestropie:
As requested by @LeeReidCSIRO.

Previously, some tracking code was heavily reliant on the use of templating in `cmd/tckgen.cpp`. While the code would behave as expected in this context, some aspects appeared as though they had been designed with polymorphism in mind, but would not exhibit the expected behaviour if actually used in this fashion. This change introduces the correct polymorphism concepts such that the code can be used in this way if desired. It however does not have any significant influence on execution speed of `tckgen`, since it still exploits the template parameter of the `Tracking::Exec` class and hence does not use the underlying polymorphism.

Closes #1234.

Have confirmed that there is no measurable effect on `tckgen` performance.

### DIFF
--- a/src/dwi/tractography/algorithms/fact.h
+++ b/src/dwi/tractography/algorithms/fact.h
@@ -90,20 +90,20 @@ namespace MR
 
 
 
-      bool init()
+      bool init() override
       {
         if (!get_data (source)) return false;
         if (!S.init_dir.allFinite()) {
           if (!dir.allFinite())
             dir = random_direction();
-        } 
-        else 
+        }
+        else
           dir = S.init_dir;
-        
+
         return do_next (dir) >= S.threshold;
       }
 
-      term_t next ()
+      term_t next () override
       {
         if (!get_data (source))
           return EXIT_IMAGE;
@@ -118,7 +118,7 @@ namespace MR
       }
 
 
-      float get_metric()
+      float get_metric() override
       {
         Eigen::Vector3f d (dir);
         return do_next (d);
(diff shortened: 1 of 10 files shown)


Lestropie:
@jdtournier: Any particular objections to this (when you get the chance)? Just trying to reduce the length of the list a little.

PR #1237, opened 2018-01-19
run.command(): Support native Windows Python

Lestropie:
Unlike the Python versions provided by either MSYS or MinGW, the native Windows versions of Python are not POSIX compliant. As a result, `shlex.split()` does not appropriately handle the file system separator symbol. This change performs the necessary alterations to the `shlex.split()` call under such an environment.


jdtournier:
This PR is currently based on `master`. I guess you'd need to rebase to `dev` before merging if you want this to be part of `3.0_RC3`?

PR #1241, opened 2018-01-29
mrcalc: More informative error message

Lestropie:
Prevents erroneous image file paths from resulting in an "error converting string" error message (due to erroneous string-to-float conversion).

As discussed on forum [here](http://community.mrtrix.org/t/displaying-fixl-based-analysis-results/1433).


jdtournier:
:+1: pretty much exactly what I had in mind...

PR #1242, opened 2018-01-30
Scripts: Small fixes to reflect mrinfo change

Lestropie:
`mrinfo`'s command-line option `-shellvalues` was changed to `-shell_bvalues` in 29dbf025, merged to `dev` branch in #1211, but a couple of usages of this option in *MRtrix3* scripts were not updated.


thijsdhollander:
:+1:  I think I remember correcting the `dwi2response fa` one in another branch either on GitHub or on my machine as well, but that definitely didn't end up on `dev` yet.  All good to update `dev` already via this pull request, so things aren't broken.

PR #555, opened 2016-04-18
tcksample

Lestropie:
Opening this up to devs to have a look around and comment if they see fit. Full description of the changes both in the commit messages, in the relevant private community forum post, or a lot is in the `tcksample` and `tckresample` help pages.

Would like to have this merged along with #511 and registration, just to form a nice little tag point, and justify a forum post about everything we've worked on since the `updated_syntax` merge. And apart from merging in `rcsd`, these may form the last set of 'new features' before we focus on bug fixes in order to release.

PR #1248, opened 2018-02-05
tckresample: Fix -line, -arc options

Lestropie:
Fixing issues reported on forum [here](http://community.mrtrix.org/t/tckresample-and-tcksample-for-along-the-tract-statistics/1368).

Example data:

screenshot0000

Current `master` code, using `-line` option:

screenshot0001

Current `master` code, using `-arc` option:

screenshot0002

PR code, using `-line` option:

screenshot0003

PR code, using `-arc` option:

screenshot0004

### DIFF
--- a/src/dwi/tractography/resampling/arc.cpp
+++ b/src/dwi/tractography/resampling/arc.cpp
@@ -34,28 +34,27 @@ namespace MR {
 
           // Determine which points on the streamline correspond to the endpoints of the arc
           idx_start = idx_end = 0;
-          size_t a (0), b (0);
+          size_t a (in.size()), b (in.size());
 
-          int prev_s = -1;
+          state_t prev_s = state_t::BEFORE_START;
           for (size_t i = 0; i < in.size(); ++i) {
-            int s = state (in[i]);
+            const state_t s = state (in[i]);
             if (i) {
-              if (prev_s == -1 && s == 0) a = i-1;
-              if (prev_s == 0 && s == -1) a = i;
-              if (prev_s == 1 && s == 2)  b = i;
-              if (prev_s == 2 && s == 1)  b = i-1;
+              if (prev_s == state_t::BEFORE_START && s == state_t::AFTER_START)  a = i-1;
+              if (prev_s == state_t::AFTER_START  && s == state_t::BEFORE_START) a = i;
+              if (prev_s == state_t::BEFORE_END   && s == state_t::AFTER_END)    b = i;
+              if (prev_s == state_t::AFTER_END    && s == state_t::BEFORE_END)   b = i-1;
 
-              if (a && b) {
+              if (a != in.size() && b != in.size()) {
                 if (b - a > idx_end - idx_start) {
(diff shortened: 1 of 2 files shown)


Lestropie:
Opening review of this one to anyone given the evidence presented. Though I probably should have clarified: Control points are either end of the corpus callosum pre-centrum-semiovale, with the arc having the midpoint in the centre of the CC (which is slightly inferior to the two endpoints). So having 10 equidistant points along those two paths should appear exactly as the two bottom images are.

src/dwi/tractography/resampling/arc.cpp
```diff
@@ -64,16 +63,19 @@ namespace MR {
...
 
           for (size_t n = 0; n < nsamples; n++) {
-            while (i != idx_end) {
+            do {
               const value_type d = planes[n].dist (in[i]);
               if (d > 0.0) {
                 const value_type f = d / (d - planes[n].dist (in[reverse ? i+1 : i-1]));
-                out.push_back (f*in[i-1] + (1.0f-f)*in[i]);
+                assert (f >= 0.0 && f <= 1.0);
+                out.push_back (f*in[reverse ? i+1 : i-1] + (1.0f-f)*in[i]);
```
jdtournier: Seems all very sensible - there was clearly an issue with the indexing here. But I'm struggling a bit to see what was causing the obvious failure in the arc case, which looks like it's using the wrong cut-plane normals - I can't see any of the current changes impacting on that... (?)

Lestropie: I borked the arc sampling all the way back in #555. Arc initialisation was completely ignoring the midpoint provided by the user. :-/

src/dwi/tractography/resampling/arc.h
```diff
@@ -64,12 +66,12 @@ namespace MR {
             Arc (const size_t n, const point_type& s, const point_type& w, const point_type& e) :
                 nsamples (n),
                 start (s),
-                mid (0.5*(s+e)),
+                mid (w),
```
jdtournier: > I borked the arc sampling all the way back in #555. Arc initialisation was completely ignoring the midpoint provided by the user. :-/

Ah, so this is essentially the fix for that, then...? But I still don't get why it looked like it was sampling along a completely different arc, if presumably all that happened was that the midpoint specified by the user was ignored, and the default midpoint used instead... (?)

Lestropie: If I had to guess, I'd say that in `init_arc()`, it's solving for `c` as the centre point of a circle containing the three points provided by the user, but when the midpoint is the bisection point the circle can't be defined, so Eigen is setting `c` to `[0.0, 0.0, 0.0]`, and that's therefore the real-space point that all the sampling planes are passing through. But I didn't bother actually interrogating it that deeply, since it was clearly an egregious fault.

jdtournier: OK, I guess that makes sense. Happy for it to be merged.

PR #1249, opened 2018-02-06
Added GIF algorithm for 5ttgen

matteomancini:
This pull request add the option of using the Geodesic Information Flow (GIF) segmentation algorithm for 5ttgen. The algorithm is described [here](http://ieeexplore.ieee.org/document/7086081/) and is implemented in the [NiftySeg](http://cmictig.cs.ucl.ac.uk/research/software/software-nifty/niftyseg) package.


Lestropie:
Hi @matteomancini,

Awesome to see :+1: I've been tinkering with a couple of alternative `5ttgen` algorithms myself recently, but haven't found the time to polish them off...

- Generating the WM image could be simplified as:
```
run.command('mrconvert input.mif - -coord 3 3,5 | mrmath - sum WM.mif -axis 3')
```

- For users, it would be more convenient if the `5ttgen` algorithm itself received as input a T1 image, internally ran the requisite NiftySeg command(s) (after having ensured that they are properly installed), and then converting the output from NiftySeg into the 5TT format, rather than requiring that NiftySeg be run explicitly first. Do you think that this would be possible? Or are there internals to NiftySeg that can't be abstracted away from the user?

matteomancini:
Hi Robert,
thanks for the feedback. So I updated the script to simplify the WM image generation. For the actual input, the issue is that on a good machine or on a dedicated cluster node the segmentation takes 5-6 hours. Although it is not as time-consuming as freesurfer, I thought that using the same approach (i.e. feeding the segmentation results rather than the T1) would be more convenient. Let me know what you think,

Lestropie:
I suppose it depends on the user: If they want to use the GIF output for more than 5TT generation, then yes it makes more sense to run them as two separate steps; whereas if the only intended output of GIF is the 5TT image, then having it all encapsulated as a single step may be more convenient.

But there's actually nothing stopping the script from supporting both use cases: By loading the input image size using the `mrtrix3.image` module, and testing whether the image is 3D or 4D with 6 (?) volumes, the script could decide whether the input is a T1 (in which case it would run GIF plus the 5TT conversion) or a GIF output image (in which case it just does the 5TT conversion).

Up to you whether or not you want to go to this effort; and there's no reason why the script can't be committed then modified later to add such functionality. I would though suggest that the script should check the size of the input image and give a meaningful error message either way; otherwise, if a user attempts to run this algorithm on a 3D T1-weighted image, they'll get a slightly obscure `mrconvert` "`axis provided with -coord option is out of range of input image`" error.

matteomancini:
I agree that it is a matter of user's choice, although NiftySeg has usually at least two indended outputs (segmentation and parcellation, two distinct files). There may be a way to embed the generation of both the files in 5ttgen, but at the moment it would be great to have at least the freesurfer-like conversion from segmentation to 5TT.
Thank you for the suggestion, I added an input validation step in the script.

Lestropie:
If there are two distinct outputs, one way would be to add a command-line option that additionally writes the parcellation image (and/or the raw segmentation image) from the GIF execution to the file system, rather than having them disappear with the script temporary directory. But I'm happy to add the algorithm without going down that route.

@jdtournier: Thoughts on whether this should go to `master` or `dev`? I'm hoping to knock some `3.0_RC3` requirements off once the fellowship is done (due 8th March), so it probably wouldn't delay the release of this algorithm a great deal if it were to go to `dev` instead. This would require some changes:

- Double-whitespace indentation

- Addition of any `pylint` pragmas required

- Modification of image header access code:
`
dim = [ int(i) for i in image.headerField(image_path, 'size').strip().split() ]
`
becomes:
`
dim = image.Header(image_path).size()
`

jdtournier:
No particular thoughts on the matter. If there's no huge rush, I'd merge to `dev` - we need to issue a release soon anyway, so given that there is a bit of polishing to do on this pull request, it'll probably not delay this going out a great deal. 

@matteomancini: is there any urgency to get this out?

Lestropie:
@matteomancini: Are you happy to make the requisite changes in order to make this change in the `dev` branch? I've already changed the base for this pull request; but if I make the corresponding changes in `/mrtrix3/_5ttgen/gif.py`, you'll lose some of the credit because I'll be over-writing your additions :-/

The new script `run_pylint.py` that resides in the root directory will automatically detect the presence of this new file, and instruct what needs to be done to pass the tests. Mostly this requires changing the indentation size, but you'll also see from the other script algorithms that some [explicit test disable pragmas](https://github.com/MRtrix3/mrtrix3/blob/dev/lib/mrtrix3/_5ttgen/fsl.py#L1) are required to satisfy `pylint`.

matteomancini:
So I surely can do the last changes, just not sure about pylint, I don't think there's need to add anything specific, how would I check that? Should I just do the other changes and give a go to run_pylint.py? Anyway, in the mean time I'll proceed to adjust indentation and header. Thank you.

Lestropie:
Yep; script `run_pylint` will tell you exactly what it's unhappy about, and therefore what you need to change in order to pass the tests (the Continuous Integration back-end runs this very script; you can see it currently failing by clicking on the "Details" hyperlink below and to the right of the bright red "All checks have failed" near the bottom of this webpage).

One other thing you can do that I forgot about: After running the `build` script on this branch, run the `docs/generate_user_docs.sh` script. This auto-generates certain parts of the *MRtrix3* documentation. After doing so, running `git status` should show that file `docs/reference/scripts/5ttgen.rst` has changed: this is due to the `gif` algorithm being automatically added to the documentation page for the `5ttgen` script. Commit this change along with your other changes; this way your GitHub account will additionally get credit for the updates to the documentation :+1:

thijsdhollander:
Hi @matteomancini,

Thanks again for your hard work and contributions.  While I was working on related stuff anyway,  I fixed those `pylint` issues for you.

> So I surely can do the last changes, just not sure about pylint, I don't think there's need to add anything specific, how would I check that?

There wasn't anything *wrong* with your script; `pylint` is merely quite OCD about style, layout, and a particular definition of "good practice" (which is indeed subjective).

I've also generated the necessary documentation, and made a minor adjustment in the synopsis (making the "GIF" acronym explicit: people may not know it, and in an image-world it may also confuse people, since there's a ".gif" image format of course).  Definitely check that the script still works as intended (and that, while making `pylint` happy, I haven't broken anything).

About credit on Github, if you're worried about that sort of thing at least: don't worry, your original commits are entirely left untouched and will eventually end up on MRtrix3's `master`, the script is still entirely authored by you (author field of the thing), the pull request is still entirely created by you (I think that ends up somewhere somehow in your stats too, should that be something you care about), and the commits that I made to fix things up for you I've created as such that they acknowledge you as a co-author, which adds all of that (lines, commits, whatever) to your profile as well when it lands on `master`, just as if you would've done all of that yourself.  :pray: Amen. :stuck_out_tongue_winking_eye: 

In principle, the tests should now succeed; but don't immediately worry if they don't: they time out sometimes, or run into other hick-ups.  We can simply restart them if that's the problem.

Cheers,
Thijs

thijsdhollander:
Just added this final tweak:

> Modification of image header access code:
> dim = [ int(i) for i in image.headerField(image_path, 'size').strip().split() ]
> becomes:
> dim = image.Header(image_path).size()

matteomancini:
Hi Thijs, thank you for taking care of this, I'll test the current version of the script and I'll get back at you.

matteomancini:
So I pulled the changes and tested it, it works fine.

thijsdhollander:
Great! It's merged into dev now. If there's anything further at any point, just let us know.

Lestropie:
@matteomancini: Would it be possible for you to add a custom avatar to your GitHub profile? The addition in this PR will be included in the reported changelog for the next *MRtrix3* update; we are currently working on formatting for that announcement, but your use of the default GitHub avatar is causing issues. Plus it'd be nice to have your image embedded in there next to details of your contribution :+1:

matteomancini:
Done!

PR #1244, opened 2018-01-30
Userdocs update

thijsdhollander:
If we link to this for response function selection in the workshop topics list, it has to be up to date.

PR #1250, opened 2018-02-07
Slightly alter image transform to RAS

Lestropie:
Arose from [review](https://github.com/MRtrix3/mrtrix3/pull/1244#pullrequestreview-93962779) in #1244. I think this should be more correct...


jdtournier:
OK, all checks out on my DICOM data - happy for it to merge.

PR #1246, opened 2018-01-31
Configure remove linker optimisation flags

jdtournier:
as discussed in #1243

Issue #1251, opened 2018-02-08
AppVeyor


Lestropie:
Apparently some multi-OS projects are using TravisCI for Linux / OSX testing, and [AppVeyor](https://www.appveyor.com/) for Windows testing, which includes [MSYS2 support](https://help.appveyor.com/discussions/suggestions/615-support-for-msys2). Free for OSS projects.

Lestropie:
Closed by #1387.

PR #1254, opened 2018-02-16
TravisCI: Try to add OSX test

Lestropie:
Just like #1181, this will likely require a lot of tinkering, and requires a PR to figure out what works and what doesn't.

While they give some [documentation](https://docs.travis-ci.com/user/multi-os/) describing how to branch for conditional installation / execution, I haven't found anything describing how to branch in the `addons:` section. So I'm expecting that to fail spectacularly at first try for OSX. But maybe it'll give me some hints on how to fix it...

Trying to avoid issues like #1246. Idea of OSX testing initially raised in #1181.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,7 +90,7 @@ before_script:
   # Since this is not needed for the installation section, only for the travis.sh script, moved this to "before_script" #
   #######################################################################################################################
   - |
-    if [[ -v TRAVIS_CXX ]]; then
+    if [[ -n "${TRAVIS_CXX}" ]]; then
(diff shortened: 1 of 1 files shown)


Lestropie:
TravisCI now successfully additionally testing OSX :+1:

If anyone thinks that the list of tests needs to be further expanded in light of multi-OS support, best raise it here.

jdtournier:
Good effort! Looks good to me - feel free to merge. 

I guess the only remaining OS is Windows, as per #1251. But I don't think you should delay merging this, it'll probably require quite a bit more effort to sort out the Windows testing...

PR #1261, opened 2018-02-28
Pull latest master changes into dev

Lestropie:


Lestropie:
Cancelling; forgot can't directly merge since it requires pulling changes from `dev` into `master` as part of the PR. Will create a new branch.

PR #1263, opened 2018-03-02
Docs to dev

thijsdhollander:
`doctest` remains the branch that builds on `master`, but since docs updates are piling up and not getting through, they may just as well come as part of the next release candidate, if they aren't on `master` before that already.


thijsdhollander:
Note part of the stuff that was/is piling up on `doctest` had already made its way into `dev` before; with a similar logic in mind.

PR #1264, opened 2018-03-06
Stats: Fix segfault when using -nonstationary

Lestropie:
Memory for storing empirical statistics was not being allocated either before or during `Stats::PermTest::precompute_empirical_stat()` function.


Lestropie:
Have verified fix of initial fault with `gdb`, and have performed a test run through to command completion in release mode without a segfault, so we should be good. @ansk: let us know if you still encounter any issues.

PR #1267, opened 2018-03-14
configure: Alter debug optimisation on Windows

Lestropie:
On Windows, when performing a debug build (i.e. '`-g -O0`'), compilation fails in multiple locations due to file size limitations. While compilation of particular commands has been impossible in debug mode on Windows for some time now, currently it's impossible to compile *any* command, as the *MRtrix3* library cannot be constructed.

This partially reverts the change of optimisation level when compiling in debug mode from commit f341021, discussed in #1072 and merged in #1143, reverting back to `-O1` only for Windows.


jdtournier:
Fine by me. But all this feels like we need to start considering a major code restructure to split up the code over a many more smaller files, both for (re-)compile speeds and to avoid this kind of issue...

Lestropie:
I just spent *ages* splitting code across multiple files. However that's not the sole cause this time around, and it was going to lead to non-sensible changes.

Previously, compilation errors on Windows have been related to having too many sections in the object file. That can be solved pretty directly by splitting code across multiple files. This time, the error is instead about a string table overflow, specifically where debugging symbols are being stored. There were multiple culprits, but mostly they were files with a lot of Eigen objects, and deep in the filesystem hierarchy (so the namespace path is long), but not actually *complex*. E.g. [src/registration/transform/initialiser.h](https://github.com/MRtrix3/mrtrix3/blob/master/src/registration/transform/initialiser.h) was an issue: it's got a couple of simple structs and a few non-templated functions, that's it. Not really the sort of thing you'd look at and think "this file needs to be broken up". But the problem is the string length of the symbols involved. So re-arranging the code to satisfy this eventually felt to me like an exercise in band-aiding a bad compiler rather than making sensible structural changes to our own code.

As long as release mode is fine, and debug in OSX / Linux is fine, I'm content to write this off as a limitation imposed by MinGW's `g++`. (That reminds me, I should test `clang` on MSYS2...)

PR #1268, opened 2018-03-14
Fix tckconvert vtk to tck

blezek:
Fixes 2 edge cases.

1. when a VTK file does not have any lines, an uninitialized variable caused segfaults on Linux (but not Mac, hmm)

2. A .tck file with 0 streamlines now produces a valid RIB file.

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -530,7 +530,7 @@ private:
 
 class RibWriter: public WriterInterface<float> { MEMALIGN(RibWriter)
 public:
-RibWriter(const std::string& file, float radius = 0.1, bool dec = false) : out(file), writeDEC(dec), radius(radius) {
+RibWriter(const std::string& file, float radius = 0.1, bool dec = false) : out(file), writeDEC(dec), radius(radius), hasPoints(false), wroteHeader(false) {
   pointsFilename = File::create_tempfile(0,".points");
   pointsOF.open(pointsFilename );
   pointsOF << "\"P\" [";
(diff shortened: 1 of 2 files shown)


jdtournier:
Happy for this to go in - just want to make sure this has been tested thoroughly first. I don't think we have tests for this part of the code. Have you checked that it still behaves as previously for a range of 'normal' inputs?

blezek:
@jdtournier I did test it out on data that ran fine before and it works as expected.  Is there a document that describes how to add testing?  I'd be more than happy to write a few tests.

jdtournier:
That would be great! Instructions in [testing/README.md](https://github.com/MRtrix3/mrtrix3/blob/master/testing/README.md) :+1:

blezek:
@jdtournier, I added tests for the two cases that I fixed.  Thanks for the help.

jdtournier:
Looks good. I can see the tests only check that the conversion works for an empty track file, right? 

One suggestion: you could simplify your first test like this - no need to use `testing_diff_matrix` here:
```
echo 1 2 3 > tmp.txt && tckconvert -force -quiet tmp.txt tmp.tck && tckconvert -quiet -force tmp.tck tmp.rib && [ $(wc -l < tmp.rib) == 4 ]
```

blezek:
@jdtournier  Thanks for the bash tip, wasn't sure what level of "script-fu" was acceptable!  Push this change.

Yes, second test is for the edge case where an empty track file produces bad a poorly formed RIB.  Is there a way to put comments in the testing scripts?

jdtournier:
> Is there a way to put comments in the testing scripts?

Not currently, but it's a trivial fix. I'll push something for this in a minute...

blezek:
Thanks @Lestropie for the review.  `tckconvert`s limitation for reading binary VTK and writing ASCII is due to the very specific problem that I needed to solve.  I can add the other two permutations if needed, but I'd prefer to avoid doing so.

I'd be happy to add comments in a later pull request.  Please go ahead and merge, and thanks for the hard work making MRTrix a great tool!

PR #1269, opened 2018-03-14
shbasis: Remove 'native' conversion

Lestropie:
Since support for the old non-orthonormal SH basis was removed in #1209, the 'native' conversion type - which was based off of the current build configuration with respect to the SH basis - is no longer necessary.


thijsdhollander:
I would go further and wonder if `shbasis` is actually still needed.  I'm often getting the feedback that some of these things, and certain pages in the documentation, are dragging our "history" a bit too much along.  Especially with the release in mind, and since we've now also removed a few other (truly) deprecated commands in this RC...  There was at some point(s) mention of a separate repository for a few other commands that are either not really actively supported, or really only very sparsely used... maybe this command would fit better in there?

jdtournier:
Removing `shbasis` altogether is worth considering. But maybe we can do this for the official 3.0 release? I.e. leave it in for now in the form that @Lestropie suggests, and then remove it altogether in the full release? I guess it's in the history, if people really need it, they can always install it by checking out the relevant tag...?

Lestropie:
I'd be happy with that.

thijsdhollander:
Sounds like a good strategy! :+1:

PR #1270, opened 2018-03-14
dwibiascorrect: Scale N4 bias field output

### LINKED ISSUE #1265: dwibiascorrect: De-mean
Realised following [this forum post](http://community.mrtrix.org/t/mtnormalise-huge-variation/1509) that this issue hadn't been listed yet.

The N4 algorithm seems to not constrain the magnitude of the constant term of the estimated bias field. This can result in large fluctuations in global intensity scaling across subjects, probably proportional to the severity of the bias field and/or aggressiveness of field optimisation parameters.

Specifically for the `-ants` branch of `dwibiascorrect`, can compute the geometric mean of the scaling factor within the brain mask, and then explicitly subtract that term from the bias field before applying it to correct the DWIs.

Lestropie:
N4's estimated bias field can contain a substantial yet unpredictable constant component, drastically changing the overall signal intensity across all image volumes and all voxels. This change scales the contant component of the estimated bias field in order to remove this effect, by computing the total image integral within the brain mask of the images before and after bias field correction, and then scaling the bias field to be applied by the ratio of these two values.

This should reduce (likely minor) differences in CSD behaviour in FBA studies due to differential scaling of estimated FOD amplitudes, and reduce the discrepancy between reported scaling factors by `mtnormalise`.

Closes #1265.

### DIFF
--- a/bin/dwibiascorrect
+++ b/bin/dwibiascorrect
@@ -107,8 +107,19 @@ elif app.args.ants:
   #   will have a smoothly-varying bias field correction applied, rather than multiplying by 1.0 outside the mask
   run.command('mrconvert mean_bzero.mif mean_bzero.nii -strides +1,+2,+3')
   run.command('mrconvert mask.mif mask.nii -strides +1,+2,+3')
-  bias_path = 'bias.nii'
-  run.command('N4BiasFieldCorrection -d 3 -i mean_bzero.nii -w mask.nii -o [corrected.nii,' + bias_path + '] -s 4 -b [100,3] -c [1000,0.0]')
+  init_bias_path = 'init_bias.nii'
+  corrected_path = 'corrected.nii'
+  run.command('N4BiasFieldCorrection -d 3 -i mean_bzero.nii -w mask.nii -o [' + corrected_path + ',' + init_bias_path + '] -s 4 -b [100,3] -c [1000,0.0]')
+
+  # N4 can introduce large differences between subjects via a global scaling of the bias field
+  # Estimate this scaling based on the total integral of the pre- and post-correction images within the brain mask
+  input_integral  = float(run.command('mrcalc mean_bzero.mif mask.mif -mult - | mrmath - sum - -axis 0 | mrmath - sum - -axis 1 | mrmath - sum - -axis 2 | mrdump -')[0])
+  output_integral = float(run.command('mrcalc ' + corrected_path + ' mask.mif -mult - | mrmath - sum - -axis 0 | mrmath - sum - -axis 1 | mrmath - sum - -axis 2 | mrdump -')[0])
+  app.var(input_integral, output_integral)
+  bias_path = 'bias.mif'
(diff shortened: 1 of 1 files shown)


thijsdhollander:
It's (as mentioned in the issue) indeed (always) an arbitrary approach up to some extent, but this solution serves it purpose of keeping the effects of drift by the algorithm under reasonable control; so happy with it.  I'm wondering though: why only apply it for the bias field (correction) of N4, and not just any algorithm in the `dwibiascorrect` script; i.e. a bit later in the script?

On a related note, and while this pull request is still open anyway: I've just gotten a collaborator who attempted (while constructing their preprocessing pipeline) the `-fsl` approach.  The problem I spotted in the outcome of that, was that the bias field *correction* was only applied in the brain mask, and outside of the brain mask it had either copied the values of the original image (before correction), or at least scaled things quite differently.  Essentially, I clearly noticed the boundaries of the brain mask in the corrected output.  In other words, the same effect you'd get with N4 if you provide the mask to its actual mask option (rather than the weight option we provide it to, to avoid this behaviour).  Maybe we need to fix this immediately as well (even though the `-fsl` option isn't frequently used)?

thijsdhollander:
> Other algorithms may have their own built-in control of global scaling; fast certainly appears to. That scaling may additionally be "superior" to the arbitrary approach I've chosen here.

Yep, fair enough. :+1:  In practice, we currently only offer these 2 approaches anyway; so if their individual outputs are ok (either due to integral scaling or another custom approach), that's fine.

> But it'd be a lot of effort band-aid-ing somebody else's command, especially when we don't really advocate using that algorithm for this purpose anyway

Yeah no, I'd definitely not go there.  However, knowing this now, it actually makes the `-fast` approach useless in that (at least multi-tissue) pipeline.  It's only purpose there is to improve the homogeneity for a better mask estimation afterwards... but this approach makes the data in anything beyond the mask inconsistent with the data within.  I'll make this point more strict in the FBA docs then.  But that immediately begs the question: shouldn't we just do away with the `-fsl` approach in `dwibiascorrect` altogether then...?  I can't see myself advising in *any* scenario, if it works like this.

Lestropie:
It's mostly there because of prior stories of ANTs being difficult to install / manage; but I think those may have arisen due to specific versioning required by Dave's old commands / scripts. In my latest experience it was very easy.

Happy to have a big fat warning on the `-fsl` option. But I also see the benefit of having inferior algorithms available. E.g. To demonstrate that the framework for performing a particular task is not stringently tied to a particular command or approach, and that alternatives can be demonstrated / proposed for integration. And so that users can prove to themselves that one is better than the other by actually running the command with multiple algorithms and seeing the results for themselves, rather than feeling "forced" to use the approach that we have chosen to support (even if it's better understood as what we "advocate" to use). Inevitably, if such a functionality is removed, at some point somebody will ask why we don't simply use `fast` to estimate the bias field given we depend on FSL elsewhere...

thijsdhollander:
A warning might not be a bad idea.  For the FBA pipeline specifically, I've actively mentioned in their this one is not appropriate for the goals the `dwibiascorrect` needs to serve there.

PR #1271, opened 2018-03-15
Fix parsing of -strides option

Lestropie:
Was not updated with other changes in 2258b08 that were merged to `dev` branch in #1211, resulting in `-strides` command-line option being ignored.

This highlights that the current testing suite assesses the underlying image data, but not the image header properties. It might be worth considering what we want to test & where / how.

### DIFF
--- a/core/stride.cpp
+++ b/core/stride.cpp
@@ -59,12 +59,12 @@ namespace MR
           in_max = std::abs (current[i]);
       in_max += desired_max + 1;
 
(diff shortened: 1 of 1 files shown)


jdtournier:
Well spotted.

> This highlights that the current testing suite assesses the underlying image data, but not the image header properties.

Yes, I agree. As always, one of those that relies on someone finding the time...

PR #1274, opened 2018-03-23
run_tests: handle comments in test files

jdtournier:
Simple change. As requested in #1268. 
Allows comments in test files. Anything that after a `#` will be stripped out and ignored during testing.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -72,7 +72,7 @@ EOD
   rm -rf testing/data/tmp*
   ((ntests=0))
   ((success=0))
-  while IFS='' read -r cmd; do
+  while IFS='' read -r cmd || [[ -n "$cmd" ]]; do
     cmd="${cmd%\#*}"
     [[ -n "$cmd" ]] || continue
     echo -n '# command: '$cmd >> $LOGFILE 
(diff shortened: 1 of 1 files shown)


Lestropie:
This skips the commented lines appropriately in the reporting of number of tests performed, but there's an unwanted side-effect: If the final line in the test file does not end with a newline character (which is ill-advised, but occurs nevertheless), this change results in that test being skipped. See e.g. `amp2sh` on this branch, which reports 0 of 0 tests successful.

jdtournier:
Indeed, well spotted. Should be fixed now...

Lestropie:
Couple of tests are currently commented out:

- `mrhistogram`: The test data are not present anywhere in the sub-module online, so I'd need to regenerate it.

- `sh2power`: Gets result that differs too much from `mrcalc | mrmath`. Would have thought after #532 that this would be correct, but the test was disabled in c66b684c; unfortunately without details as to why, but it probably had to do with modifications to `testing_diff_data` using absolute v.s. fractional differences. Would want to be sure there isn't some other underlying issue before re-enabling this.

PR #1276, opened 2018-03-26
Docs to dev

thijsdhollander:
Another one, so far for a very small, though important, correction.  I'll leave this one open for now, should other `doctest` commits still come along before RC3 wraps up, so I don't have to keep creating these pull requests.


thijsdhollander:
Ok, I'm finishing this up now.  As RC3 nears, I'm pretty sure I won't be adding to the docs via `doctest` --> `docs_to_dev` --> `dev` any more (also don't really have the time anyway).  Several collaborators that I've provided wit the updated response function estimation as well as FBA docs haven't had any issues so far, and reported better understanding and success with some of the changes.

PR #1282, opened 2018-04-04
Command-line usage: Warn on likely erroneous option usage

Lestropie:
If one of the inputs to a command-line option starts with a dash character, it is likely that the contents of that command-line entry are intended as its own command-line option, and are not supposed to be an input argument to some other command-line option. If this occurs, the most likely scenario is that the user has not provided enough arguments to the first command-line option, and hence the second command-line option is being 'consumed' as one of the requisite inputs to the first option.


jdtournier:
I like the idea! But I think I'd rather change the warning itself to something like:
```
Value "-other" is being used as one to the 3 expected arguments to option "-opt"; is this what you intended?
```

jdtournier:
OK, just came across an issue with this idea:
```ShellSession
$ mrconvert in.nii -stride -1,2,3 out.nii
mrconvert: [WARNING] Value "-1,2,3" is being used as the expected argument for option "-strides"; is this what you intended?
mrconvert: [WARNING] Value "-1,2,3" is being used as the expected argument for option "-strides"; is this what you intended?
mrconvert: [100%] copying from "in.nii" to "out.niif"
```
Not sure why it comes up twice. But in this case it shouldn't come up at all... Maybe we should reconsider this feature...?

Lestropie:
D'Oh; didn't think of that.
But it should be adequate to just restrict the argument types where it applies; i.e. exclude the test for anything numerical.
I did look into why it popped up twice; it wasn't "erroneous", it just wasn't worth the effort it would require to suppress one of the two.

PR #1288, opened 2018-04-06
tckmap: Fix segfault with -dec -stat_vox mean combination

### LINKED ISSUE #1285: tckmap Invalid Memory Access with -stat_vox mean and -dec
tckmap throws an exception with the following command:
`tckmap -template someImage.nii.gz -dec -stat_vox mean trackFile.tck outputFile.nii.gz`

The message is 
[SYSTEM FATAL CODE: SIGSEGV (11)] Segmentation fault: Invalid memory access

This only occurs if the -dec flag is specified

Lestropie:
Fixes issue introduced in 17e22166. The specific use case of combining `-dec` with `-stat_vox mean` was altered to no longer rely on keeping track of a voxel-wise track density, and so memory for this buffer is no longer allocated; yet the receiving functor would still attempt to write to this buffer.

Closes #1285.

### DIFF
--- a/src/dwi/tractography/mapping/writer.h
+++ b/src/dwi/tractography/mapping/writer.h
@@ -102,8 +102,8 @@ namespace MR {
 
           public:
           MapWriter (const Header& header, const std::string& name, const vox_stat_t voxel_statistic = V_SUM, const writer_dim type = GREYSCALE) :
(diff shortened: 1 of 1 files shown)


Lestropie:
@LeeReidCSIRO: Problem should now be solved with an update on the `master` branch.

PR #1291, opened 2018-04-11
Strides option accepts image

jdtournier:
as [discussed on forum](http://community.mrtrix.org/t/mrtransform-does-not-output-same-data-strides-as-template-volume/1577)


jdtournier:
Test fails with an assert failure:
```
# command: mrconvert mrconvert/in.mif -strides 2,-1,3 - | testing_diff_image - mrconvert/in.mif [ ERROR ]
mrconvert: ./core/app.h:181: vector<int> MR::App::ParsedArgument::as_sequence_int() const: Assertion `arg->type == IntSeq' failed.
```
I'll fix that now.

Lestropie:
To avoid a clash with #1293, the `-strides` option will need to be changed to a `.type_variable()`.

PR #1295, opened 2018-04-12
CI tweaks

Lestropie:
Cherry-picked the implementation of persistent build mode from #1177 (which shows as merged but was removed in the process of resolving with other changes), and activates its use for TravisCI. Let's see if this is a good solution for g++7 regularly running out of memory during CI tests.

PR name has been left vague in part because it looks like a number of PR tests are failing at the shared library step of the build process, so this PR can serve as a CI testbed for trying to sort that one out as well.

### DIFF
--- a/build
+++ b/build
@@ -16,7 +16,7 @@ DESCRIPTION
       $ ./build
 
     If no targets are provided, the command will default to building all
-    applications by scanning through the cmd/ folder. 
+    applications by scanning through the cmd/ folder.
 
     The target executables will be located in the bin/ folder, and the shared
     library (if requested - the default) will be located in the lib/ folder (or
@@ -25,20 +25,20 @@ DESCRIPTION
 
 SPECIAL TARGETS
 
-    clean 
+    clean
        used to remove all compiler-generated files, including objects,
-       executables, and shared libraries. 
-  
+       executables, and shared libraries.
+
     bash
        used to update the bash completion script. Note that automatic updating
        of this script can be enabled at the configure stage, by running
        './configure -dev' prior to invoking './build'.
-  
+
     doc
        used to update the command documentation, so that any modifications to
        the inline documentation in commands can be propagated through to the
        user documentation site.
-  
+
     select name
        used to switch between configs / builds. This stores the current config
        and all compiler-generated files in a folder (called "build.oldname"),
@@ -55,7 +55,7 @@ PARALLELISED BUILD
 
       $ NUMBER_OF_PROCESSORS=1 ./build
 
-OPTIONS 
+OPTIONS
 
     -verbose
        print each command as it is being invoked
@@ -197,7 +197,7 @@ def disp (msg):
   logfile.write (msg.encode (errors='ignore'))
   sys.stdout.write (msg)
   print_lock.release()
-  
+
 def log (msg):
   print_lock.acquire()
   logfile.write (msg.encode (errors='ignore'))
@@ -214,7 +214,7 @@ def error (msg):
   else:
     sys.stderr.write (msg)
   print_lock.release()
-  
+
 
 
 def split_path (filename):
@@ -226,7 +226,7 @@ def get_git_lib_version (folder):
   log ('''
 getting short git version in folder "''' + folder + '"... ')
 
-  try: 
+  try:
     process = subprocess.Popen ([ 'git', 'describe', '--abbrev=0', '--always' ], cwd=folder, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     ( git_version, stderr ) = process.communicate()
     if process.returncode == 0:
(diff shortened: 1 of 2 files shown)


Lestropie:
OK, so, as I thought, there's something funny going on with TravisCI. All of our PRs are failing tests, and they are all failing at the shared library generation step, for Mac-Clang, Linux-Clang and Linux-G++. Unfortunately there is no useful information in the logs: It says that the shared library generation has an error, prints the actual compiler invocation arguments, but doesn't print any details of the error; the job gets aborted by TravisCI due to no terminal output after 10 minutes.

The most recent PR to pass testing and be merged to `dev` was #1290, for which I don't see any way it could influence the outcomes of testing for PRs based on `dev`. I also `diff`'d the logs between the successful and failed builds and there's nothing glaring.

If nobody has any good ideas, either @jdtournier contact TravisCI or I'll do it tomorrow. We'll probably need [CI debug mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/).

jdtournier:
Looks like all the commands are being truncated at byte 1600, and then some random temporary filename is being tacked on for some reason. Here's the 3 failing link commands, edited to skip close the 1600 character mark, across 3 different travisCI jobs (`clang` on Linux, `g++` on Linux, `clang` on macOS), along with the same command, truncated to the same extent, as run on my system:
```
# note things go funny at the 1600 byte mark
#                                         v
# on TravisCI:
clang++-5.0 tmp/core/im ... ore/file/mmap./tmp/tmp86g7x0po
g++-7 tmp/core/exceptio ... mp/core/file/o/tmp/tmp9p5ULp
clang++ tmp/core/except ...  tmp/core/file/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpZZdN5E

# on my system:
clang++ tmp/core/format ... ile/ofstream.o tmp/core/file/nifti1_utils.o tmp/core/image_io/sparse.o tmp/core/formats/pipe.o tmp/core/formats/list.o tmp/core/formats/analyse.o tmp/core/math/stats/glm.o tmp/core/formats/mgh.o -shared -Wl,--sort-common,--as-needed -pthread -lz -ltiff -lfftw3 -o lib/libmrtrix-3.0_RC2.so
```

Not sure why this is suddenly happening... We're nowhere near the maximum command-line length on any OS. 

One option might be to use `gcc`'s `@file` syntax to keep that one nice and short. From the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html):
> `@file`
> 
> Read command-line options from file. The options read are inserted in place of the original @file option. If file does not exist, or cannot be read, then the option will be treated literally, and not removed.
> 
> Options in file are separated by whitespace. A whitespace character may be included in an option by surrounding the entire option in either single or double quotes. Any character (including a backslash) may be included by prefixing the character to be included with a backslash. The file may itself contain additional @file options; any such options will be processed recursively.

I'll give that a crack now...

jdtournier:
Guess that didn't work... But we have a different error message now:
```
ERROR: (396/504) [LD] lib/libmrtrix-3.0_RC2.so
clang++-5.0 @tmp/objlist -shared -Wl,--sort-common,--as-needed -pthread -lz -ltiff -o lib/libmrtrix-3.0_RC2.so
failed with output
tmp/core/file/nifti1_utils.o: In function `bool Eigen::numext::equal_strict<double, double>(double const&, double const&)':
core/file/nifti1_utils.cpp:(.text+0x20): multiple definition of `bool Eigen::numext::equal_strict<double, double>(double const&, double const&)'
tmp/core/formats/mrtrix.o:core/formats/mrtrix.cpp:(.text+0x20): first defined here
tmp/core/file/nifti1_utils.o: In function `bool Eigen::numext::equal_strict<float, float>(float const&, float const&)':
core/file/nifti1_utils.cpp:(.text+0x0): multiple definition of `bool Eigen::numext::equal_strict<float, float>(float const&, float const&)'
tmp/core/formats/mrtrix.o:core/formats/mrtrix.cpp:(.text+0x0): first defined here
tmp/core/file/nifti1_utils.o: In function `bool Eigen::numext::not_equal_strict<double, double>(double const&, double const&)':
core/file/nifti1_utils.cpp:(.text+0x60): multiple definition of `bool Eigen::num/tmp/tmpf2mrvfym
```
That last one is particularly odd...

Lestropie:
> Not sure why this is suddenly happening...

Would have to be a Travis internal change. No other way it could suddenly hit on both Mac and Linux builds within CI.

> That last one is particularly odd...

Not really: that looks exactly like the truncation of command invocations you posted above. The `clang++` text output is being truncated at:
```
                                                                               V
core/file/nifti1_utils.cpp:(.text+0x0): multiple definition of `bool Eigen::numext::equal_strict<X, X>(X const&, X const&)'
```
, and then the temporary file path `/tmp/tmpf2mrvfym` being appended at the end.

Lestropie:
Between this PR and the most recent successful test (#1290), neither OSX deployment target, nor `clang++` version, nor Eigen version have changed.

jdtournier:
OK, seems to have magically fixed itself - at least for #1296. I've restarted the relevant jobs for all the others, hopefully it'll hold for now... 

If this works, we could drop my last two commits, they were really only at attempts at fixing a problem that most likely for some momentary glitch on the TravisCI end.

Lestropie:
> If this works, we could drop my last two commits ...

Done.

I had another thought with respect to this PR: If multiple build jobs fail due to memory limitations, then re-running them all using `-persistent` may not solve the issue, since they may all be run again in parallel. Theoretically a further "persistence" loop could be added to `build`, whereby if no jobs succeed, the number of threads could be e.g. halved and the loop repeated. Hopefully what's already in this PR will be sufficient to make CI happy for now, but this would probably be the next step if in the future that's no longer the case.

Lestropie:
[That got it.](https://travis-ci.org/MRtrix3/mrtrix3/jobs/366417344).

Issue #953, opened 2017-04-04
decode('utf-8') falls over on alternative encodings


bjeurissen:
I came across DICOM data sets that contain non utf-8 meta information, (e.g. accute accents stored as latin1), causing run.command() calls to fall over:

```
'utf-8' codec can't decode byte 0xc9 in position 85: invalid continuation byte
```
As an ugly work-around, I have added ``errors='replace'`` to all decode('utf-8') calls, however, not sure what the best way of properly handling this would be...

jdtournier:
What command is this? I assume it's a script, but I struggle to see where the interaction with DICOM happens... Are you talking about the filename, or the actual contents of the DICOM?

bjeurissen:
None of the currently available scripts are affected, but the problem pops up when one calls dcminfo using the script API, e.g.:
```
...
run.command('dcminfo file.dcm -tag 0010 0010)
...
```

Lestropie:
So it's specifically decoding the `stdout` / `stderr` result from `subprocess` that's falling over, rather than the *file name* (which would fall over at the `POpen()` step)?

I was never happy using `decode('utf-8')` there, but it was the only way to get `run.command()` to work on both Unix and Windows, and on both Python 2 and 3 for each OS. Happy for the addition on the premise that it's tested in all four scenarios.

jdtournier:
Yes, text encoding is a wonderful minefield. I specified `utf-8` since I anticipate this is what most systems and terminals will eventually converge to. But when reading DICOM tags, it's going to be tricky. I have a vague feeling that there might be a DICOM tag that specifies the encoding used in text fields. If that's the case, it might be worth trying to catch this in the DICOM handling code, and return properly formatted UTF8. I think C++11 has better facilities for handling Unicode that we might be able to leverage. But otherwise, I'm not sure how we can possibly handle this in any sensible way, `errors='replace'` might be the most sensible way to go... I'm surprised there wasn't already an `errors='ignore'` there, that's what the `build` and `configure` scripts both rely on...?

Lestropie:
> I'm surprised there wasn't already an `errors='ignore'` there, that's what the `build` and `configure` scripts both rely on...?

[...]

PR #1296, opened 2018-04-13
run.command(): More robust text decoding

Lestropie:
Relates to discussion in #953.

Having thought about the situation a bit, I think this is the best solution in terms of the script library. Even if decoding of DICOM text data were improved as described in #953, there's an external tool invoked by `run.command()` generating data that's genuinely invalid if interpreted as unicode, so the function needs to be robust to that.

This solution works for what I've tested it on thus far; I've got a couple of other cases to test, but would appreciate testing on problematic DICOM data as well.

### DIFF
--- a/lib/mrtrix3/run.py
+++ b/lib/mrtrix3/run.py
@@ -148,7 +148,7 @@ def command(cmd, exitOnError=True): #pylint: disable=unused-variable
         do_indent = True
         while True:
           # Have to read one character at a time: Waiting for a newline character using e.g. readline() will prevent MRtrix progressbars from appearing
-          char = process.stderr.read(1).decode('utf-8')
+          char = process.stderr.read(1).decode('cp1252', errors='ignore')
           if not char and process.poll() is not None:
             break
           if do_indent and char in string.printable:
@@ -181,7 +181,7 @@ def command(cmd, exitOnError=True): #pylint: disable=unused-variable
     if tempfiles[index][0] is not None:
       tempfiles[index][0].flush()
       tempfiles[index][0].seek(0)
-      stdout_text = tempfiles[index][0].read().decode('utf-8')
+      stdout_text = tempfiles[index][0].read().decode('utf-8', errors='replace')
       return_stdout += stdout_text
       if _processes[index].returncode:
         error = True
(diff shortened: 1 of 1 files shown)


jdtournier:
I'm not sure how to handle this... I _think_ the issue here is about the encoding used for filenames (rather than inside a DICOM file)? And that seems rather murky to me (see e.g. [this thread] (https://stackoverflow.com/questions/12237195/how-to-know-character-encoding-of-file-names-depending-on-the-filesystem)). There's the encoding used for the filesystem, and that probably gets translated to the encoding used be the the OS and passed to the application (us). Not sure whether this is in fact what happens though... But it seems the encoding assumed by the system is not the same on Windows (cp1252) and Linux (_probably_ utf8) - no idea about macOS. Based on this, I _think_ this should at the very least be dependent on the OS...?

Lestropie:
There's three separate cases that are all leading to the same type of error:

- Special characters contained within filenames, as reported by Hamed on the forum. Importantly, that only causes an error *after the command has completed*; it's the script library reading and decoding the contents of `stderr` that results in an exception, which is most likely just `mrconvert` echoing the name of the file within a progress bar.

- Special characters stored within subject names inside DICOM data, as reported by Ben in #953. Handling of such could plausibly be improved within the C++ code, but ultimately the problem that was reported was the reading of such a file within a script using the `run.command()` function, which probably means the subject name being stored under a header comments field and then the header contents being echoed to `stderr`.

- Either special characters or bytecode data being detected in `stdout` / `stderr` from any arbitrary command that could be executed by a script within `run.command()`, like reported by Derek on the forum. I was able to reproduce this by convincing the CUDA version of `eddy` to make a segmentation violation (and that's probably what is happening to Derek as well). In this case there's no way of knowing *a priori* what might be received on `stdout` / `stderr`.

This change should deal with all three cases. If OS-based filename encoding is a problem above and beyond this change, I'd need to see that to justify pursuing it.

Lestropie:
While I'm merging this now, being fairly confident that it'll deal with the reported issues in the context of the script library, will need to keep an eye open for any cases where either exotic file names or subject names within DICOMs still cause issues.

PR #1301, opened 2018-04-17
mtnormalise bug fix

thijsdhollander:
In a very unlikely course of events, I happened to spot this one while staring at this bit of code for other reasons.  @rtabbara : looks like our masterpiece still had a little imperfection. :wink: 

For the RC3 release notes; can keep it short: "mtnormalise: bugfix related to outlier rejection"

For future reference: the bug made it a bit less likely for negative outliers to be detected (as they should've), and a bit more likely for positive outliers to be detected (as they shouldn't have).  Even though we intended for this step to be in the log domain, due to a small mistake in the looping, the log was essentially not applied (apart from one voxel, where is was wrong in another way).

### DIFF
--- a/cmd/mtnormalise.cpp
+++ b/cmd/mtnormalise.cpp
@@ -331,8 +331,8 @@ void run_primitive () {
   auto outlier_rejection = [&](float outlier_range) {
 
     auto summed_log = ImageType::scratch (header_3D, "Log of summed tissue volumes");
-    for (size_t j = 0; j < n_tissue_types; ++j) {
-      for (auto i = Loop (0, 3) (summed_log, combined_tissue, norm_field_image); i; ++i) {
+    for (auto i = Loop (0, 3) (summed_log, combined_tissue, norm_field_image); i; ++i) {
+      for (size_t j = 0; j < n_tissue_types; ++j) {
         combined_tissue.index(3) = j;
         summed_log.value() += balance_factors(j) * combined_tissue.value() / norm_field_image.value();
       }


jdtournier:
I must be missing something... I don't see how changing the loops around make any difference to the output...?

thijsdhollander:
Look closer... It's about what's in the outer loop at the end of it... (the log). That (accidentally) wasn't in the current outer loop when they were reversed.

jdtournier:
Sorry for being dense. I'm looking very hard at [the diff](https://github.com/MRtrix3/mrtrix3/pull/1301/files), and there is nothing that looks like what you're talking about in there. It just looks like a straight swap of lines 334 & 335, with no actual changes to their contents..?

Lestropie:
Presumably the difference is line 339 being within the outer loop only. So previously it would add a tissue volume to the result, then take the log, then add another tissue volume, then take the log again, and so on. Whereas now it takes the sum of tissue values, then takes the log. Gotta look beyond the +- 3 lines diff report :-P

thijsdhollander:
Beat me to it. :wink:  Yes, that's it indeed.  Subtle difference though: what you describe, did indeed happen, but only for the last voxel (somewhere in a corner of the FOV).  For the other voxels, it ended up only taking the (weighted) sum.  So for all but that one voxel, the difference before/after the bug fix is essentially the log.  This (log) matters, because of the interpretation of what we regard as an outlier here (it has to be an outlier in the multiplicative sense; hence the log).

Lestropie:
Ah yep, so it would.

rtabbara:
👍good catch

jdtournier:
> Gotta look beyond the +- 3 lines diff report :-P

Words to live by... 

Ok, got it, thanks for holding my hand on this one!

PR #1299, opened 2018-04-16
Fix compiler warnings

Lestropie:
Details & links in commit message.

Closes #1247.

Issue #1247, opened 2018-02-04
Address compilation warnings on Windows GCC7.1


Lestropie:
Example:

```
In file included from ./core/mrtrix.h:36:0,
                 from ./core/cmdline_option.h:27,
                 from ./core/app.h:27,
                 from ./core/command.h:22,
                 from cmd/mrthreshold.cpp:18:
./core/types.h:284:40: warning: optimization attribute on 'uint64_t std::abs(uint64_t)' follows definition but the attribute doesn't match [-Wattributes]
   FORCE_INLINE uint64_t abs (uint64_t x) { return x; }
                                        ^
In file included from C:/msys64/mingw64/include/c++/7.2.0/cstdlib:77:0,
                 from C:/msys64/mingw64/include/c++/7.2.0/ext/string_conversions.h:41,
                 from C:/msys64/mingw64/include/c++/7.2.0/bits/basic_string.h:6159,
                 from C:/msys64/mingw64/include/c++/7.2.0/string:52,
                 from C:/msys64/mingw64/include/c++/7.2.0/stdexcept:39,
                 from C:/msys64/mingw64/include/c++/7.2.0/array:39,
                 from C:/msys64/mingw64/include/c++/7.2.0/tuple:39,
                 from C:/msys64/mingw64/include/c++/7.2.0/bits/stl_map.h:63,
                 from C:/msys64/mingw64/include/c++/7.2.0/map:61,
                 from cmd/mrthreshold.cpp:16:
C:/msys64/mingw64/include/c++/7.2.0/bits/std_abs.h:61:3: note: previous definition of 'long long int std::abs(long long int)' was here
   abs(long long __x) { return __builtin_llabs (__x); }
```

Tried to address in 2a508473, but reverted in f4a96a85.

Need to experiment again to try to find appropriate pre-compiler conditions to use in `core/types.h` that will work silently with all compilers / OS's.

Lestropie:
Boosting this to `3.0_RC3` milestone. Exceptionally frustrating, and hasn't gone away as of `g++` 7.3.0.

jdtournier:
closed with #1304

PR #1304, opened 2018-04-23
Alternative handling for abs

jdtournier:
This is an alternative solution to #1247, which was already addressed in #1299. My concern is that relying on the built-in `std::abs()` calls means casting from unsigned to signed versions of each variant. This is probably fine for most operations, but it is [technically ill-defined](http://en.cppreference.com/w/cpp/numeric/math/abs) if the integer can't be promoted to something that can't handle the precision of the original. This solution means we have our own `MR::abs()` call, and rely on SFINAE to call the built-in `std::abs()` call unless the type is unsigned. This seems to work on Linux, currently building on Windows (no errors so far... :crossed_fingers:)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,11 +46,11 @@ matrix:
       env: py=python2 test=sphinx
     - os: linux
       env: py=python3 test=sphinx
-    #############################################################
-    # Run the check_memalign script; only needs to be done once #
-    #############################################################
+    ###########################################################
+    # Run the check_syntax script; only needs to be done once #
+    ###########################################################
     - os: linux
-      env: test=memalign
+      env: test=syntax
     ##############################################
     # Run PyLint tests; both Python2 and Python3 #
     ##############################################
@@ -114,8 +114,8 @@ after_failure:
       "sphinx")
         cat sphinx.log
         ;;
-      "memalign")
-        cat memalign.log
+      "syntax")
+        cat syntax.log
         ;;
       "pylint")
         cat pylint.log
--- a/check_syntax
+++ b/check_syntax
@@ -102,13 +102,13 @@ if [[ $retval == 0 ]]; then
   echo "OK"
   echo "no issues detected" >> $LOG
 else
-  echo "FAIL"
+  echo "FAIL (see syntax.log for details)"
 
   echo "" >> $LOG
-  echo "Please add MEMALIGN() or NOMEMALIGN macro to the class declarations identified above,
-replace all occurrences of std::vector<> with MR::vector<> (or just vector<>),
-avoid use of std::make_shared(),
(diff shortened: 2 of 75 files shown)


jdtournier:
OK, confirmed this builds with no warning on MSYS2. Checks out OK on macOS as well, according to TravisCI, but can't tell whether there are any warnings, since it's build with `-nowarnings`...

maxpietsch:
:+1: All good in macOS, no warnings and tests pass.

Lestropie:
Figured if you're going to go to the effort of editing every single usage, might as well use CI to make sure it doesn't get used again.

> ... can't tell whether there are any warnings, since it's build with `-nowarnings`...

Should these be `Werror` rather than `-nowarnings`? I'd rather be cleaning up compiler warnings as soon as they arise.

jdtournier:
> Figured if you're going to go to the effort of editing every single usage, might as well use CI to make sure it doesn't get used again.

:+1:

> Should these be Werror rather than -nowarnings? I'd rather be cleaning up compiler warnings as soon as they arise.

I think `-Werror` is a bit OTT... It would be nice to be notified of warnings though. The reason we introduced this flag was that sometimes there were so many warnings that the job output exceeded the max size allowed by TravisCI. One way around this would be to at least have the build script record the fact that warnings were produced (i.e.the gcc stderr/stdout was non-empty), make a note of that in the output, and `cat build.log` so it can be inspected (all warnings are always captured in there). Worth doing?

jdtournier:
> One way around this would be to at least have the build script record the fact that warnings were produced (i.e.the gcc stderr/stdout was non-empty), make a note of that in the output, and cat build.log so it can be inspected (all warnings are always captured in there). Worth doing?

Actually, thinking about this, I don't see that this would be any different from compiling without the `-nowarnings` flag... Maybe that's what we should do?

maxpietsch:
I think the SFINAE fails for MR::Image types. I get warnings for lines 307, 308, and 325 in `fixel2voxel`:

```
cmd/fixel2voxel.cpp:307:29: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
        if (!f.padding() && abs (data.value()) > absmax)
```

Lestropie:
> I think the SFINAE fails for MR::Image types.

[Reproduced](https://travis-ci.org/MRtrix3/mrtrix3/jobs/376650684). Given the compiler hint:
```
cmd/fixel2voxel.cpp:307:29: note: use function 'std::abs' instead
```
, it seems that on OSX there is a function `abs<int>()` defined in default namespace (i.e. *not* in `std::`), which is being selected in preference to `MR::abs<float>()`. Solution can be pursued as part of #1329.

PR #1316, opened 2018-04-30
mrview: Fix memory error in Lightbox mode

Lestropie:
If the lightbox mode is set up to display different volumes, and the focus point is placed within a visible volume that subsequently becomes no longer visible due to a reduction in the number of rows / columns in the lightbox, then a memory violation would occur.

I'm not 100% sure that this is the logically optimal solution: it's not immediately clear to me the purpose of that particular code block, it's updating a "slice" index but checking against "volume indices" for each pane; so perhaps the variable named "`current_slice_index`" is being used for different purposes depending on whether the lightbox is in "Cycle through volumes" mode. But this change at least prevents the assertion failure.


jdtournier:
Quick update before I head off to bed: 

I can reproduce this, there's clearly a bug. I've had a look into it today, and I have to say the lightbox mode generally feels a bit clunky

thijsdhollander:
I think this bit (show volumes using lightbox mode) was originally designed by @rtabbara.  Not sure if he has the time still (and is mentally still into these bits), but he may be able to explain the intended logic there.

jdtournier:
Yes, it was. No problem with the logic, I think I can follow all of that. I'm just simplifying a few things... Shouldn't be long now.

jdtournier:
OK, I think this is good to go :crossed_fingers: 
Please try to break it as much as possible...

maxpietsch:
Focusing on volumes does not work with volume cycle and volume increment 2: images jump by 1 volume

maxpietsch:
I can't find any other problem other than somewhat sluggish performance (100% CPU on a single core) when clicking and dragging the cursor. This is not the case in the other 2D and 3D viewers.

jdtournier:
Thanks for that last fix. What system is being sluggish? Is the sluggishness specific to these changes, or was it already sluggish on `dev`?

maxpietsch:
My ubuntu desktop. CPU load has been close to 100% at least since 3.0_RC2-35-gdefb3f0f. The version on dev is slightly faster than this version. The difference is not a deal breaker at all (on my desktop) but noticeable: when scrolling while clicking at a pace so that one sub-image gets traversed in roughly 3s, then the cross-hair lags behind by about a character's width.

jdtournier:
That's unexpected... Do you have any other tools open? What does the help->OpenGL report about your implementation? Alternatively, the first few lines when running with `-debug`? I'm wondering whether you're running with software rendering...

maxpietsch:
The view tool was open. Might also be a shotty graphics driver. I've had trouble with slow scrolling under high load in sublime text, which other users with nvidia cards have as well under linux, possibly related to [cairo](https://github.com/SublimeTextIssues/Core/issues/803#issuecomment-353606018).

```
mrview: [INFO] GL renderer:  GeForce GTX 760/PCIe/SSE2
mrview: [INFO] GL version:   3.3.0 NVIDIA 378.13
mrview: [INFO] GL vendor:    NVIDIA Corporation
```

screenshot 2018-05-09 21 05 22

jdtournier:
OK, that's odd... It looks like the regular NVIDIA drivers, and on my system (quite similar specs), it's pretty smooth....

Good to merge otherwise? I'll give it a spin on my Win10 laptop, just to be sure.

jdtournier:
Failed to build on my Win10 laptop, but I think that's because of Qt4 (signals are protected in Qt4, public in Qt5, apparently...). This next commit deals with it OK, but there's a very faint chance it might stuff things up a little - would appreciate testing before merging this in... Seems fine on my Win10 laptop though, running on Qt4.

jdtournier:
By the way, @maxpietsch: performance seems fine on my laptop, using an Intel HD 620. Just ran it with a 10x10 split, no noticeable lag at all...

maxpietsch:
All good on macos with Qt 5.10.1.

PR #1320, opened 2018-05-03
mrview Connectome tool: Edge visibility by nodes change

Lestropie:
See commit message for f0ef4d6.

Also cherry-picked commit 3b93cf20 from branch `mrview_connectome_tool_ui_fix` so that it can be cleared & deleted; that fix is to do with a pretty esoteric feature, so no surprise that nobody has come across it.

### DIFF
--- a/src/dwi/tractography/ACT/gmwmi.cpp
+++ b/src/dwi/tractography/ACT/gmwmi.cpp
@@ -41,18 +41,6 @@ namespace MR
         }
 
 
-        bool GMWMI_finder::is_cgm (const Eigen::Vector3f& p) const
-        {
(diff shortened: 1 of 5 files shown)


Lestropie:
If anyone wants to go to the effort of testing this, let me know, otherwise I'll merge it; don't know that many people get this deep into the connectome visualisation capabilities...

jdtournier:
OK, I'm _very_ unlikely to have any time to look into this. If no-one else can have a look, I'm happy for you to merge directly. In general, I'm much more comfortable doing that on the viewer than for the apps that do the analysis proper - bugs in the viewer are not going to affect the results of any study...

PR #1327, opened 2018-05-06
mrhistmatch: New matching modes

Lestropie:
In preparing a presentation and demonstration material for a local connectome course, and trying to make changes to my connectome script to update that to match `3.0_RC3`, and seeing [this forum thread](http://community.mrtrix.org/t/multi-tissue-csd-and-pooled-b0s-across-shells/1593) about relative scaling of *b*=0 volumes between DWI protocols, I decided this functionality might be worth the time.

In addition to non-linear CDF-based histogram matching, `mrhistmatch` can now additionally perform *linear* histogram matching. This can be either a scaling factor only, or a scaling and offset; the former might be better for using *b*=0 images to determine scaling factors between protocols than using the median as [suggested](http://community.mrtrix.org/t/multi-tissue-csd-and-pooled-b0s-across-shells/1593/7) in the forum thread, whereas the latter will hopefully improve capability of registration of T1-weighted images to templates using the squared difference metric.


jdtournier:
Sounds like a good idea, but just a couple of questions:

- Are you hoping to have this merged within `dev` prior to release? I'm hoping to release that today, so it's a bit a last minute... No problem in principle, but:

- I'm struggling a bit following the logic (not spent a great deal of time trying to follow, mind you). It looks like you're creating a sorted vector of amplitudes in the input, and trying to match that to the target vector by matching each entry in input to the closest ranked entry in target, in the least-squares sense...? 

- It might help to separate that out of `main()` and into its own function/class...

- Ideally, we'd need tests for this new functionality... It's not a critical aspect of any pipeline currently, so no big deal at this stage, but still...

Lestropie:
> Are you hoping to have this merged within `dev` prior to release?

Only if it slips in hassle-free; not going to delay the update in any way.

> I'm struggling a bit following the logic ...

You've more-or-less got it right. Each (sorted) input datum maps to a position in the vector of sorted amplitudes of the target image; an interpolated position (& consequently value) is taken to account for input & target images maybe not having the same number of voxels. It then looks for the total least-squares solution. So while the goal is to make the two histograms as close as possible (within the domain of the solution space), there isn't actually any histogram 'binning' in deriving this solution.

> It might help to separate that out of `main()` and into its own function/class

Done.

> Ideally, we'd need tests for this new functionality

Done.

Lestropie:
> switch testing for absolute value differences, since otherwise minor differences in low intensity regions may cause failures.

:rofl: The test is of the regression coefficients, not image intensities :-P

jdtournier:
> :rofl: The test is of the regression coefficients, not image intensities :-P

Yeah, funny. :neckbeard: 

On a different note: I note you're still setting non-finite values to zero. Any particular reason?

Lestropie:
> Yeah, funny. :neckbeard:

My neckbeard is ginger, *thankyouverymuch*.

> I note you're still setting non-finite values to zero.

Good point; fixed.

jdtournier:
Ginger beard. Nice.

All looks good to me! Merge when ready...

Lestropie:
Firstly, a [win](https://travis-ci.org/MRtrix3/mrtrix3/jobs/377087832) for multi-OS CI.

Secondly, it [appears](https://travis-ci.org/MRtrix3/mrtrix3/jobs/377096473#L462) that `clang-5.0` is no longer a package provided with `llvm-toolchain-trusty-5.0` :man_shrugging:

PR #1328, opened 2018-05-08
Dwibiascorrect options

maxpietsch:
Added options to pass arguments to N4. The hardcoded resolution in mm is not appropriate for rodent data.


thijsdhollander:
:+1: Good idea.  `pylint` is still complaining though.  Also, not that it's all *that* crucial or important, but still: the interface (to set the N4 parameters) seems to be slightly non-standard with respect to MRtrix "standards" (insofar those exist; so admittedly subjective).  But I entirely support the intention of the pull request.


-----
Each time this comes up though, I tend to start to question (again) whether the FSL option in `dwibiascorrect` is still relevant.  Every time someone has issues with `dwibiascorrect`, the first thing I naturally ask is what algorithm they selected, and if it's the FSL one, that almost always explains any issue they had.  Its output, which only corrects in the mask and leaves the intensities outside the mask unchanged, is also inappropriate for almost any use of the data in most of our pipelines.  I've got the feeling we have almost more warnings up everywhere, from code to docs, to advice on the forum to deal with this "issue" and essentially tell people to not use that option; compared to the actual length of the code dealing with that. :man_shrugging:  So again, wondering whether `dwibiascorrect` should be simply focusing on interfacing with ANT's N4 only, for clarity sake.  But well, this is completely separate from this pull request.  I'll just keep on bringing it up whenever it comes to mind. :wink:

maxpietsch:
I used the algorithm as prefix for its parameters to make sure there is not going to be a conflict with our options. I separate the nesting levels using periods as in `mrregister`. Happy to change if you can think of a more consistent alternative! The interface of the added options is:
```
Options for ANTS N4BiasFieldCorrection

  -ants.b [100,3]
     N4BiasFieldCorrection option -b. [initial mesh resolution in mm, spline
     order] This value is optimised for human adult data and needs to be
     adjusted for rodent data.

  -ants.c [1000,0.0]
     N4BiasFieldCorrection option -c. [numberOfIterations,convergenceThreshold]

  -ants.s 4
     N4BiasFieldCorrection option -s. shrink-factor applied to spatial
     dimensions
```

thijsdhollander:
Well, the only inconsistency in a way is the use of the dot in there.  But, admittedly, it's quite hard to come up with anything better that doesn't become strange in other ways.  It's also not all that unjustified to have it like this, since it's really a case of trying to unify the interfaces of 2 different softwares (ANTS being the other one), and clearly letting that other interface "shine through" enough.  I'm quite happy to just have it this way.  It's also for more specialised scenarios any way, so for most "casual" users, it won't cause any problems.  Those that look into these parameters will need a better technical understanding any way.  Let's stick with it. 👍

PR #1329, opened 2018-05-09
TravisCI: Use -Werror

Lestropie:
Checking to see if this reproduces the compiler [warning](https://github.com/MRtrix3/mrtrix3/pull/1304#issuecomment-387451181) reported in #1304.


Lestropie:
Lazy fix in 9844dc3. Would be better to SFINAE on classes that have an `operator value_type()` defined, and cast to that before running the relevant `abs()`; but with `-Werror` used in TravisCI from this PR onwards, similar issues shouldn't be able to make it to `dev` at least.

jdtournier:
Can't think of much we can do other than what you've done here... Even with better SFINAE, the global scope `abs<>()` call will take precedence over the `MR` namespace scope version, even after the `using namespace MR;` directive. Happy to merge.

Issue #1223, opened 2017-12-23
Incorrect version tag reported (RC1 instead of RC2)


ansk:
Dear experts,
my `mrtrix` binaries are reporting incorrect version tag (RC1 instead of RC2), despite building of master branch of most recent commit 7d68b748b8af1f078d4d1ffb3f1725bbdad92bb2 .
In my `build` script
`git describe --abbrev=0 --always`
is executed.
Looking to
https://stackoverflow.com/questions/4154485/git-describe-ignores-a-tag
adding `--tags` parameter is showing correct version:
```
git describe --abbrev=0 --always
3.0_RC1
git describe --abbrev=0 --always --tags
3.0_RC2
```
Antonin

ansk:
I have encountered another problem with version reporting: Due to the fact that my most recent build of `mrview` (built against Qt5) does not work in my server with `vglrun` ( XCB error), I temporarily reverted `mrview` to version from August 2017 (built against Qt4), which was working with `vglrun`. Nevertheless, the reported git commit sha even for this older version is the most recent commit I built yesterday. When looking at `about MRView`, the build date is correct (August), when invoking `mrview -version`, build date is yesterday. Commit sha is wrong in both cases (pointing to yesterday build).

As a side note, rebuilding most recent commit against Qt4 resolved my primary issue - it is again working with `vglrun`.

Lestropie:
I'd noticed some unusual behaviour here myself, though it's still unclear to me exactly what's going on.

Running `git pull --tags` before building also solves the problem; which suggests that our standard recommendation for people to update the software is somehow not pulling / resolving tag information.

The [`git describe` documentation](https://git-scm.com/docs/git-describe) suggests that the `--all` and `--tags` options are for enabling use of references other than annotated tags. But all of our tagged versions, including `3.0_RC2`, are annotated tags; hence I'm unsure why either of these options should be required in our case, or why adding these command-line options should fix the problem.

Lestropie:
@jdtournier: Did we want to try to fix this for `3.0_RC3`?

jdtournier:
OK, had a quick look into this, I don't think I'd wrapped my head around fully around this one at the time. There was a bunch of issues going on here, and I'm not sure they're all related. 

[...]

PR #1332, opened 2018-05-09
changes to ensure consistent version and build date

jdtournier:
As discussed on #1223


Lestropie:
Slightly concerned that `git pull --tags` will not do what we want it to.  Indeed the [manual](https://git-scm.com/docs/git-pull) for `git-pull` doesn't even offer a `--tags` option.

On older versions of `git`, `git fetch --tags` *only* pull tags, and not branch updates, and therefore will not bring in the update for anyone currently on `master`; e.g. [here](https://stackoverflow.com/questions/1204190/does-git-fetch-tags-include-git-fetch).

I tend to think the safest recommendation would be something like: `git fetch; git fetch --tags; git pull`. But I don't claim to be an expert here; have just experienced in the past `git` not doing what I expected it to in the presence of a new tag.

Lestropie:
It may also be worthwhile getting the `build` script to remove any shared library file that includes the tag in the file name, similarly to how it removes any old `release/bin/` directory from pre-FHS.

Lestropie:
Another follow-up thought:

There is a bit of ambiguity about exactly what the "build date" refers to, as reflected in the prior behaviour as reported in #1223. With these changes, `build_date` will be updated whenever there is any detected change in the `git` version string; but that doesn't necessarily mean that the command for which `command -version` is being run was re-built on that date, it might just be linking against a shared library that has been re-built more recently.

Therefore I wonder if two separate build dates are required: One for when the shared library is compiled, which can be triggered by a change in `git` version string and concomitant update to `lib/version.cpp`, and one for when that specific command was built. `-version` would then report both.

jdtournier:
Ok, I'll add a note to the instructions on Linux with alternative instructions in case `git pull --tags` isn't supported.

I'll also add an explicit delete if any matching libraries that aren't the one we're about to build.

Regarding the build date: this is a concern I've had since I introduced it. Current behaviour sets the build date within each command's `main()`, so it's currently set to whenever the application `.cpp` file was recompiled (with a minor bug in the version string actually produced, which was set to when `core/app.cpp` was recompiled - fixed here). Ultimately, neither of those dates is robust: you can change the code in the main library without triggering a recompile of the application cpp, you can change lots of code in `core/` or `src/` without triggering a recompile of `core/app.cpp` (or wherever we decide to store it) _or_ the application cpp, including code that might affect the operation of that executable. All of this is fragile.

As far as I can tell, the only thing guaranteed to update the build date in (almost) all cases is this proposal: any change in the git version tag will trigger a rewrite of `core/version.cpp`, which will get recompiled with today's date, and linked into the main library. Even updates to the scripts would trigger a fresh build date. The point being not so much that all the code was actually compiled that day (which nothing but a clean build could guarantee), but that the build script deemed it up to date on that day - it's indistinguishable from having been built that day. 

The only way to 'fool' it is if the main library is updated but individual application executables that should be recompiled are not. For example, if `cmd/mrconvert.cpp` needs to be recompiled, and the user/dev called `./build lib/libmrtrix.so`, or maybe `./build bin/mrcalc`, leaving the `bin/mrconvert` executable out of date relative to the stated build date. Note that its version string would also be incorrect in this case, at least for those parts of the code that are specific to the executable. But this is a rare use case, and I'd only expect developers to end up in this position. 

The only other case I can think of is when making changes to the code: the version string will change as soon as `git` detects changes and the `dirty` label is added to the version string. But further changes won't trigger further updates to that, so if these additional changes are made on subsequent days, the build date will indeed be out of date. But that would be rectified as soon as a commit is made (or a `git reset --hard HEAD`, or similar). So again, that ought to be a transient mismatch, and should only affect developers.

What do you reckon?

jdtournier:
> I'll also add an explicit delete of any matching libraries that aren't the one we're about to build.

OK, this is done. 

As stated earlier, I'm pretty confident with the build date issue, and the git docs strongly suggest that we don't need to update our instructions, so I've reverted them to what they were. So this one is good to go as far as I'm concerned. Any final thoughts before I merge (once testing completes)?

Lestropie:
Trying to wrap my head around this...

Fundamentally, the issue is that it's not really the *compilation date* of a `.o` file corresponding to a `.cpp` file in `cmd/` that we'd like reported, but the *link date* for the executable itself. Only way I can currently see this possibly working is if the build script created a `version.cpp` / `version.h` pair *for each individual executable*, compiled into a `version.o` file unique for each executable, and then on executable linking, providing the relevant `version.o` file for each in turn.

Short of doing that, trying to have a correct date of shared library compilation reported is probably the best thing; but maybe the version reporting text should be more explicit that that is the date that is being reported. And if it's OK for the bulk of users and the dates are slightly funky for developers that's fine by me.

> ... according to the official docs for `git pull`, a `git pull` will bring in all tags that point to things that are being fetched.

I was reasonably confident that I'd observed the reported version string as `3.0_RC1_*` after `3.0_RC2` had been merged and I'd pulled in the latest `master`, and that following a `git fetch --tags` it was rectified (and that the `git fetch --tags` call reported the new tag as having been fetched). Testing this would require pushing a temporary tag to the remote from a test branch on one machine, pulling the relevant branch updates to a second machine (with an earlier version of that branch already on that second machine), and seeing if the tag comes along with it. Maybe what's in my memory was in fact something other than this going wrong; I don't know.

jdtournier:
> Fundamentally, the issue is that it's not really the _compilation_ date of a `.o` file corresponding to a `.cpp` file in `cmd/` that we'd like reported, but the _link_ date for the executable itself. 

Nope, I don't think that what we want either. The library can happily be updated and recompiled without triggering a re-link of the executables - just try making a change to any `.cpp` file in `core/`: you'll note that only the library is re-linked, the executables are left untouched. So any version etched into the executable is fundamentally unable to report changes in the library - and such changes could potentially be quite substantial to its operation.

What we could do though is check for version mismatches between the app and library in the `App::init()` call...?

jdtournier:
About the `git pull` issue: this is straight from the [docs for `git fetch`](https://git-scm.com/docs/git-fetch):

> By default, tags are auto-followed. This means that when fetching from a remote, any tags on the remote that point to objects that exist in the local repository are fetched. The effect is to fetch tags that point at branches that you are interested in. This default behavior can be changed by using the --tags or --no-tags options, by configuring remote.<name>.tagopt, or by using a refspec that fetches tags explicitly.

So if for any reason you didn't have a _local_ `master` branch at the time you issued the `git fetch` or `git pull`, then this says the corresponding tags would not have been fetched. I guess if you subsequently check out the `master` branch, then you won't have the tags at that point. Could that be what happened? Seems a pretty unusual edge case to me...

jdtournier:
> What we could do though is check for version mismatches between the app and library in the App::init() call...?

That last commit performs version checking between executables and main library. I think that nails the issue: any git version change will trigger an update to both command and library versions (both very quick compiles, since only a single string included in the cpp), and a relink of both library and applications, which is again fast. Also checked that if version of library is updated, but not all executables (e.g. by building only `bin/mrinfo`), issuing some other executable will trigger the mismatch.

Merge?

Lestropie:
> > Fundamentally, the issue is that it's not really the compilation date of a .o file corresponding to a .cpp file in cmd/ that we'd like reported, but the link date for the executable itself.

> Nope, I don't think that what we want either. The library can happily be updated and recompiled without triggering a re-link of the executables - just try making a change to any .cpp file in core/: you'll note that only the library is re-linked, the executables are left untouched.

It would need to be this for the executable, *in conjunction with* a separate build date for the library.

> Could that be what happened?

More likely my memory is off with respect to exactly what happened. The help page is pretty clear, so let's proceed on the assumption that's right and I'm wrong. Absolute worst case scenario, we suggest people call `git fetch --tags` if they want the reported version to be correct.

jdtournier:
OK, we'll proceed with the assumption that the git issue is something else, and leave the instruction as is. 

Regarding versions, the commit I just pushed ensures that the versions match, which more or less automatically ensures that the build date is also the same, since any change causes all apps and the library to be re-linked with correct git version and today's date. I personally don't see much point in having two dates - although I see why technically you might want that. In practice, I'd like the build date to be the date that the codebase as a whole was last recompiled - even if that recompilation didn't touch the executables as such, as least I have confidence that the build was last brought up to date on that day. But that would be moot with that last commit anyway, I think.

In other news: looks like [this issue](https://github.com/MRtrix3/mrtrix3/pull/1295#issuecomment-380749494) is back... I'll try re-running [that particular job](https://travis-ci.org/MRtrix3/mrtrix3/jobs/377244798).

jdtournier:
OK, the weird failure to link with truncated command thing is back on TravisCI. I've tried setting stdout to non-blocking, as suggested [here](https://github.com/travis-ci/travis-ci/issues/8920) and [here](https://github.com/travis-ci/travis-ci/issues/4704), but no joy. No trying after clearing the TravisCI cache, see what happens...

jdtournier:
Ok, that didn't work... Not sure what to do at this point.

jdtournier:
OK, I think I know what the problem is...

Based on debugging additions in e224d95:
```
TTY detected - errors will be displayed in paginator
stdout is non-blocking: 0
stderr is non-blocking: 0
```
Problem is not that the IO is non-blocking, but that the build script thinks it's running in an interactive terminal... We check using `sys.stderr.isatty()`, [which is the recommended way](https://stackoverflow.com/questions/6108330/checking-for-interactive-shell-in-a-python-script) as far as I can tell. I note this happen on both Linux and macOS, so it's not specific to the macOS container. But I guess this issue occurs only when the build script produces errors to feed to the paginator (`less -X`). I'll add an option to `build` to disable that, see if that fixes the issue...

jdtournier:
OK, that did the trick, we can now see the error that was generated. Let's see if that next commit fixes things up...

jdtournier:
Ok, that sorted that problem out, but it looks like someone needs to rethink their testing :wink:
```
# command: M=$(bc <<< "scale=9; $RANDOM/16384") && C=$(bc <<< "$RANDOM-16384") && echo $M > tmp1.txt && echo $C > tmp2.txt && mrcalc b0.nii.gz $M -mult $C -add tmp1.mif -force && mrhistmatch linear b0.nii.gz tmp1.mif tmp2.mif -force && mrinfo tmp2.mif -property mrhistmatch_scale > tmp3.txt && mrinfo tmp2.mif -property mrhistmatch_offset > tmp4.txt && testing_diff_matrix tmp1.txt tmp3.txt -frac 1e-2 && testing_diff_matrix tmp2.txt tmp4.txt -abs 10.0 [ ERROR ]
mrcalc: [WARNING] existing output files will be overwritten
mrcalc: [WARNING] qform and sform are inconsistent in NIfTI image "b0.nii.gz" - using qform
mrcalc: computing: ((b0.nii.gz * 0.034668) + -10272)...  [=====]
mrhistmatch: [WARNING] existing output files will be overwritten
mrhistmatch: [WARNING] qform and sform are inconsistent in NIfTI image "b0.nii.gz" - using qform
mrhistmatch: Loading & sorting image data...  [=]
mrhistmatch: Estimated linear transform is: 0.035194397x + -10272.5068
mrhistmatch: Writing output image data...  [==================================================]
testing_diff_matrix: [ERROR] matrices "tmp1.txt" and "tmp3.txt do not match within fractional precision of 0.01 (0.0346679688 vs 0.035194397)

## ERROR: 1 tests failed for "mrhistmatch"
```
I'll restart that job, hopefully it'll come good this time...

jdtournier:
Success at last...

If you guys are happy with this, please approve and merge, and we can get on with the release!

jdtournier:
Ok, that sorted that problem out, but it looks like someone needs to rethink their testing :wink:
```
# command: M=$(bc <<< "scale=9; $RANDOM/16384") && C=$(bc <<< "$RANDOM-16384") && echo $M > tmp1.txt && echo $C > tmp2.txt && mrcalc b0.nii.gz $M -mult $C -add tmp1.mif -force && mrhistmatch linear b0.nii.gz tmp1.mif tmp2.mif -force && mrinfo tmp2.mif -property mrhistmatch_scale > tmp3.txt && mrinfo tmp2.mif -property mrhistmatch_offset > tmp4.txt && testing_diff_matrix tmp1.txt tmp3.txt -frac 1e-2 && testing_diff_matrix tmp2.txt tmp4.txt -abs 10.0 [ ERROR ]
mrcalc: [WARNING] existing output files will be overwritten
mrcalc: [WARNING] qform and sform are inconsistent in NIfTI image "b0.nii.gz" - using qform
mrcalc: computing: ((b0.nii.gz * 0.034668) + -10272)...  [=====]
mrhistmatch: [WARNING] existing output files will be overwritten
mrhistmatch: [WARNING] qform and sform are inconsistent in NIfTI image "b0.nii.gz" - using qform
mrhistmatch: Loading & sorting image data...  [=]
mrhistmatch: Estimated linear transform is: 0.035194397x + -10272.5068
mrhistmatch: Writing output image data...  [==================================================]
testing_diff_matrix: [ERROR] matrices "tmp1.txt" and "tmp3.txt do not match within fractional precision of 0.01 (0.0346679688 vs 0.035194397)

## ERROR: 1 tests failed for "mrhistmatch"
```
I'll restart that job, hopefully it'll come good this time...

thijsdhollander:
> Regarding versions, the commit I just pushed ensures that the versions match, which more or less automatically ensures that the build date is also the same, since any change causes all apps and the library to be re-linked with correct git version and today's date.

So I haven't read the whole avalanche of messages from the last 2 days regarding all this, but I reckon the above quoted passage is exactly what I agree with.  It looks like that was also the final outcome of all of this?  If so, awesome, and I fully agree! :+1: 

> I personally don't see much point in having two dates

I couldn't agree more.  It's confusing for most regular users.

> although I see why technically you might want that.

Maybe, but only really for developers.  The word "technically" in such arguments *almost* always precedes something that's going to go against a great user experience.  Most users don't tend to think "technically", but rather intuitively.  A mix of dates isn't very intuitive, and indeed requires a (imo way too) technical explanation.

> In practice, I'd like the build date to be the date that the codebase as a whole was last recompiled 

Yes, please. :slightly_smiling_face:

Lestropie:
Made some suggestions; I found the `mrtrix_version` and `mrtrix_command_version` not entirely intuitive, so changed to `library_version` and `command_version` (the fact they're in `MR::App` namespace should mean the "`mrtrix_`" part isn't required). Also moved `core/version.cpp` to `core/library_version.cpp` to echo `src/command_version.cpp`.

I've tried tweaking things a little for compilation of external projects:

- If a developer reads from `MR::App::command_version`, they will get the version string of the external project (i.e. a copy of `MR::App::project_version`), rather than the version of those commands provided with *MRtrix3*.

- If the shared library is re-compiled through the build script executed from an external project directory, attempting to run an *MRtrix3*-provided command will fail due to the version mismatch, as expected. Whereas if the shared library is re-compiled by the build script run from the main *MRtrix3* directory, any potential version string mismatch between the *MRtrix3* library and the external project doesn't actually matter and so isn't tested.

- Note that with or without these changes, the paginated help page quotes the *MRtrix3* *library* version alongside the command name at the very top of the page, for both internal and external-project commands. There's still text clarification above the synopsis as to whether the command is part of the MRtrix3 package or is an external project, so this is probably satisfactory as it is; after all, the help page is being generated by the library. Just thought I'd point it out.

Lestropie:
... so ...
Linking the shared library fails on my Windows 8.1 machine with:
```
tmp/core/app.o:app.cpp:(.rdata$.refptr._ZN2MR3App15command_versionE[.refptr._ZN2MR3App15command_versionE]+0x0): undefined reference to `MR::App::command_version'
```
Only way to compile is to add `#include "command_history.cpp` to `app.cpp`. But it wouldn't surprise me if that throws off compilation on other machines...

jdtournier:
OK, I'm not sure I'm following these latest changes, and maybe there's some misunderstandings about how this all fits together. I'm going to give my description of how I see things working, which will hopefully clear up why I don't agree with these changes. 

Before that though: the unresolved symbol link error also happens on macOS, but there's a command-line option to deal with that. I'll come up with a cleaner solution, it's no big deal. 

So the way I envisaged this working is:

- a version string is stored in the main library, previously labelled `App::mrtrix_version`

- a (new) version string is stored in the command itself, which refers to the version of the _MRtrix3_ library that it was build against, labelled `App::mrtrix_command_version` (until the last couple of commits). It is _not_ intended to be used for anything other than checking for library mismatch. If the command is part of the main _MRtrix3_ repo, then that also happens to be the version of the command itself (by definition, since the git hash is the same). 

- a _project_ version string is _additionally_ stored in commands compiled as separate modules, and this is used purely for version reporting. These commands _still_ need to store the _MRtrix3_core library version to check for mismatch (I note [this change](https://github.com/MRtrix3/mrtrix3/pull/1332/commits/c42a3bec03c4eb82139541b8fe3c2ce9034bd35e#diff-ec573ddb349211acb3d6d90777394996R1141) disables version checking altogether for external projects).

- A recompile of the main library _should_ cause a mismatch for external projects that haven't also been recompiled. It's simply not safe to assume that the command will run as expected when making such changes to the library. We've never strived for ABI stability, which would be required if we were expecting to provide guarantees that newer versions of the library were still usable with older executables. I've had plenty of crashes in external modules caused by exactly this type of problem.

So just about the only change I agree with in 324a1a1 and c42a3be is that the label `App::mrtrix_command_version` is not great. It would be better labelled as `App::mrtrix_version_as_stored_in_command` or something like that. Otherwise, the label `App::mrtrix_version` is I think entirely appropriate, that's exactly what it refers to. I know technically that is the version of the library, but it is also the version of all components in the _MRtrix3_ core. So I would prefer to keep it as-is. The `App::project_version` should stay as-is, primarily because the `App::mrtrix_command_version` was never intended to store the command version as such. 

The project version not being displayed in the help page is unexpected though, given [the code](https://github.com/MRtrix3/mrtrix3/blob/c42a3bec03c4eb82139541b8fe3c2ce9034bd35e/core/app.cpp#L220). I'll look into that (and fix the link error issue).

In the meantime, let me know what you think about the above - we clearly need to agree before pushing this out...

jdtournier:
OK, I think those last few changes should be it. Runs fine on my Linux system, for the core itself and for external projects. I'm testing this on my Win10 laptop now, but otherwise, I reckon once testing is done, we merge and release. All good?

jdtournier:
> are you fussed about still having `'-DMRTRIX_MODULE'` despite "module"->"project" in cd8f4b9?

Ah, well spotted. Let's fix that in `dev` once merged - I don't want to delay the release for this. It's invisible to users, and almost invisible to developers too...

I'll merge now, and hopefully release soon after that.

Issue #1339, opened 2018-05-17
Typo in configure script, static option fails


thijsdhollander:
See http://community.mrtrix.org/t/standalone-installation-on-linux-fails/1642 ; confirmed the bug, and confirmed the fix, line 231:

`if noshared: build_options.application ('noshared')`

should be changed to

`if noshared: build_options.append ('noshared')`

Lestropie:
Beyond the hot fix, this highlights that `build` and `configure` should be added to `run_pylint` from #1098.

thijsdhollander:
Hmm yes, that's maybe a good idea indeed.  I haven't done that in #1340 , but being pragmatic, I'm definitely not against adding that in there as well.  @MRtrix3/mrtrix3-devs , opinions?  I'm fine either way; anyone who feels authorised, go and click stuff in #1340, be it after adding (somewhat related) commits or not.  I'm all good with it.

jdtournier:
> Beyond the hot fix, this highlights that build and configure should be added to run_pylint from #1098.

Agreed. Not sure I'm looking forward to what that'll throw up though...

jdtournier:
I've merged the hotfix, the `pylint` tests should also be added, but there's no reason to delay the fixes in #1340 for that.

Lestropie:
Closed by #1340.

Issue #1337, opened 2018-05-16
Untracked file testing/src/project_version.cpp


thijsdhollander:
I haven't properly investigated, so this may just be a by-product of me switching between branches, but anyway: since the release of RC3, I've now properly checked out `master` again, nicely re-run `./configure` and `./build doc` to be entirely up to date, followed by `./run_tests` to, well, run the tests and get the testing data in, etc..

After all that git status brought up `testing/src/project_version.cpp`  as an untracked file, probably generated during the above process.  I haven't been following all the discussion on the version strings end of last week; but I suspect that might have been introduced then...?

thijsdhollander:
...so well, I suppose it still had/has to be added to `.gitignore`?

jdtournier:
> so well, I suppose it still had/has to be added to .gitignore?

Yep, spot on. Sorry, that got left off in the last minute rush...

thijsdhollander:
Ok, no worries.  I've labelled this "bug" for now; it's not really a bug that breaks stuff of course, but more of a bug in the "proper" user experience.  Essentially just trying to indicate that this is "merely a technical thing"; also fixing this wouldn't change behaviour of the software or methods of course.

thijsdhollander:
Fixed in #1340.

PR #1340, opened 2018-05-17
RC3 hot-fixes

thijsdhollander:
The first (lesser and... less lesser) bugs have surfaced.  This pull request should deal with #1337 and #1339.  Doesn't change behaviour of any methods, and the fix to the configure script is actually a bit critical, since it enables static builds again.  So well, beautiful cases of hot-fixes in GitFlow speak; so should probably go straight to `master` ASAP.

Sorry for combining both into a single pull request.  If anyone feels strongly about this, happily split it all up; but I'm just trying to enable a swift resolution of both issues.  Separate commits though, so easy enough to look at separately, and will be nice and verbose in the git commit history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,5 +48,6 @@ build.*
 testing.log
 testing/build.log
 testing/bin/
+testing/src/project_version.cpp
 testing/src/project_version.h
 .__tmp.log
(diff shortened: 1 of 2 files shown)


thijsdhollander:
Tests done, successful.  If anyone gives it the thumbs up review, please don't hesitate to push the merge button along with that! :pray:

PR #1342, opened 2018-05-17
Prevent erroneous command-line usage warning

Lestropie:
In #1282, a terminal warning was introduced in an attempt to catch cases where users failed to provide enough input arguments to command-line options.
This was refined in #1293 to prevent erroneous warnings from being displayed when having an input argument begin with a dash character is entirely reasonable.
This change further refines the logic dictating when a terminal warning should be issued, by identifying the use of piped images with command-line option inputs and not issuing the warning in that circumstance.

### DIFF
--- a/core/app.h
+++ b/core/app.h
@@ -243,7 +243,9 @@ namespace MR
             opt (option), args (arguments)
         {
           for (size_t i = 0; i != option->size(); ++i) {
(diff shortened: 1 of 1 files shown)


jdtournier:
Ah. Does this mean that the use of pipes or tmpfiles generates the warning...? Remarkable this hasn't come up as an issue so far... 

Just wondering what the list of exclusions leaves as types of arguments that will trigger a warning...? 

Also, I think if the type is `ImageIn` or `ImageOut`, it might be better to explicitly check that the argument is a simple `-`, rather than merely starting with `-`.

jdtournier:
> that's already handled by the strcmp() call.

Whoops, sorry, I missed that. All good!

PR #1346, opened 2018-05-21
dwi2response: Fix for absent -mask and no gradient table

Lestropie:
Came across this one myself. Details in commit message of 19eaa28.


Lestropie:
I tested different combinations of `-mask` / `-grad` / `fslgrad` & lack thereof, so should be good.

Issue #1347, opened 2018-05-22
image.Header(): UTF decoding error in JSON library


Lestropie:
As reported on forum [here](http://community.mrtrix.org/t/dwipreproc-eddy-cuda-slspec-unable-to-open-file-multibandgroups/1573/23).

Somehow a non-UTF8 byte is being exported by the `mrinfo -json_all` option. I'm guessing it's something in a comment field or the like; maybe a MGZ that was converted prior to #970, or something like that. Will know more with sample data.

civier:
I think I experience a non-UTF8 in dcminfo -all with data collected at the Florey.
I'm not familiar with the underlying code, but may it be related?

These are the problematic fields:
[DCM] 0008 0016 UI       26      104   SOPClassUID                            [ 1.2.840.10008.5.1.4.1.1.4^@ ]
[DCM] 0008 1140 SQ      306      656 > ReferencedImageSequence                
[DCM] FFFE E000 UN       94      664   - Item                                 
[DCM] 0008 1150 UI       26      672     ReferencedSOPClassUID                [ 1.2.840.10008.5.1.4.1.1.4^@ ]
[DCM] 0008 1155 UI       52      706     ReferencedSOPInstanceUID             [ 1.3.12.2.1107.5.2.19.45056.201712051133052344508765^@ ]
[DCM] FFFE E000 UN       94      766   - Item                                 
[DCM] 0008 1150 UI       26      774     ReferencedSOPClassUID                [ 1.2.840.10008.5.1.4.1.1.4^@ ]
[DCM] 0008 1155 UI       52      808     ReferencedSOPInstanceUID             [ 1.3.12.2.1107.5.2.19.45056.2017120511331296101608773 ]
[DCM] FFFE E000 UN       94      868   - Item                                 
[DCM] 0008 1150 UI       26      876     ReferencedSOPClassUID                [ 1.2.840.10008.5.1.4.1.1.4^@ ]
[DCM] 0008 1155 UI       52      910     ReferencedSOPInstanceUID             [ 1.3.12.2.1107.5.2.19.45056.2017120511331689997308777 ]

Lestropie:
`dcminfo -all` only ever shows the result of handling by the C++ code; which admittedly there are sometimes fields stored within DICOM to indicate the use of alternative encodings, and in those situations our DICOM handling may not be ideal, but it's not something that's ever become sufficiently problematic for us to address.

[...]

PR #1349, opened 2018-05-22
strip() function now also strips null characters

jdtournier:
This should sort out dangling null characters in DICOM fields - see #1347.

Not sure whether this needs to be merged to `master` though. It's a non-critical bug fix, might be better going to `dev`...


jdtournier:
I should point out that most of these changes are cosmetic (update code to C++11 syntax). The only line that matters here is [this one](https://github.com/MRtrix3/mrtrix3/pull/1349/files?utf8=%E2%9C%93&diff=unified&w=1#diff-862ff24ad0989a05134c5351fc903920L165).

thijsdhollander:
> It's a non-critical bug fix, might be better going to dev...

Yes, agreed.

jdtournier:
Ok, rebasing onto `dev`. Note this brings all the hotfixes from `master` into `dev` too.

jdtournier:
Everything also checks out on my local DICOM tests. :+1:

thijsdhollander:
Looks good; I'll hit the button. :+1:

Issue #552, opened 2016-04-15
dwiextract: Remove output image dimensionality reduction


Lestropie:
Change `dwiextract` back to produce a 4D image even if there's only a single volume in the input image.

I think I originally did this because I was getting frustrated with the prevalence of [x y z 1] images going all over the place. But it has to be handled explicitly in scripts, and it's confusing users.

This will require corresponding modifications in both scripts and documentation.

jdtournier:
Relates to #344

Issue #344, opened 2015-09-07
Handle image mismatches when only difference is additional singleton axes


jdtournier:
As discussed l

thijsdhollander:
In this context: I just had a case with an "SH coefficient" file of lmax=0.  Depending on whether the 4th singleton dimension is present or not, Math::SH::check will accept or reject it, even though the actual data is off course exactly the same.

jdtournier:
Yes, that's exactly the kind of daft behaviour I was worried about... But how should this be handled? You'd face the same problem in say Matlab, routinely needing to `squeeze()` or something to massage the data to fit the expectations of the next stage. We could support some form of what Python calls _broadcasting_ (`mrcalc` already does), where missing dimensions are treated as having size 1, but I'm not sure that'll be appropriate in all cases. 

Feeling a bit short of inspiration on this one. Anyone got any brilliant ideas they'd like to share...?

Lestropie:
Not sure that we want to accept _any_ 3D volume as input to _any_ command that expects SH volumes, but clearly the case needs to be handled. My instinctive response is that in this specific case the user could run the file back through `mrconvert` with `-axes 0,1,2,3`, which would add the singleton dimension; and `Math::SH::check()` could make the 'suggestion' that if an input 3D image is intended as an `lmax=0` volume, this step is required. But that may or may not make sense once I read it back to myself...

jdtournier:
I've been thinking a bit about this issue, given the [recent issues users have been facing](http://community.mrtrix.org/t/error-cannot-perform-operation-along-axis-3-image-only-has-3-axes/164), and #552. How about we add a method to the `Image` class, called say `.with_ndim(int)`, which behaves similarly to `.get_image()` and `.with_direct_io()`, in that it returns a new fully formed `Image` with the number of dimensions fixed up if needed to make the image have the requested dimensionality? The idea would be to use it at `Image::open()` time, like this:

~~~c++
auto in = Image<type>::open (filename).with_ndim(4);
~~~

and optionally with `.with_direct_io()` tucked in after that. All it would do is return a valid image with the missing dimensions added with size 1 when the dimensionality of the input is less than expected, and throw an `Exception` if it's higher and the excess dimensions are non-singleton. That should handle case like this OK, right...?

draffelt:
Sounds like a nice solution to me. I've been caught up with singleton dimensions elsewhere in the past.

thijsdhollander:
[...]

Issue #55, opened 2014-06-05
Bootstrap CSD tracking algorithm


Lestropie:
Apparently a number of people have asked Ben whether his bootstrapping tracking algorithm will be implemented in MRtrix. It will undoubtedly be much slower than other FOD-based algorithms, but still, it would be nice to have it there to play with. It may also be not too hard to implement with the Bootstrap class already implemented.

Lestropie:
https://onlinelibrary.wiley.com/doi/epdf/10.1002/hbm.21032

Lestropie:
FYI @bjeurissen: Have a Masters intern student who is going to make an attempt at this.
Feel free to jump in here with any suggestions.

One confound I'm thinking about now is that this issue was listed prior to MSMT. It will potentially be quite tricky to get a command-line interface for `tckgen` that supports use of MSMT in bootstrap tractography. So we might omit that prospect for now and just focus on soft-constraint single-tissue single-shell CSD.

Below is what I think is a reasonable ordering of steps. May edit as we go.

- [ ] Initial addition and compilation of "new" algorithm as duplicate of existing algorithm
    - [ ] Duplicate `SD_STREAM` algorithm with new name
    - [ ] Make new algorithm accessible from the `tckgen` command-line interface
    - [ ] Ensure that `tckgen` compiles
    - [ ] Ensure that the "new" algorithm can be executed from the command-line, and produces the same results as the `SD_STREAM` algorithm

- [ ] Have algorithm perform CSD internally
    Unlike `SD_STREAM`, which takes as input a pre-calculated FOD image, this new algorithm will take as input a DWI series, the CSD algorithm will be executed internally across the entire image, and this will then be used as the input for the `SD_STREAM` algorithm.
    - [ ] Change algorithm code to reflect fact that input is DWI rather than FOD
    - [ ] Add command-line option allowing the user to provide the response function file as an input to `tckgen` (this will be a requisite input whenever this algorithm is used)
    - [ ] Algorithm constructs a single instance of the `MR::DWI::SDeconv::Shared` class
    - [ ] Algorithm performs CSD, storing results in an internal scratch image
    - [ ] Algorithm executes the relevant `SD_STREAM` functions, but using the internally-calculated FOD image

- [ ] Introduction of bootstrapping to new algorithm
    - [ ] Introduce `MR::DWI::Bootstrap` class as exemplified in the `tensor_prob` algorithm
[...]

PR #1352, opened 2018-05-25
fixel2voxel: Fix dimensionality of output image

Lestropie:
When the operator chosen to map fixel data to a voxel grid results in a single scalar value per voxel, the output image should logically be a 3D image. However, due to use of the fixel format index image as the template, many such images come out as 4D. This change explicitly sets the output image dimensionality for each choice of operator.

The test data need to be updated also; but I'm posting here first to see people's opinions on base branch for the merge before I do so. I personally consider this a "bug", since when writing a script wrapping this functionality I had to explicitly pipe the result to `mrconvert -coord 3 - -axes 0,1,2` in order to get what would intuitively seem the "correct" result. But conversely, the command is not fundamentally "broken" in terms of crashing or producing outright erroneous results, and this change does "alter behaviour", which would justify instead going to `dev`... Thoughts?

### DIFF
--- a/cmd/fixel2voxel.cpp
+++ b/cmd/fixel2voxel.cpp
@@ -524,6 +524,7 @@ void run ()
   H_out.datatype().set_byte_order_native();
   H_out.keyval().erase (Fixel::n_fixels_key);
   if (op == 7) { // count
+    H_out.ndim() = 3;
     H_out.datatype() = DataType::UInt8;
   } else if (op == 10 || op == 11) { // dec
     H_out.ndim() = 4;
@@ -542,6 +543,8 @@ void run ()
       // 3 volumes per fixel if performing split_dir
       H_out.size(3) = (op == 13) ? (3 * max_count) : max_count;
     }
+  } else {
+    H_out.ndim() = 3;
   }
 
   if (op == 10 || op == 11 || op == 13)  // dec or split_dir


jdtournier:
I'm not sure I fully appreciate all the subtleties and side-effects of these changes, but it reminds of similar discussions we had on stripping out singleton dimensions in commands that _generally_ produce 4D output, but can sometimes produce 3D output in the form of a 4D image with a singleton trailing dimension - issues #55, #343, #344 & #552 in particular. Do [the concerns I'd raised ](https://github.com/MRtrix3/mrtrix3/pull/343#issuecomment-138291941) about these kinds of changes also apply in this context? Things like [this specific example](https://github.com/MRtrix3/mrtrix3/issues/344#issuecomment-143935196) spring to mind...

Lestropie:
I get the concern, but I don't think they're relevant here. In the `dwiextract` example, the issue was that the output image could be either 3D or 4D depending on the input image. In other examples, the issue was that this outcome could vary depending on a user input to a command-line option (e.g. single value v.s. list). Here, the fixel-to-voxel operator is compulsory, the expected output is unambiguous for all cases (despite the fact that it is different for different operators), and the case where a user could potentially provide either a single value or a list of values as input to an option / argument doesn't apply. Conversely, having an image of "fixel count" that contains 2 volumes, and the second volume is zero-filled, is clearly not intuitive. I don't see a sensible use case where a user would be free to provide either "count" or "split_dir" to `voxel2fixel`, and a script would be constrained to expect the outputs of both of those images to be 4D, which is what would be required for similar issues to those previous examples to arise.

jdtournier:
OK, that's good to hear. :+1:

thijsdhollander:
Ok, (have to admit) I haven't had the time to look into the details, but I've seen problems in scripts (both my own as well as other peoples') being introduced due to this dimensionality aspect.  One important script that at least uses `fixel2voxel` that comes to mind, is the `dwi2response tournier` algorithm, which is quite important (so should definitely not be risked to be broken).  It may be worthwhile to at least test that one, since there's no CI tests for that one.  However, due to (I think? Correct me if totally wrong.) at least a potential for this to break anyone's existing script(s), I think it might at least be safer to rebase this to `dev` for now...?

Lestropie:
`dwi2response tournier` is unaffected since it only uses `split_data` and `split_dir`. 

Anyone wrapping this functionality within a script would need to explicitly rely on the fact that the output of the command is currently inappropriate (i.e. try to access the second zero-filled volume for an operator that one would expect to produce a single volume) in order for the change proposed here to trigger an issue in their script.

But it's behaviour modification nevertheless; so let's go to `dev`. (Just getting that out of the way in order to clear my list)

PR #1353, opened 2018-05-25
5ttgen: Fix printing output of 5ttcheck

Lestropie:
If the text output from `5ttcheck` was a single-line warning, then "`for line in stderr:`" would print a warning line *per word* :-/

### DIFF
--- a/bin/5ttgen
+++ b/bin/5ttgen
@@ -49,7 +49,7 @@ alg.execute()
 stderr = run.command('5ttcheck result.mif')[1]
 if stderr:
   app.warn('Generated image does not perfectly conform to 5TT format:')
-  for line in stderr:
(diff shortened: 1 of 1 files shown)


Lestropie:
No, `run.command()` always returns a single string per stream (i.e. `stdout`/`stderr`). I just used the wrong function from the outset. I suspect my memory of discovering the issue is off; this is just something I found today when dumping out my git stash, but for whatever reason I seem to remember it having worked in the past, so thought maybe it worked OK if there was multiple warnings split by a newline but not for a single one. Would not think `split()` would actually change behaviour like that though; so more likely just my memory is off.

jdtournier:
OK, guess it's all good then!

PR #1335, opened 2018-05-11
fix README.md and macro MRTRIX_MODULE -> MRTRIX_PROJECT

jdtournier:

PR #1354, opened 2018-05-25
Update README.md

jdtournier:
Current info is wrong. This fixes the obvious mistakes, feel free to correct anything I might have left off.


thijsdhollander:
As far as I can see, this makes all good sense.  I can't spot any obvious mistakes...

In practice though, wouldn't we want to combine this with #1335 altogether?  Or am I overlooking something?

jdtournier:
Good point. I've just merged #1335 into here, I'll close that other pull request...

PR #1355, opened 2018-05-27
Fba docs finalfixes

thijsdhollander:
Ok, as I'm now working essentially through the weekends for our workshop preparation; we ran into a finding, and at literally almost the exact same time an external collaborator on another data ran into the same thing; resulting in a Sunday-morning investigation into the template analysis fixel mask for single-tissue FBA change.  I can happily explain over a beer or two at ISMRM, or use half an hour or so during our meeting tomorrow, but in the interest of our workshop data being processed in a standardised manner, and the manual corresponding to it; please believe me: the single-tissue FBA template analysis fixel mask derivation process that still dated from long long long ago is not robust and a bit broken.  I investigated whether it is possible to simply bring it into line with the multi-tissue FBA instructions for that step, and it appears that's entirely feasible and actually quite good even.  Only the threshold (0.06 for 3-tissue CSD vs 0.10 for single-tissue CSD) is customised to either individual pipeline, for obvious reasons.

With the workshop in mind, I would suggest to merge this into `master` ASAP; this should also finalise everything about the FBA docs (for real this time).  If this is in there, we can get the FBA data for the workshop in theory done by somewhere Monday-Tuesday nighttime Melbourne time; pending a fingers-crossed moment for significant results to appear at a significance threshold of 0.05.  If the latter fails, this would leave us with just nearly enough time for a second all-or-nothing attempt (as failing at that stage would mean the data (populations) themselves would have been insufficient; so we need to make strategic changes there and rerun everything; taking a week of time).

If anyone can give the thumbs up ASAP, that would be highly appreciated. :pray:


Lestropie:
Given the recommendation is changing in between tagged updates, should the revised documentation include a note or warning indicating that the change has occurred post-`3.0_RC3`, so that anyone referring to the online document is more likely to recognise the change? And/or should an announcement be sent (maybe just a reply to the `3.0_RC3` announcement topic) to alert anyone for whom the change is relevant?

Lestropie:
> At the moment, those kind of things aren't mentioned anywhere on the blog or in/with release posts (I think?)

I have been including hotfixes that have gone to `master` between tags within the tag changelogs, since they apply for anyone who does a direct checkout of tag versions rather than accepting rolling updates on `master`. Plus it makes the list longer :blush: 

Thus far hotfixes have only really been reported as replies to particular threads where issues have been raised. Additionally putting them a centralised location (either as replies to the tag update announcement, or collected as their own "hot fix" topic) shouldn't take too much work, so I'm for it (even if it only draws even more attention to all of my bugs :-/ ).

thijsdhollander:
> I have been including hotfixes that have gone to master between tags within the tag changelogs, since they apply for anyone who does a direct checkout of tag versions rather than accepting rolling updates on master.

Yeah, makes full sense.  So at least, eventually, they were always documented upon the next tag.

> Additionally putting them a centralised location (either as replies to the tag update announcement, or collected as their own "hot fix" topic) shouldn't take too much work, so I'm for it 

:+1:  When I merge this one, I'll write a reply to the initial tag announcement; short and simple.

If nobody has any objections, can someone give this one the :+1: review it needs to allow for merging?

thijsdhollander:
While the merge is done, I haven't added the reply to the RC3 announcement yet.  As I was almost starting on it, I realised there's been a few other "hot fixes" since RC3, and not everyone may agree yet on how to e.g. format such a reply to the announcement.  I'll open an issue for this to continue discussion, if needed, and to basically not forget to do this (in any agreed format).

PR #1358, opened 2018-05-31
Yet another FBA fix

thijsdhollander:
Recent collaboration activity as well as preparations for the workshop data revealed a few other things; one (major) thing in particular with respect to / caused by RC3's `tckgen` changed behaviour.  The impact of this was quite a bit more drastic than you'd expect; blowing memory requirements way out of proportion, and (should everything not start swapping forever) almost certainly affecting results badly.  This is an important hot fix, both for all our FBA users as well as with the planning of the workshop in mind; so there is some urgency to merge this to `master` for that reason.

This is probably also a good lesson to us all, in that we need to tread more carefully when doing any kind of major release.  Maybe we should be working a bit more consistently on things; not delay and then rush.


thijsdhollander:
> So if I've got this right, the issue is that the default threshold is now slightly too liberal for MSMT-CSD when applied to the template, and quite a bit too liberal for regular CSD?

I can elaborate on Monday, but there is an important added nuance (that matters a lot) here: to use your words and amend the nuance:

the issue is that the default threshold is now slightly too liberal for MSMT-CSD when applied to the template *in the context of a whole brain tractogram for the purpose of CFE in an FBA*

That last bit is what matters.  And for single-tissue CSD (and 2-tissue CSD is *mostly* like that *for this particular aspect*), yes, "quite a bit too liberal" (in the context of that nuance).

Note that the peak thresholds for the fixel analysis mask were already different between both single- and 3-tissue pipelines (and they have to be, for similar reasons); and think about it some more: those peak thresholds (and their difference) should appear again for that same template, when you do the tractography on it, for the purpose of CFE, together with those fixels that exist under the same threshold.  There could be an _added_ nuance to the original nuance that the latter threshold *can be* a bit lower than the fixel mask one, because of reasons that I can explain but not type up now, but that's a side note really.

Note that those thresholds were already in the past somewhat a hint that something was not right in the code (but we never noticed): as the peak thresholds to select fixels from the template weirdly differed, or led to different results, than those used in tractography on the same template; even though both thresholds are defined as acting on the (same) amplitudes.  But the SH bug led them to be evaluated as grossly 35-40% different, compared to how the peak threshold in `fod2fixel` evaluated them.  The latter was correct, the former was the effect of the bug.

...the challenges of long pipelines where everything interacts with everything are never easy or trivial. :wink:

thijsdhollander:
Small addition on this:

>  I'm slightly uncomfortable with different recommendations for the MSMT-CSD vs. regular CSD.

I agree and also share the uncomfortable feeling about that; but then again it's also clear (to me / us) why that is the case.  It's a bit (or a lot) annoying to explain to more casual users; but there's no way around it.  The outputs of the one and the other CSD algorithm differ in a few ways, and the range/distribution of peak amplitudes, even within "pure" WM, is one of those aspects.  I should note (and you're not going to understand that at this point, I think, but I can elaborate on Monday if you wish), that in a funny twist of events, in the FBA pipeline, `mtnormalise` counteracts a small bit of those differences, and (coincidentally) brings both ranges of amplitudes actually a bit closer to each other.  But a significant gap does remain.

PR #1359, opened 2018-06-01
DICOM: Fix slice timing import for mosaic data

### LINKED ISSUE #1252: DICOM slice timing import for mosaic sequences
I've seen [this warning](https://github.com/MRtrix3/mrtrix3/blob/dev/core/file/dicom/mapper.cpp#L215) pop up a couple of times after #1175. There's something about CSA field "`MosaicRefAcqTimes`" that is maybe not quite what I thought it was, or has some kind of confound.

Lestropie:
The wrong member function of class `File::DICOM::CSAEntry` was being used to pre-allocate the memory required to store slice timing, resulting in a mismatch in length between this data vector and the number of slices in the image.

For the sake of web searches, the issue that this fix corrects looks something like:
```
[WARNING] Number of entries in mosaic slice timing (82) does not match number of images in mosaic (60); omitting
```

Closes #1252.

Should probably wait for @jdtournier to run against private DICOM test suite.

### DIFF
--- a/core/file/dicom/csa_entry.h
+++ b/core/file/dicom/csa_entry.h
@@ -88,7 +88,8 @@ namespace MR {
 
             const char* key () const { return name; }
 
-            uint32_t size () const { return num; }
+            uint32_t num_items() const { return nitems; }
+            uint32_t size() const { return num; }
 
             int get_int () const {
               const uint8_t* p = start + 84;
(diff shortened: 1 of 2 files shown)


Lestropie:
Note: This only affects the ability of *MRtrix3* to create the key-value entry "`SliceTiming`" in the image header, which may then be used by `dwipreproc` when instructing `eddy` to use slice-to-volume motion correction. Otherwise, it's just an annoying erroneous warning.

thijsdhollander:
Thanks for looking into it! Should definitely go to master once tested by Donald. It's making a lot of people panic / worry.

jdtournier:
OK, just put this through its paces on my DICOM test suite, it's all good!

Issue #1318, opened 2018-05-03
Inconsistent qform / sform on image write


Lestropie:
Since merge of #1212, I seem to be getting a lot of warning messages related to inconsistent qform / sform, despite those NIfTIs having just been generated using *MRtrix3*. Ideally want to figure out what's going on here before tagging; otherwise there'll be a lot of questions on the forum.

Lestropie:
Some observations:

- NIfTIs generated by either `mri_convert` or `mrconvert`, which `fslhd` report as having identical qform and sform matrices (save for a `-0.0` v.s. `0.0`), yield warnings in *MRtrix3*.

- Using `mrconvert` rather than `mri_convert` to go from MGZ to `.mif` results in a different header transform matrix than does `mrconvert` going from MGZ to `.nii`. If I set `NIfTIUseSform` to 1 however, they converge.

Result: Think the issue is in reading of `qform`...

Lestropie:
Yes, but I don't think that thread is wholly relevant here; apart from having no involvement of `bvecs` / `bvals`, `mrconvert` has behaviour consistent with *its own use* independent of the NIfTI format if reading from `sform`, but behaves differently if reading from `qform`.

Maybe expressed another way, in order to remove the confound of `mri_convert`:
- `mrconvert` MGZ to MIF gives transform T.
- `mrconvert` MGZ to NIfTI then `mrinfo` with `NIfTIUseSform: 0` gives a transform other than T.
- `mrconvert` MGZ to NIfTI then `mrinfo` with `NIfTIUseSform: 1` gives transform T.
- `mrconvert` MGZ to NIfTI then `fslhd` reports that `sform` and `qform` matrices are identical.

Lestropie:
Commenting [this line](https://github.com/MRtrix3/mrtrix3/blob/master/core/file/nifti1_utils.cpp#L198) solves the issue for some test data, but introduces the issue in other data that behaved appropriately without any code modification. So that's the spot. Not sure if the difference is due to Eigen handling quaternions differently to pre-Eigen code.

jdtournier:
OK, [the line you refer to](https://github.com/MRtrix3/mrtrix3/blob/master/core/file/nifti1_utils.cpp#L198):
```
              Q.w() = std::sqrt (std::max (1.0 - Q.squaredNorm(), 0.0));
```
[...]

PR #1363, opened 2018-06-08
Nifti cleanup

jdtournier:
Follows on from discussion on #1318

### DIFF
--- a/core/header.cpp
+++ b/core/header.cpp
@@ -465,12 +465,6 @@ namespace MR
       WARN ("transform matrix contains invalid entries - resetting to sane defaults");
       transform() = Transform::get_default (*this);
     }
-    else {
-      auto R = transform().rotation();
-      if (!R.isApprox (transform().linear(), 1.0e-6))
-        WARN ("transform matrix contains non-rigid rotation - setting to nearest rigid-body rotation");
-      transform().matrix().topLeftCorner<3,3>() = R;
-    }
   }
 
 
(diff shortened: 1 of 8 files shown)


core/header.cpp
```diff
@@ -465,6 +465,12 @@ namespace MR
       WARN ("transform matrix contains invalid entries - resetting to sane defaults");
       transform() = Transform::get_default (*this);
     }
+    else {
```
jdtournier: This bit is not strictly related to this, so might be best not to include... We'd had discussions in the past as to how best to handle non-orthogonal transforms - I guess this might be one way...

jdtournier:
OK, this has thrown up another issue: exceptions thrown by threads in ThreadedLoops were not caught - although the error was displayed. This means that just about all of the tests passed when they shouldn't have, apart from a single lone failure. It's only because of that failure I went to have a look at the log and realised there should have been more... 

I've fixed that, and I'll go and figure out why the others are failing. So far it looks like mismatched transform matrices...

jdtournier:
OK, hopefully that'll fix it...

jdtournier:
@Lestropie, can you give this the thumbs-up when you can? This needs merging ASAP if it's to make it onto the laptops. It's good to go as far as I can tell...

Lestropie:
`Thread::run()` is also used in `dirflip`, `dirgen`, `dirsplit` and `tckglobal`, so a similar effect could occur there (though `dirgen` is the only one of those with a test).

core/header.cpp
Lestropie: I seem to recall Dave having commented proudly at some point about being able to apply affine transformations directly to the image header and being able to operate on those images / even open them in `mrview` without necessitating image regridding. So I don't think this is something that should be "slipped in" without more careful consideration.

jdtournier:
> Thread::run() is also used in dirflip, dirgen, dirsplit and tckglobal, so a similar effect could occur there (though dirgen is the only one of those with a test).

Good catch. Let's make those changes on `dev` for now, they're not so critical. 

About the rigid-body transform:

> I seem to recall Dave having commented proudly at some point about being able to apply affine transformations directly to the image header and being able to operate on those images / even open them in mrview without necessitating image regridding. So I don't think this is something that should be "slipped in" without more careful consideration. 

Agree with removing the rigid body test from this PR, it does need further discussion. But in general, I'm not sure I agree that we should allow non-rigid transforms... I'll open another issue for that discussion.

Issue #902, opened 2017-02-06
Release distribution platforms


Lestropie:
* GitHub releases

* Docker

* Snap

* NeuroDebian

* Tracking download / installation numbers: Make use of whatever mechanisms are in place within individual distribution platforms, but don't bother putting effort into trying to track e.g. unique git clones

* Encouragement to join community forum, regardless of installation mechanism

* Windows: Try to package MSYS2 & QT into single installer, removing unnecessary components to minimize size. No separate Windows 10 Ubuntu Bash install since it doesn't support QT.

jdtournier:
Just a couple of comments here:

- the Neurodebian side of things will be managed by them independently. They've done a great job with MRtrix 0.2, I expect that'll continue just fine with MRtrix3

- On the Windows side: by all means try to reduce the size of the download by removing unnecessary components, but we should make sure the MSYS2 install is nonetheless fully functional. We should also figure out a way to support updates on that platform that don't require downloading the whole thing all over again (it'll be close to a 1GB download!). Best thing would be to set up a pacman repo for this so it can simply be included in the package sources list, and updated along with everything else with a simple `pacman -Syu`. It might also make installation much easier on Windows, we'd just need to get users to install MSYS2, then issue `pacman -Syu mrtrix3`, and they'd be set. No idea how much work that might be though...

bjeurissen:
Regarding windows: I agree that the simplest option for the user would be an installer which bundles msys2 and qt. Stripping the unneeded Qt stuff is easy, but stripping the unneeded msys2 not so much... Also note that this approach should also include python.  I kind of gave up on this route for now because I always ended up breaking some part of the msys2 system and the full msys2 system is just huge... Another option is to let users use their terminal of choice, i.e. providing only the qt and mrtrix binaries in the installer, but this will be a nightmare to support. Should we still want to follow the installer route, we can reuse a lot of the code from the NSIS installer that I created for the 0.2 release.

Creating msys2 packages for mrtrix releases on the other hand is super-easy. I only had to modify the PKGBUILD script that I had lying around for arch linux. If you want to give it a try: https://www.dropbox.com/s/hk7006bgyzhej96/mrtrix3-0.3.16-1-x86_64.pkg.tar.xz?dl=0 . Curious to see if it will work on your systems as well. Ideally, one would test it on a clean msys2 install.

[...]

PR #1372, opened 2018-06-25
ADD: ubuntu 18.04 and centos 6.9 dockerfiles

kaczmarj:
Hello,

This PR proposes adding specifications to build MRtrix Docker images. Containerization will make installation easier for some users, especially on HPC.

Two Dockerfiles are included in this PR. The file `Dockerfile` is meant for users to build when they fork the project and is based on Ubuntu 18.04 (the latest stable, long term support release). This image includes support for the GUI components.

This Ubuntu 18.04 image would be built as follows:

```shell
$ git clone git@github.com:MRtrix3/mrtrix3.git
$ cd mrtrix3
$ docker build --tag=mrtrix3 .
```

The file `Dockerfile.centos6` includes the compile-time dependencies of MRtrix3 on Centos 6.9. If MRtrix3 is compiled on that image, the binaries should work on Linux systems after 2010 (glibc newer than 2.12). I realize that MRtrix3 releases macOS and Windows binaries, and this image can be used to release static, non-gui Linux binaries.

The Centos 6.9 Docker image would built as follows

```shell
$ git clone git@github.com:MRtrix3/mrtrix3.git
$ cd mrtrix3
$ docker build --tag=mrtrix3-build -f Dockerfile.centos6 .
$ docker run --rm -it mrtrix3-build
[in-container]$ ./configure -nogui -static
[in-container]$ ./build
[in-container]$ cd ..
[in-container]$ tar czf mrtrix3.tar.gz mrtrix3
```

Thank you,
Jakub


jdtournier:
Thanks for this, I'm sure it'll be appreciated by users out there (there's already [one request on the forum](http://community.mrtrix.org/t/docker-container-for-mrtrix3/1720)). It's also been suggested before in #446, and we'd already agreed it would be good back then, so it's high time we did something about it... 

Unfortunately, I'm not sufficiently familiar with docker to really review this, but hopefully some of the other team members will be happy to give it the thumbs-up. I'm personally all for this, with my only concern being one of ongoing support (since I'm not familiar enough with the technology to commit to maintaining it). Is this something you're happy to keep up to date when required...?

Strictly speaking, this should go into the `dev` branch, since it's not a bug fix as such. But since there is no scope for interaction with the existing code, I'd be happy to merge this to `master` directly, if everyone else agrees...?

kaczmarj:
@jdtournier - I would be happy to keep these Dockerfiles up to date.

Lestropie:
Hi @kaczmarj,

Thanks for the PR. I've had a little experience with Docker now from developing my [BIDS App](https://github.com/BIDS-Apps/MRtrix3_connectome), so I can help with getting this through; I just want to raise a couple of options, and I have a couple of other post-ISMRM priorities I need to churn through first.

- I've not yet attempted to get UI elements working through a container, though I've at least seen it done. If it is indeed possible to run `mrview` inside the container (bearing in mind the OpenGL 3.3 requirement), then any additional instructions required for users to achieve this should be documented somewhere.

- A couple of crucial processing functionalities in *MRtrix3* are still just wrappers around commands provided in the FSL software package. Therefore without this software additionally being installed, those functionalities would be absent; most importantly, DWI motion / eddy / EPI distortion correction. It may be necessary to include FSL within the containers if they are to provide adequate standalone functionality.

- While I'm not super-confident when it comes to version / tag marking for CI / automated deployment, I think this is something we would want to be 'on top of' before adding Dockerfiles to the repo. Most likely case here, we would want container tags as accessed on e.g. DockerHub to mirror the underlying *MRtrix3* version.

- Given the apparent purpose of the CentOS container, i.e. building static binaries, I think we'd be interested as to whether the GUI commands were omitted for simplicity or due to fundamental limitations. As part of the official release of the software, we would like to have a controlled environment in which to compile binaries for distribution via direct download. A container seems a logical choice for this task, but I don't know whether or not this is in line with your intended design / use of this particular container.

- If the Dockerfiles are included as part of the main *MRtrix3* repository, they probably also need to be put through continuous integration testing. I've only done container-based CI testing via CircleCI, whereas *MRtrix3* uses TravisCI, so this is something I'd need to look into further.

maxpietsch:
@kaczmarj Thanks for the docker files! I am quite new to docker so bear with me if this is obvious. I tried to run mrview from the ubuntu image but can not get mrview to display. Host is ubuntu 16.04.

```
$ docker run -i --tty --rm -e DISPLAY=$DISPLAY mrtrix3
root@4d07dd525f36:/opt/mrtrix3# mrview
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
qt.qpa.screen: QXcbConnection: Could not connect to display :1
Could not connect to any X display.
```

Also, I think the first line of your centos instructions should read `docker build -f Dockerfile.centos6 --tag=mrtrix3-build .` Otherwise it creates an ubuntu image.

kaczmarj:
@Lestropie - Here are my responses to each of your points.

> I've not yet attempted to get UI elements working through a container, though I've at least seen it done. If it is indeed possible to run mrview inside the container (bearing in mind the OpenGL 3.3 requirement), then any additional instructions required for users to achieve this should be documented somewhere.

It can sometimes be tricky to run a GUI through a container. The following works on Linux (Debian testing). I was not able to get this to work on macOS. @maxpietsch - this should answer your question. (And thank you — I will make that edit to the `docker build` command)

```shell
$ xhost +local:root  # allow non-network local display connections
# When running the container, set the DISPLAY variable and mount the tmp X11 path.
$ docker run --rm -it \
    -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
    -e DISPLAY=$DISPLAY \
    mrtrix_container mrview
$ xhost -local:root  # revert permissions
```

> A couple of crucial processing functionalities in MRtrix3 are still just wrappers around commands provided in the FSL software package. Therefore without this software additionally being installed, those functionalities would be absent; most importantly, DWI motion / eddy / EPI distortion correction. It may be necessary to include FSL within the containers if they are to provide adequate standalone functionality.

I did not realize this. FSL can definitely be installed in the container. In this case, I would opt to use a NeuroDebian base Docker image, install FSL through NeuroDebian, and then compile MRtrix3.

> While I'm not super-confident when it comes to version / tag marking for CI / automated deployment, I think this is something we would want to be 'on top of' before adding Dockerfiles to the repo. Most likely case here, we would want container tags as accessed on e.g. DockerHub to mirror the underlying MRtrix3 version.

Yes, it would be good for the tagged containers on DockerHub to mirror MRtrix3 version. But those containers do not have to be uploaded immediately. With a `Dockerfile`, users can build MRtrix3 without having to worry about dependencies for their system. Even better, in my opinion, one can checkout to any commit and build the project at that point.

> Given the apparent purpose of the CentOS container, i.e. building static binaries, I think we'd be interested as to whether the GUI commands were omitted for simplicity or due to fundamental limitations. As part of the official release of the software, we would like to have a controlled environment in which to compile binaries for distribution via direct download (Edit: #902). A container seems a logical choice for this task, but I don't know whether or not this is in line with your intended design / use of this particular container.

I omitted the GUI components because I did not have time to delve into the Qt installation on Centos 6 (though here is [Qt's documentation](https://wiki.qt.io/How_to_Install_Qt_5_and_Qwt_on_CentOS_6) on this). I don't think there are any fundamental limitations. One consideration is that Qt might have to be statically compiled to be part of static MRtrix binaries. If that is not the case, we could include the Qt libraries in the release tarball. I believe FreeSurfer does this.

My intended use case of the Centos 6 container would be to compile MRtrix on an old system, so the binaries could be used on most current Linux distributions. In #902 I think there was talk of compiling several releases for different Linux distributions, but that might not be necessary.

> If the Dockerfiles are included as part of the main MRtrix3 repository, they probably also need to be put through continuous integration testing. I've only done container-based CI testing via CircleCI, whereas MRtrix3 uses TravisCI, so this is something I'd need to look into further.

The containers can be built in TravisCI (they have good support for Docker). At the very least, CI can build the containers. But a more complete testing could include running MRtrix3's tests inside the container.

maxpietsch:
mrview crashes on ubuntu 16.04 as host:
```
xhost +local:root
docker run --rm -it -v /tmp/.X11-unix:/tmp/.X11-unix:ro -e DISPLAY=$DISPLAY mrtrix3 mrview
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
libGL error: No matching fbConfigs or visuals found
libGL error: failed to load driver: swrast
No XVisualInfo for format QSurfaceFormat(version 3.3, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize 0, redBufferSize 1, greenBufferSize 1, blueBufferSize 1, alphaBufferSize 0, stencilBufferSize -1, samples -1, swapBehavior QSurfaceFormat::SwapBehavior(SingleBuffer), swapInterval 0, profile  QSurfaceFormat::OpenGLContextProfile(CoreProfile))
Falling back to using screens root_visual.
Could not initialize GLX

mrview: [SYSTEM FATAL CODE: SIGSEGV (11)] Segmentation fault: Invalid memory access
```

kaczmarj:
@Lestropie - maybe instead of statically compiling MRtrix3 in the CentOS 6 docker container, we dynamically compile and then package with `package_mrtrix`. Those binaries _should_ work on most recent Linux distributions and would include the GUI dependencies. I can update `Dockerfile.centos6` for that if it sounds like a good idea.

@maxpietsch - I'm not sure why that is happening. I will try to debug on Ubuntu 16.04.

Lestropie:
Given the work being invested in release & distribution, we should re-invigorate this discussion.

I think the generation of binaries for distribution is being managed through other means, so we should be able to constrain this discussion to just providing a container within which users can execute *MRtrix3* commands.

Thinking about this fresh, I'm slightly put-off from the idea of providing container files within the main *MRtrix3* repository. In order to be fully functional, such a container needs to include FSL (and ideally ANTs), so it's questionable as to whether or not it's "distribution of *MRtrix3*". What might make more sense is if there's a separate repository within the *MRtrix3* GitHub organisation that contains only the requisite files for container construction: it may then be easier to appropriately advertise this as being "a container for using *MRtrix3*", but that does not contain *only* *MRtrix3*. It should be possible to trigger any extenal sites like DockerHub off of updates to that repository instead of this one. It should also be possible to give write access to such a repository to any community members willing to maintain it.

While this might seem like extra overhead, it might be *necessary* overhead. Containers should check out a specific tag / commit from software repositories in order to be reproducible: if your `Dockerfile` installs `master`, the contents of the container will vary depending on *when* you build it. Therefore, if these files were in the main repository, and somebody were to push a bug fix, increment the patch version and tag, but not explicitly update the container file accordingly, the software version downloaded by the container during construction would not match the software version that the container file shipped with. For these files to reside in the same repository and still be version-matched, we'd need some kind of arrangement whereby either the contents of `Dockerfile` were somehow automatically updated to reflect an upcoming tag prior to creation of that tag, or `Dockerfile` were to read from a text file containing the upcoming tag string and insert that into its `git checkout` call. Unless there's a clean solution to this, delays due to having to perform explicit updates of container files in a second repository may be lesser consequence than introducing version mismatches within one repository due to failure to perform such an update when performing changes to the repository unrelated to containers.

jdtournier:
Fully agree with everything here. Happy to support - but I have very little experience with docker. Would anyone be willing to maintain this? Might be worth a call-out on the forum...?

kaczmarj:
@Lestropie - i agree with many of these points, and i wanted to provide some feedback. i do not see the advantage of having a separate repository for dockerfile-related things. i've outlined my thoughts below.

>Containers should check out a specific tag / commit from software repositories in order to be reproducible: if your Dockerfile installs master, the contents of the container will vary depending on when you build it. Therefore, if these files were in the main repository, and somebody were to push a bug fix, increment the patch version and tag, but not explicitly update the container file accordingly, the software version downloaded by the container during construction would not match the software version that the container file shipped with.

typically the `Dockerfile` should not checkout the code (it should not contain a `git checkout` call for mrtrix3. outside of the container, the developer should checkout to whichever commit and then build the docker image. the `Dockerfile` should copy the project into the container and then compile the software. so if a bugfix were pushed and the container were rebuilt, the code (with the bugfix) would be copied into the container and then compiled. does that make sense?

>In order to be fully functional, such a container needs to include FSL (and ideally ANTs), so it's questionable as to whether or not it's "distribution of MRtrix3"

that's fine to include fsl and ants, instead of only mrtrix3. for example, [fmriprep's dockerfile](https://github.com/poldracklab/fmriprep/blob/master/Dockerfile) installs many neuroimaging dependencies in addition to fmriprep. i don't think there's an expectation that a `Dockerfile` in a project repository only installs that project. the `Dockerfile` is meant to standardize the installation process, so if that process should include installing fsl and ants, that's perfectly acceptable.

>What might make more sense is if there's a separate repository within the MRtrix3 GitHub organisation that contains only the requisite files for container construction.

this is also fine. it's not done as often as having the `Dockerfile` in the source repository, but it is doable. if another repository were created to house docker things, i would suggest having a `Dockerfile` (or a few) in the root of the repository and then attach mrtrix3 as a submodule. you can pin the submodule to a specific commit. at docker image build-time, mrtrix3 sources files would be copied into the docker image and compiled. though i don't see a benefit to this over including a `Dockerfile` in the mrtrix3 repository.

Lestropie:
Thanks for coming back for the reincarnation @kaczmarj :-P

In my own experience, the purpose of a container was not just to provide a *working* version of a software tool within a container environment, but to provide a *fixed & reproducible* version of *all* main underlying softwares, so that the exact same container could be rebuilt at any location at any time. As such, specific versions of the main softwares should be explicitly requested during container build. With what you're proposing, essentially the software version of *the main* software in the container (*MRtrix3*) is obtained by checking out the code corresponding to the version of interest of that software, which includes the `Dockerfile` intended for that version of the software, prior to invoking the container build.

On reflection I think this style of container version control works when there is a single primary software package for which the container is intended. I would however, when installing major external software packages to the container (e.g. FSL, ANTs), install specific versions based on what is available & known to be compatible with that version of *MRtrix3*. Otherwise, if in the future someone were to check out an older version of *MRtrix3* and build the corresponding container, certain commands may not work because of alterations to the behaviour or interface of external packages outside of our control.

Would you like to have a go yourself at adding FSL and ANTs to the container build?

kaczmarj:
glad to be back @Lestropie ! 

>Otherwise, if in the future someone were to check out an older version of MRtrix3 and build the corresponding container, certain commands may not work because of alterations to the behaviour or interface of external packages outside of our control.

this is a great point that i had not considered. in this case, the dockerfile could be stored in some separate repository, or dependencies could be versioned within the dockerfile as well. i have updated the `Dockerfile` to install specific versions of ANTs and FSL from neurodebian. please build with

```
docker build --tag mrtrix3 .
```

of course use whatever tag you desire.

Lestropie:
Mucked around with the container a little; some things to resolve:

- Installing `libpng-dev` now will provide support for #1541 once the divergent code paths are merged.

- Scripts that depend on FSL abort immediately because `FSLDIR` is not set. I don't know if the convention is different for NeuroDebian installations, but usually one executes `$FSLDIR/etc/fslconf/fsl.sh` to get everything set up.

- Some scripts additionally require installation of `fsl-first-data`.

- The FSL installation does not contain normative template data (`$FSLDIR/data/standard/`), which is required for some scripts.

- The FSL installation does not contain `eddy`, which is one of the key components required to make the container functional. It may be preferable to install FSL via their new installation script rather than through NeuroDebian.

For reference, I'm assessing this by going into the container, checking out `dev`, and utilising what was added in #1614 to test both the C++ binaries and the scripts. For the binaries only `mrconvert` fails a couple of tests because of the absence of PNG support, but for the scripts there's a lot of failures that need to be addressed before merging.

kaczmarj:
thanks for the feedback @Lestropie.

>Installing libpng-dev now will provide support for #1541 once the divergent code paths are merged.

i have added this to the `Dockerfile`.

> Scripts that depend on FSL abort immediately because FSLDIR is not set. I don't know if the convention is different for NeuroDebian installations, but usually one executes $FSLDIR/etc/fslconf/fsl.sh to get everything set up.

my fault. i updated the `ENTRYPOINT` to source `/etc/fsl/fsl.sh` (which sets FSL environment variables) at container runtime.

> Some scripts additionally require installation of fsl-first-data.

i added this to the `Dockerfile`.

> The FSL installation does not contain normative template data ($FSLDIR/data/standard/), which is required for some script.

where would these data come from? could it be one of these packages? http://neuro.debian.net/pkglists/toc_pkgs_for_field_datasets.html

> The FSL installation does not contain eddy, which is one of the key components required to make the container functional. It may be preferable to install FSL via their new installation script rather than through NeuroDebian.

i added this to the `Dockerfile`

jdtournier:
Just thought I'd chip in with my 2 cents, following a brief discussion with @Lestropie this morning. My personal preference here would be for this Dockerfile to reside in a different repo if possible, with all dependencies (or least some) listed as submodules (or fetched within the Dockerfile directly). I don't know much about the technology, but some of the concerns I have about a Dockerfile in the main repo are:

- If the Dockerfile resides in the repo, we need to be 100% certain that it is fully functional before we can tag and release - once tagged, it's immutable, we can't modify it. I personally don't feel too confident with Docker, but I would feel much happier knowing that the Docker file can be tweaked post-release if need be, using the exact same version of MRtrix3 itself, but fixing whatever external issues might be specific to the Docker container. 

- If we have issues post-release with the docker container, using e.g. more up to date versions of other packages (FSL / ANTs), then we can't update the Dockerfile if it resides on the same repo since it's tagged and immutable. If it's on a separate repo, we can create as many Dockerfiles as we want, each set to use whatever versions of all the dependencies we might want to target. For instance, we may want to create a Dockerfile for current versions of MRtrix3 using future releases of FSL or ANTs, or even use a different base Linux distribution. I'm not sure how big a deal that is though, but it strikes me that there's a lot of different combinations of the various components that people might legitimately be interested in, and having a single 'official' Dockerfile in the main repo would make that near-impossible. 

- If the Dockerfile is maintained on a separate repo, we can be much more relaxed with permissions on that repo. I'd feel a bit uncomfortable giving someone write permissions on the main repo purely to maintain the Dockerfile, but I'd have no problem at all with them having full admin access on a separate repo within the MRtrix3 organisation, for instance. 

So that's my naive brain-dump... I don't have enough experience with this to know whether any of these concerns are legitimate, so don't hesitate to correct my _a priori_ prejudices if need be!

kaczmarj:
@jdtournier - separate repo is fine. you raise valid concerns.

Lestropie:
If this is to be done using a separate repository, I'll switch the milestone, since it will no longer require components to be in place within the repository at time of release.

jdtournier:
OK, I've just set up another repo to host these filles:
https://github.com/MRtrix3/docker

@kaczmarj: I've granted you admin rights on that repo. Feel free to maintain it as you see fit (within reason!). 

One thing I'm not sure about is whether the name is appropriate. It's fine if we only want to host DockeFiles, but we may also want to host files for other types of containers (singularity, snapcraft, ...).  It might be better named to something compatible with that in the future, e.g. 'containers' or something. I'll leave this up to you - but if we do want to rename it, I suggest we do this now before people actually start to use it.

Lestropie:
When I first started using this stuff, you set up a Dockerfile, and then used [`docker2singularity`](https://github.com/singularityhub/docker2singularity) to convert the container to Singularity. Singularity seems to have been developed since then to [more directly support Dockerfiles and Docker containers in general](https://sylabs.io/guides/3.5/user-guide/singularity_and_docker.html), but they also recommend using their own [definition file](https://sylabs.io/guides/3.5/user-guide/definition_files.html); I'm guessing the latter will provide some extended functionality, and might be consequential in getting `eddy_cuda` working in a container that can be executed on HPC systems.

So `docker` might indeed be too constrictive a name. I *think* SnapCraft could fall within the definition of "container"? Otherwise something like "deployment" could work, as long as the readme file also provides links to the compiled binary downloads.

kaczmarj:
@Lestropie - i agree with naming the repository `containers`.

on one hand, it would be nice to maintain on container specification (i.e., `Dockerfile`), but i guess we could also singularity recipe files. though it's easy to pull images from dockerhub and convert to singularity (the conversion is done automatically by singularity)

jdtournier:
@kaczmarj: happy with `containers` too. You should have the right privileges to rename, let me know if that's not the case.

kaczmarj:
Yes I have write privileges. Thank you. I will work on this next week — school is busy at the moment.

Best,
Jakub

jdtournier:
No problem, take your time!

PR #1379, opened 2018-07-03
New commands: mrcentroid, labelstats

### LINKED ISSUE #1370: New commands: mrcentre, labelcentre
Adding this explicitly since it's come up a number of times in quick succession.

I have a [gist](https://gist.github.com/Lestropie/06afe8c2eedfd48b9d34ab41539d337f) for a stand-alone command that calculates the centre of mass of an image (taking into account image intensities). This can be useful for e.g. reporting MNI coordinates of statistically significant elements upon transformation to the MNI template.

I could quite easily write a similar command `labelcentre`, which would provide the centre of mass of each parcel within a label image, for applications such as [this forum thread](http://community.mrtrix.org/t/node-coordinates-in-mni-space/1657).

Given a few people have wanted the functionality, it makes sense to me to add it to the main repo. Open to suggestions for alternative command names.

Lestropie:
Has come up more than one on the forum and in discussion, so just did it.

Command names are up for revision. Also not sure about explicitly requiring that the space be defined, as opposed to defaulting to scanner space and having a command-line option to switch to voxel space.

Closes #1370.


Lestropie:
Following [this](http://community.mrtrix.org/t/get-areas-and-orientations-of-nodes-in-parcellated-image-for-the-virtual-brain/1741) forum post: Conceivably a command could be written that provides the volume of all parcels in a label image. This could be a separate command; or should `labelcentre` be modified to become `labelstats` and provide both centre-of-mass *and* volume (and potentially anything else)?

maxpietsch:
Just my 2 pence:

Not sure how you feel about it, but having a dedicated binary to calculate the centre of mass of an image might inflate our list of commands unnecessarily? Centre of mass calculation in scanner and voxel space could easily be an additional output of `mrstats`, which seems like the natural command to use.

Also, you could rename `labelcentre` to `labelstats` as well. This keeps the naming more consistent and more statistics could be added later on if needed.

Lestropie:
A concern with embedding in `mrstats` would be discoverability: Previously we had the functionality of `mrhistogram`, `mrdump` and `maskdump` all within `mrstats`, but they were never used because nobody knew they were there.

That code also became too difficult to manage when `mrstats` was [modified](https://github.com/MRtrix3/mrtrix3/commit/c84e45cd4c68a2b3e795433edb0c8cd58d4dfbcd) to enable computation across all volumes rather than being forced to operate on each volume independently; incorporating centre of mass calculation in there might have a similar effect.

jdtournier:
I have to admit, I'm leaning towards @maxpietsch's suggestion here... I hear what you're saying in terms of discoverability, but I'm not sure the names `mrcentre` or `labelcentre` would really help here. I'm sure we've had similar discussions before (and I can't remember what I argued for...), but I was personally quite happy having the functionality of `mrhistogram` and `mrdump` in `mrstats`, I think they make a lot of sense there. I'm slightly concerned about having too many commands - and I think if people know about `mrstats` and have used it before, it's not a bad first port of call for this type of operation...

The handling of labels is a bit trickier, since there's quite a few operations now that are specifically tailored for this kind of image. From that point of view, it makes sense to call it `labelcentre`, but equally, if we did want to amalgamate the CoM calculation into `mrstats`, we could similarly add generic capabilities to handle label images with an additional `-labels` option. So getting the CoM of all labels would look like:
```
mrstats -labels parc.nii -output com 
```
But I agree it's not necessarily as immediately obvious as `labelcentre` - but it's something that we should be able to document relatively easily, so at least it would come up on a Google search... 

Otherwise, the `labelstats` idea is also appealing, it leaves the door open for further operations.

> That code also became too difficult to manage when mrstats was modified to enable computation across all volumes rather than being forced to operate on each volume independently; incorporating centre of mass calculation in there might have a similar effect.

Yes, I can see that might be the case. But equally, if we have two apps for this currently, it should be trivial to merge into the one app and branch as early as necessary to handle things as expected - it would be just as clean in terms of the code, and if anything, reduce some duplication...?

Lestropie:
Personally I place the power of discovering capabilities based on command names at a higher precedence than the outright number of commands. Having a set of commands with the prefix `label*`, but also having a functionality specifically tailored for label images hiding within `mrstats -labels`, feels like a backwards step to me. `labelstats` to me works: It's somewhat similar to `mrstats`, but has a requisite assumption of a specific form of input data for the types of statistics it is calculating. I do however need to modify it to both report masses, and also better deal with images where the labels are not contiguous.

`mrcentre` I'm not wholly sold on, but a standalone is still my preference. The calculation requires more information than is currently fed through the statistics calculation functor; so either that functor would need to also receive voxel locations (which would reduce the generality of that statistics code), or the centre of mass would need to be generated through a completely different code branch. There's also the question of reporting. I don't think that the centre of mass should be reported to the terminal under default usage: for instance, running `mrstats` on a fixel data file is an entirely reasonable & not uncommon usage, yet this measure is meaningless for such data. So then the centre of mass would need to only be available via the `-output` option. But currently, *all* of the `mrstats` command-line outputs are available via the `-output` option; so adding the centre of mass to the list of choices would break this one-to-one mapping. This to me feels like an unnecessary obscure solution.

jdtournier:
Yep, I'm guessing this is one of those things where we're going to have to agree to disagree... There's going to be no clear consensus, and no overwhelmingly compelling case either way. I'm happy enough with `labelstats`, it matches our naming convention, etc. 

As to `mrcentre`, I hear what you're saying. I'd also assumed it would probably be available if explicitly requested by `-output`, which would indeed make it more difficult to find. But I think this is where #1394 would come into its own. 

I'm also not 100% convinced by the argument that because it would need to operate differently from the current implementation of `mrstats`, it would then warrant a different command - there's no additional burden or overhead to a branch in the same command (the discoverability argument is more compelling). But given that you have to deal with these issues more than I do, I'm happy enough to leave this up to you. However, the name still irks me - I'm not convinced I'd immediately figure out what `mrcentre` does out of context...

Lestropie:
I'm also not wholly convinced by the naming of `mrcentre`, I just haven't come up with a better alternative yet. Plus there's always the command synopsis for finding out what it does.

To add to the anti-`mrstats` argument: 

- `mrstats` permits multiple uses of the `-output` option. Therefore it'd be possible to request the centre of mass in addition to other statistics. But the centre of mass is three values rather than one, breaking the mapping from `-output` command-line option ordering to output data column. You could comma-delimit the centre of mass while retaining space-delimiting for everything else to fix that, but that would then be inconsistent with everywhere else in *MRtrix3* that outputs vector / matrix data space-delimited.

- With what's currently `mrcentre`, the centre of mass can be reported in scanner space or voxel space. While I'd rather have it default to scanner space and have voxel space available via command-line option, it's nevertheless a command-line option that would apply to only one specific functionality within `mrstats` if the functionality were moved there. That sort of option is fine to have for nuanced behaviours that apply only to a specific aspect of a command when those aspects are well-integrated, but when the functionalities are already seemingly disparate it provides greater justification for instead keeping them separate.

jdtournier:
Yep, I can see all of the arguments here. Like I said, this isn't going to be a case of right and wrong... 

I reckon since you'll be maintaining that functionality, it's probably best if you take that decision. I think there's enough consensus anyway. My only slight issue is the `mrcentre` name itself; I reckon you might as well call it `mrcentreofmass` to make it properly unambiguous...

thijsdhollander:
Bit late to the party, but just my 2 cents: this really feels like something that shouldn't be a separate command to me.  Thematically, I reckon, it fits much better with the purpose of `mrstats`.

@jdtournier , separate to that, about `mrdump` functionalities having been split off; I feel that's still fine though.  For a simple dump, I'd not per se go looking in something that has "`stats`" in its name.  Nothing is really computed or being summarised in a dump, so I'd feel no "stat" is output... if that makes sense.  However, for a histogram generation then again, I probably *would* go looking in `mrstats`.  Given someone's time and motivation, we could even add quartiles and outliers to `mrstats` (i.e. combined with the median, essentially everything needed for a box plot).  That not all can/will be shown as a default output makes sense as more "stats" would be added to `mrstats`.

PR #1381, opened 2018-07-03
Docs: New page on handling external modules

Lestropie:
As raised on [forum](http://community.mrtrix.org/t/node-coordinates-in-mni-space/1657). Apart from providing my own Gists, which may become more frequent, I think this is an important detail to have visible in the docs for when the *MRtrix3* publications are sent out.


dchristiaens:
Fully agreed!

Lestropie:
Have added a more explicit walk-through for the common use-case of a single `.cpp` file in the hope it results in less support requests. If anyone has other suggestions, go for it.

Also raising whether this should go to `master` since it's instructions that may be worth having visible now.

jdtournier:
I've added this branch to the mrtrix3-dev readthedocs site for easier reviweing - [direct link here](https://mrtrix3-dev.readthedocs.io/en/docs_module_instructions/tips_and_tricks/external_modules.html). 

Looks good, but main concern is how to handle the path... If you follow these instructions to the letter, you'll still be unable to run the commands. I'm not sure how to handle this. Maybe we can modify the `set_path` script to _also_ add the module's `bin/` folder to the path, in addition to the core MRtrix3 one - but that'll be fiddly. The current script wholesale replaces the line in `~/.bashrc` if it doesn't match what it was expecting, and as it currently stands, would replace the core MRtrix3 folder's entry with the module's entry - it wouldn't _add_ as such. We could fix that by detecting when we're running within a module, to make sure we also list the core, but that would screw up when multiple modules are installed... I guess one option is to modify `set_path` to detect when it's running in a module, in which case it would add another entry altogether with a different marker (the comment line that immediately precedes our `export` entry) specific to the enclosing folder's basename, so that subsequent runs wouldn't match the core, and only match that module. 

Anyway, food for thought...

Lestropie:
> If you follow these instructions to the letter, you'll still be unable to run the commands.

But:

> ... You can then invoke such commands either by providing the full path to the executable file, or ...

I don't think we should be going down the path of automatically manipulating user config files based on their having downloaded an external module that may not even have come from one of the core *MRtrix3* devs. If someone can obtain an external module and set up a softlink, they should be capable of understanding how `PATH` works, and either updating their own config files or using the full path to the executable.

The docs page could be updated to more explicitly state something like "just typing the name of the command won't work, because the location of your newly compiled executable file won't appear in your `PATH`, and therefore your system won't know where to find the executable with that name; here are your options for dealing with that".

jdtournier:
OK, maybe I didn't express that right. Two things: 

- main thing is that there needs to be some information on that page about the fact the PATH needs to be set. 

- secondary consideration is do we want to make it easier for users to do that. Clearly we can tell them to  set their own PATH, and that's how we rolled until recently. What I was suggesting was that we could amend the `set_path` script to also handle adding the location of the current module to the PATH. I'm _not_ suggesting this should be done automatically (I'm not even sure that's possible...?), all I'm saying is that we could suggest to users to run the `set_path` script from within the module folder once they've built it, and it would add the corresponding `bin/` folder to the PATH. 

The instructions would then boil down to:
```
cd mrtrix3/
git clone https://github.com/someone/module.git
cd module/
../build
../set_path
```
Currently the only non copy-pastable bit is that last path modification...

Lestropie:
> But I think it would be better to make that more prominent ...

Done.

> we could relatively easily simplify the process for users.

Potentially. But as you said earlier, you'd need to deal with the prospect of multiple modules; I haven't thought through the whole process to decide how much complexity that would introduce. I'm not against it being done, I just don't want to put my own time into it given that currently it's still fairly rare for non-developers to use this capability.

jdtournier:
Looks good, thanks. 

Check out #1401 for a potential addition to `set_path`. Pretty simple really, all it does is simplify the process of updating the PATH for the vast majority of users. Again, not a strict requirement, but it's such a simple modification of the original script, I guess why not...

jdtournier:
And while I was at it, I also had a crack at putting together a [little script for module handling](https://gist.github.com/jdtournier/1550eb4a830239e1cd278a962634679f). Would allow things like:
```
./module install jdtournier/mrtrix3_extras
./module update jdtournier/mrtrix3_extras
./module remove jdtournier/mrtrix3_extras
```
and takes care of cloning, building, and setting the PATH.

Only thing that bugs me at the moment is that it won't cleanup the `~/.bashrc` on removal. But that could be handled relatively easily by modifying `set_path` to accept a `remove` argument.

PR #1385, opened 2018-07-04
mrtrix3.file.delTemporary(): Compatibility with list input

Lestropie:
Had this floating in a feature branch. Should also check to see whether it can simplify the use of this function in existing scripts.


Lestropie:
`dwipreproc` tested OK.

PR #1386, opened 2018-07-04
mesh2voxel improvements

Lestropie:
- Various improvements to the internals of `mesh2voxel` that were necessary to get the algorithm to yield accurate results when applied to more complex surface shapes.

- Multi-threads the bottleneck part of the algorithm.


Lestropie:
Small point on this one for historical purposes: While the outcomes of `mesh2voxel` are altered quite a lot by these changes for complex geometry, the surface provided in the `test_data` repo is sufficiently simple that the output of `mesh2voxel` for that particular file is in fact identical; hence the test data didn't need to be updated here.

PR #1387, opened 2018-07-05
CI for Windows using AppVeyor

Lestropie:
Need to create a PR to hopefully set up AppVeyor to automatically build requests.

If history is anything to go by, will likely require a lot of tinkering before anything actually works...


Lestropie:
Made some progress, but have gotten stuck on one part in particular.

In `tckconvert`, function `File::create_tempfile()` is used in a couple of formats for writing geometry information to multiple files before combining them in a later step. However, seemingly no matter what I do, on MSYS2 these files are never created with more lenient permissions than 0644, and the subsequent `File::unlink()` call fails with permission denied. This results in one failed CI test of the RIB format; but more generally, converting to either this or the PLY format will currently throw an exception on Windows, as will any future functionality relying on this function.

In addition to replacing numerical 0666 with `(S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH)` in the `open()` call, I've also tried calling `umask()` beforehand (could only make resulting files *more* restrictive, not less), and calling `chmod()` on the temp file after creation. No success.

jdtournier:
> seemingly no matter what I do, on MSYS2 these files are never created with more lenient permissions than 0644, and the subsequent File::unlink() call fails with permission denied.

Confirmed here. Lots of other failures, mind you, but not altogether unexpected on Windows... Not sure what's going on, I can remove the file just fine with a regular `rm` call. I think this may be related to Windows not allowing files to be deleted if they're still in use. We'll have to see whether the order of operations is right for Windows to actually operate as expected...

jdtournier:
>> seemingly no matter what I do, on MSYS2 these files are never created with more lenient permissions than 0644, and the subsequent File::unlink() call fails with permission denied.
> 
> Confirmed here. Lots of other failures, mind you, but not altogether unexpected on Windows... Not sure what's going on, I can remove the file just fine with a regular rm call. I think this may be related to Windows not allowing files to be deleted if they're still in use. We'll have to see whether the order of operations is right for Windows to actually operate as expected...

OK, I think I got the bastard... As I'd suspected, these files weren't closed by the time of the `unlink()` call. It wasn't obvious from the code, but there is a code path where that can happen, and this is obviously where it was going during testing. In case you're wondering, the fix is in commit 272f05e - the rest is cosmetic or unrelated changes that I made while trying to figure this out that seem to make sense.

Lestropie:
:tada:

jdtournier:
Just looking at the `appveyor.yml` file as it stands, a few thoughts:

- do we really need to test under python2 & python3 on Windows? Our instructions currently lead to a python3-only install on MSYS2...

- Might it be easier to add a script in `testing` that performs all of the steps, and pushes artifacts as necessary? I'm pretty sure you can invoke Windows executables directly in MSYS2, so invoking `appveyor` directly shouldn't be an issue

- I would also suggest we push all artefacts by default if any step fails. The `configure.log` might for instance be useful to figure out a failure in the build, etc. 

I might give this a go if you don't mind...?

Lestropie:
> Our instructions currently lead to a python3-only install on MSYS2...

If everyone followed the installation instructions to the letter we wouldn't keep getting complaints from people who are invoking the wrong version of the shell -.-

> Might it be easier to add a script in testing that performs all of the steps, and pushes artifacts as necessary?

I toyed with this approach; however with AppVeyor actually receiving terminal output with the extra command-line options to `msys2_shell.cmd` this should be less of an issue.

> I would also suggest we push all artefacts by default if any step fails.

Same as above.

> Any chance you could give me admin access to AppVeyor...?

As in #1427 (sorry I'm doing my e-mails most recent first :-/ ): If you can't get to this using GitHub credentials we might need to re-map the repo to an explicit "MRtrix3" AppVeyor user account. This is the extent of permissions control I currently have within AppVeyor:

appveyor_permissions

PR #1384, opened 2018-07-04
fixelcfestats: Remove erroneous warning

Lestropie:
When constructing a scratch fixel mask due to non-use of the `-mask` option, give sensible values to the voxel sizes and transform matrix so as to suppress terminal warnings.

Eventually I would prefer to modify the fixel format handling more deeply so that header information such as spacing and transforms are ignored entirely for ndim < 3 (e.g. for fixel data files). But the change in this PR should at least reduce user feedback about irrelevant warnings.

Closes #1102.

PR #1388, opened 2018-07-05
run.command(): Fix app cleanup with exitOnError=False

Lestropie:
In rare instances, `run.command()` may be run with the argument `exitOnError=False`, allowing the underlying command to fail without aborting the script. When this occurs, only a warning-level message is issued. However, due to misplacement of code, such a circumstance would lead to the temporary directory not being erased on script completion, even if the script goes on to completion. This fix ensures that forcing the temporary directory to be retained only occurs if the underlying command both fails and is not permitted to fail.

### DIFF
--- a/lib/mrtrix3/run.py
+++ b/lib/mrtrix3/run.py
@@ -203,8 +203,8 @@ def command(cmd, exitOnError=True): #pylint: disable=unused-variable
   _processes = [ ]
 
   if error:
-    app.cleanup = False
     if exitOnError:
+      app.cleanup = False
       caller = inspect.getframeinfo(inspect.stack()[1][0])
       script_name = os.path.basename(sys.argv[0])
       app.console('')


jdtournier:
... apart from the AppVeyor tests. Relevant error message is:

> The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.

Lestropie:
AppVeyor tests will disappear upon merging of #1384 and then merging `master` into this branch. I instructed AppVeyor to only run tests for those branches that contain the requisite configuration file in the repository, but then failed to hit the "save" button. So there's a few PRs that will show failing AppVeyor tests, but as soon as the tests are triggered again, AppVeyor should disappear (for all but #1387, obviously) now that the setting has been saved properly.

Issue #1390, opened 2018-07-10
./configure -static fails with traceback


tutufan:
Despite the documentation, the `-static` flag seem not to work (at least in RC3).
```
./configure -static -nogui
Traceback (most recent call last):
  File "./configure", line 231, in <module>
    if noshared: build_options.application ('noshared')
AttributeError: 'list' object has no attribute 'application'
```

Lestropie:
Hi Michael,

That one in particular was reported in #1339 and fixed in #1340; however while the hotfix is on `master`, it came after the `3.0_RC3` tag. So if you're determined to get a static compilation, you'll need to check out a version of the code that includes the fix. We've lately tended to move away from static compilation in favour of other solutions, which is probably how the mistake crept in there unnoticed in the first place.

Rob

tutufan:
Thanks Rob.  No, I'd definitely prefer _not_ to use a static build--I was just grasping for any way at all to get it built on our system (RHEL 7.5 plus gcc 7.3).  After a lot of hacking about, I discovered that the configure/build scripts' rewriting of `-I` compiler flags to `-isystem` was triggering an obscure GCC bug.

Lestropie:
Interesting: I've had issues myself in the past with `-isystem`, but was under the impression that it was something specific to Windows. Are there any pages you can provide a link to describing the bug?

jdtournier:
> I was just grasping for any way at all to get it built on our system (RHEL 7.5 plus gcc 7.3)

I've had to do this once, and vowed never to do it again... RHEL is simply not designed for scientific computing, it's difficult to get many of the dependencies installed. I feel your pain.

> After a lot of hacking about, I discovered that the configure/build scripts' rewriting of `-I` compiler flags to `-isystem` was triggering an obscure GCC bug.

This is something you could probably get rid of if necessary - it's only there to prevent warnings from external headers that don't relate to our code as such. There was a time when the output of `./build` was full of compiler warnings from Eigen, which made it very difficult to identify warnings that we could actually do something about. Unfortunately, the `pkg-config` file shipped with Eigen uses a `-I` flag, which is why I'd resorted to modifying that in our scripts. You'd just need to [get rid of these lines](https://github.com/MRtrix3/mrtrix3/blob/master/configure#L526-L528), and maybe [these lines too](https://github.com/MRtrix3/mrtrix3/blob/master/configure#L1192-L1194) if the problem occurs with the Qt step. 

[...]

Issue #1341, opened 2018-05-17
run_pylint: Include build and configure


Lestropie:
As raised in #1339: Script `run_pylint` executed as part of CI testing should additionally test scripts `build` and `configure`. Any issues raised by incorporating such testing will need to be resolved.

PR #1393, opened 2018-07-11
pylint: Test build and configure scripts

Lestropie:
Having #1390 reported a second time reminded me of #1341, particularly as to whether or not there may be other errors hiding in those scripts that we haven't come across. Most first-time users would likely immediately give up trying to install our software if hitting a critical error at such an early stage.

One thing I want to raise explicitly is the nature of the file generated by `configure`. Currently the parameters stored in there are loaded via an `exec()` call. Given it only involves variable values and doesn't actually require *execution*, maybe a more secure solution would be to export the relevant variables to a JSON, and have `build` load values from that file into a dictionary?


jdtournier:
Woah, are all these changes directly related to the pylint tests...? Why the aversion to the term 'folder'? 

Otherwise, I agree that storing our config as JSON might be cleaner, but I don't see a great deal of point in changing at this stage. It works fine as-is, and there's really no security risk here when the users have access to the main script itself... Unless there's an actual issue with the approach, I'd just leave it as-is.

Lestropie:
> Woah, are all these changes directly related to the pylint tests...?

Believe so. You can easily on `dev` / `master` make just the change to `run_pylint` to enable testing of `build` and `configure` and see the list it generates. Some of the variable renamings that seem kinda pointless are preventing renaming of a variable in the outer scope within function scope (similar to "ghosting"). But there were so many of these in `build` that I just disabled the check there.

> Why the aversion to the term 'folder'?

"Strictly speaking, there is a difference between a directory which is a file system concept, and the graphical user interface metaphor that is used to represent it (a folder)."

Personally I've always thought of "folder" as specifically a Windows concept / term.

> Unless there's an actual issue with the approach, I'd just leave it as-is.

:+1:

Lestropie:
@jdtournier: One I did want you to double-check is the [`build_next()` function](https://github.com/MRtrix3/mrtrix3/pull/1393/files#diff-b0da275520918e23dd615e2a747528f1L958): Previously it was [receiving](https://github.com/MRtrix3/mrtrix3/pull/1393/files#diff-b0da275520918e23dd615e2a747528f1L1265) a single argument "`id`", which was seemingly not being used; "`id`" is also a Python reserved keyword. I'm not quite sure historically what this may have been used for, so I just want you to check it and see if it flags anything in your memory.

jdtournier:
I like folders :open_file_folder:. Directories sound like the phone book :books:...


> @jdtournier: One I did want you to double-check is the build_next() function: Previously it was receiving a single argument "id", which was seemingly not being used

Yes, that must be some historical leftover. If it's unused, there's genuinely no reason for it to be there... :+1:

build
```diff
@@ -176,21 +174,19 @@ def colorize(s):
...
 def pipe_errors_to_less_handler():
   global error_stream
-  if len (error_stream):
+  if error_stream:
     try:
       [ fid, name ] = tempfile.mkstemp()
       try:
         fid = os.fdopen (fid, 'wb')
-        [ fid.write (x.encode (errors='ignore')) for x in colorize(error_stream) ]
+        fid.write ([ x.encode (errors='ignore') for x in colorize(error_stream) ])
```
jdtournier: OK, just noticed that this fails when there is output to be written - i.e. a compilation failure. Once I'd relaxed all the exception handling, the error was:
```
Error in atexit._run_exitfuncs:
  File "./build", line 181, in pipe_errors_to_less_handler
    fid.write ([ x.encode (errors='ignore') for x in colorize(error_stream) ])
TypeError: a bytes-like object is required, not 'list'
```
We'll need to revert that back...

jdtournier: Should be fixed in #1412 - if & when it gets merged...

Lestropie: One of my suggestions was going to be the solution you've got in 6d4f8383.

But I'm wondering if instead it would make more sense for `colorize()` to return a string rather than a list?

Lestropie: Just had a look, and every usage of `colorize()` was a weird syntax specifically to deal with that function returning a list rather than a string. Changed in #1412, diff of bcad65e should demonstrate what I mean.

jdtournier: 👍

Issue #1391, opened 2018-07-10
./configure doesn't abort on KeyboardInterrupt


michaelkarlcoleman:
The configure script fails to properly trap a keyboard interrupt, apparently giving a bogus check result instead.

```
./configure -nogui

MRtrix build type requested: release version with nogui

Detecting OS: linux
Looking for compiler [clang++]: not found
Looking for compiler [g++]: g++ (GCC) 7.3.0
Checking for C++11 compliance: ok
Checking shared library generation: ok
Detecting pointer size: 64 bit
Detecting byte order: little-endian
Checking for variable-length array support: ok
Checking for non-POD variable-length array support: ^C('Unexpected error:', "(<type 'exceptions.KeyboardInterrupt'>, KeyboardInterrupt(), <traceback object at 0x2aaaabdbf290>)")
not found
Checking for ::max_align_t: 16 bytes
Checking for std::max_align_t: 16 bytes
Checking for Eigen3 library:
...
```

Lestropie:
Michael,

Thanks for the report. Setting up signal handlers in both the `configure` and `build` scripts would be worthwhile I think; I sometimes myself get caught mashing Ctrl-C during builds trying to kill the multiple processes that are running. I have a little experience with such in Python since #858, so hoping it won't be a huge amount of effort.

Rob

jdtournier:
I don't think it'll be necessary to implement full signal handling here. I reckon we just need to _not_ catch the [KeyboardInterrupt exception](https://docs.python.org/2/library/exceptions.html#exceptions.KeyboardInterrupt) at [this point in `configure`](https://github.com/MRtrix3/mrtrix3/blob/master/configure#L462)...

tutufan:
Yes, I believe @jdtournier 's approach is correct.  (And I'd fix all of the bare `except:` clauses.)

On a somewhat related note, it'd be really nice if programs like `mrview` didn't try to catch and swallow segfaults themselves.  This behavior makes it harder to debug the issue, and arguably doesn't really add much value.

jdtournier:
> On a somewhat related note, it'd be really nice if programs like mrview didn't try to catch and swallow segfaults themselves. 

[...]

PR #1402, opened 2018-07-18
remove naked except: statements from python scripts

jdtournier:
As discussed in #1391
    
Now using 'except: Exception' as a catch-all instead. This will not catch KeyboardInterrupt, SystemExit, or GeneratorExit exception, since by design these derive from BaseException directly - precisely to ensure they don't get caught by an `except: Exception` statement.


jdtournier:
included in #1393 - closing.

Issue #206, opened 2015-04-08
Sync focus between multiple instances of MRview. 


draffelt:
Had somebody request this recently. We have discussed this before. Should be able to handle cross platform interprocess communication with Qt. 

http://doc.qt.io/qt-5/ipc.html

jdtournier:
Yes, it's been discussed multiple times. The issue has always been about how to implement this in a clean and intuitive fashion. Main thing is that you don't necessarily want your multiple instances of MRView to sync, so we need some mechanism to tell the relevant instances that they need to sync. Maybe if we had a big sync button on the toolbar, it might be easy enough to use?

Other options are whether we have a global sync (i.e. once sync'd, they all sync together), or a more fine-grained approach, where we can say window A syncs with window B, and window C with window D. Personally, I think the latter is going to be very unwieldy for no tangible gain...

Other thing we discussed is _what_ to a sync aside from the focus. There's all kinds of things we might want to keep consistent across instances: projection plane (when locked to plane), full projection matrix (when not locked), intensity window, FoV, window size... And also slightly less important ones: current volume (although to be fair we need that in the view toolbar first), colourmap, interpolation, what annotations are shown, etc. Ideally we'd chose sensible defaults and provide a mechanism to select what is to be synced...

Admittedly some of these probably seem unnecessary, but I reckon by default you'd probably want most of these on - I guess it all depends on what you want this for...

Lestropie:
Closed by #1940.

PR #1403, opened 2018-07-19
Sync mrview

LeeReidCSIRO:
Hi Guys,

Sometimes we use overlays, and sometimes we find ourselves repeatedly flicking between images to see how they overlap. Great sometimes, very awkward other times (e.g. when looking at three images). When awkward, users of viewers such as ITKSnap can just open 2+ windows. Those applications synchronise the focus crosshairs between windows, and so exploring the images together is a simple as dragging the cross hair around the image. 

This pull request adds this focus-syncing functionality to MRView. The option is switched off by default so as to maintain existing behaviour. It can be turned on by passing `-sync.focus`  to MRView, or through the view menu. Obviously, it needs to be turned on in at least two windows to have any effect.
e.g.
`mrview someImage.nii.gz -sync.focus & mrview someImage.nii.gz -sync.focus`

The viewers do not need to be started at the same time. E.g.
```
mrview input.nii.gz -sync.focus & 
./someBlockingScript.sh input.nii.gz output.nii.gz
mrview output.nii.gz -sync.focus
```

This is implemented in a fairly modular way that could easily allow other data to be passed between windows too; it should be very easy to extend if others have ideas later. This is not limited to just pushing 'sync' data, but could theoretically include commands, polling or... whatever comes up in the future. 

Beware if you decide to pull into master, this was forked from **dev** last week, and you will pull in unrelated commits. 

Limitations:
------------
- I have tested this on Ubuntu 12.04 (Virtual Box) with Qt 4.8, Windows Qt 5.9.2, and OSX Darwin Qt 5.10.1. These are the platforms I have easily accessible to me. Further testing might be warranted before release? (**hint ;)**) The best way to test performance and stability is to open multiple windows simultaneously (i.e. in one command using `&` ) as this will force them to compete for unique ids. Then set one window to volume render, and move the focus rapidly in another window. 
- This is implemented in a way that limits the number of windows open to 32 (Search for `MAX_NO_ALLOWED`). At present, when additional windows are loaded, they will not sync, and will print a warning when first starting. Thirty-two could be be changed to any number but if it is raised significantly, this may start to hurt start-up performance. I suppose, ideally, for the 33rd+ window, SyncManager::SetWindow would uncheck and de-activate the option in MRView window's menu (which it doesn't) but... surely noone needs 33 synced windows?  
- The current implementation does not clear data passed between processes after it has been read, which means that memory usage will slowly increase with use. However, the volume of data passed is so small (20 bytes), that 10h of constant clicking with two windows is likely to cause a memory leak of only 1.2MB. Qt doesn't support clearing data passed through a connection; it only takes place when destroying a connection. I thought it was best that the code remained simple, performant, and readable. I expect that this is unlikely to ever become a problem unless people start passing diffusion datasets through this mechanism. If that occurs, I'm sure a signal could be sent that requests the connection is deleted and re-opened. Doing this for every mouse movement, however, is overkill and likely to hit performance of the app, especially in things like virtual machines. 
- I'd like to move the actions of SyncManager::SetWindow into the constructor, but Qt makes it difficult to make custom constructors, especially for someone without much C++ experience. If anyone with more experience feels like cleaning this up...

Code Explanation:
----------------

Added classes are in gui/mrview/sync and I'm happy to answer questions. Roughly, from bottom to top:
- `Client` sends data (effectively a wrapper; message agnostic)
- `LocalSocketReader` receives data (a wrapper; message agnostic)
- `InterprocessCommunicator` searches for, connects to, and conducts two-way communication with other InterprocessCommunicators in other processes. To do this it uses multiple `Client`s and a single `LocalSocketReader`. It understands the concept of syncing, but is agnostic to the type of data being synced and has room to support messages that are for purposes other than syncing.
- `SyncManager` listens to signals from window, and sync signals from `InterprocessCommunicator`. It translates them and passes to one another.
- `window` has had the sync option added for the user to turn on/off via the gui/command line, but otherwise is unchanged. I.E. Syncing is conducted by `SyncManager`, not window, and is not built into this class at all.
- mrview.cpp creates a `SyncManager` upon start up, and links it with `window`.


Lestropie:
Hi Lee,

This kind of functionality would be *awesome* to have, and was initially listed all the way back in #206. Thanks for putting in the effort; we're really eager to have more incoming contributions from the community.

I'm currently stuck in a mad writing rush, but I promise when I find the time I'll get around to having a look at both this and #1238 (assuming nobody else beats me to them).

Rob

jdtournier:
@LeeReidCSIRO: sorry about the (long) delay in responding to this - it's a big addition, and I'll need to set a whole day aside to go through it in detail... 

However, before I even start reviewing this, I'd thought I'd share my thoughts on how I was envisaging this being implemented when the issue was originally brought up. Unfortunately, it's quite different from your implementation, which I realise is not great given the amount of time you've undoubtedly spent on this... I never bothered to write my thoughts down at the time, so I apologise for this. 

Anyway, as you've already discovered, the big issue with your general approach is that it requires some kind of client/server model. This is not a problem per se, and would most likely be what everyone would recommend starting with. However, I think this client/server model is problematic since in my opinion all instances of MRView are equal: they're all servers and clients at the same time. 

So what I had in mind was to synchronise via the filesystem. The process would be:

- when synchronisation is enabled for an instance of MRView (whether through UI or command-line), it creates a file in a known location (say `~/.mrview.sync`) using `O_CREAT | O_RDWR | O_EXCL` flags to guarantee no race conditions. It immediately truncates the file to the size of the structure that will be shared between instances. If the file already exists, skip this step.

- memory-map the file, and cast a pointed to the shared structure so it can be easily manipulated. Basically, at this point we have a memory region that will be shared between processes, and its contents will be a direct binary dump of a class/structure containing the primitives to be synchronised. Ideally, the class would contain atomic types, which I _think_ work across processes on modern CPUs, providing _some_ degree of protection against race conditions.

- Each process has access to this shared region, but also stores its own state. It can also track which primitives it wants to synchronise with. So one instance might want to sync on focus, the other on brightness/contrast, etc. Either that or they all sync on the same things, in which case the shared structure itself keeps track of which primitives are being synchronised (I'm not sure which is best...).

- Each process also watches the file for changes using the [QFileWatcher](http://doc.qt.io/archives/qt-4.8/qfilesystemwatcher.html) class: this emits a signal when the file is changed, so we can set a slot/callback to be called if anything changes.

- In corresponding slot/callback, we check the contents of the shared structure against current parameters, copy over any changes into the processs-specific structures, and call `updateGL()` if necessary. 

- When the user changes anything in one instance, if sync is enabled, those parameters that are being sync'd are compared to those in the shared structure, and copied over if needed. This should immediately trigger an update on all other instances. When this happens, it would probably be necessary to disconnect the QFileWatcher signal to prevent a double update in the original process.

This would I think be a much cleaner and simpler implementation, which I think deals with a lot of the issues. 

However, one of the things I'm slightly worried about is the scope for race conditions, if two processes happen to be writing to the shared structure at the same time. Not something I anticipate would be a huge problem given that most people can only interact with one instance at a time, but there may be scope for weirdness if for example the screen capture is active, playing over a sequence of frames, and the user decides to interact with another instance at the same time, _and_ they are all set to sync... Not a common use case, I reckon, but hey. 

In any case, the problem with the memory-mapping is that changes to the parameters could trigger updates at any time, so maybe another instance could grab changes in some of the parameters before the first one has a chance to modify some of the other parameters, but the second instance may not be notified of these further changes. To get around this, it might be safer to use regular `read()` & `write()` calls, rather than memory mapping. Basically, when a process needs to update the shared primitives, it would:

- issue a single `read()` to load the structure, and cast it to the right type.

- update all sync'd parameters

- issue a single `write()` call to dump the structure back to file. 

The other processes, upon receiving notification, would:

- issue a single `read()` call to load the structure, and cast it to the right type

- update its own parameters if there have been any changes, as per the memory-mapping approach. 

The advantage here is that these `read()` & `write()` calls are system calls, and using a single one each time provides relatively strong guarantees of atomicity (the whole write happens in one go), and the other processes are likely to be notified of a change _after_ the write is complete (it's unlikely to be before...). So I reckon that might be a better approach.

In any case, that's just my 2 cents, and what I've in mind to do for a while. I feel a bit guilty for not sharing that information sooner, it might have saved you some effort if we'd discussed this upfront... Please let me know what you think, and let's discuss this.

LeeReidCSIRO:
@jdtournier: Thanks for the feedback. Perhaps we set up a video call at some point and discuss with a whiteboard handy? This is quite complex and it might save a lot of time. For the interim, some things to consider:

Your suggestion was actually my original plan as it seemed simpler to implement, but after doing a bit of reading / playing around, I hit a number of snags. I'll try to convey why I picked pipes. Some of these reasons are related to your concerns and I have to admit that it was surprisingly easy to induce these problems.

Ultimately
======
The reason I picked pipes comes down to multiple OS support. Memory mapping requires that we have some way of locking that memory to avoid race conditions, but there is no OS-independent way to do this that I could find.... at least without setting up another interprocess communication channel (and why implement two?). Locking itself can also cause issues when it works (see below). Pipes don't have any such need. 

**If you decide you'd really like it done a different way and can tell me how to avoid the issues below in a platform-independent way, I am happy to implement it that way, but just no guarantees on when :)**

Some scenarios I hit:

Scenario A
----
MRView ('Process A') goes into an error state and is killed by the OS, meaning it can't clean up the sync file. A few issues now exist. 
1) Junk file sitting on the HDD. Not a big deal, but kind of annoying.
2) Process B now starts. It will see a file to sync with, and try to copy its instructed state. This state puts the focus completely out of the range of the loaded image and the image appears black. The average user probably will have no idea what is wrong with their "image". Opening and closing mrview repeatedly does not solve the problem. Similar issues could exist for things like intensity scaling etc.
3) Process B now starts. It will see a file to sync with, and try to copy its instructed state. Potentially, this file is corrupt _and/or_ contains some sort of sync state that triggered a bug and caused Process A to crash in the first place. Process B and any new MRView processes now either crash upon opening or have many settings set to obscure values. 
4) Process B now starts. It will see a file to sync with, but is smarter and checks whether any other MRView processes are open. If no, ok, it wipes the file and starts in default state. However, if any others are open, it needs to ask them whether they have sync turned on (i.e. whether they are using/watching the file). To do this, it needs to communicate directly with them (yes?) and we end up back using pipes or another mechanism outside the file system anyway. A similar problem exists in cleaning up this sync state file. If two processes are open and one turns off sync or closes, how does it know whether to clean up this sync file without polling the other process directly? Perhaps you end up writing to the file, waiting for a period for it to be edited by the other process, and then deleting it if nothing happens, but this is pretty messy; how long do you wait? That other process may be busy loading a large dataset or taking a while to start up. We don't want to have a ten second delay in waiting for MRView to shut down or respond after clicking a check box.

Contrast pipes: Process B simply searches for a named pipe. If it exists, then it is connected to an MRView which is knows is running. If either process crashes, the pipe closes. Messages are read in full or not at all.

Scenario B
----
If we have non-atomic file writing
Process A starts writing to the file... 
Process B loads up in the background (e.g. a script running in another window launches it). 
Process B reads the half-written file, many settings are out of date, junk, or zeros. 
Process B starts applying these partially out of date settings
Process A finishes writing. 
Process B receives an signal to sync, but delays this update because it hasn't finished applying the last one. 
Process B finishes applying the wrong settings. 
Process B finds it does not match the sync file, so it updates it 
Process A receives the signal and updates itself to these incorrect/corrupt settings

Contrast pipes: Messages have a stated size and are read in full or not at all.

Scenario C
----
If we have atomic file writing
Process A and B are both loaded 
Process A starts to write to the file. It locks it to avoid Scenario B
Process A hangs for some reason
The user changes the focus for Process B. It tries to write to the file and waits for it to be available, causing it to also hang
The user gets frustrated and loads a third window (Process C). This tries to read the file upon loading but it is locked, causing it to hang.
If the user kills Process A, it releases the file, but now we have a corrupted state and are in some situation similar to the previous two scenarios

Contrast pipes: If either process crashes, the pipe closes. If a pipe hangs during writing, the message will never be read by any client. Processes B and C can continue to talk to one another.


Client/Server
==============
Regarding _"I think this client/server model is problematic since in my opinion all instances of MRView are equal: they're all servers and clients at the same time."_
I might misunderstand here. To me, ultimately, server/client just means that there is a class for sending the data and a class for listening/reading it. Regardless as to the IPC mechanism you have to have a class that reads and one that writes. You can make a super-class if you'd like but it will internally still have methods to listen/read, methods to write.
In the current implementation, connections are set up in all directions, but only one window at a time acts as the client (being "in charge"). This is important to avoid infinite loops between processes, or to avoid needing lots of code/chance to avoid race conditions. This is because if you ask mrview to do some heavy lifting (e.g. playing with 3D views) the processes each end up with a backlog of instructions while also trying to syncronise with others.

Process A and B are active
Process A gets a backlog of move instructions (10,20,30) and (11,22,33)
Process A processes "move (10,20,30)", sends to sync
Process B syncs with      (10,20,30), and sends to sync, while Process A processes "move (11,22,33)" and sends this to sync
Process A syncs with      (10,20,30), and sends to sync, while Process B syncs with (11,22,33) and sends this to sync
Process B syncs with      (10,20,30), and sends to sync, while Process A syncs with (11,22,33) and sends this to sync
Process A syncs with      (10,20,30), and sends to sync, while Process B syncs with (11,22,33) and sends this to sync
etc

With the file system model you end up with a similar loop or one of the previously mentioned scenarios instead (depending on atomicy, chances with read/write timing etc). To avoid this, you need a lock, but again, this isn't simple to do across OSs. There are also checks you can place to not send signals unnecessarily but the code grows very quickly.

With the pipes model, this issue exists because the messages are posted around the same time. The current implementation avoids this simply. When changes are made, a window just needs to check whether it's the focussed window or not. If it is, then it sends sync information. If not, it does no such thing. There's no need to check whether values have changed etc. This is a platform-independent - all OSs only have one window focussed at a given time. This is not suitable for memory mapping, though, because window focus can change while a file is still being written; it's not a true "lock" on the memory region. Perhaps with atomic write, atomic read, and window focus it might work? I'd need to think...

Also...
=====
The implementation could be changed to more similarly reflect your "one file" approach, and just have that passed through the pipe. I suppose while this looks simpler I see a few limitations:

- This limits you to sharing visualisation states; trying to extend to issue commands ("take screenshot and send to me", "update 3D renders of ROI", "next frame") or pass data (e.g. ROI drawing commands as the user performs them) will require that code becomes quite hack-ish. Maybe this is not likely to happen anyway, so you might not care. Understandable.
- It does add an additional state layer to MRView, and implementing checks as to whether values have changed etc, which would need to be implemented (win some/lose some)

Anyway, happy to have a chat at some point if you'd like. Again, if you'd really like things done in a different way, I'm happy to oblige but I just can't guarantee certain time frames.

jdtournier:
@LeeReidCSIRO: thanks for your thoughts, this is very informative, and indeed comes up with a few edge cases that I hadn't considered... 

I also hadn't twigged how your model works exactly. If I understand correctly, each process has one receiver, which all other sync'd processes can send to, and one sender for each separate sync'd process, right? So when the user updates the window for one instance, that instance notifies all others via its multiple senders. On the receiving side, each instance only needs to monitor one incoming socket for updates. So that basically invalidates my comment about the client/server issue - I'd assumed one instance was the 'master' server, and all others sync'd through it. My apologies. 

However, before I throw in the towel, I'd just like to mention a few simple tricks that might get around many of the issues you rightly highlight. 

- Most of the issues around scenario A can be side-stepped by simply changing what happens when an instance enables sync. All of the examples you mention imply that turning on sync on a given instance makes that instance sync to the current state. But we could also make it so that the state of the current instance becomes the new state - in many ways, that kind of makes sense, since it leaves the current window unaffected, and all other sync'd instances would update to match. I'm not convinced that this is what the user might have in mind 100% of the time, but it would certainly avoid the issue of stale or corrupt state files, since it would be overwritten immediately anyway. 

- Race conditions between read & write could be avoided relatively easily, either using atomic write operations (essentially a write to temporary followed by atomic rename - though not sure whether that's guaranteed atomic on all OS's), or by making sure the first thing written is a flag that is initially set to say zero (signifying that the file is undergoing update), and once the update is complete, that flag is then set to some non-zero value, at which point other processes can update. So other instances would only need to iterate the read until the flag is non-zero. 

   About that last idea: I used a similar technique in the old MRtrix 0.2 version to update the streamlines file while `streamtrack` was generating it, and simultaneously updating its display in `mrview`: I'd set a barrier consisting of an `Inf` triplet, write the data for the next streamline, including a new `Inf` triplet barrier, and then go back to overwrite the barrier with a `NaN` triplet (the streamline delimiter). The viewer would then not try to read beyond that `Inf` barrier - avoiding the bus errors that would otherwise happen. 

- You make a good point about when instances should emit their new state, and avoiding infinite loops. I like your idea of only emitting when the window has focus. Another safeguard might be to write the sender's PID (maybe as the initial flag I was talking about above?) as a way to prevent an instance from syncing with itself (although I'll admit that's only necessary using the filesystem approach). That would be in addition to temporarily disconnecting the signal handler for the `fileChanged()` signal. 

- I hadn't considered the idea of sending commands that way. It's an interesting idea, but not impossible to implement with the filesystem approach. For instance, the snapshot idea only requires an additional flag in the state file to request the snapshot be taken for that frame - the harder part is making sure the filenames for the screenshots don't conflict between instances... (I guess we could just prefill the default filename entry with the instance's PID?). Actually, there's also the potential for frames to be skipped if the primary instance renders faster than the other(s), which might be difficult to deal with. Your approach avoids that, but on the other hand it might introduce significant lag in interactive use (i.e. not via the screenshot tool) if one of the sync'd instances is much slower to update than the one being interacted with (e.g. if it's displaying 1M streamlines while the primary is only displaying ROIs). 

   As an aside, I'm not sure what you mean by "update renders of 3D ROI" - they're currently not rendered in the 3D view (although they should...), but are otherwise held in RAM (so I can see the difficulty in syncing there). If you are indeed referring to ROIs that are being edited, that would be better served by saving it, loading the file as an overlay in a separate instance, and watching the file for changes (something we should implement anyway, truth be told). The added benefit would be that the image would be updated regardless of how it was changed - it could have been overwritten by a separate command, etc. 

   I'm also not clear on what you mean by 'next frame' - are you referring to the screenshot tool? In which case it should already be served by the sync backend anyway, right? Otherwise, if you mean 'next slice', then that would mean that the focus won't in general be guaranteed in sync anymore (unless the images were identical in voxel size and orientation). That might be acceptable in some circumstances, but I'm not sure I'd be too keen on it...

   So while I can see the appeal long-term, like you say, it's not something I can foresee being required for the foreseeable future - though others might disagree... 

---

Finally, I should add that some of this is influenced by changes I'd like to make long term, where this idea of each instance holding its own state would be much more central. But it's a long term goal, and one I'm unlikely to find the time to do any time soon... I might add a couple of issues to document my thoughts on that front though - I'll add references to these here in due course.

LeeReidCSIRO:
@jdtournier 

I few bits and then I think we will be in a good position to make a judgement call. Remember that the actual method of IPC will be abstracted away from the idea of a internal-state, so this internal-state idea perhaps will affect what is passed between processes, but not how it is passed.

> Most of the issues around scenario A can be side-stepped by simply changing what happens when an instance enables sync.
Yes, that's a workaround. With focus I'd argue that resetting like this maybe (?) is even expected behaviour. It does mean that future features added to sync may have to be careful though: I certainly would be frustrated to find that I absent-mindedly open MRView and find my other open window closes things I've taken time to set up (e.g. overlays, rois, tractograms) or changes settings for colour palettes/contrast/views. I guess it depends on what you and others envisage being useful to sync in the future? 

Another option is to use shared memory rather than a file but otherwise work as you envisage. This wouldn't avoid all my mentioned issues but it would avoid leaving corrupt files on the system after a process is killed. (I actually would need to check whether this memory leaks or is cleaned up on a process kill operation). Is there an event to listen to changes in memory like `fileChanged()`?

>  Another safeguard might be to write the sender's PID (maybe as the initial flag I was talking about above?) as a way to prevent an instance from syncing with itself (although I'll admit that's only necessary using the filesystem approach)

It has actually just occurred to me (sorry about this) that I have already implemented a multi-process lock in the pull request at
mrtrix3/src/gui/mrview/sync/processlock.cpp
it uses shared memory and works a charm. It doesn't solve issues with knowing whether any other process exists (thinking corrupted files, above), or potential issues with a process hanging, but it does solve the access issue. Sorry, I guess my memory had gone a bit stale! It would still need to be combined with the window focus and a disconnection. The disconnection part really relies on the fact that the final operation (whether renaming or writing to the file) is guaranteed to be completed before the next line of code (reconnect to the signal) executes. Do we have this guarantee on all OS's without writing platform-specific code?

**Command Stuff**
I was just listing potential future commands as examples, no need to worry about my specifics. I think I was just trying to make sure it's understood that this is something that is easily possible with the current implementation, less so with the file system. That said, I'm not 100% convinced that commands will ever be desired, so I guess I just wanted to throw that your way for you to make a judgement call on. Your workaround sounds possible. The issue with the file system here is that there is no guarantee that any receiver has read that command before it is overwritten by another sync/command. I think for commands a queue in each process, or history, is truly required. By the sounds of it, though, commands are not a priority?

LeeReidCSIRO:
@jdtournier 
Thanks for noting #1437, I really feel I understand your motivation now. I'm happy to get #1437 started (sufficiently to have this sync working), but just note again that the decision regarding IPC mechanism (file/pipe/shared memory) can be independent of this.

How would you see this organised, in terms of layers? 
In the past when I've made sizeable GUI programs I've used a fairly traditional MVP design. Tier 1 is the data. Tier 2 holds an abstract visual representation of the data (i.e. settings like contrast, show/do-not-show, etc) and accepts commands from Tier 3. Tier 3 is the view (here, the Qt controls). Higher tiers have knowledge of lower tiers but not vice versa. In this case we're slightly different because you're looking to synchronise aspects which might traditionally be left as responsibility of the view, but it could still work and might look something like that below:

Let T be some generic type. Consider the following a 'signalling command' and 'signalling property' pattern or class:
   - `T Get()` method (for properties)
   - `void Set(T)` method (for properties)
   - `execute()` or similar method (for commands)
   - `bool getSetLegal()` (set/execute is legal)
   - signal `getLegalityChanged` (whether `getSetLegal()` has changed)
   - `T[] getLegalValues()` (for properties with ranges etc) 


**Layer 1.** Image/Tractography Data
- Exposes signalling properties, as above

**Layer 2.** Presenter, containing:
- Abstracted Visual Representations (VR) of Data
   - Expose signalling properties, as above
- Abstracted VR of GUI Scene (which is what our sync manager would listen to for signals)
   - Expose signalling properties, as above
- Methods to accept command requests (from GUI/sync manager etc) which do not fall under the VRs e.g. `setFocus(x,y,z)`, `openImage(path)`
- Exposes command, as above for commands which do not fall under VRs, such as
    - VRs are added or removed
    - Program-wide settings changes (e.g. sync on/off)

**Layer 3.** View (Qt, etc)
- Binds to `Presenter` so:
   - When VRs are added/removed to the `Presenter`, they are bound/released
   - Actions on controllers (buttons, shortcuts etc) bind to the signalling property/command, listening for legality in order for them to be disabled or enabled appropriately
   - Changes in VR properties cause the view to update

View objects would include things like the toolbars, menus, and scene itself. The window would be the top level of this (as it is now already). 

The sync manager would sit outside of this, binding to the Presenter's events just like Layer 3 objects do. 

On the right track, or did you have something different in mind? This should be possible to implement property-at-a-time. Given that only one property is synced at the moment I don't foresee this as a huge task to get started and then assess whether the design is likely to scale well for everything else.

LeeReidCSIRO:
@jdtournier 

> Lack of a broadcast support, which requires each instance to maintain a list of all other instances to broadcast to. This feels messy and error-prone somehow (but we might be able to get around that - more on that below).

As far as I'm aware this is the case. Ultimately they all have something keeping track, but with sockets, yes, the OS doesn't help you out here and you need to do it manually. This is fairly simple though, essentially a list of things which it is connected to. When a new connection arrives, it is added to the list. If sending fails because a connection is closed, it is removed. I'll admit that there is a slightly messy component to it (searching for multiple connections with a loop, rather than just one, upon opening), but I think this still has much less room for error than anything else we've discussed.

> What happens when the instance sending updates renders them faster than the other instances? Should all instances discard intermediate updates if newer ones are available, at the risk of dropping frames, or should all updates be rendered, at the risk of potentially very significant lag? I'd argue the former would be appropriate in all circumstances other than screen capture - but as far as I can tell, the current approach processes all updates, right?

No, the current implementation checks for the most recent complete message for a given property (remember that current implementation is property-at-a-time, not state-at-a-time) and skips to it. Originally, I found no lag when it was set up to process everything in order. I changed this, however, to avoid race conditions or unexpected behaviour whereby a backlog in the current window sees old states and pushes them back to other processes. It would be very easy to have different behaviour for commands.

**Memory Mapping**

> However, it would be trivial to have each instance poll for updates at regular intervals (say 20-30 times per second), which would provide smooth frame rates and hopefully negligible CPU usage. Not ideal, but certainly an option.

I guess this is what I was saying previously. Regardless as to the inner workings, with memory mapping, ultimately an additional communication port needs to be set up or we need to consider a bit of a hack-solution. I can see something like a timer being simple to implement, but making debugging this and unrelated code a bit of a nightmare in the future. Just imagine someone clicking a button 20 times a second while trying to figure out why (e.g) some load/save operation is running slowly....

**UDP**

*  UDP isn't something I've used before but how you're describing it sounds good. 

*  In terms of gotchas, what I have read this morning is that UDP does not guarantee packets arrive. There are plenty of forum questions where developers are saying that their packets are dropped regularly on loopback. So, for example, it's not suitable for commands which depend on a particular state or previous commands, because there's no guarantee those previous messages arrived in the first place. e.g. capturing all frames and saving as an image series may see some of them missing. It's hard to tell without testing on many PCs whether this is likely to become a real or extremely-rare issue. 

*  As it stands, it might not be a huge amount of work to change the current branch to use this in place of sockets

*  Does QT really restrict to 512 bytes? Will this be enough room for this in 24 months time? It will be, I'm sure, if you are sending one property at a time, rather than a full state, but with a full state... well, I think it's important at this stage to make a list of things that you might want to sync in the future. If a string is likely at any point (e.g. file location) then 512 is not going to be ok. Packets can be concatenated together, of course, but at that point this solution starts to look more and more like what I've implemented, because data will all be pushed into a stream anyway. Not the end of the world...

> I've not yet figured out whether multiple instances can all concurrently hold the socket open for broadcast. If so, then hopefully the system preserves the integrity of each datagram. If not, a hopefully trivial modification would be to only open the socket for broadcast when writing to it, allowing other processes to broadcast in the meantime. It'll need some handling in case of race conditions when opening the socket, but hopefully nothing more than iterating on failure...


*  I'm not sure either. What I've read implies that you can do it on some platforms (at least), but you need to be a little careful. I strongly suspect the OS can enforce access to one-process-at-a-time without us needing locks. If we do, like I've said before, we already have existing code that allows a multiprocess lock, but the issue with this is if one process hangs with the port locked, they all do.

jdtournier:
Sorry about delay - I'm officially on leave, so I can only respond when I find a spare minute...

My apologies, I hadn't twigged your implementation checked for pending updates - all good!

Otherwise, I've had a brief look into UDP, given the concerns you've raised. Personally, I reckon it's just the ticket...

From my (limited) reading up on the topic, UDP is not a reliable protocol, in that:

 - it doesn't guarantee delivery. This is mostly because there is no formal acknowledgement of receipt sent back to the sender, so the sender cannot be certain that all the data sent have been received. TCP does guarantee receipt.

- it doesn't guarantee in-order delivery of datagrams. They are handed over to the receiving application as they come in. And there's plenty of reasons why they'd be out-of-order. However, on a single host, the only reason that I've come across that seems plausible is multi-threaded handling in the network stack - makes sense. So it's up to the application to ensure the datagrams are in the right order, presumably by writing an index for each datagram (which also allows detection of dropped packets). 

- As a consequence of the lack of acknowledgement, there is no mechanism within the network layer itself to ensure delivery of all datagrams. Again, it's up to the application to verify all datagrams have arrived. On a single host, the only plausible explanation that I've come across for packets being dropped is buffering: if the data rate is faster than the receiving application can handle, of course packets will be dropped. Also, if the target application stalls for a significant amount of time, packets could similarly be dropped. 

But despite the above, it's apparently still the protocol of choice in most online gaming platforms, apparently. This is mostly because it avoids the latency that would be caused by acknowledgement packets and the occasional resend. But clearly, if gaming engines are happy to rely on it, it can't be _that_ unreliable... 

So given the above, I reckon UDP might still be our best bet. I don't think our rate of update will ever be fast enough that we need to worry about dropped packets, or even out-of-order reception - we're unlikely to send more than ~50 packets / second at any time. Even at full rate, it doesn't matter that much if a few packets are out of order, we'll probably discard a lot of intermediate updates and keep only the latest anyway...  And if for some reason the display ends up out of date on rare occasions, it should right itself the very next update anyway. If we're really worried about packets dropping due to stalling (e.g. a massive render), we can just set up another thread to listen for updates, and schedule an update to the main thread with the latest state. That shouldn't be too difficult. But only really worth doing if it turns out to be a problem. Hopefully, if the system decides to drop packets, it'll drop the oldest ones rather than the most recent (?), in which case it doesn't really matter at all: we're only after the latest update anyway...

Other reasons I like the idea of UDP is it allows worry-free broadcast (we don't need to worry about who's receiving), and it allows us to post standalone full-state updates each time: since a datagram is either delivered or it's not, we can't get partial or corrupted updates. Also, any new instances that start listening immediately get the latest updates. We can easily insert a timestamp in there to figure which is the most recent, which will be accurate to within microseconds, and furthermore work across instances. An advantage of a timestamp over a counter/index is there is no reliance on each instance having received the latest update before it can send its state with the correct counter - this would I reckon be particular useful when turning on sync on an instance, since it won't have received any updates at all by that stage. 

Also, a few clarifications:

- Qt doesn't limit the size of the datagram to 512 bytes - the largest packet size as far as I can tell is 65,535 bytes. But 512 bytes is small enough to guarantee no fragmentation during transit, which might be introduced by the various links the packet might have to travel through. I reckon this is only relevant across hosts, but I just mentioned it here to say that we can post plenty enough data in one packet for our purposes. 

- Also, I reckon 512 bytes is plenty enough for us for the foreseeable future, given that currently the only parameters I think we're thinking about synchronising are:

  - the focus
  - the target
  - the projection plane (for lock-to-axes display)
  - the orientation (for off-axis rendering)
  - the viewport size
  - the volume indices
  - brightness/contrast (note we should enable sync _within_ instance / across images too...)

  And ideally we'd be able to select which of these get synchronised. My guess is the user will be able to select which aspects to synchronise in any instance, and that would apply to all others (so the state to send out should contain flags to indicate which of these parameters is 'active', and the UI for each instance would be updated to match when receiving a state update. 

  While we may in future decide to synchronise other aspects, I reckon we can worry about that if & when the need arises. I reckon anything more than this and it will start to morph into something it wasn't intended to be... Maybe there is demand out there for other features that I'm not aware of? In which case let's discuss them, by all means!

thijsdhollander:
...candidate for RC4?

LeeReidCSIRO:
Hi @thijsdhollander 

As it was initially submitted, this is stable and tested to work on Win/OSX/Ubuntu. However, it's perhaps not set up in the way Donald originally envisaged; whether it's likely to clash with near-future plans for the code base is not something I can answer. Embarrassingly, within about a week of saying "I'll have a crack at making some changes" I became snowed under with work that I have to prioritise; I haven't managed to revisit it yet :( 

Whether you guys want to merge it and have these changes made later, or leave it out of the product until it is rewritten is obviously up to you. I do intend on getting back to this, but I can't put a time line on it just yet. Papers to write, grants to apply for etc

LeeReidCSIRO:
(sorry, wrong button!)

thijsdhollander:
No worries.  The RC4 milestone was just added during a meeting we had, basically to make sure we don't overlook it.  @jdtournier was considering maybe adding it "as is" to the current RC4, and then indeed maybe revisiting later, potentially with other changes or a larger redesign of `mrview` related stuff.  I initially just brought it up because I vaguely remembered it sitting there (and it *is* a very useful feature, I reckon!).  So well, no pressure at all.  We'll just see what happens as we near the completion of RC4. :wink:

LeeReidCSIRO:
Great 😄

jdtournier:
@LeeReidCSIRO: we had a chat recently about outstanding PRs that really should have been merged a while back... We decided yours is good to go, and while I might like to change the approach down the track, there's really no reason to block this given that it's there and it works. So I've had another look through to refresh my memory and make sure it all still works against the latest `dev` branch, and it seems all good. 

I made a few modifications to bring it up to date and get past a couple of conflicts. You'll find them on the `LeeReidCSIRO-sync_mrview` branch. If you can merge them into yours and push again, I'm hoping that'll allow this PR to pass all the CI testing and we can then press the button.

Hope that's OK?

LeeReidCSIRO:
Hi @jdtournier,

Of course that's OK. I'll try to see to it today/tomorrow.

LeeReidCSIRO:
Merged, pushed, tested on Windows. Looks like the automated secondary-checks fail, but not sure if that's to do with these commits specifically.

jdtournier:
@LeeReidCSIRO: I'm closing this PR in favour of #1940 - simply so I can push to it... I've just cleaned a few things up and and made the changes necessary to pass all the tests. 🤞

PR #1406, opened 2018-07-21
amp2response speedup

### LINKED ISSUE #1231: amp2response performance
As raised [here](http://community.mrtrix.org/t/problem-with-dwi2response-msmt-5tt-amp2response-command-for-ex-vivo-monkey-data/1402) (even though there's an upstream problem in that particular case). 

- Don't bother constructing transformation matrices when using the `-isotropic` option.

- Pre-determine the number of voxels in the mask, and pre-allocate `cat_transforms` and `cat_data` rather than using `conservativeResize()`.

Lestropie:
Pre-allocate memory for ZSH transformations and image data, rather than using Eigen's `conservativeResize()` function. According to a user, their `dwi2response` call went from 40 minutes to 8.

Also should fix the segfault on providing an XxYxZx1 mask image as reported in #1231.

Closes #1231.


jdtournier:
Let's see if the recent merge of #1393 will help with the testing here... 

In the meantime, any chance you could take me through the maths behind `amp2response`? Maybe they're already described in some other issue? I'm just surprised it would still take 8 minutes - I reckon there's scope for some further optimisation, but I'd like to get a better feel for what you're trying to do before suggesting something daft...

Lestropie:
That timing would have been for `dwi2response` as a whole, not just `amp2response`.

Re-acquired some HCP data to check it myself: `dwi2response dhollander` was 2:40 on `dev`, `2:10` on this branch. Guessing that if the system starts swapping at all the difference will be more drastic.

Also confirm that this worked with a XxYxZx1 mask image, whereas `dev` segfaulted; should push that to `master` as well I think.

PR #1407, opened 2018-07-21
Changes to temporary directory handling

### LINKED ISSUE #1033: TmpFileDir -- Environment Variable Option
Disclaimer: I've not run MRtrix, I'm attempting to install and configure it for a user on a cluster I support.

Also disclaimer: I'm not that hot on C++.

Our cluster uses a setup whereby each batch job has its own allocated tmpfs temporary storage, whose path includes a numeric job ID.

This doesn't really work well with MRtrix's default of putting all temporary files into `/tmp` or a path set in a file in the user's home directory, since no two jobs, which could be running simultaneously, will have the same temporary storage path.

Would you accept a small change to `core/file/utils.h` that makes it accept either an MRtrix-specific environment variable, or a common environment variable to use for such things, `$TMPDIR`?

Something like (based on TmpFilePrefix):

```C++
      //CONF option: TmpFileDir
      //CONF default: `/tmp` (on Unix), `.` (on Windows)
      //CONF The prefix for temporary files (as used in pipelines). By default,
      //CONF these files get written to the current folder, which may cause
      //CONF performance issues when operating over distributed file systems. 
      //CONF In this case, it may be better to specify `/tmp/` here. Note that 
      //CONF this option can also be set using the `MRTRIX_TMPFILE_DIR`
      //CONF environment variable or the `TMPDIR` environment variable,
      //CONF without editing the config file. If both are set, 
      //CONF `MRTRIX_TMPFILE_DIR` takes precedence.

      const std::string __get_tmpfile_dir() {
            const char * from_env_specific = getenv("MRTRIX_TMPFILE_DIR");
            if (from_env_specific) return from_env_specific;
            const char * from_env_general = getenv("TMPDIR");
            if (from_env_general) return from_env_general;
            return File::Config::get ("TmpFileDir", 
#ifdef MRTRIX_WINDOWS
                  "."
#else
                  "/tmp"
#endif
                  );
      }
            
      const std::string& tmpfile_dir () {
            static const std::string __tmpfile_dir = __get_tmpfile_dir();
            return __tmpfile_dir;
      }
```

### LINKED ISSUE #1405: run.command(): Move temporary file directory
Proposed solution for issue reported on [forum](http://community.mrtrix.org/t/problem-with-parallelizing-mrtrix-pipeline/1685/).

- Allow the temporary file directory `MR::File::tmpfile_dir()` to be set via an environment variable.

- Within the `mrtrix3.run` module, set this environment variable to be the temporary directory created for the script: `app.tempDir`.

Piped images generated during `run.command()` calls will be created in the script's temporary directory rather than e.g. `/tmp/`, and will additionally be deleted as soon as the piped command is completed.

Lestropie:
- New environment variable `MRTRIX_TMPFILE_DIR`, which has precedence over the MRtrix config file in setting the location of the temporary directory used for piped images. If this is not set, environment variable `TMPDIR` will also be tested and used if set.

- In Python scripts, use environment variable `MRTRIX_TMPFILE_DIR` to instruct any invoked MRtrix commands to place piped images within the temporary directory created by the Python library, rather than the default location.

Closes #1033.

Closes #1405.


jdtournier:
@Lestropie, see if you're happy with my latest commit...

Lestropie:
Yeah that's better.

PR #1410, opened 2018-07-27
TravisCI and pylint updates

Lestropie:
Trying to resolve a CI issue in order to prepare #1409 for merge ended up requiring more commits than the issue that PR is intended for, and those commits had nothing whatsoever to do with the title of the PR, so I've reset that branch back to include only the relevant commits, and am here making a separate PR in order to directly address the issue of CI currently failing on `master`.

Fixing the CI installation issue led to an updated `pylint` with new tests flagging previously unidentified "issues" (nothing behaviour-changing, more just removing redundant logic and using more Pythonic syntax). I've addressed all but one of them here, which I'm raising for closer inspection.

The Python documentation [here](https://docs.python.org/3/library/subprocess.html#popen-constructor) indicates that:

> The `preexec_fn` parameter is not safe to use in the presence of threads in your application. The child process could deadlock before `exec` is called.

When it is available ([Unix only, at least for Python2](https://docs.python.org/2/library/subprocess.html#popen-constructor)), I currently use `preexec_fn` to call `os.setpgrp()`, which ensures that any child processes get the same process group ID and are therefore killed if the parent is killed ([example documentation](http://ptspts.blogspot.com/2012/11/how-to-start-and-kill-unix-process-tree.html)).

This will have been the result of me faffing about for some time with `run.command()` and the signal handlers, trying to find a solution that demonstrates the correct behaviour on both Windows and Unix, both Python 2 and 3, and for different uses (e.g. Ctrl-C, piped commands where one of them fails, etc..). I've no idea if what I have is "right", it's just what behaved how I wanted it to do for the tests I did. So I'm tempted to just ignore the `pylint` error, but I felt it warranted raising in case there were any concerns about disobeying a bright red warning in the official Python documentation.


Lestropie:
I believe this usage was first introduced in 2956b244.

jdtournier:
Most of these changes look like those on #1393...?

Just checking though: is the issue with the signal handlers down to trying to set up custom handlers in the scripts? And if so, why were custom handlers required? Is that to ensure deletion of temporary files & folders?

jdtournier:
> Most of these changes look like those on #1393...?

My mistake - I think I'd gone through most of these changes last week, but I must have forgotten to push them or something, can't find them anymore... 

I'm happy with all of this. I've just added a directive to the remaining problematic call to ignore the warning. This isn't changing behaviour in any sense, it's only required to get `pylint` to pass the tests. And we need the tests to pass, there's a lot of pull requests pending that are currently stalling due to these tests failing... 

Hopefully this PR will pass its testing, in which case we merge, and worry about whether this use of `preexec_fn` is problematic or not later...

Issue #1398, opened 2018-07-13
Info on remote display issues page is outdated


tutufan:
The Linux and Display Issues documentation pages state that `mrview`/etc. won't work for remote displays.  This might have been true in the past, but it's not true for the current Debian release (for example), which has a software renderer that supports OpenGL 3.3.

Although `mrview` doesn't work natively on our RHEL release, I was able to get it work by running the Singularity/Debian container here: https://singularity-hub.org/collections/729  There'a pretty good chance that this would provide remote X11 display for almost any platform, and this alternative would be good to mention in the docs.

jdtournier:
Thanks for the info - that's the first I'd heard of it. My understanding of OpenGL handling was that it was handled by the GLX library, which explicitly didn't support anything beyond version 2. Maybe that's changed, or maybe the setup you refer to is different altogether? 

I'm also not familiar with these singularity containers (first time I've heard of these). Any chance I could get access to an instance to have a rummage...?

jdtournier:
Yes, I've seen the `mesa` software rasteriser used on different environments, for example VMs running in VMware, or using something like MobaXterm in combination with the Windows Subsystem for Linux. I'm not sure how to set that up reliably though. I don't think it's a server-only setup, since for instance MobaXterm works using the mesa software rasteriser, but fails when hardware acceleration is enabled - but this is a _client_-dependent parameter. I've just tried to render from my Arch Linux setup (up to date  as of last week, mesa 18.1.3-1), but that won't render over `ssh -X` to an Ubuntu 14.04 system (might because it's running the NVIDIA drivers, so using a different `libGL.so` - might look into that). So it sounds to me like there's some non-trivial interaction in there that I'm not confident I fully understand, and until I do, I'm hesitant to update the docs substantially... 

We could however mention that the mesa libraries can work with software rendering, but that setting this up properly is still unclear at this stage. Is any of this documented anywhere, BTW? I can't find anything relevant about OpenGL3+ remote rendering with mesa in a quick Google search, other than the links you provided...

tutufan:
As you say, there are things going on that seem under- or un-documented, and I also noticed on both the client and server side, merely having 3D hardware around seemed to effect whether this option worked.  This stuff seems fiendishly complicated, and even after all these years not really something we can expect to "just work".

[...]

PR #1411, opened 2018-07-27
docs: update remote display recommendations

jdtournier:
as discussed in #1398


jdtournier:
Suggest committing to `master` as doc only and applies to current (and previous) versions anyway...

PR #1412, opened 2018-07-29
Build tweaks

jdtournier:
Addresses issues raised in #1390. 

This changes the `configure` script so that it does _not_ modify `-Ipath` include directives to `isystem path`. This means we get all warnings from Eigen & Qt headers. To silence these, I've added `#pragma GCC diagnostic` directives around the inclusion of problematic headers (at least those that were causing warnings on my Win10 MSYS2 install). 

I've also fixed a few issues with the `build` script itself, see individual commits for details.

The main thing to check here is whether this silences _all_ warnings on all platforms - there's a good chance a few warnings might remain here and there. Please give this branch a shot with a full `./configure && ./build`, and send me the `build.log` if it produces any warnings at the `./build stage`...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
     # (Note: Run this one first, since it'll be the longest job)        #
     #####################################################################
     - os: linux
-      env: CFLAGS="-Werror -Wno-inconsistent-missing-override" TRAVIS_CXX=clang++-6.0 py=python3 test=run
+      env: CFLAGS="-Werror -Wno-inconsistent-missing-override -Wno-deprecated-register" TRAVIS_CXX=clang++-6.0 py=python3 test=run
     #######################################################################
     # Build binaries (without optimisation): GCC, Python2                 #
     # (Also ensures both Python 2 and 3 are tested for configure & build) #
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
     # (Note: Run this one first, since it'll be the longest job)        #
     #####################################################################
     - os: linux
-      env: CFLAGS="-Werror" TRAVIS_CXX=clang++-6.0 py=python3 test=run
+      env: CFLAGS="-Werror -Wno-inconsistent-missing-override" TRAVIS_CXX=clang++-6.0 py=python3 test=run
     #######################################################################
     # Build binaries (without optimisation): GCC, Python2                 #
     # (Also ensures both Python 2 and 3 are tested for configure & build) #
--- a/build
+++ b/build
@@ -180,7 +180,7 @@ def pipe_errors_to_less_handler():
       [ fid, name ] = tempfile.mkstemp()
       try:
         fid = os.fdopen (fid, 'wb')
-        fid.write ([ x.encode (errors='ignore') for x in colorize(error_stream) ])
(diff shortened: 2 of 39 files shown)


jdtournier:
OK, I think I've hit the end of the road with the `#pragma GCC diagnostic` idea. I currently get this error on the TravisCI build using `clang++6` on Linux:
```
In file included from src/gui/mrview/window.cpp:28:
In file included from src/gui/mrview/mode/base.h:24:
src/gui/mrview/window.h:55:11: error: 'metaObject' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
          Q_OBJECT
          ^
/usr/include/qt4/QtCore/qobjectdefs.h:162:32: note: expanded from macro 'Q_OBJECT'
    virtual const QMetaObject *metaObject() const; \
                               ^
/usr/include/qt4/QtGui/qmainwindow.h:65:5: note: overridden virtual function is here
    Q_OBJECT
    ^
/usr/include/qt4/QtCore/qobjectdefs.h:162:32: note: expanded from macro 'Q_OBJECT'
    virtual const QMetaObject *metaObject() const; \
```
This is where the Q_OBJECT macro gets expanded, and the error is raised in the injected code... To get around this, two options:

- relax the `-Werror` flag. Where does it even come from in the first place? No hint of it in `configure` or `build`. And I don't get it in my config on my Arch Linux system with `clang++6.0.0`. How did it get there?!?

- add the `-Wno-inconsistent-missing-override` option - not my preference, since that disables the warnings altogether.

jdtournier:
OK, I see that the `-Werror` flag is set in `.travis.yml`. We may need to add the `-Wno-inconsistent-missing-override` for the clang/Linux combination specifically then. It's not supposed to moan about this given that it's in the list of  `#pragma` ignores, but clearly that's not working as advertised...

jdtournier:
OK, good to see that things are actually working OK with this approach, but before merging, I reckon it's worth discussing the fact that it doesn't work properly with clang 6 + Linux. It makes me wonder whether it might fall over in other cases that are more relevant to us in the long term. I guess this only affects developers, so maybe we don't need to worry about it a great deal. The main thing is the only reason it's a problem during testing is because of the `-Werror` flag. Even if these warning do pop up on other people's system, at least they shouldn't cause compilation failures as such... 

I'm also wondering whether we need to `./build` with the `-nowarnings` flag during testing, if we're going to also have the `-Werror` flag in there anyway...?

Lestropie:
Historically build warnings have caused a fair bit of anxiety for users, especially when they're e.g. huge lists arising from templated header code as opposed to e.g. the odd signed-unsigned comparison. So would rather try to get on top of it properly, and not rely on build flags for Travis.

The other option would be to use "`-isystem `" for Eigen and Qt, `-I` for everything else... but the `#pragma`s should be the more appropriate solution, so maybe I should at least put a second set of eyes on that.

Lestropie:
No warnings for me on Linux, clang 6.0.0 / clang 6.0.1 / gcc 7.3.1, Eigen 3.3.4, Qt 5.10.1. So no idea why the pragmas are failing to suppress the warnings specifically on TravisCI. In which case maybe some `CFLAGS` just to shut Travis up *is* in fact the way to go...

PR #1416, opened 2018-08-01
mrconvert -vox: Permit single value input

Lestropie:
Prior to this change, providing a single floating-point value as input to the `-vox` option in `mrconvert` would result in rescaling of the voxel size along the first axis only. This change makes this behaviour more consistent with the `-voxel` option in `mrresize`: If a single value is provided rather than a comma-separated list, set the voxel size to that value along all three spatial axes.


cmd/mrconvert.cpp
```diff
@@ -259,9 +260,11 @@ inline vector<int> set_header (Header& header, const ImageType& input)
 
   opt = get_options ("vox");
   if (opt.size()) {
-    vector<default_type> vox = opt[0][0];
+    vector<default_type> vox = parse_floats (opt[0][0]);
```
jdtournier: Not a fan of implicit type-casting...? 😉

Lestropie: Generally yes; but when it's possible to provide either a single value *or* a comma-separated list to the same option, I'd rather be more explicit than less.

Plus it was like that in the `mrresize` code :man_shrugging:

Lestropie:
Was worried for a brief moment, but can confirm that "`-vox 1`" sets all three spatial axes to 1mm, whereas "`-vox 1,,`" modifies just the first axis; so there's no loss of capability in that respect.

cmd/mrconvert.cpp
jdtournier: Oh well... It is clearer that way 👍

PR #1418, opened 2018-08-03
set_path: add -remove option

jdtournier:
As suggested in #1381 - [this comment](https://github.com/MRtrix3/mrtrix3/pull/1381#issuecomment-405202733) in particular. Seems to work as expected in my hands, for both core and module folders.


jdtournier:
I'll merge this once testing completes if there are no objections? Just trying to clear things up a little...

PR #1422, opened 2018-08-08
Disable signal handlers with env var

jdtournier:
This allows signal handlers to be disabled with an environment variable. However, as noted in #1391, this may not necessarily be all that useful. 

A more useful feature on this pull request is that environment variables can now be documented in the same way as config file options, and will be updated with a call to `generate_user_docs`. 

This pull request also includes a few changes to quieten warnings with GCC 8.2.0 on MSYS2.

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -77,6 +77,9 @@ namespace MR
     std::string NAME;
     vector<ParsedArgument> argument;
     vector<ParsedOption> option;
+    //ENVVAR name: MRTRIX_QUIET
+    //ENVVAR Do not display information messages or progress status. This has
+    //ENVVAR the same effect as the ``-quiet`` command-line option.
     int log_level = getenv("MRTRIX_QUIET") ? 0 : 1;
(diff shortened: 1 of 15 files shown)


jdtournier:
Preview of environment variable docs [available on the mrtrix3-dev.readthedocs.io site](https://mrtrix3-dev.readthedocs.io/en/disable_signal_handlers_with_env_var/reference/environment_variables.html). 

I haven't documented other environment variable like `HOME` and `TMPDIR` since these are not MRtrix-specific, but maybe we should document the fact that we do interpret them...?

Lestropie:
> ... maybe we should document the fact that we do interpret them...?

Would take a bit of work since you'd need to differentiate between "used" and "defined" envvars in both their definitions and the parser. Only if you're willing to put in the effort, I don't think it's a huge necessity.

jdtournier:
> Only if you're willing to put in the effort

Guess that settles it then 🤣

PR #1425, opened 2018-08-13
docs: update install instructions, notably for Windows

jdtournier:
This brings Windows install instructions in line with Linux & macOS. 

As I'd [suggested in this forum post](http://community.mrtrix.org/t/installation-error-in-windows/1833).


jdtournier:
AppVeyor seems a bit stuck: status is _Expected — Waiting for status to be reported_. What do we do about this...?

Lestropie:
AppVeyor is configured to only attempt building PRs that have an `appveyor.yml` file (which `master` / this PR don't), and GitHub is configured to require a successful report from AppVeyor to enable a merge. Unfortunately it seems AppVeyor is not reporting "success" to GitHub in this instance.

Can either temporarily disable GitHub requiring AppVeyor success status for the sake of merges to `master`, or cherry-pick #1387 into `master`.

jdtournier:
OK, I think we can disable the AppVeyor tests temporarily - merges to `master` are not frequent. Cherry-picking #1387 into `master` is a pretty big change...

Other option might be to add a bogus `appveyor.yml` into `master` that simply reports success? It would then be overwritten by the correct one for the next release. Thoughts?

jdtournier:
In any event, a quick thumbs-up would be nice :grin:

Issue #1424, opened 2018-08-10
mrconvert: question about extracting from 4D volume


nooralsh:
Hi -- I have a set of fmri volumes in which every other timepoint is phase-encoded oppositely to its preceding timepoint (ie vol 0 is A-P, vol 1 is P-A, vol 2 is A-P, etc). I need to use all similarly encoded timepoints as two separate volumes, ie all A-P as one volume and all P-A as another. 

I know I can extract every other volume using `mrconvert` with the option `-coord 3 0,2,4,6,8,10,...,n` but is there an easier or more versatile way of defining which alternating volumes to extract? Especially since not all volumes might have the same number of timepoints.

Any help would be appreciated! Thanks.

nooralsh:
Actually.. I think I solved my own issue already, oops.

Lestropie:
Hi @nooralsh,

Just a quick follow-up:

In an ideal world, it would be possible to select these volumes using [the `-pe` option in `dwiextract`](http://mrtrix.readthedocs.io/en/latest/reference/commands/dwiextract.html#options-for-selecting-volumes-based-on-phase-encoding). However my suspicion is that while I've done my best to ensure that *MRtrix3* will correctly identify from DICOM if the phase encoding direction varies between volumes and [represent such information appropriately](http://mrtrix.readthedocs.io/en/latest/concepts/pe_scheme.html), the *sequences* used to acquire such data will unfortunately fail to properly encode the flipping of phase encoding direction within the DICOM headers.

This was certainly the case in my communication with the person responsible for Siemens DWI WIP development; are you able to add any information here from your own fMRI data?

Rob

jdtournier:
As another quick follow-up: the `-coord` option in `mrconvert` actually accepts a [number sequence](https://mrtrix.readthedocs.io/en/3.0_rc2/getting_started/command_line.html#number-sequences-and-floating-point-lists), and also understands the `end` keyword. This allows you to specify `-coord 3 0:2:end` to do what I _think_ you're trying to do. Could do with being better documented though, I agree...

nooralsh:
@jdtournier -- That is _exactly_ the solution I was looking for! Thank you for the suggestion, made my code much simpler. 

[...]

Issue #1086, opened 2017-08-10
Example usage in help page


Lestropie:
Each command could have one or more examples of how the command could be used. This is particularly relevant for commands that don't have a fixed argument list; e.g. `dwi2fod msmt_csd`, `tckedit` (for concatenating track files), `mrmath` with wildcard input, some commands e.g. `mrcalc`, `transformcalc` basically already have this but in the `DESCRIPTION` field.

Up for debate as to whether this field should be compulsory.

portokalh:
this would be very helpful IMHO. best!

Lestropie:
Just received another +1 for this due to `meshfilter` being designed like `mrfilter`, but currently only having one filter option, hence the interface looking kinda weird.

`mrconvert` is another that would I think benefit from an explicit example usage section.

jdtournier:
OK, that would require a bit of effort... Also, I'd suggest having a triplet, for operation, command, description, e.g:

---

## Examples

#### Extract every other volume from 4D dataset
```
$ mrconvert in.mif -coord 3 0:2:end out.mif
```
This makes use of number sequences and the 'end' keyword to select coordinates 0, 2, 4, ..., end along the volume axis (axis 3). 

#### Flip an image upside down
```
$ mrconvert in.mif -coord 2 end:0 out.mif
```
This makes use of number sequences and the 'end' keyword to select coordinates end, end-1, ... 1, 0 along the z-axis (axis 2), effectively flipping the image upside down. 

---

My only concern about this is that there are commands like `mrconvert` where the list of examples could grow _very_ large...

Lestropie:
> OK, that would require a bit of effort

A bit, sure; but it's not all that different from other command interface functionalities that are already there.

> I'd suggest having a triplet

:+1:

> My only concern about this is that there are commands like mrconvert where the list of examples could grow very large

Well it's already at a decent length; it's just all shoe-horned into the description section. Which is better than previous, when each individual option had a massively-long description.

jdtournier:
OK, let's do it. One more question though: where should the examples show up? Between description and options? Or after options? In most other software help pages, it tends to come after the options...

thijsdhollander:
Naturally, I'd have thought of them appearing right after the description (i.e. before the options)... however, with this in mind:

> My only concern about this is that there are commands like mrconvert where the list of examples could grow very large...

[...]

PR #1426, opened 2018-08-14
mrconvert update doc

jdtournier:
prompted by #1424

### DIFF
--- a/cmd/mrconvert.cpp
+++ b/cmd/mrconvert.cpp
@@ -119,7 +119,7 @@ void usage ()
   + Option ("coord",
             "retain data from the input image only at the coordinates "
             "specified in the selection along the specified axis. The selection "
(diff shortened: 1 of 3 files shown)


Lestropie:
Looks to me like another reason to raise #1086.

jdtournier:
> Looks to me like another reason to raise #1086.

Agreed. 

How would you envisage this working in practice, by the way? Essentially another variable like the current [App:DESCRIPTION](https://github.com/MRtrix3/mrtrix3/blob/master/core/app.h#L294), labelled `App::EXAMPLES`, which would allow for relatively free-form examples, or something more prescriptive than that? 

I'll merge this in the meantime...

jdtournier:
> I'll merge this in the meantime...

Assuming there's no objection, that is?

Issue #1345, opened 2018-05-21
build script writing to buffered stdout


Lestropie:
Currently, script `build` is writing some information to `stdout` and some to `stderr`, with higher-verbosity / higher-severity messages going to the latter. The primary terminal feedback during typical usage, i.e. the build item number and target file, is in fact going to `stdout`.

Where I've noticed this is when compiling *MRtrix3* inside a container. With local usage, the build progress typically only displays once every 100 items or so; not the expected behaviour, but not hugely problematic. Where this buffering is however annoying, which I'm encountering at the moment, is when using a cloud service that enforces a job timeout period when there is no output on `stdout` or `stderr` for some period of time. It's also kinda strange using the `-verbose` option when `stdout` is buffered, since the text output on `stdout` and `stderr` no longer properly align with one another.

While there's quite a number of solutions to the particular issue I've got, thought I'd raise the general issue here for opinions. E.g. Should *all* `build` output go to `stderr`? Or if retaining the separation between `stdout` and `stderr` is desired, we could define a [wrapper class](https://stackoverflow.com/a/107717) that calls `sys.stdout.flush()` after every `sys.stdout.write()` call.

jdtournier:
It's probably not a bad idea to throw everything out on `stderr`, and should be trivial to do. All terminal output is already wrapped through various calls (`disp()`, `error()`, `log()`, `report()`), so changes should be minimal.

In general, I guess we've always relied on `stdout` being line-buffered

thijsdhollander:
> It's probably not a bad idea to throw everything out on `stderr`

Yep, I also agree with this. :+1:

maxpietsch:
I think the buffered output of `build` is also an issue for building the test commands inside homebrew's `test` function, which deadlocks and aborts occasionally.

Lestropie:
Closed by #1348.

PR #1427, opened 2018-08-14
Appveyor take2

jdtournier:
Just experimenting a bit...


jdtournier:
OK, sorry about all the closing & reopening of this PR, it's the simplest way I could think to stop a build on AppVeyor... 

The changes I've made here seem to work, apart from the fact that the `./build` step produces no output. I'm not sure why. Either it's stalled for some reason, or its output is being lost or buffered somewhere (it's all going to `stderr` normally, although I've no redirected it to `stdout`). I'll wait to see if the build actually completes. In the meantime, I'm guessing the simplest way to debug this would be to log in directly via RDP? @Lestropie, I note you've already set things up for that, would it be easy for me to log in too?

jdtournier:
OK, seems to work. Clearly the output of the `./build` step is being buffered somewhere, and dumped in one go at the end. This is most likely due to the fact that it's pushing all its output to `stderr`. Maybe we could/should change this `stdout`? Didn't we change this recently...? 

I also need to check that errors actually get caught... It works on my system, in that a non-zero exit status does actually propagate to the `cmd` shell, but I'd like to be sure...

Lestropie:
Where did you find the function of those other `msys2_shell.cmd` options? Could have saved me a lot of gymnastics work.

> ... it's the simplest way I could think to stop a build on AppVeyor

I can stop & start builds from the AppVeyor site. I don't have an AppVeyor account, I just log in using GitHub credentials. I don't see any way of adding / removing users either. So I'd have thought that once you yourself log in there using GitHub, it should identify that you're part of the GitHub project, and hence give you access to the same controls?

If not, might need to create an "MRtrix3" user on AppVeyor and get the GitHub repo "re-registered" over to that; currently it may be too strongly linked to my account.

> In the meantime, I'm guessing the simplest way to debug this would be to log in directly via RDP? @Lestropie, I note you've already set things up for that, would it be easy for me to log in too?

You have to un-comment the "`ps:`" line in `appveyor.yml`; the build terminal output then provides you with RDP address & credentials.

> Clearly the output of the `./build` step is being buffered somewhere, and dumped in one go at the end. This is most likely due to the fact that it's pushing all its output to `stderr`. Maybe we could/should change this to `stdout`? Didn't we change this recently...?

Well `stderr is being redirected to `stdout`; I'd have thought `stdout` being buffered would be more likely? Nevertheless, `build` was altered to use only `stderr` in #1348, but that's only on `dev`, not `master` (merge to `dev` was 22 May, `3.0_RC3` tag was May 11).

> I also need to check that errors actually get caught

Testing now via 5d0acf8: https://ci.appveyor.com/project/Lestropie/mrtrix3/build/1.0.81

jdtournier:
> Where did you find the function of those other msys2_shell.cmd options? 

I read through that script... And tried a few of the options on my laptop - one of them seemed to do the trick :wink:

About access: I can log in with GitHub, but that doesn't give me access to your builds. Take a look here though: https://ci.appveyor.com/team

Finally: yes, I remember we changed all output to stderr (couldn't remember whether it was stderr or see stdout...). Might need to try setting it to stdout exclusively instead then, see if that helps. It's the only real difference I can see between the build and configure scripts...

And no problem with it being `dev` only - there's no appveyor testing on `master` anyway currently.

Lestropie:
> Take a look here though: 

Invitation sent. Though it's inviting you to *my* "team"; so an *MRtrix3* AppVeyor account might be preferable.

>...  no problem with it being dev only ...

Got it slightly mixed up with #1425, which was `master`. Nevertheless, in `appveyor.yml` you're currently explicitly redirecting `stderr` to `stdout`, presumably because `stderr` otherwise isn't echoed back to the Windows command prompt; so changing `build` to write to `stdout` is unlikely to have an effect on the buffering. From memory trying to control buffering of streams on Windows is kinda funky too. Maybe the better solution would just be to add `sys.stderr.flush()` calls? Or we could try the `PYTHONUNBUFFERED` environment variable.

-----

CI build appropriately [fails](https://ci.appveyor.com/project/Lestropie/mrtrix3/build/1.0.82) when a syntax error is introduced, so your changes haven't affected that.

Lestropie:
> Or we could try the `PYTHONUNBUFFERED` environment variable.

AppVeyor terminal output is unbuffered with this change.

jdtournier:
OK, I have access, no problem. Let's stick with that for now. 

I've just pushed a couple of commits: 

- stream all output from `build` & `configure` to `stdout`. These are not commands that will ever be used in pipes, so the motivation to write to `stderr` is mostly gone. It'll hopefully mean that these scripts play nicer in testing environments, and other similar contexts where they're wrapped further in other scripts. This was discussed previously on #1345, where we opted to push all output to `stderr`, but the idea of pushing to `stdout` with immediate flushes was also raised - this is what this commit does. Hopefully it won't also cause issues in other contexts...?

   This works for AppVeyor by the way: I can see the output of `build` updating in real time as I write this.  

- don't build with `-persistent`, since that clearly led to massive runtimes and an enormous logfile... We hopefully won't run into the out-of-RAM issues we had previously on TravisCI. If we do, we can always re-enable it.

jdtournier:
OK, it all seems to work. What do you reckon, @Lestropie? It's a relatively large changeset for `build`, but ultimately mostly tidying up loose ends that should have happened a long time ago. Pushing to `stdout` instead of `stderr` fixes the buffering issue (not sure how AppVeyor handles this, to be honest), and is unlikely to create any problems downstream - if anything it might avoid some future headaches. Otherwise, it's probably a cleaner solution all round for AppVeyor. 

Merge?

Lestropie:
Not sure about switching to `stdout` TBH. It seems to run counter to the rest of *MRtrix3*, and seems weird to go to `stdout` in order to *prevent* buffering. The buffering has been an issue historically in Travis / homebrew / containers, so need to check those as well. Ultimately I think the typical user needs to be a higher priority than the behaviour in CI; the latter was fixed trivially with that environment variable (did you by chance miss [this comment](https://github.com/MRtrix3/mrtrix3/pull/1427#issuecomment-413085993)?), so I don't fully see the justification for the flip on #1345.

> Don't build with `-persistent`

:+1:

jdtournier:
I hear you concerns. I did see the commit with that variable, but I think it's cleaner just to use `stdout`, and IMO is probably the right thing to do in the long run. 

> It seems to run counter to the rest of MRtrix3

I don't think the design decisions implicit in the software itself should necessarily translate to the scripts used to build it. In particular, the only concrete reason for our heavy use of `stderr` in our commands is because pipelines are intrinsic to the design. This is not the case with `configure` and `build` - there is no expectation that the output of these scripts might be piped or redirected anywhere. 

But if you think about it, the only thing that makes sense to pipe or redirect is what would show up on the terminal anyway, so that would argue that the whole output _should_ in fact go to `stdout`. Currently, if you typed `./build > log`, you'd get an empty log and all output going to the terminal anyway - probably not what would be expected. And as we're currently finding out, as soon as these scripts are used on other platforms where their output is redirected, these platforms will overwhelmingly cater for `stdout`, sometimes treating `stderr` as a bit of an afterthought.

> seems weird to go to stdout in order to prevent buffering. 

Well, it seemed to handle `stdout` as expected for the other commands (and now for the updated `build`),  which is why I started thinking about this approach in the first place. But I wouldn't have tried this if I thought it was important to stick to `stderr`, given you already had a fix in place. It did however prompt me to think about the issues above, and on balance I reckon it might be a better long-term approach - it happens to fix this issue, but that's actually not the main reason I think we should adopt it.

> The buffering has been an issue historically in Travis / homebrew / containers, so need to check those as well.

If I recall, it wasn't the buffering that was the issue on Travis, but the attempt to pipe the errors to `less`, which caused truncated outputs and stalls - might be remembering wrongly though. Not sure about homebrew, but if anything, I'd expect it would handle a pure `stdout` output better than a pure `stderr`. On containers, the issue you mentioned was the lack of line-buffering, but hopefully the `flush()` calls will deal with this as expected. But yes, it would need checking. 

> I think the typical user needs to be a higher priority than the behaviour in CI

Fully agree, but I don't see how this applies here...? If anything, the ability to pipe or redirect the output of `build` without having to fiddle about with `stderr` redirection sounds like a bonus for users too? .

jdtournier:
Any further thoughts regarding pushing all output from `build` and `configure` to `stdout`...? I'd like to merge this, since AppVeyor CI is currently the bottleneck for merging (it only starts a job at a time, and we currently have a matrix of 2).

Lestropie:
> If I recall, it wasn't the buffering that was the issue on Travis, but the attempt to pipe the errors to `less`, which caused truncated outputs and stalls

There was that, but I think there was also output buffering earlier than that.

> Any further thoughts regarding pushing all output from build and configure to stdout...?

I think it's OK; though I'd have preferred to [disable buffering for all `sys.stdout.write()` calls](http://algorithmicallyrandom.blogspot.com/2009/10/python-tips-and-tricks-flushing-stdout.html) rather than relying on devs to always include the extra `flush()` call.

> I'd like to merge this, since AppVeyor CI is currently the bottleneck for merging 

Depends on how confident you are that any divergent behaviour between Python 2 and 3 on Windows will be echoed in the Linux tests. It's probably safe, but given some of the gotchas we've fallen for of late I wouldn't put it past ol' Uncle Murphy.

> it only starts a job at a time, and we currently have a matrix of 2

Free accounts don't get parallel builds, full stop.

jdtournier:
> I think it's OK; though I'd have preferred to disable buffering for all sys.stdout.write() calls rather than relying on devs to always include the extra flush() call.

You'll find that all output in both `configure` and `build` should now go through the appropriate calls (`report`, `log`, `error`, `fail`) rather than by direct writes to `stdout`. This also guarantees terminal output  is additionally written to the logfile, as well as thread-safe behaviour. So devs really shouldn't ever need to flush manually... If they are, they should update their code to use these convenience calls. It also makes it trivial to switch back to `stderr` (or any other suitable strategy) if this turns out to be problematic... 🤞 

> Depends on how confident you are that any divergent behaviour between Python 2 and 3 on Windows will be echoed in the Linux tests. 

I don't think it's critical to test behaviour under different conditions than our instructions would give rise to. If you follow our MSYS2 install guide, you should end up with Python 3 only. If we could run tests concurrently, I'd be happy enough to test both versions (even if I reckon it's overkill), but if it now extends our test time even further, I don't think it's worth it... 

In general, I'm not convinced we need to test Python 2 behaviour anyway: Windows-specific problems are (probably) more likely to arise on the C++ side, and Python issues are likely to be:
- syntax-related (so should be picked up on the Linux tests);
- due to unavailable modules, which _may_ be possibly be available in v3 but not v2 on Windows, but available in both versions on Linux - unlikely to be a common occurrence (?);
- or down to OS idiosyncrasies, which I reckon would manifest with either version.

Another option would be to additionally run the _tests_ only under Python 2? That would only add a couple of minutes, and at least verify correct runtime operation... We could also run the `configure` script under Python 2 after the initial build, and check that the `config` matches the Python 3 version? The run `./build` again under Python 2, which shouldn't build anything, but will at least run through all dependencies and build the todo list. We could additionally `touch` one cpp file in `core/` to force it to be re-built, which should test most parts of the script (at least most of those that the Python 3 test will test...). What do you reckon? I'll give that a shot now...

jdtournier:
OK, I've just pushed a couple of extra lines to the current appveyor config to re-run a quick re-build and configure under python2: this all checks out fine. But I'd also tried to run the `run_test` script under python2, when it's a bash script... 🤦‍♂️  No point in re-running the tests in that case. Is there a simple way to force the python version for our scripts? 

I note that the same situation exists on travisCI: the current testing matrix will also not check that our scripts work under python2. Do we want to rectify this?

jdtournier:
One more point of note: with the latest changes, which try to exercise `build` and `configure` under Python 2, the `configure.log` and `build.log` files that will be uploaded as artefacts will be the Python 2 versions. Not necessarily a big deal since a failure at the original configure or build stage should mean that those correct versions get uploaded (I assume it won't try to carry on running through the tests when a failure is encountered...?). Also, the second `python2 ./configure` is a full run, so its log file should be the same as the python3 one (that's what we test for here anyway). But the last build is deliberately much shorter, so we'll lose any information for the full run under Python 3. Not sure how big a deal this is. Might be worth renaming the python3 build.log and loading it as a separate artefact...?

jdtournier:
I think it's time this got merged (or canned). I've stripped out the Python2 tests that I'd added a while back, I think they needlessly complicate things. We really only need to catch issues unlikely to have already been picked up with the Linux tests, I don't think we need to cover all possible scenarios on Windows too...

The other issue that was left unresolved in this PR was the output going to `stdout` (with explicit flushes) rather than `stderr`. I'd made the case for it in [this post](https://github.com/MRtrix3/mrtrix3/pull/1427#issuecomment-413511431), but there's not been any actual debate about it since. Can I assume everyone is happy enough this the idea...?

Lestropie:
Confirm no buffering when building in a container, & AppVeyor gives unbuffered build updates. So good to go, as long as people stick to the convenience functions in `build` and `configure` and don't go adding direct `sys.stdout` calls.

Lestropie:
OK It's coming back to me now...

One of the reasons for the gymnastics was that if any stage of testing fails, *none* of the artifacts get uploaded. So if e.g. one of the runtime tests fail (like it has here), you only get the terminal output saying which test(s) failed, you can't access `testing.log` to find out *why* it failed.

Let's see if that change does as required without the same degree of clumsiness as before.

jdtournier:
OK, I reckon it might indeed be best to stick with artifacts for the logs. The output is already verbose enough... I've tried to push the output of the `pacman` package install to its own log too, to avoid the dump from the Qt install. I've also tried to use an environment variable to clean up the scripts a bit, let's see if that works... 

Problem now is AppVeyor doesn't seem to trigger when I push. Looking at the settings, it looked like it was set to build `master` only. I've tried to change that on the AppVeyor side, but I'll also set it to build `dev` and `master` in the config file too - if only to see if that triggers a re-build...

jdtournier:
Right, all my attempts at using environment variables have been a miserable failure. I've dumped all these attempts to keep things clean. Otherwise, I've stuck with the artifacts to store the logs, I think it's the best way to do this, even if it's not what it was designed for. 

@Lestropie, feel free to merge if you're happy with the changes as they stand.

Lestropie:
The `branches:` directive in `appveyor.yml` is intended for building particular branches in order to produce distributables; it should not have had any effect on the building of PRs by my understanding, that's a completely different webhook within the GitHub settings. I don't know whether that change has just happened by chance to correlate with some independent fix to the web hooks / AppVeyor, making it only appear as though it was the fix to some other issue.

jdtournier:
You're most likely right re branches. I did notice that pushes to `dev` weren't triggering a fresh build, and saw in the AppVeyor settings that it was set to build `master` only - so I changed that, _and_ added the `branches` directive. Probably not needed...

No idea why you'd want to push the one-line `python --version` output to the artifacts though, it's perfectly fine (in fact, probably better) straight in the console log...

jdtournier:
> No idea why you'd want to push the one-line python --version output to the artifacts though, it's perfectly fine (in fact, probably better) straight in the console log...

??? I swear I saw a commit that pushed the python version to file... Seems to have disappeared though. Must be losing the plot - ignore me...

Lestropie:
> No idea why you'd want to push the one-line `python --version` output to the artifacts though

Was thinking that it would be better if going back through build histories to be able to find that information alongside the other key information as stored in the other artifacts, given that the terminal output can be exceptionally long particularly due to `pacman`. But because the latter is now piped to a file, the console output is now much cleaner. Hence the force push above.

jdtournier:
> Was thinking that it would be better if going back through build histories to be able to find that information alongside the other key information as stored in the other artifacts

Sounds sensible - until you find out that [AppVeyor by default deletes artifacts after 6 months](https://www.appveyor.com/docs/packaging-artifacts/#artifacts-retention-policy)... Best to keep it in the console output from that point of view alone. :+1:

PR #1429, opened 2018-08-16
amp2response: Exception on invalid transform

Lestropie:
Mitigation of reported errors such as [here](http://community.mrtrix.org/t/version-3-0-dwi2response-gives-nan/1232) and [here](http://community.mrtrix.org/t/nan-in-response-wm-txt/1767) due to misuse of `-lmax` option.


cmd/amp2response.cpp
```diff
@@ -297,6 +297,8 @@ void run ()
 
         // Generate the ZSH -> amplitude transform
         Eigen::MatrixXd transform = Math::ZSH::init_amp_transform<default_type> (rotated_dirs_azel.col(1), lmax[shell_index]);
+        if (!transform.allFinite())
+          throw Exception ("Unable to construct A2SH transformation for shell b=" + str(int(std::round((*shells)[shell_index].get_mean()))) + "; lmax (" + str(lmax[shell_index]) + ") may be too large for this shell");
```
jdtournier: You could break that up over multiple lines if you want, it might be cleaner:
```C++
if (!transform.allFinite()) {
  Exception excp ("Unable to construct A2SH transformation for shell b=" + str(int(std::round((*shells)[shell_index].get_mean()))));
  excp.push_back ("  lmax (" + str(lmax[shell_index]) + ") may be too large for this shell");
  throw excp;
}
```

thijsdhollander:
In practice, this is 99% of the time going to happen due to specifying a non-zero lmax for b=0, I reckon (as in both of the referenced cases).  Technically, the exception text is entirely correct, but maybe we could add a little something that would help users specifically in the non-zero lmax for b=0 case (otherwise, I'm afraid most of them will *still* end up on the forum; if they made that mistake to begin with, the error message may still be too confusing for them, I reckon).

Maybe this is an idea: in case of non-zero lmax for "b=0" (using the bzero definition of either >10 or the config entry of course), show a warning at the start of the run.  If it then runs into this error subsequently, at least the warning will also sit there as a strong suggestion.  Still just a warning though; there *could* be a case where someone has b=5 or something data for actual different gradient directions, and would still be keen to get an lmax=2 or something response for that (let's say, in a multi-shell context or something).  Seems unlikely, but you never really know... could be a genuine scenario.

So well, maybe add that kind of warning at the start; I reckon it may help a lot of users who run into the scenarios like those reported on the forum.

Lestropie:
I thought of that, but had a surge of laziness. The threshold for classifying *b*=0 can behave identically to the C++ code using [this snippet](https://github.com/MRtrix3/mrtrix3/blob/7f97033fdf28e11a64706a3ac0c1a2b76765d056/bin/dwipreproc#L434-L436).

The help text for `-lmax` could potentially be tweaked also. It's not immediately clear that it corresponds to the *lmax* at different shells of the WM response function, as opposed to e.g. the *lmax* of different tissues.

thijsdhollander:
> I thought of that, but had a surge of laziness. The threshold for classifying b=0 can behave identically to the C++ code using this snippet.

Yep, aware of that, I'm actually using that snippet now in a part of my improvements to the `dwi2response dhollander` algorithm (in an offline branch for now).

Here though, this would be in C++, right?  I'm referring here to the `amp2response` command/binary itself (to give this warning).

> The help text for -lmax could potentially be tweaked also. It's not immediately clear that it corresponds to the lmax at different shells of the WM response function, as opposed to e.g. the lmax of different tissues.

Similar here (although it could be further clarified in the dwi2response help test for this), this could (also) be done for the `amp2response` command itself.  However, the confusion there may be inherently less, since there's no different tissue types when interfacing with `amp2response` directly.

Lestropie:
I interpreted "at the start of the run" to mean upon running `dwi2response`, i.e. before the script algorithm is even invoked. Generating a warning at the start of `amp2response` rather than at the transform construction step wouldn't provide much of an advance warning.

thijsdhollander:
Hmm, yes, that would also not show in a script, right?  I was somehow reasoning it would also be visible at the stage the script run stops when the error happens.  That's actually what I was after in general: something that shows very close to the line of the error, which is typically (always) at the end of the dwi2response algorithms.  Algorithms like `tournier` and `dhollander` generate a massive output that advances quite fast from the start.  If that errors at the `amp2response` step, typical users would probably still not spot the warning that sits all the way at the top of the command line output at that stage.

Lestropie:
For any changes in Python I was expecting that identification of *l*>0 for a shell classified as *b*=0 would cause an immediate error, rather than a warning. In the unlikely case that someone actually wants to do such a thing with their data, they would need to alter the config file in order for other *MRtrix3* commands to work anyway.

Perhaps instead of having a warning at the start of `amp2response`, the exception in `amp2response` should be made multi-line as @jdtournier suggested, and an additional line added if *MRtrix3* considers the shell to be *b*=0, e.g. "This shell appears to correspond to no diffusion weighting, and hence lmax must be set to zero for this shell".

thijsdhollander:
> For any changes in Python I was expecting that identification of l>0 for a shell classified as b=0 would cause an immediate error, rather than a warning. In the unlikely case that someone actually wants to do such a thing with their data, they would need to alter the config file in order for other MRtrix3 commands to work anyway.

If they provide the lmax option consistently (and in a scenario where lmax>0 makes sense for their "b=0" or "b = very low" data of course), a lot of commands would still work perfectly, I reckon.

> Perhaps instead of having a warning at the start of amp2response, the exception in amp2response should be made multi-line as @jdtournier suggested, and an additional line added if MRtrix3 considers the shell to be b=0, e.g. "This shell appears to correspond to no diffusion weighting, and hence lmax must be set to zero for this shell".

I had missed that suggestion, but that would be a good solution indeed.  It actually makes more sense than my warning suggestion: if it doesn't lead to actual problems, the warning I suggested earlier may be misleading.  If they do request an lmax>0 and the data actually allows for it, then it might be less likely they are mistaken (and probably consciously intended to do this).

thijsdhollander:
I think this one is all good now.  @Lestropie , @jdtournier any further comments/changes planned?  If not, let's merge to `dev`.

Lestropie:
Outstanding was whether or not we want to detect the use of *l*>0 for *b*=0 in `dwi2response` as opposed to deferring entirely to `amp2response`.

jdtournier:
This is looking good. On the outstanding issue: I think it's sensible to provide a hefty warning if we find _l_>0 for _b_=0, but I don't think we necessarily want to prevent it altogether. There's a few exotic situations that pop up every so often when it would be nice to allow it (can't think of any right now, but I _think_ it has come up on occasion...). However, as @thijsdhollander, a mere warning at the start is likely to go completely unnoticed. One way to deal with this might be to abort in this case, but suggest a way to proceed if that is actually what is desired - maybe using an option to force it to run anyway (could even be `-force`? Not convinced). Thoughts?

thijsdhollander:
I'm actually 100% ok with how it is now in the pull request. The warning I suggested earlier on was just meant to be somehow contextual; i.e., people would have to see it in case the command fails due to invalid transform.  I first suggested to show the warning in general up front, so by the time it *potentially* errors out, there's a real-world explanation (not merely a very technical one about the transform).  However, that goal is now reached (but differently) by having the multi-line exception; and it's even better than my first suggestion, because the explanation only pops up when it actually fails due to invalid transform.  Also, with the current script environment, the exception should then indeed also shown upon the relevant failure (and that's only where it's relevant indeed).

So I'd keep it as it's now in this pull request.  I wouldn't abort just when someone asks for lmax>0 for b=0; there's indeed valid scenarios for that that shouldn't become any more difficult to run when needed.

thijsdhollander:
Oh sorry, I may have mis-seen what's currently in the suggested code; it is in fact a warning (I think).  So I'd make the error / exception itself multi-line.  I.e. when it effectively fails due to the transform, then (and only then) show an additional error / exception line *if* this happens for a b=0 shell.

jdtournier:
OK, I think it makes sense to leave the handling of this issue within `amp2response` - as @thijsdhollander says, this means the warning & error pop up at the point of failure, which will be the first thing users will see. Seems reasonable. 

> So I'd make the error / exception itself multi-line.

It already is. Both the error and the initial warning are multi-line. But the warning will provide additional context for the error as well, which I think is helpful. 

> when it effectively fails due to the transform, then (and only then) show an additional error / exception line if this happens for a b=0 shell.

Are you suggesting _not_ showing a warning when _l_>0 for a _b_=0 shell? I'm not so sure. I think that warning should be there, in case users request that by mistake, but their acquisition actually has 0<_b_<10 as the _b_=0 (which is the case for most Siemens acquisitions, if I'm not mistaken?) and non-zero gradient vectors. There's a good chance that the command will succeed in this case, but it would still make no sense to use _l_>0. If we leave the warning in place, the only people who might see it when they shouldn't are those that are doing something unusual anyway - in which case I think it's appropriate that they see it anyway...

Lestropie:
> However ... a mere warning at the start is likely to go completely unnoticed.

Currently yes; though with #1024 such warnings would be repeated to the user upon script failure.

> There's a few exotic situations that pop up every so often when it would be nice to allow it

My thinking was that if there were a use case that necessitated such, `BzeroThreshold` would need to be modified by the user in order for any other commands to work properly anyway; hence requiring this variable to be changed for `dwi2response` wouldn't actually be an inconvenience. But I'm not too fussed.

> ... when it effectively fails due to the transform, then (and only then) show an additional error / exception line if this happens for a b=0 shell.

Already there: the raised exception itself gets an additional line if the shell is classified as *b*=0, over and above the warning that's printed earlier in such a use case.

thijsdhollander:
:+1:

PR #1431, opened 2018-08-16
mtnormalise tweaks

Lestropie:
Bunch of minor things I did while mucking around with this command. Details in commit message of 28848f6.

### DIFF
--- a/cmd/mtnormalise.cpp
+++ b/cmd/mtnormalise.cpp
@@ -24,7 +24,7 @@
 using namespace MR;
 using namespace App;
 
-#define DEFAULT_NORM_VALUE 0.28209479177
+#define DEFAULT_REFERENCE_VALUE 0.28209479177
 #define DEFAULT_MAIN_ITER_VALUE 15
 #define DEFAULT_BALANCE_MAXITER_VALUE 7
 #define DEFAULT_POLY_ORDER 3
@@ -55,8 +55,6 @@ void usage ()
     + Argument ("input output", "list of all input and output tissue compartment files. See example usage in the description.").type_various().allow_multiple();
 
   OPTIONS
-    + OptionGroup ("Options that affect the operation of the mtnormalise command")
-
     + Option ("mask", "the mask defines the data used to compute the intensity normalisation. This option is mandatory.").required ()
     + Argument ("image").type_image_in ()
 
@@ -66,15 +64,15 @@ void usage ()
     + Option ("niter", "set the number of iterations. (default: " + str(DEFAULT_MAIN_ITER_VALUE) + ")")
     + Argument ("number").type_integer()
 
-    + Option ("value", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
-                       "(default: " + str(DEFAULT_NORM_VALUE, 6) + ", SH DC term for unit angular integral)")
+    + Option ("reference", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
+                           "(default: " + str(DEFAULT_REFERENCE_VALUE, 6) + ", SH DC term for unit angular integral)")
     + Argument ("number").type_float (std::numeric_limits<default_type>::min())
 
     + Option ("balanced", "incorporate the per-tissue balancing factors into scaling of the output images "
                           "(NOTE: use of this option has critical consequences for AFD intensity normalisation; "
                           "should not be used unless these consequences are fully understood)")
 
-    + OptionGroup ("Options for outputting data to verify successful operation of the mtnormalise command")
+    + OptionGroup ("Debugging options")
 
     + Option ("check_norm", "output the final estimated spatially varying intensity level that is used for normalisation.")
     + Argument ("image").type_image_out ()
@@ -315,8 +313,8 @@ void run_primitive () {
     throw Exception ("Mask contains no valid voxels.");
 
 
-  const float normalisation_value = get_option_value ("value", DEFAULT_NORM_VALUE);
-  const float log_norm_value = std::log (normalisation_value);
+  const float reference_value = get_option_value ("reference", DEFAULT_REFERENCE_VALUE);
+  const float log_ref_value = std::log (reference_value);
   const size_t max_iter = get_option_value ("niter", DEFAULT_MAIN_ITER_VALUE);
   const size_t max_balance_iter = DEFAULT_BALANCE_MAXITER_VALUE;
 
@@ -470,7 +468,7 @@ void run_primitive () {
           combined_tissue.index(3) = j;
           sum += balance_factors(j) * combined_tissue.value() ;
         }
-        y (index++) = std::log(sum) - log_norm_value;
+        y (index++) = std::log(sum) - log_ref_value;
       }
     }
 
--- a/cmd/mtnormalise.cpp
+++ b/cmd/mtnormalise.cpp
@@ -52,9 +52,11 @@ void usage ()
    + "Example usage: mtnormalise wmfod.mif wmfod_norm.mif gm.mif gm_norm.mif csf.mif csf_norm.mif -mask mask.mif.";
 
   ARGUMENTS
-    + Argument ("input output", "list of all input and output tissue compartment files. See example usage in the description.").type_image_in().allow_multiple();
+    + Argument ("input output", "list of all input and output tissue compartment files. See example usage in the description.").type_various().allow_multiple();
 
   OPTIONS
+    + OptionGroup ("Options that affect the operation of the mtnormalise command")
+
     + Option ("mask", "the mask defines the data used to compute the intensity normalisation. This option is mandatory.").required ()
     + Argument ("image").type_image_in ()
 
@@ -64,6 +66,16 @@ void usage ()
     + Option ("niter", "set the number of iterations. (default: " + str(DEFAULT_MAIN_ITER_VALUE) + ")")
     + Argument ("number").type_integer()
 
+    + Option ("value", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
+                       "(default: " + str(DEFAULT_NORM_VALUE, 6) + ", SH DC term for unit angular integral)")
+    + Argument ("number").type_float (std::numeric_limits<default_type>::min())
+
+    + Option ("balanced", "incorporate the per-tissue balancing factors into scaling of the output images "
+                          "(NOTE: use of this option has critical consequences for AFD intensity normalisation; "
+                          "should not be used unless these consequences are fully understood)")
+
+    + OptionGroup ("Options for outputting data to verify successful operation of the mtnormalise command")
+
     + Option ("check_norm", "output the final estimated spatially varying intensity level that is used for normalisation.")
     + Argument ("image").type_image_out ()
 
@@ -71,9 +83,8 @@ void usage ()
                             "This mask excludes regions identified as outliers by the optimisation process.")
     + Argument ("image").type_image_out ()
 
-    + Option ("value", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
-                       "(default: " + str(DEFAULT_NORM_VALUE, 6) + ", SH DC term for unit angular integral)")
-    + Argument ("number").type_float (std::numeric_limits<default_type>::min());
+    + Option ("check_factors", "output the tissue balance factors computed during normalisation.")
+    + Argument ("file").type_file_out ();
 
 }
 
@@ -224,7 +235,7 @@ void run_primitive () {
   vector<Header> output_headers;
   vector<std::string> output_filenames;
 
-  ProgressBar input_progress ("loading input images");
+  ProgressBar input_progress ("loading input images", 3*argument.size()/2);
 
   // Open input images and prepare output image headers
   for (size_t i = 0; i < argument.size(); i += 2) {
@@ -482,7 +493,7 @@ void run_primitive () {
 
   progress.done();
 
-  ProgressBar output_progress("writing output images");
+  ProgressBar output_progress("writing output images", output_filenames.size());
 
   opt = get_options ("check_norm");
   if (opt.size()) {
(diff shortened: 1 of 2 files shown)


cmd/mtnormalise.cpp
```diff
@@ -64,16 +66,25 @@ void usage ()
...
 
+    + Option ("value", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
+                       "(default: " + str(DEFAULT_NORM_VALUE, 6) + ", SH DC term for unit angular integral)")
+    + Argument ("number").type_float (std::numeric_limits<default_type>::min())
+
+    + Option ("balanced", "incorporate the per-tissue balancing factors into scaling of the output images "
+                          "(NOTE: use of this option has critical consequences for AFD intensity normalisation; "
+                          "should not be used unless these consequences are fully understood)")
+
+    + OptionGroup ("Options for outputting data to verify successful operation of the mtnormalise command")
```
jdtournier: That feels a big long. How about simply "debugging options"?

cmd/mtnormalise.cpp
```diff
@@ -52,9 +52,11 @@ void usage ()
    + "Example usage: mtnormalise wmfod.mif wmfod_norm.mif gm.mif gm_norm.mif csf.mif csf_norm.mif -mask mask.mif.";
 
   ARGUMENTS
-    + Argument ("input output", "list of all input and output tissue compartment files. See example usage in the description.").type_image_in().allow_multiple();
+    + Argument ("input output", "list of all input and output tissue compartment files. See example usage in the description.").type_various().allow_multiple();
 
   OPTIONS
+    + OptionGroup ("Options that affect the operation of the mtnormalise command")
```
jdtournier: I'd rather the first section was left to default unless there's a very strong argument against. Here, these are the 'regular' options that users will likely want to use, the next section is indeed separate. But I reckon the standard section header is fine here.

cmd/mtnormalise.cpp
```diff
@@ -64,16 +66,25 @@ void usage ()
     + Option ("niter", "set the number of iterations. (default: " + str(DEFAULT_MAIN_ITER_VALUE) + ")")
     + Argument ("number").type_integer()
 
+    + Option ("value", "specify the (positive) reference value to which the summed tissue compartments will be normalised. "
```
jdtournier: I find `-value` a bit nondescript. I'd rather use `-reference` or `-sum` or something more expressive?

cmd/mtnormalise.cpp
Lestropie: Trouble is they may not be used solely for debugging :smiling_imp:

cmd/mtnormalise.cpp
Lestropie: I tend to try to organise *all* groups into options as soon as there are option groups defined within the `cmd/` file. Most likely cause is that if you either locally define or import from a header file an option group, and then subsequently define a "root-level" option, that option will *appear* in the docs / help page to belong to the option group - or at least it has in the past, I don't know whether this has changed. I'll look into this more directly instead.

cmd/mtnormalise.cpp
Lestropie: :+1: for `-reference`

cmd/mtnormalise.cpp
thijsdhollander: I agree with @jdtournier here; I'm keen on seeing a better justification for this than just ":smiling_imp:".  People can be creative however they like, but it doesn't mean we have to endorse using certain outputs for certain purposes without a justification.  There's a certain context within which these numbers exactly live (Dave, Rami and myself *very* carefully thought about this and discussed this at length at the time; and for very good reasons), and using them outside that context can actually be quite misleading in this very case.  So well, happy to understand the justification before I can endorse this.

cmd/mtnormalise.cpp
thijsdhollander: Happy with '-reference'; definitely *not* '-sum' though: that suggest too much that the output will strictly sum to that value, which `mtnormalise` by definition of its very purpose does not intend to achieve.  The reason for the original name '-value` was discussed among Dave, Rami and myself: to indicate that it's merely a reference value (in a way, this is a sort of "necessary" evil for a normalisation step like this to act across subjects and populations; the reference value is the (not so) hidden means by which they communicate to each other indirectly, yet without requiring a script/command like the old strategy that needs to take all data at once and doesn't allow for future addition of individual subjects without rerunning everything).  So well, long story short: "`-value`" intended to come across as "just reference value".  `-reference` does that as well, so happy with that if that makes things more clear.

cmd/mtnormalise.cpp
thijsdhollander: In general, I agree with @jdtournier on this point; the very usage of the "default" section communicates an important point to the user, I reckon (as Donald essentially says).  However, in this very case, if we're reorganising the option categories for this command any way, I'd argue a slightly different organisation: the `-mask` and `-order` options as the only ones in the default option section, and the other ones (yet still the *non debugging ones*, those should be in the last debugging section, that is) in a section called something like "advanced options" (happily suggest a better name, this is just the general idea of it).  The reason is two-fold: first, `-mask` is actually mandatory for this command since the complete overhaul of the algorithm that Rami and I spent quite some time/effort on (the one that quite fundamentally changed it by taking it to the log domain, but also changed the nested iteration structure, added outlier rejection, etc...).  Second, the other options, in the case of this algorithm, you'd "*almost*" never touch (`-niter`, and `-value` or `-reference` as suggested).  The `-balanced` option is another discussion; I'm *strongly* against having that one available, we originally *took it out* for good reasons as well.  I'll argue that later as a separate comment on that change, so things don't get too messy here.

....but!  That would then only leave 2 default options, and 2 "advanced" ones, which is maybe a bit silly to have separate categories for...?  If so indeed, and if still putting them together in a single one, I'd argue for the default category, mostly because `-mask` is mandatory, but also because `-order` may in some scenarios be considered carefully, so shouldn't be "hidden" *too* far away. (as I at some point, I think, detailed when I made a very lengthy post about how `mtnormalise` works and why, `-order` can even be set to 0, which results in intensity normalisation without bias field correction: a very well defined mode of operation that can be useful in some pipelines, or even when using `mtnormalise` in different spots in existing pipelines)

So far my 2 (or more) cents on the categories.

cmd/mtnormalise.cpp
```diff
@@ -64,16 +66,25 @@ void usage ()
...
+                       "(default: " + str(DEFAULT_NORM_VALUE, 6) + ", SH DC term for unit angular integral)")
+    + Argument ("number").type_float (std::numeric_limits<default_type>::min())
+
+    + Option ("balanced", "incorporate the per-tissue balancing factors into scaling of the output images "
```
thijsdhollander: So, strongly against including this: I argued in the past why we *removed* it, and this was at length discussed among Dave and I, and later Rami, when we originally designed the original algorithm and later on redesigned it from the ground up.  I sadly don't have the time today to write up a very lengthy text on this (again), as I'm dealing with a few other urgent deadlines sadly, but wanted to mention this already now in an attempt to be open and clear about this.  The short, somewhat simplified version of this is essentially: this strategy (i.e. switching this option on) *breaks* the requirement of using a single set of (average, or other) response functions across a group / groups of subjects in quantitative (A)FD studies, such as an FBA and others.  If there is in fact a strong kind of (overall) "miscalibration" of the response functions, or more specifically their magnitude, relatively speaking for the different tissue types (e.g. the CSF response function is 10 times to large in signal magnitude or something, relative to the WM/GM ones), then the only way to deal with this is also "globally": I'm not sure if my words are entirely clear here, but what I consequentially mean by this, is that (in a certain way of looking at this) in principle the balance factors within this algorithm should be the same for all subjects that one wants to normalise with respect to each other.  But this is the annoying struggle that Dave and I originally discussed so long about: there is no simple or elegant way to have such a property in place in a context where this step would have to be able to run on individual subjects.  The latter desire originally mostly arose due to the Docker / BIDs pipelines kind of design, where Dave tried to avoid such points of "the data having to come together across subjects" as much as possible in the FBA pipeline.
So well, balance factors are still a necessary evil for the basic strategy of the `mtnormalise` algorithm, and they have to be used internally at some point to have that flat-looking (or flat-wanting-to-be) map that the bias field is fit to, but they are only and internal construct, and exporting it "into" the result breaks the very definition of the kind of operation `mtnormalise` aims to perform, which is "applying a single bias/"normalisation"-field globally to the dMRI data, represented as the set of tissue types being fed to the algorithm".  Changing internal contrasts to this input (e.g. among the tissue types relatively), and doing so differently for different subjects, defeats the normalisation purpose that `mtnormalise` specifically aims for.  We're already having more than trouble enough to get this normalisation thing (and "single unique set of responses" thing) across; so I'm strongly against introducing confusion in this.

As a side note, I suppose everyone here also understands that there *are* indeed limitations to `mtnormalise`, if you hadn't thought all of the above through yet.  We know / knew this right from the start though, all the way back to the original design, even before when we introduced the log domain into it; it's nothing new (just in case the above led to some surprises for anyone, while reading it...?).  If it wasn't clear yet in any way, as mentioned above, this is fundamentally unavoidable if all subjects are "independently" put through this process.  There are ways to mitigate this up to a certain extent, but they would typically be rather involved for your average user.  I most settings, these limitations also shouldn't have a biasing effect on the outcome of studies (there's a few rare exceptions; but they're rare, and there's ways to mitigate some limitations up to a certain specific extent there).

Well, looks like my short explanation got a bit lengthy after all...  Just to clarify: I know the above may come across as somewhat blunt here and there, but rest assured there's no bad or angry intentions here.  I don't have a lot of time to write this up now, and I'd rather be as clear as possible; so nuance may be lacking a bit for sure.  I will add a "request changes" kind of review to this pull request for now; **_again don't mean it in any offensive way_**; I fully appreciate the effort of trying to improve commands and options of course (and I'm happy about the debate on the re-categorisation that's going on in the other comments; this should lead to user interface improvements for sure). However, due to the story on quantifying 3-tissue CSD I'm working on with different groups in a few different studies now, that'll have text and theory submitted in the very short term, I feel quite strongly on not introducing confusion for users on this topic.  I may elaborate later, but this is all time I can spend on this today...

cmd/mtnormalise.cpp
jdtournier: Looks like we go for `-reference`, then.

cmd/mtnormalise.cpp
jdtournier: OK, I get the rationale, and I _think_ I agree with it... @Lestropie, what specific use case did you have in mind? Was this for debugging? I note the balance factors were already written out as `INFO` level messages, so presumably even the `-check_factors` option is somewhat redundant - assuming it genuinely is only supposed to be inspected in the (hopefully) rare cases where we/users need to figure out whether something went funny? Likewise, if the `-balanced` option is only there to facilitate debugging, I think there's already enough info there to debug if need be - they're simple scale factors that can retrospectively be applied to each ODF, right? What I'm saying is that if this is something that you anticipate might be needed for other reasons, by all means let's consider including it, but otherwise I think I'd rather keep things lean (not to mention avoid users doing something they shouldn't, although that's less of a concern for me, we can always document clearly that this is _not_ to be used in practice).

cmd/mtnormalise.cpp
Lestropie: I'd have thought that if sub-dividing further, one might instead put `-mask` into its own option group, so that it appears at the top of the options list with its own bold banner highlighting that it's compulsory (which is rare for *MRtrix3* command-line options). But unless users have actually not understood that, it's probably not worth changing.

Lestropie:
Once the `-check_factors` option is added, then the `-balanced` output images can be generated after command completion. So it's not super critical. It's just a trade-off of convenience of its generation v.s. needing a sufficiently bright warning to prevent naive users from using it. It to me comes off as strange that the command explicitly *doesn't* output its native internal result, *only* a modification thereof based on assumptions around its use; but I fully appreciate the need to avoid its mis-use. Content either way.

thijsdhollander:
Ok, so if everyone's ok with that, let's just remove the `-balanced` option.  Happy to have the `-check_factors` option still in there; it can serve a control purpose (even beyond debugging).  There are a few scenarios where using those factors to generate the "balanced" output can find a use indeed, but they have little to do with "this" definition of normalisation that `mtnormalise` aims to target (a *single* field of normalisation factors ("bias field" times global constant)).

> based on assumptions around its use

Well, I'd rather say "based on its intended use".  Yes, other things are effectively possible, but having them under the umbrella of "normalisation" the way we typically refer to it in an FBA context, would be confusing.  That said, maybe I should add at some point a few extra words of advice in the FBA manual on what can be done to actually keep those balance factors more "balanced across subjects", if it would be important for a specific study setting.  It doesn't involve *using* those balance factors though.  I just have to find the right words to describe it well at some point; but I'll be writing quite a bit about this topic soon anyway. :grin:

jdtournier:
OK, let's remove the `-balance` option then. Happy to leave `-check_factors` in there, although I reckon the bulk of that is [already available via the `-info` flag](https://github.com/MRtrix3/mrtrix3/blob/f3619d8a8c8750936113c4a3cf665986780fec82/cmd/mtnormalise.cpp#L435) (unless there is a requirement to use this in a script?). If we do keep it though, may I suggest we rename it to something a touch more explicit, like `-check_balancing` or something similar...?

thijsdhollander:
> OK, let's remove the -balance option then. 

:+1: 

> Happy to leave -check_factors in there, although I reckon the bulk of that is already available via the -info flag (unless there is a requirement to use this in a script?). If we do keep it though, may I suggest we rename it to something a touch more explicit, like -check_balancing or something similar...?

Fully agreed.  For all common use cases, the `-info` output was designed to deal with that.  If anything goes funny during a run, the `-info` output should clearly reveal that.  Unless there's a well-defined (and maybe subsequently well documented) use for the `-check_factors` option, I'd rather remove it as well, to keep things simple and not overcrowd the options list.  If a use case can be argued / justified, then I agree something like `-check_balancing` is probably a better and more explicit name.

thijsdhollander:
Removed the `-balanced` option as discussed.  Thinking usability of the command through (which is facilitated by removing clutter from the options, if possible), I'm now also erring much stronger on the side of getting rid of the `-check_factors` option.  Any solid not-too-exotic use cases that can be justified?

Lestropie:
Superseded by #1756 and #1766.

PR #1434, opened 2018-08-18
Handle dicom variable scaling

jdtournier:
To handle data with inconsistent offset/scale, as [discussed on the forum](http://community.mrtrix.org/t/enhanced-dicom-no-gradient-table/1790/4?u=jdtournier). This adds a new `ImageIO` handler to load the data as `float32` into RAM with appropriate rescaling. Required to read some Philips multiframe data produced on R5.4.


Lestropie:
> Looks like I don't permissions to stop or restart jobs on AppVeyor ...

appveyor_pending

:man_shrugging: 

Want to create an *MRtrix3* account on there instead? Alternatively it seems I can grant access to the "MRtrix3 devs" GitHub team.

Lestropie:
Mmm tremendous fun. I presume you've incorporated this into your private DICOM test suite / made sure the output makes sense / run it in debug mode just to be sure?

Can really only look at raw code changes from this end, and it all looks pretty sensible, no red flags jumping out at me.

jdtournier:
Yes, it all checks out OK on my DICOM test suite. I've also added this new dataset in there while I was at it. Realistically, the scope for interaction with the existing code is minimal, this new code path only gets triggered when it detects inconsistent scaling, in which case it'll use the new `ImageIO` handler instead of the default one. Note that this won't handle Siemens mosaic data if the scaling is inconsistent, but I've never encountered any such data without consistent scaling, and they're being phased out anyway. 

Thanks for looking at it :+1: - I'll update and merge.

jdtournier:
OK, I've set up a new MRtrix3 AppVeyor account, and granted all members of the MRtrix3-dev Github team access. Still finalising a few issues with webhooks, etc. Hopefully I'll have it all working shortly...

Lestropie:
If you're still having issues I should delete the project that's within my AppVeyor account.

PR #1435, opened 2018-08-18
Silence compiler warnings take3

jdtournier:
OK, faced with yet more new warnings following an update of MSYS2 to `gcc` 8.2.0, I've decided to revisit the changes I pushed recently in #1412 (to use `#pragma GCC diagnostic` preprocessor directives rather than `-isystem`), since I don't fancy the idea of having to update the list of warnings to ignore every time compilers decide to get stricter. A quick Google reveals this other GCC option: [-idirafter](https://gcc.gnu.org/onlinedocs/gcc-4.9.4/gcc/Preprocessor-Options.html#index-idirafter-1029), which seems to be _exactly_ what we're after... 

So I've gone and reverted (most of) #1412 (leaving unrelated changes), and simply changed `-isystem` for `-dirafter` in `configure`. It seems to work fine on my laptop (no more warnings), and no problem on `clang` 6.0.1 on Linux. And it's a lot simpler to manage than the `#pragma` route...


Lestropie:
As far as I can tell from online documentation, `-idirafter` was available in pre-C++11 compiler versions, so we shouldn't have to worry about compatibility in that respect.

Issue #927, opened 2017-03-02
scripts: run.command blocked if command produces lots of output


maxpietsch:

maxpietsch:
`run.command ('cat /tmp/text | echo')` also fails to finish if /tmp/text is larger than about 100KB  (b7424ccf79cae3b5597bd8ccdd13e) The reason I am looking at this is that `population_template` using many subjects can create that much output which causes scripts that call it to freeze.

maxpietsch:
[This](https://stackoverflow.com/questions/375427/non-blocking-read-on-a-subprocess-pipe-in-python/) suggest it can be done in a cross-platform compatible way using [queues](https://docs.python.org/3/library/queue.html#Queue.Queue.get) and a separate thread to do the reading.

maxpietsch:
Ok, I think I got it to work on my system. Below code generates a 12MB text file and pipes it into a list without locking up. The trick is that the reader is in a separate thread (and defined as demon so that it is killed with the process x): 

    #!/usr/bin/env python
    import subprocess
    from threading import Thread
    import time
    
    from random import choice
    import string
    def gen_line(length=8, chars=string.letters + string.digits):
       return ''.join([choice(chars) for i in range(length)])
    
    print "generating text file"
    with open('/tmp/text','w') as f:
      for i in range(200000):
        f.write("%i: %s\n" % (i, gen_line(50,string.ascii_letters)))
    print "done"
    
    def reader(f,buffer):
       while True:
         line=f.readline()
         if line:
            buffer.append(line)
         else:
            print "reader has finished"
            break
    
    linebuffer=[]
    x=subprocess.Popen(['cat', '/tmp/text'],stdout=subprocess.PIPE)
    t=Thread(target=reader,args=(x.stdout,linebuffer))
    t.daemon=True
    t.start()
    
    # below is for near real time printing the output, could just use t.join() here
    while True:
      time.sleep(0.1)
      if len(linebuffer):
        print "line " + linebuffer.pop(0),
      else:
        if not t.isAlive() and len(linebuffer) == 0:
          print "finished"
          t.join()
          break
        print "waiting for output"
        time.sleep(1)

Lestropie:
[...]

Issue #692, opened 2016-06-24
Scripts documentation


maxpietsch:
Scripts that do not use the lib.app framework are currently not in the documentation.`foreach` and `notfound` contain a `HELP_PAGE` which could also be copied to the docs.

I would also suggest to add a brief HELP_PAGE to the other scritpts (average_response, blend, convert_bruker). This would help in the homebrew installation to identify scripts to link to /url/local/bin. Currently, I link foreach, average_response, blend, convert_bruker and notfound manually.

Lestropie:
Alternatively, those scripts could be adapted to the library framework? That way users would get a help page just like with everything else.

jdtournier:
I think some of those scripts could probably be moved out of the main repo, and made available via a different repo. There is no way I want to support the `convert_bruker` script to the level that we support the rest of the code, it really is there as a convenience on a best-effort basis... I'm guessing the same applies for `blend`?

draffelt:
Agreed `convert_bruker` could go.  I was planning on writing a blog post/tutorial on the screenshot tool. Which would include a section on how to make spinning brains videos that blend different images together while rotating. 

Happy to move blend into a personal repo of mine. However I think it's a pretty neat way of presenting FBA results, so it may be nice to keep it in. 

I'm also planing on writing a blog post on batch processing images with `foreach`. Something like I used to have on our internal wiki. I guess we could reimplement both `foreach` and notfound in python if someone has spare time. But that won't be me any time soon.

Lestropie:
Not sure I'd be removing scripts: only takes one user to be using it and discover it's disappeared to cause a complaint. I reckon just leave them there, maybe add to the docs FAQ section something like 'why aren't all scripts listed in the documentation?', and just say that these are provided in case they're useful to someone 'on good faith', but aren't fully supported. I just think any effort here would be better spent on porting them than hacking the documentation to try to include them as-is.

SebastianoF:
Small comment about the convert_bruker and related documentation, as I just found the script by accident after having recently implemented my version of the same purpose converter. 
I believe as well it should be better advertised/documented as it is invisible from an "outside" google search for who is looking only for that function.

Issue #1445, opened 2018-08-30
"gateway" cmdline or may be more unique commands prefix?


yarikoptic:
I am updating mrtrix package to 3.x in Debian (will come up with mrtrix3 to make it possible to install 0.2.x series or this brand new mrtrix3).  Nearly all toolkits (FSL, AFNI, ...) in neuroimaging which provide a number of cmdline tools are hindered by proliferation of commands with not unique enough names.  E.g. `dirsplit` already would conflict with the one provided by genisoimage. And that is why such commands with as generic name as  `blend` provided by special toolkits would also not be advisable to be installed system wide.
That is why when packaging for (Neuro)Debian we then typically install them under a path not within $PATH and just symlink only a handful of main ones and then demand users to adjust their PATH one way or another . And that is why you have `./set_path` helper to tune those PATH variables user-wide, which could also overload the other cmdline tools.
Since 3.x is still RC I wondered if you might consider in the long(er) run to either adopt the strategy
- alike git and many other tools (including [cmtk](http://www.nitrc.org/projects/cmtk/) in neuroimaging),  of having a "gateway" command line "mrtrix" or "mrtrix3" which would accept a "command" argument which then would execute the corresponding mrtrix3 command, e.g. `mrtrix calc` (instead or current `mrcalc`).
- of having a uniform and somewhat unique prefix to all commands.  Some already have `mr` prefix, which already makes it less likely to overlap with some other tools (although there is an `mr` tool on its own). `mrt` or `mrx`  would've been even more 'unique' but that would be too much of a change

Just ideas ;) for now I guess I would symlink similarly to before the `mr*` ones

thijsdhollander:
Hi @yarikoptic,

Thanks for the feedback; I've been wondering about the same issue at times.  I do think *most* MRtrix3 commands are quite unique, due to the use of a prefix strategy we have in place (but the prefix each time refers to an action, or a "kind" of data, .e.g "`tck`" or "`dwi`", etc...  So the problem may be mostly limited to a few particular ones.  I knew (and ran into, at times) the `dirsplit` one.  `blend` indeed seems like another reasonable candidate to clash with other stuff.  Are there any other ones you can easily see being troublesome?  Here's a complete list:

```
5tt2gmwmi         dwinormalise         mrdegibbs            tckedit
5tt2vis           dwipreproc           mrdump               tckgen
5ttcheck          dwishellmath         mredit               tckglobal
[...]

PR #1449, opened 2018-09-01
Script changes

### LINKED ISSUE #928: scripts: allow list input to run.command
I'd suggest allowing to support passing a list of arguments to run.command (on top of the current approach of passing a whitespace separated string that is then split to not break current scripts.).  I have written an example (d4aab3d33561652c4e444d2335dad306a008f87c) of what I was thinking about: It bypasses the `shlex` splitting if the input is a list and bypasses the pipe splitting if the input is a nested list.

This would help for more complicated pipelines such as in `population_template` to allow lists as input to `run.command` where elements of that list can be changed in place - for instance depending on the result of the command. Another benefit of lists is that we don't need to rely on shlex to do the right thing (for instance with funny character encodings).

With those changes, all the commands below would be valid and equivalent:
```
output1 = run.command ('cat /tmp/text | cat')
output2 = run.command (['cat', '/tmp/text', '|', 'cat'])
output3 = run.command ([['cat', '/tmp/text'],['cat']])
```

### LINKED ISSUE #1096: average_response: De-mean before averaging

### LINKED ISSUE #1275: Python: Renaming "file" module
I'm increasingly disliking the naming of the "`file`" submodule in the Python script libraries. This is a builtin in Python; so not only does it light up in syntax highlighting, it also requires the use of "`#pylint: disable=redefined-builtin`" wherever it is imported. I'd like to rename this before `3.0_RC3`, since that's where the `pylint` tests become a part of the `master` branch and so having good conformity out of the gate would be preferable.

I was thinking `filesys` or `fsys` could work. `fs` might be too short, and clash with the `fsl` module. `filesystem` is a little too verbose for my liking, given in some instances it's used quite regularly. Other suggestions welcome.

I suppose included in here would be: Whether the contents of the `file` and `path` submodules should be merged together (was initially suggested in #898); keeping them split made the interface more like the C++ code, but if `file` is renamed then merging them may make more sense.

### LINKED ISSUE #1338: mrtrix3.run: Capturing help pages of external commands
I'm using `run.command()` to access the help page of non-*MRtrix3* binaries, in order to parse the available command-line options of those binaries. Unfortunately, whether commands are executed with no additional command-line arguments, or explicitly specifying the relevant help command-line option, the commands return a non-zero exit code. While `run.command()` has an argument `exitOnError` which can prevent the whole script from failing when this occurs, it nevertheless produces a warning about failed execution of the command, which isn't ideal.

Either the `mrtrix3.run` module needs a separate convenience function specifically for this purpose, or `run.command()` needs to intelligently determine that it's the contents of a help page that are being requested, and not produce a warning about a non-zero return code.

### LINKED ISSUE #1377: run.command(): Provide return code in tuple output
In some instances, capturing the return code of a particular command (in addition to the stdout / stderr contents) would assist in determining how a script should proceed.

### LINKED ISSUE #1400: Scripts: Alter library to enable use of exceptions
Something that's increasingly catching me out is using functions I've provided in the Python libraries in cases where failure of that function is permissible. Already in a couple of functions I've tacked on a boolean argument to specify whether or not to call `app.error()` if the function can't serve its purpose. I could throw an `if not function(): app.error()` around every single usage of such functions, but that'd be a lot of unnecessary expansion.

I'm thinking of modifying fundamentally how scripts are executed while making use of the library. Instead of execution proceeding down the main script file, each script would instead define `usage()` and `run()` functions, just like the C++ library.

Principally, this would allow placement of a `try: except:` clause around the `run()` function, that could catch a tailored `MRtrixException`. If caught, this would perform similar to what the current `app.error()` function does. But it would mean that individual library functions could throw an `MRtrixException` (instead of calling `app.error()`), which could potentially be caught and dismissed / alter the behaviour of the calling code.

Ideally I'd like to define the code that executes in the case where `MRtrixException` is not explicitly caught; that would enable both the catching of errors within library functions and the correct cleanup / messaging if not caught, but wouldn't necessitate the other library execution changes. However I haven't been able to find a way to do this.

Lestropie:
Mostly addressing #1400, but a few other things in there too; it was kind of a long steady accumulation of various things, I've tried to chop it up into a few commits as best I could.

They're significant library changes, and I don't like making changes that break compatibility with any stand-alone scripts out there, but I think there's strong justification for making the changes. After all, how many times have we modified the C++ library code in getting to the point we're at now?

I'll try to enumerate everything of interest here for comments. Not going to be merging this PR quickly, I should probably be implementing more tests before accepting such large changes; but I wanted to open the proposed changes up for discussion.
(Also I need to get this stuff out of my head so I can focus on more pertinent demands...)

1. Completely changed how executable script files invoke the core application library.
Previously there was a set of compulsory library functions that needed to be executed in the correct order, e.g. `app.initialise()`, `app.parse()`, `app.complete()`. Now, one defines two functions: `usage()` and `execute()` (couldn't have `usage()` and `run()` like in C++, since the latter would clash with the `run` module). After definition of these functions, `app.execute()` is invoked. `app.execute()` then sets everything up, using the `inspect` module to find those two defined functions and run them when appropriate.
The biggest advantage of this is that it enables the use of exceptions. Instead of calling `app.error()`, library functions will throw a custom *MRtrix3* exception, which can either be caught by the script programmer, or result in similar error message & cleanup behaviour as `app.error()`. This substantially improves the capability for scripts to catch when things don't quite work as intended and adjust appropriately.
`run.command()` gets its own exception, which includes all information about the command and the `stdout` / `stderr` contents. This means that the error messaging and cleanup can occur entirely within `app.execute()` when the exception is called; as opposed to the current solution, which is partially managed by `run.command()` writing a text file that `app.complete()` then reads, which is kinda clunky.

2. Moved some basic capabilities, e.g. listing available *MRtrix3* binaries / importing the config file(s), to `__init.py__`. Anyone using the library functions but not invoking the `app` module in its entirety should still be able to access this information.

3. Changes to `run.command()` to support multi-threaded usage.
Note that this required the use of `tempfile.mkstemp()` rather than `tempfile.TemporaryFile()` due to an apparent race condition in the latter on Python3 on Windows (not present on Windows Python2; not yet tested on Linux). This is despite the [Python documentation suggesting](https://docs.python.org/2/library/tempfile.html#tempfile.TemporaryFile) that `TemporaryFile()` *uses* `mkstemp()`... The side-effect of this is that the `stdout` / `stderr` contents of executed commands are written to files in `/tmp/` and deleted after they are read; there could potentially be issues with users not having write access to that location and/or accumulation of stale files due to cleanup issues.

4. Porting of `foreach` to Python. In addition to appearing in the online documentation, this also means that the terminal will present a simple progress bar showing the execution of however many instances of the command are run, and at completion only invocations that failed, along with their `stdout` / `stderr` contents, will be printed. Note that the command-line usage is fractionally different to the current bash script.

5. New script `responsemean`, which has substantial changes from the basis `average_response`, but I think the name is more consistent with the rest of *MRtrix3*. In addition to making use of the libraries (and hence appearing in the online documentation), having better error messages, and being a little more Pythonic in places, the (default) averaging process is different, as proposed in #1096.

6. Merge `file` and `path` modules into `fsys`. Really didn't like having to use `pylint` directives to hide the fact that a Python built-in was being over-ridden.

7. *Much* nicer terminal feedback when running `population_template`, as you don't get humongous `run.command()` strings when running e.g. `mrmath` across all subjects (implemented the solution proposed in [this comment](https://github.com/MRtrix3/mrtrix3/issues/928#issuecomment-410120565) in #928).

Partially addresses #692.
Hoping that #927 can be closed, but point 3 may raise other issues regarding the solution.
Closes #928.
Closes #1096.
Closes #1275.
Closes #1338 (script author can simply catch the exception along with `stdout` / `stderr` contents).
Closes #1377 (command returncode is returned in exception raised by `run.command()`).
Closes #1400.


thijsdhollander:
I suppose this will still sit here for a while, so I'll comment in more detail at a later point in time, when I, well, find the time. :slightly_smiling_face: 

That said, now that I thought of it when commenting on #1445, something that might be considered at this stage: getting rid of the `notfound` script as a separate script, but incorporate its functionality in `foreach` in some way (which could then immediately also lead to a cleaner UI to use the `notfound` functionality within a `foreach`).
Now that I'm talking about `foreach` and its UI here anyway; I haven't looked into the details (yet), but about that UI: was the change from the colon to the string input (technically) necessary for the command to use the scripting library, or is there another motivation for the UI change for this script in particular?

jdtournier:
Sorry, a bit late to the party, but my 2 cents for what it's worth:

- more than happy for `foreach` to absorb the functionality in `notfound` - I think it makes sense.

- I'd personally rather stick with the colon syntax, but happy enough with a consensus. I think the colon makes it very clear which options apply to the `foreach` command, and which to the command it will invoke. Note this is _not_ a distinction that would be required in any other regular _MRtrix3_ comand, so I don't think it would actually make sense to allow arbitrary positioning of arguments or options. I also reckon handling of special characters in filenames (e.g. a space) would get messy if the command itself is quoted - you'd need to escape the quotes within the quotes, or some other trick, in a way that I don't think is particularly obvious. I do agree that it would be nice not to have to escape or quote `|`, `&&` or `||`, though. But on balance, I prefer the current syntax... 

- :+1: for `responsemean`

Lestropie:
> I also reckon handling of special characters in filenames (e.g. a space) would get messy if the command itself is quoted - you'd need to escape the quotes within the quotes, or some other trick, in a way that I don't think is particularly obvious.

I'd thought of escaping whitespace characters in filesystem paths, but the potential use of quotes to encapsulate paths (e.g. coming from "`command_history`"), thus necessitating escaping of those quotation marks within the command string, would indeed be pretty messy, and so tips the balance toward the colon syntax.

I'll still have to wait and see whether it can be done easily without re-implementing `subprocess` though. Colon is technically just a no-op in bash, so this usage is very non-standard. I think I know how to do it though.

Lestropie:
... Was too curious, so had to try the colon syntax. The easy solution worked. :+1: Python is awesome sometimes.

(see 5c859f5 for details)

Lestropie:
Another criticism I'll throw in the ring specifically for `foreach` is the "`NAME`" substitution. This extracts specifically the *basename* of a filesystem path. Given that `foreach` *technically* isn't only applicable to using filesystem paths as inputs, "`NAME`" to me feels too non-specific: "`BASE`" would be better.

thijsdhollander:
Just to chip in as the discussion is happening: after giving it some thought, and also reading @jdtournier's arguments, I'm similarly in favour of the existing colon syntax.  I get what you tried to do before regarding consistency with other commands, but I don't think consistency (to other mrtrix commands) really matters here: foreach is used very differently and for different purposes anyway.  Also, the colon is visually very clear, and I like how it helps to read / navigate / ... someone's long pipeline script.  And finally also, I hadn't even thought yet about the issues arising with having to escape quotes in quotes.  That's indeed one to really avoid, I think.  So well, all that to say that I'm in support of the "original" colon syntax.

> Another criticism I'll throw in the ring specifically for foreach is the "NAME" substitution. This extracts specifically the basename of a filesystem path. Given that foreach technically isn't only applicable to using filesystem paths as inputs, "NAME" to me feels too non-specific: "BASE" would be better.

Not sure whether "BASE" is really more clear than "NAME" (even though I do appreciate the argument), but I'm not to fussed about it either; happy to change if some people like it more.  I'm more sensitive about breaking peoples' existing scripts, or habits/experience; but that chance should be quite low for this substitution option... Most (if not all?) of the steps in the pipelines in the manuals only use "IN", and a few "PRE".  I actually don't recall myself using "NAME" or "UNI" (...well, maybe once or something, but not frequently for sure).  So safe to change, I'd say.

When I find some (more) time, I still have to look into `responsemean` though.  I'm ok with the name change, but I've got at least one minor concern with the mechanism.  I've brought this up ages ago, I think, but can't find the post: it's indeed about the fact that this mechanism *in principle* requires to process all (different) tissue responses at once, so no relative rescaling happens.  So by introducing this (pull request proposed) mechanism without that (all response functions processed together), you gain some (the different weighting due to input scaling without normalisation; i.e. the purpose you indeed introduced this mechanism for), but you also lose something else (harder to describe what, but I've got a good idea about it).  I don't want to get ahead of myself at this time (so will still look into it), but I emphasised this "*in principle*" phrase above because it may not be much of an issue *in practice*.  However, I believe it indeed "won't be much of an issue" under the conditions that the original problem _also_ wasn't much of an issue (i.e. when the responses for a given tissue don't vary that much in shape (contrast, not scale) among themselves).  But well, I'll do some thinking for a few typical scenarios and evaluate whether the risk of my minor concern is real (and significant), or whether it may not be much of an actual concern in practice.

That said, I've actually got a few other ways of computing a combined ("average", but not really) response or set of responses in store; but it's a bit too early to introduce them at this stage.  So if my above minor concerns with the proposed change in this pull request turn out to be not really an issue in practice for the properties of most realistic scenarios, I'm happy to go with that approach for the time being though. (if used correctly, I of course do fully support the use of geometric means / arithmetic in log-space :wink: )

thijsdhollander:
Now that we're talking script library here anyway, and when I noticed this again in the original pull request post:

> Much nicer terminal feedback when running population_template, as you don't get humongous run.command() strings when running e.g. mrmath across all subjects (implemented the solution proposed in this comment in #928).

...can someone remind me again what our motivation actually is to, by default, print all the run.command strings?  I'm not overly fussed about it, but maybe still: for some scripts, that really unleashed *the Matrix* onto peoples' terminals... I do get we're trying to be clear to the user what is exactly going on, in a transparent fashion, but it's a bit weird in contrast to (binary) commands, right?  It's not like we're by default printing the full `-info` output for those (that's exactly why `-info` exists as an *option*).  And developers of scripts can still choose to print (by default) some useful information anyway (and they are probably in the best position to decide what is useful / informative terminal output for a given script).  So well, why not, e.g., only print the run.command strings when the `-info` option is provided to a script?  Or even have a separate option altogether to print the run.command strings... i.e., a level between the default terminal output and the `-info` output (the latter indeed _also_ shows the full `-info` per binary command within the script).

Maybe I'm overlooking a good reason to show the run.command strings by default though... so just thought I'd bring it up to figure this out (again).

Lestropie:
> can someone remind me again what our motivation actually is to, by default, print all the run.command strings?

Because people get nervous and think that something has crashed if nothing prints on the terminal. 

> for some scripts, that really unleashed the Matrix onto peoples' terminals

Use a [`progressBar`](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/app.py#L800-L877). If you have a list of commands of known length, you could make a wrapper class that initialises the `progressBar` and provides a member function that both invokes `run.command()` and increments. When providing `-info` command-line option, individual `Command:` prompts will still be shown.

Lestropie:
Other point to consider with the `foreach` substitutions is the danger of clashes: e.g. currently any filesystem path that happens to contain the two sequential characters "`IN`" will be erroneously substituted. "`INPUT`" would be less dangerous; "`NAME`" could be changed to "`BASENAME`" rather than "`BASE`", which is more clear than either alternative; "`PRE`" could be "`PREFIX`"; "`UNI`" could be "`UNIQUE`". But I've no idea how frequently people may or may not have encountered such clashes in order to justify such extensions.

jdtournier:
Yes, I've also wondered whether this might cause trouble. I'd be happy to come up with alternatives, but I'd favour an explicit de-referencing syntax of some form, rather than a longer name. Something like `:IN:` or `[IN]` (although there is scope for wildcard expansion with the latter), or `+IN+`, or ...?

I'm not a fan of using longer names even if it might makes the intention more obvious on casual inspection. I think as long as it's relatively well documented, users will appreciate shorter substitutions in routine practice. The current description in the `foreach` help page is clear enough, in my opinion (although it could be clearer still):
```
AVAILABLE SUBSTITUTIONS:
  IN:   the full matching pattern, including leading folders
        For example, if the target list contains a file "folder/image.mif",
        any occurrence of "IN" will be substituted with "folder/image.mif".
  NAME: the basename of the matching pattern
        For example, if the target list contains a file "folder/image.mif",
        any occurrence of "NAME" will be substituted with "image.mif".
  PRE:  the prefix of the matching pattern
        For example, if the target list contains a file "folder/image.mif",
        any occurrence of "PRE" will be substituted with "image".
  UNI:  the unique prefix of the matching pattern after removing the common suffix
        For example, if the target list contains files:
        "folder/001dwi.mif", "folder/002dwi.mif", "folder/003dwi.mif"
        any occurrence of "UNI" will be substituted with "001", "002", "003".
```

Lestropie:
> I'd favour an explicit de-referencing syntax of some form

That could work. Probably the closest in 'spirit' would be to use e.g. "`$IN`", but the dollarsigns would need to be escaped, and so would probably lead to a lot of mistakes. Given we're already kind of mis-using the colon operator in the usage of `foreach`, "`:IN:`" might be the least objectionable. Still, this PR is a fair way off merge, so further alternatives welcome.

thijsdhollander:
> I'm not a fan of using longer names even if it might makes the intention more obvious on casual inspection.

I don't always agree with this, but in this context, I do.

> I'd favour an explicit de-referencing syntax of some form

I'm not even sure this is really needed, to be honest.  `foreach` has now existed already a long time, and I've never heard anyone running into any issues caused by these keywords.  Maybe one quick question though: when we detect them, is that detection case sensitive?  If so, that may just be more than sufficient to avoid even a rare possible problem.  If not, I do suggest it'd be case sensitive (i.e. only detect these keywords if they're full caps).

thijsdhollander:
> I reckon most people would conventionally use lower-case when naming their own files, hence why this hasn't really been an issue.

That's one reason indeed.  But also: typically the folders at those levels (e.g. subjects) wouldn't be mentioned in most `foreach` scenarios; they'd actually be the ones captured by `foreach` to iterate over e.g. all subjects... so in this case, there's also no issue.

> I've also tried to explicitly catch and issue a warning when such an erroneous substitution might be happening here.

That should indeed help, just in case.

thijsdhollander:
Just about this, because it was brought to my attention again today:

>> can someone remind me again what our motivation actually is to, by default, print all the run.command strings?

> Because people get nervous and think that something has crashed if nothing prints on the terminal.

>> for some scripts, that really unleashed the Matrix onto peoples' terminals

> Use a progressBar. If you have a list of commands of known length, you could make a wrapper class that initialises the progressBar and provides a member function that both invokes run.command() and increments. When providing -info command-line option, individual Command: prompts will still be shown.

Ok, if the only motivation to show those `Command` and `Function` lines by default is to give some kind of progress feedback to the user, I think they're better *not* shown *by default*.  I get that a progress bar can somewhat hide some of them, but I see this really as to up to the developer (in this case, of an MRtrix python script) to decide for their script: they can add progress messages, bars, ... depending on what makes sense for their script.  This would make it much more in line with our `C++` "command writing" library: here, no internal function calls are shown by default; developers show progress lines or bars or... based on what their app requires.  When I mentioned the "unleashing of the Matrix" above, it was not so much the amount of text actually, but the level of abstraction of it: typically the `Command` lines, and even more so the `Function` lines, are part of the implementation of the script, and are (most of the time) not relevant to confuse the user with (and if so, a developer can always *show* what's relevant, rather than trying to *hide* what's not).  This might also help more users see the *relevant* messages, e.g. when something goes wrong.  I'm often seeing users provide a script output of a script that went wrong somewhere, and even though the information/error is in the output, they'd overlooked it because there was so much else (less relevant) being shown on screen.

So well, long story short: in line with our binary commands, I definitely feel it makes sense to not show `Command` and `Function` lines by default and leave it up to the developer to indicate progress and information in any way they think makes sense for their app.  `Command` and `Function` lines, IMHO, belong at the `-info` level.  If there's a good case to show them while not showing all the "deeper" level `-info` output, it might even make sense to offer an intermediate level for that (which can make good sense, since our scripts do provide an additional level of abstraction by definition).

Lestropie:
It's about development effort. When a script is first written, printing the underlying commands gives the user an indirect sense of progress of the script, without placing additional demands on the person who wrote the script to explicitly code such. Sure more explicit progress information can be added on a per-script basis (including use of the `progressBar` and/or something like [this class](https://github.com/MRtrix3/mrtrix3/pull/1449/files#diff-c889bd0f414f8b36acd5f0383cb6550eR474) that I suggested above), but most devs simply won't bother with such. This kind of "free" progress information protecting against lazy script writers does not have an analogue in C++ commands.

It also means that if the user is unsure as to whether the script is progressing or whether it has stalled, they can read the name of the command most recently executed, and so can cross-reference this against e.g. `top`.

Lestropie:
Suggested changes following online discussion:

- Rename `fsys` module back to `path`

- Rename `delTemporary()` to `path.cleanup()`

- For `fromUser()` and `toTemp()`, rename `is_command` parameter, make it default to True, and remove all explicit usages where the value is True

- Rename `run` module to `call` / `exe` / `invoke`

- For `newTemporary()` function, add description in comment regarding how to create a temporary in a race-free manner (or provide such wrapper function; could decide whether the input path is a directory or a file based on the final character)

jdtournier:
I reckon this all sounds sensible - nothing jarring in that list as far as I can tell. Still need to settle on the name for the `run` module, but I'm pretty happy with any of the options proposed - preference for `invoke` for clarity, `call` for brevity, not so hot on `exe`, but not too worried either way. 

One more thing while we're at it: I've tended to reserve `CapitalisedNames` for class names, and `names_with_underscores` for function / method calls. I note you're using a different convention in your code, closer to the Qt convention. Not a big deal, but if you want to keep everything consistent, we may want to discuss this...

Lestropie:
Class / variable naming in Python hasn't really been considered in great detail, it just kind of came about organically at the time (and bearing in mind that I knew no Python whatsoever when the library started). We're already not fully conforming with PEP8 due to using two spaces rather than four for indentation & permitting long lines; but that's probably nevertheless a good reference to use. 
https://www.python.org/dev/peps/pep-0008/#naming-conventions
They suggest `CamelCase` for classes, `lowercase_with_underscores` for both functions & variables, `ALLCAPS_WITH_UNDERSCORES` for constants; which is consistent with your preference. `mixedCase`, which I've gravitated toward at some point (for Python only), seems to be discouraged.

Lestropie:
Maybe starting to lean toward `call` for the module currently known as `run`, considering that devs may have some familiarity with `subprocess.call()` of which the most commonly-used function in that module is essentially a supercharged version.

Lestropie:
P.S. If anybody has strong thoughts on naming conventions, best speak up now: am strongly considering updating the whole library & all scripts to conform to what is described above.

thijsdhollander:
I don't mind personally what happens with naming conventions within the library, but I strongly prefer to keep my own naming within the `dwi2response dhollander` algorithm code.

thijsdhollander:
(performed a manual `dev` merge; so things don't start to diverge or otherwise get messy with the copyright update in mind)

Lestropie:
Should mention here in addition to the commit message:

In the process of 8f6dcb7 I made a change to `dwibiascorrect` when using `fsl fast`, which is related to discussion in #1270. Instead of accepting `fast`'s bias field estimate of 1.0 outside the provided brain mask, which introduces a 'step' in image intensities at the edge of the mask, it now instead zero-fills the output bias-corrected DWI outside of the mask. This hopefully makes the influence of this limitation slightly less 'insidious'. I had the idea to instead use `mrmodelfield` to parameterise the bias field inside the mask and hence enable extrapolating it beyond the mask, but it seems that command was removed at some point.

thijsdhollander:
> Instead of accepting fast's bias field estimate of 1.0 outside the provided brain mask, which introduces a 'step' in image intensities at the edge of the mask, it now instead zero-fills the output bias-corrected DWI outside of the mask. This hopefully makes the influence of this limitation slightly less 'insidious'.

:+1: Very happy with that; it's indeed quite misleading otherwise.

> I had the idea to instead use mrmodelfield to parameterise the bias field inside the mask and hence enable extrapolating it beyond the mask, but it seems that command was removed at some point.

I wouldn't do this in any case; it makes it potentially even more unclear on "what method" is exactly run.  `fast` does what it does, nothing more, nothing less; let's leave it at that.  I say it's up to them, if they want to offer this kind of functionality.

thijsdhollander:
Probably aware and/or because this branch is still unfinished, but just for the record / in case: currently generating the user docs generates some errors related to dwibiascorrect:

```

  File "/home/thijs/mrtrix3/bin/dwibiascorrect", line 82, in <module>
    app.execute()
  File "/home/thijs/mrtrix3/lib/mrtrix3/app.py", line 90, in execute
    module.usage(CMDLINE)
  File "/home/thijs/mrtrix3/bin/dwibiascorrect", line 20, in usage
    algorithm.usage(cmdline)
  File "/home/thijs/mrtrix3/lib/mrtrix3/algorithm.py", line 42, in usage
    module = importlib.import_module(path.script_subdir_name() + '.' + package_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named dwibiascorrect.ants
  File "/home/thijs/mrtrix3/bin/dwibiascorrect", line 82, in <module>
    app.execute()
  File "/home/thijs/mrtrix3/lib/mrtrix3/app.py", line 90, in execute
    module.usage(CMDLINE)
  File "/home/thijs/mrtrix3/bin/dwibiascorrect", line 20, in usage
    algorithm.usage(cmdline)
  File "/home/thijs/mrtrix3/lib/mrtrix3/algorithm.py", line 42, in usage
    module = importlib.import_module(path.script_subdir_name() + '.' + package_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named dwibiascorrect.ants
```

thijsdhollander:
> When I find some (more) time, I still have to look into responsemean though. I'm ok with the name change, but I've got at least one minor concern with the mechanism. [...]

Ok, did that in the mean time; I'm ok with `responsemean` for now, doesn't cause the concerns I was worried about sufficiently to be an issue in most cases.  So I'm not worried about its inclusion currently.  That said, the need for it compared to the "simple" averaging is now also mostly gone when `dwibiascorrect` is used in (group wise) pipelines.  So well, the overall impact of the change is minor; thus happy to have it either way.

> That said, I've actually got a few other ways of computing a combined ("average", but not really) response or set of responses in store; but it's a bit too early to introduce them at this stage. [...]

Works (but not "neatly" implemented yet; more internally/messy; tested successfully in a few collaborations though) and is on the way at some point, but probably not immediately in the public space.  So don't wait for it currently; it'll happen when the time is right.

Lestropie:
Note issue where scratch directory location contains whitespace as reported on [forum](http://community.mrtrix.org/t/dwi2response-dhollander-error/2210) fixed in 7d1537f.

Lestropie:
I've finished manually cross-checking outputs from this branch against `dev`. I'll leave the PR open until the next video call, in case anyone has comments regarding the API changes.

thijsdhollander:
Did some minor `dwi2response` UI tweaks (no changes to functionality or e.g. option names).  Further feedback from testers:

* First line of the usage section in the help page of a script with algorithms reads (e.g.): `dwi2response [ options ] algorithm ...` .  They found it a bit weird that it didn't read something like `dwi2response algorithm [arguments] [options] ...` .  Most confusion arose from it not starting with `dwi2response algorithm` before anything else appeared.

* They found it a bit weird (and somewhat "useless", in a way) that the help page of a script with algorithms, while not yet having specified the algorithm (e.g. just typing `dwi2response` and hitting enter) shows information on shared arguments and/or options; because there's no use to this information in that context.  The only relevant context to really go in detail about how to interface with the thing, is when the context of a given algorithm has already been established.  So they suggested making the help page of scripts with algorithms much more concise: only showing the basics (version, authors, copyright, etc..), a synopsis, and have usage and/or description basically only explain how to access the help page of a(ny) specific algorithm and of course list those algorithms.  So basically cutting out all sections to do with further arguments and/or options.  To put it differently: the fact or information that these arguments and/or options "happen to be" shared among algorithms is of no real value to the user.

* In principle the same kind of feedback holds for binaries with algorithms.  At the moment, our approach there is different from scripts, and in a certain way "worse": there are no individual help pages per algorithm, and all options (for any or all algorithms) are listed all in one massive page, so the user has to dig in and be bothered by whole lists of options that don't even apply at all to the algorithm they're interested in interfacing with.  This bullet is beyond the scope of this pull request though, as it has nothing to do with scripts.

thijsdhollander:
(appveyor seems to never get there within the 1 hour limit... not even close...?)

Lestropie:
> (appveyor seems to never get there within the 1 hour limit... not even close...?)

Seems to succeed if triggered at certain times during the day. But yes, is becoming increasingly problematic.

One option might be to separate building GUI and non-GUI components into two separate jobs. Another could be to generate the shared library file (& associated object files?) in one job, then pass the result to subsequent jobs via their Artifact mechanism.

thijsdhollander:
It's weird / drastic though: when they succeeded in the recent past, they typically did so in ~ 42-45 mins.  When they fail now though, they're sometimes only still building at ~ 300/505 when at the 1 hour mark.  AppVeyor is starting to get more of a (major) hassle rather than a benefit...

thijsdhollander:
...just checked; it takes about 15 mins to advance beyond this point:

```
Build started
git config --global core.autocrlf input
git clone -q https://github.com/MRtrix3/mrtrix3.git c:\msys64\mrtrix3
git fetch -q origin +refs/pull/1449/merge:
git checkout -qf FETCH_HEAD
C:\msys64\msys2_shell.cmd -mingw64 -here -defterm -no-start -c "pacman -S --noconfirm bc mingw-w64-x86_64-diffutils mingw-w64-x86_64-eigen3 mingw-w64-x86_64-fftw mingw-w64-x86_64-libtiff mingw-w64-x86_64-qt5 python > pacman.log" || (appveyor PushArtifact %APPVEYOR_BUILD_FOLDER%/pacman.log & EXIT 1)
warning: dependency cycle detected:
warning: mingw-w64-x86_64-harfbuzz will be installed before its mingw-w64-x86_64-freetype dependency
warning: no /var/cache/pacman/pkg/ cache exists, creating...
```

thijsdhollander:
...were the test data changes in 3e00b2dd9c7c12c417f9ed72015dc72c47093fef intentional? They seem to be unrelated to the commit, and (I suppose) cause some tests to currently fail.

thijsdhollander:
Would be good to try and close this one off; it's probably one of the bigger changes (if not the largest) for RC4, and merging it in would help some other ones not diverge too much.  I've poked everyone for review; maybe we can use tomorrow's meeting to get some thumbs up (or comments), so we can move on?

thijsdhollander:
@Lestropie , as I'm working with all code on this branch now, I've stumbled across a bug in the new `foreach`:  the new capability that reports how many inputs (commands/runs) of `foreach` succeeded works well with binaries, but doesn't seems to work correctly with scripts, or at the very least not with `dwipreproc` (but I *suspect* other scripts just as well).  I was running `foreach` in a folder that has a bunch of subfolders, but also a single file, and then just mostly copy-pasting the typical first steps of an FBA.  Binaries nicely report an error when trying to find `that_file.something/an_image.mif`, and in the end `foreach` correctly reports `(n+1)/(n+1) jobs completed sequentially (1 errors)` (for n correct folders and a single file-that's-not-a-folder).  `dwipreproc`, however, while correctly reporting its own error (I was running `foreach` with `-info`), `foreach` doesn't seem to pick it up, thus reporting `Script reported successful completion for all inputs`, and no `(1 errors)` to be seen.

thijsdhollander:
Little update: unlike what I guessed above, `dwibiascorrect` *doesn't* have the problem.  Also, `dwibiascorrect` retained the temp/scratch folder, whereas `dwipreproc` did not.  In fact, `dwipreproc` hadn't yet created the scratch folder yet.

thijsdhollander:
Remaining issues:

* Weirdness with test data introduced in 3e00b2dd9c7c12c417f9ed72015dc72c47093fef
* Problem with `dwipreproc` not reporting failure(s) to `foreach`

Others?  If not, @MRtrix3/mrtrix3-devs please review or flag "general ok", I *will* bring this up at next meeting to resolve and move on.

Lestropie:
[Forum thread](http://community.mrtrix.org/t/how-to-exclude-a-folder-file-in-foreach/2434) raised an issue I've encountered myself in trying to use this script, so decided to do something about it.

```text
$ ./bin/foreach -info -exclude "1*" 1 2 3 4 5 6 7 8 9 10 11 : echo IN
foreach: [  0%] 0/8 jobs completed sequentially...
Command:  echo 2
foreach: [ 12%] 1/8 jobs completed sequentially...
Command:  echo 3
foreach: [ 25%] 2/8 jobs completed sequentially...
Command:  echo 4
foreach: [ 38%] 3/8 jobs completed sequentially...
Command:  echo 5
foreach: [ 50%] 4/8 jobs completed sequentially...
Command:  echo 6
foreach: [ 62%] 5/8 jobs completed sequentially...
Command:  echo 7
foreach: [ 75%] 6/8 jobs completed sequentially...
Command:  echo 8
foreach: [ 88%] 7/8 jobs completed sequentially...
Command:  echo 9
foreach: [100%] 8/8 jobs completed sequentially...
foreach: [100%] 8/8 jobs completed sequentially
```

thijsdhollander:
In the spirit of making progress merging stuff in to `dev`... another gentle encouragement on my part: @jdtournier , I noticed you're a bit on a roll reviewing PRs for `dev`.  I absolutely realise this one may not be so inviting to review, due to the magnitude/number of changes, but I reckon it would help avoid this becoming another `dev` branch on it's own that risks diverging, or holding other stuff back, or similar.  All of this would benefit greatly from getting real world testing on `dev` too, I think.  Nothing stops us from making further tweaks or changes before RC4 goes to `master` after all.

jdtournier:
OK, I'll look into it tomorrow. This has indeed turned into a bit of a monster, but I do need to bite the bullet...

thijsdhollander:
What a delight to wake up to this! :smile: :tada:

Lestropie:
> I'd pushed a change to rename the ANSI named tuple to VT100, but looking into it a bit more, it seems ANSI is appropriate

I made the same evolution myself. :+1:

Lestropie:
> What a delight to wake up to this!

I hadn't realised that you were such a big fan of all of the work I've done here.

PR #1451, opened 2018-09-05
amp2response: Bug fix and speedup

Lestropie:
#1406 (specifically 01ebd8f) cherry-picked to `master`.

In addition to preventing a segfault in the case where the provided mask has dimensions XxYxZx1, which is really a bug and should have gone to `master` immediately, this change can apparently significantly speed up `amp2response` in some usage scenarios. I've therefore had a request for this change to be merged into `master`, so that for those where the processing time is problematic, they can simply update their `master` branch before commencing processing, rather than themselves managing a dedicated branch install with this change cherry-picked. I'm therefore requesting that the speedup also go to `master`, even though it's not strictly a "bug".

### DIFF
--- a/cmd/amp2response.cpp
+++ b/cmd/amp2response.cpp
@@ -194,10 +194,20 @@ void run ()
   auto image = header.get_image<float>();
   auto mask = Image<bool>::open (argument[1]);
   check_dimensions (image, mask, 0, 3);
+  if (!(mask.ndim() == 3 || (mask.ndim() == 4 && mask.size(3) == 1)))
+    throw Exception ("input mask must be a 3D image");
   auto dir_image = Image<float>::open (argument[2]);
   if (dir_image.ndim() < 4 || dir_image.size(3) < 3)
     throw Exception ("input direction image \"" + std::string (argument[2]) + "\" does not have expected dimensions");
   check_dimensions (image, dir_image, 0, 3);
+
+  size_t num_voxels = 0;
+  for (auto l = Loop (mask, 0, 3) (mask); l; ++l) {
+    if (mask.value())
+      ++num_voxels;
+  }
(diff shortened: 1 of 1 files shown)


thijsdhollander:
Just to confirm: this doesn't change the behaviour (i.e. result) in any way, right?  Just the bug-fix for the mask, and a speedup due to a different memory allocation strategy?  If that's indeed the case, I see no harm to get it straight to `master`.

I'm a  bit surprised that this would have such a strong impact on the running time of any analysis though.  It's a pretty fast step anyway, and the (MT- or other) CSD step that those responses would eventually be used in, is always going to take longer anyway... So just wondering what the circumstances are where the processing time was "problematic" before, but would now be so much more manageable...

Lestropie:
> this doesn't change the behaviour (i.e. result) in any way, right? 

Correct. Confirmed no difference in output across a range of uses, and confirmed no memory issues in `gdb`.

> So just wondering what the circumstances are where the processing time was "problematic" before, but would now be so much more manageable...

Will depend on the number of voxels, and the memory allocation library being used on the system (even something low-level like `malloc()` does not have a universal implementation). Doing fresh memory block allocations and complete memory block copies for every voxel when the matrices are already large could conceivably be slow. Could potentially also depend on the implementation of Eigen's `conservativeResize()` for any particular version, as to whether it can do column-wise `memcpy()`'s or whether it's limited to element-wise.

This should have been the original implementation when `amp2response` was first provided; I just forgot to do it because it wasn't problematic for anyone at the time.

jdtournier:
Sounds good to me. This is a relatively small changeset, and if it does bring substantial speedup to the extent that users can actually use it (there was a forum post about that, IIRC?), I think it gets close enough to a 'bug fix' to consider including in `master`. :+1:

PR #1456, opened 2018-09-10
tck2connectome empty output on 4-D parcellation image 

dchristiaens:
(Reported to me by Tanya Poppe)

On 4-D node images, `tck2connectome` outputs a warning for missing nodes, and produces an empty connectivity matrix. The commit in this pull request adds a trivial check for 3-D inputs, which can hopefully demystify such error cases more easily.


Lestropie:
Alternative solution in d313e597 means that "4D" images with a single volume will still work (tested).

Lestropie:
Also: May be worth sending to `master`.

dchristiaens:
Great! Thanks Rob.

Issue #1394, opened 2018-07-12
Need easy to edit, user-editable FAQ section


jdtournier:
Something that seems to keep coming up is the idea of a FAQ page with quick tips and ways to perform common operations. One barrier to getting this underway is the overhead of adding entries and maintaining such an FAQ. What might help would be to have a page/section separate from the main docs (but linked to from the docs) that would be quick and easy to update, not require a full commit/test/review cycle to the main repo, and that users could more directly contribute to. 

The obvious candidate here is the GitHub Wiki. It's currently disabled, but we could just re-enable it, provide a front page with a link to the main docs, and then point users to the sidebar for the full list of FAQ entries (this means one page per entry). We could also provide a brief set of instructions to add a new entry, to help keep it consistent. The advantages are: much easier for anyone to add content; it's markdown (simpler than ReST); the sidebar provides a full listing (no need to maintain a ToC); we can make it editable by anyone (less burden on us); it's immediate and doesn't require the full hour's delay that a push to the main repo would entail... 

Thoughts?

Lestropie:
Anything that opens the potential for user contributions and taking a load off my own back gets a :+1: from me :-P

The challenge would be getting the separation between official documentation and user-editable documentation right. Currently in the docs:

- "Tips and tricks" contains info on DICOM handling, and a description of `foreach`. The former could go either way; the latter would ideally instead be encapsulated within the help page of `foreach` after porting it to the Python library. 

- "Troubleshooting" contains: Crashes/errors/performance issues, display issues, FAQ, and advanced debugging instructions. The contents of FAQ are things that could probably be individual pages in the wiki; the others, I honestly don't know.

The other thing to consider is level of access. Open access is fine if people are adding pages with their own information; but for something like e.g. "What tracking parameters should I use for HCP data?", different users may have different opinions, and we wouldn't want to enable a back-and-forth editing war. Is there a "pull request" equivalent for the GitHub wiki?

jdtournier:
Yes, that one will always be tricky. There's 3 categories that I can see here:

- our officially-endorsed tips & tricks, including `foreach`, and most of the troubleshooting section (not FAQ)

- the user-editable FAQ on the wiki (or elsewhere), with the understanding that this is not necessarily officially endorsed, and even if a member of the team adds an item, it's not necessarily been endorsed by the rest of the team.

[...]

PR #1459, opened 2018-09-17
Change FSL autodetect behaviour

octomike:
In environments where multiple FSL versions can be loaded (HPC),
preferring the fsl5.0-X defaults leads to undesirable behaviour.

Users may explicitly activate some custom FSL either with environment
modules or manually with the default FSL setup procedure. If there is
a fsl package installed which prefixes their scripts with "fsl5.0-"
(i.e. neurodebian) mrtrix should not prefer that one over the explicit
user choice. This PR defensively changes the preference, but actually I would argue to completely drop the fsl5.0 convenience detection. If this is just to work with neurodebian then this assumption is missing the fact that neurodebian still requires users to run `source /etc/fsl/5.0/fsl.sh` to actually load and activate everything. Afterwards all binaries are in PATH and can and should be found with `find_executable(name)`.

### DIFF
--- a/lib/mrtrix3/fsl.py
+++ b/lib/mrtrix3/fsl.py
@@ -71,10 +71,10 @@ def eddyBinary(cuda): #pylint: disable=unused-variable
 def exeName(name, required=True): #pylint: disable=unused-variable
   from mrtrix3 import app
   from distutils.spawn import find_executable
-  if find_executable('fsl5.0-' + name):
-    output = 'fsl5.0-' + name
-  elif find_executable(name):
+  if find_executable(name):
     output = name
+  elif find_executable('fsl5.0-' + name):
+    output = 'fsl5.0-' + name
   elif required:
     app.error('Could not find FSL program \"' + name + '\"; please verify FSL install')
   app.debug(output)


Lestropie:
Thanks for the PR. The convenience detection was implemented simply to "get the scripts to work" for some users. I wasn't aware of any kind of precedence or hierarchy: I had instead assumed that the "`fsl5.0-`" prefix was for FSL5 installs that were intended to reside alongside FSL4.

If there's good justification for not auto-detecting such binaries (please link any relevant pages), I'm happy to make the change. Though I would be in preference of making the change on the `dev` branch rather than `master`, just in case it would result in a different version of FSL being invoked on some users' systems, which has a chance to alter outcomes: in between tag updates, we're trying to alter `master` with bug fixes only.

Also, if the presence of prefixed binaries but absence of non-prefixed binaries in the path is indicative of a "compulsory" setup script having not been run, alternative logic for `fsl.exeName()` (or appropriately renamed function) would be:

1. If non-prefixed binary is in `PATH`, that should be used.

2. If not, but the "`fsl5.0-`" prefixed binary is, then write an error message instructing that the FSL setup is *incomplete / inappropriate*, and that the appropriate setup script needs to be run.

3. If neither is present, then write an error message saying that FSL is not at all installed.

?

octomike:
> I had instead assumed that the "fsl5.0-" prefix was for FSL5 installs that were intended to reside alongside FSL4.

As far as I know the `fsl5.0-` prefix is very specific to the fabulous, but outdated Neurodebian (ND) packages. Their existence is described in this [README.debian](http://neuro.debian.net/debian/extracts/fsl/README.Debian). It turns out it is a convenience layer to run a FSL program without setting `FSLDIR` and sourcing the setup script first. People probably don't know about that layer and when using the ND packages just assumed that they have to run the commands with the prefix, just like [this person in the ND comments section](http://neuro.debian.net/pkgs/fsl-5.0.html#comment-940939860).

> If there's good justification for not auto-detecting such binaries (please link any relevant pages), I'm happy to make the change. 

Okay, I will try to find other people falling over this. But I really think this is potentially very dangerous bug concerning reproducibility. In HPC systems [[1](https://www.zedat.fu-berlin.de/HPC/EN/Modules), [2](https://hpc.nih.gov/apps/fsl.html)] people usually activate and pin their software with something called environment modules. So a job would look something like this:

```
#PBS -l mem=32gb,nodes=1:ppn=8
module load freesurfer/5.3.0
module load fsl/5.0.7
module load mrtrix3/rc3

dwipreproc foo bar baz
[...]
```

These environment modules mostly just change PATH, LD_LIBRARY_PATH and the sorts to point to a specific version. What happens right now is that users activate a specific FSL version and mrtrix3 is **revertin that back** iff the ND packages happen to be installed as well. This is probabaly not just a bug in our environment and I can imagine that some people might not even notice they're using the outdated FSL. In the worst case they would report their results with the wrong FSL version.

> Though I would be in preference of making the change on the dev branch rather than master

That makes total sense, I'll change the target branch.


> Also, if the presence of prefixed binaries but absence of non-prefixed binaries in the path is indicative of a "compulsory" setup script having not been run, alternative logic for fsl.exeName() (or appropriately renamed function) would be:

I was too fast with my judgement. The [README.debian](http://neuro.debian.net/debian/extracts/fsl/README.Debian) states that the prefixed binaries are a convenience so there is no need to output an error in the second case. However, if you want to keep the ND autodetection/convenience, I believe, as stated above, it is necessary to change the order to always prefer the default binaries currently in the PATH.

Lestropie:
> Okay, I will try to find other people falling over this.

No need to go constructing a comprehensive list; just wanted a link to anything you'd already come across / discussions that had happened that had led to the PR.

> What happens right now is that users activate a specific FSL version and mrtrix3 is reverting that back iff the ND packages happen to be installed as well.

Yep, just that scenario in isolation is a sufficient justification to alter the logic.

> But I really think this is potentially very dangerous bug concerning reproducibility.

If you're concerned about it, you're more than welcome to write a post on the [community forum](http://community.mrtrix.org/) describing to people in what circumstances the issue may arise. That way you also get full credit for it.

> The README.debian states that the prefixed binaries are a convenience so there is no need to output an error in the second case. 

Okay, in that case let's just flip the precedence. At least there's some existing opportunity for feedback in that if the prefixed binaries are invoked, the `stderr` output generated by the script will reflect as such in its "`Command: `" strings.

jdtournier:
>> But I really think this is potentially very dangerous bug concerning reproducibility.
> 
> If you're concerned about it, you're more than welcome to write a post on the community forum describing to people in what circumstances the issue may arise. That way you also get full credit for it.

Sounds like a candidate for our new FAQ wiki section...? Although we haven't officially opened it up to users yet. Now might be the time?

Lestropie:
> Sounds like a candidate for our new FAQ wiki section...?

Maybe less so than other things: Users may need to be alerted that this is already happening currently with their processing, whereas they won't get automatically alerted to the presence of a new wiki entry..

jdtournier:
>  they won't get automatically alerted to the presence of a new wiki entry

Well, actually I think they will, given that it would just be a regular post in a different category, with the ability for regular users to edit it. But it's just a thought, no big deal.

jdtournier:
Yes, there's been a few changes of direction on the FAQ front... Latest suggestion is[ this one](https://github.com/MRtrix3/mrtrix3/issues/1394#issuecomment-411755210).

PR #1460, opened 2018-09-30
mrdegibbs: disambiguate pronoun

maxpietsch:
popped up here: http://community.mrtrix.org/t/mrtrix-tutorial-available-on-osf/1942/3


thijsdhollander:
Makes sense. :+1: 

...however, your docs update is doing something odd.  It seems to have duplicated several config file option docs sections, as well as even change a few values... I suppose that's not intended?

maxpietsch:
Oops, the docs update script picked up config file parameters from old registration code I had left in the same folder. Reverted the docs update commit and pushed a clean one.

thijsdhollander:
No worries; looks all good now!

Issue #1468, opened 2018-10-11
merge list of executable and scripts


jdtournier:
But mention in listing whether they're C++ or python

Lestropie:
Closed by #1546.

PR #1471, opened 2018-10-16
Script documentation changes

Lestropie:
Separating this from the rest of #1449 for the sake of generating a clean online diff.

In the process of generating example help pages for exemplary *MRtrix3* binary & script commands, and also in light of #1468, I thought that it would be preferable to make the documentation for Python scripts even more closely mimic that of the C++ binaries. So this change:

- Re-groups the command-line options, so that "standard options" matches that of the compiled binaries and those command-line options that are available in all Python scripts but not in compiled binaries are instead placed into their own second command group.

- Slightly alters the header at the top of the help page. `git` version of the script (which may be an external project not part of the *MRtrix3* core) is reported top-left, command name bold in centre, no date in top-right (since there's no compilation date). If the script is an external project and not a part of *MRtrix3*, the version of *MRtrix3* is then reported on the second line, just as is done for C++ external projects.

### DIFF
--- a/docs/reference/scripts/5ttgen.rst
+++ b/docs/reference/scripts/5ttgen.rst
@@ -34,26 +34,31 @@ Options common to all 5ttgen algorithms
 
 - **-sgm_amyg_hipp** Represent the amygdalae and hippocampi as sub-cortical grey matter in the 5TT image
 
+Options specific to Python scripts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- **-nocleanup** do not delete temporary files during script execution, and do not delete temporary directory at script completion
+
+- **-tempdir /path/to/tmp/** manually specify the path in which to generate the temporary directory
+
+- **-continue <TempDir> <LastFile>** continue the script from a previous execution; must provide the temporary directory path, and the name of the last successfully-generated file
+
 Standard options
 ^^^^^^^^^^^^^^^^
 
-- **-continue <TempDir> <LastFile>** Continue the script from a previous execution; must provide the temporary directory path, and the name of the last successfully-generated file
+- **-info** display information messages.
 
-- **-force** Force overwrite of output files if pre-existing
+- **-quiet** do not display information messages or progress status. Alternatively, this can be achieved by setting the MRTRIX_QUIET environment variable to a non-empty string.
 
-- **-help** Display help information for the script
+- **-debug** display debugging messages.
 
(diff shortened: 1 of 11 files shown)


thijsdhollander:
Anyone has any issues with this?  If not, I'll click the merge button on Monday or so.  I'd like to get some things cleaned up before creating other stuff.

thijsdhollander:
Ah, hold on, this is a pull request against `script_changes` (as base)...?  I might be misunderstanding the goal of this (separate) pull request then...

Lestropie:
Yes, this was based to `script_changes` so as to provide a clean diff for showing the changes being made specifically to the documentation as inspired by writing the manuscript, but will be merged to `dev` as part of the wider script library changes in #1449. I just want to double-check that the docs appearance for both built-in commands and external projects works as intended. I may also address the slightly erroneous-looking appearance of ungrouped options as part of this PR. Also note that the changes here may be altered in the future as part of addressing #1479.

Lestropie:
## Help page headers

Binary, part of core installation:
```
MRtrix 3.0_RC3-290-ge1982bbb-dirty demo_binary_core                  Nov 12 2018

     demo_binary_core: part of the MRtrix3 package

SYNOPSIS
     Clamp image intensities below the specified threshold to zero
```

Python, part of core installation:
```
MRtrix 3.0_RC3-290-ge1982bbb-dirty demo_python_core

     demo_python_core: part of the MRtrix3 package

SYNOPSIS

     Clamp image intensities below a threshold to zero
```

Binary, external project:
```
Version unknown                demo_binary_project                   Nov 12 2018
using MRtrix3 3.0_RC3-290-ge1982bbb-dirty

     demo_binary_project: external MRtrix3 project

SYNOPSIS
     Clamp image intensities below the specified threshold to zero
```

Python, external project:
```
Version unknown                demo_python_project
using MRtrix3 3.0_RC3-290-ge1982bbb-dirty

     demo_python_project: external MRtrix3 project

SYNOPSIS

     Clamp image intensities below a threshold to zero
```

-----

So I think this is right (apart from the inconsistency between "SYNOPSIS" and the actual synopsis: there's a space in Python and no space in the binaries). The version in the top left corner is the git version of the core at time of compilation for files part of the core, and the output of `git version` for an external project. The date in the top right is absent for Python scripts as they don't have a build date. External project commands have the "using MRtrix3" second line, which is based on the `library_version` in the case of C++ and `lib/mrtrix3/_version.py` in the case of Python.

PR #1472, opened 2018-10-19
dcminfo: fix -tags output

jdtournier:
As discussed on [this forum thread](http://community.mrtrix.org/t/dcminfo-tag-broken/1986).

Suggest committing to `master` since this is a (admittedly minor) bug fix, but also unlikely to affect anything else. Happy to commit it to `dev` if this is deemed more appropriate.

In case you're wondering: this passes all testing on my DICOM store.

### DIFF
--- a/cmd/dcminfo.cpp
+++ b/cmd/dcminfo.cpp
@@ -76,11 +76,8 @@ void run ()
     while (item.read()) {
       for (size_t n = 0; n < opt.size(); ++n)
         if (item.is (tags[n].group, tags[n].element))
-          tags[n].value = join (item.get_string(), " ");
+          std::cout << MR::printf ("[%04X,%04X] ", tags[n].group, tags[n].element) <<  item.as_string() << "\n";
     }
-
-    for (size_t n = 0; n < opt.size(); ++n)
-      std::cout << tags[n].value << "\n";
 
     return;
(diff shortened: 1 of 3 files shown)


thijsdhollander:
Slightly related to this, but separate to the pull request itself: would it not actually be possible to (eventually) incorporate tests for this?  Of course we don't want to include massive images in the testing, but would it be possible to strip most of the data somehow from DICOM images; i.e. create very small images, but still with a full header, with the purpose of mostly testing import of the pieces of information in headers? (especially the ones we are most interested in importing correctly, e.g., gradient table, phase encoding information, etc...)

jdtournier:
> I can't imagine anyone relying on the erroneous behaviour that's fixed by this pull request.

Well, the concern is more that some of the changes I've made here actually affect the core DICOM handling code - albeit very superficially. So I would still be a bit wary of committing them to `master` unnecessarily...

But in any case, I'll merge this. 

> would it not actually be possible to (eventually) incorporate tests for this?

Possible, but a lot of work. Also checking things like DW gradient table import really needs a full dataset , since the information is spread over multiple files. And we'd need data from different versions of the major manufacturers, so we'd still be looking at a large data store for this to work, I think. It would require a lot of thought...

PR #1478, opened 2018-10-26
PBRT writer for tckconvert

blezek:
The PBRT rendering engine (https://www.pbrt.org/) is a [well documented](http://www.pbr-book.org/) and versatile open source rendering engine on the order of Pixar's RenderMan.

This pull request adds support to export `.tck` files as PBRT `curve` objects, essentially a series of
`C0` and `C1` continuous Bezier splines.

The exporter allows the radius of the curves to be set and honors the `-increment` flag to skip points along the streamlines, thus reducing file size and rendering time.

The pull request includes a simple smoke test to detect if the exporter writes a file of the correct number of lines.

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -675,7 +675,9 @@ void run ()
         writer.reset( new PLYWriter(argument[1], increment, radius, sides) );
     }
     else if (has_suffix(argument[1], ".rib")) {
-        writer.reset( new RibWriter(argument[1]) );
+        auto radius = get_options("radius").size() ? get_options("radius")[0][0].as_float() : 0.1f;
+        writer.reset( new RibWriter(argument[1], radius) );
+    }
     }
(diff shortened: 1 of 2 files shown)


Lestropie:
Hey @blezek,

I've changed the target branch to `dev`, since we now accumulate all non-bugfix changes there and eventually merge the whole lot to `master` as a tagged update. Looks like you'll need to make a merge commit with a little conflict resolution. Once that's done, we can probably accept the PR fairly quickly: This is a very specific capability, which I would trust you to vet more so than myself, and it's exceptionally unlikely to lead to issues for other users.

Rob

blezek:
Hey @Lestropie, This was a good experiment with alternative rendering methods, but PBRT won't work for my needs.  I'll leave the code in my fork for posterity (and a good reminder on how to interpolate points with Bezier curves), but I'll close the PR.

Appreciate it all the same,
-dan

PR #1482, opened 2018-11-09
Update dev with master hot-fixes

thijsdhollander:
I've just revived and reused the latest hot-fix branch for this (seemed to be the most efficient way).


thijsdhollander:
Side-note: I didn't realise we could actually simply use this strategy to stick close to the GitFlow model, without requiring an artificial branch each time we want to get a hot-fix both on `master` as well as on `dev`.  So well, everyone's encouraged to get their hot-fixes immediately on `dev` once they land on `master`, I'd say. :slightly_smiling_face:

jdtournier:
Not sure what you mean by reusing the latest hot-fix branch? Just restoring the last branch to be merged to `dev`? It sounds sensible in principle - although I'd have used one that was actually named something like 'dev_hotfix_updates_from_master` or something: the merge commit from this one currently shows _Merge pull request #1482 from MRtrix3/fix_mrview_norealign_option_ as the commit message... I don't think that's hugely reflective of what happened...? Assuming I've read this right, I'm all for re-using or creating a branch with a clear obvious name like the above, and we use that every time. 

Let me know if I've got the wrong end of the stick...

thijsdhollander:
Yep, in this case, I meant just restoring the last branch.  I agree that here, it might have been a bit weird wrt the naming due to it being a mashup of several hot fixes; but that's exactly my point: if everyone does this *immediately* for their hot fix(es), it's very explicit in naming, and `dev` remains closely in sync with `master`.  Note that, again, it was somewhat unusual to have so many hot fixes in such a short time frame: in principle, hot fixes should be quite an exceptional occurrence in GitFlow.

thijsdhollander:
Yep, sounds like we're on the same page now.  So in general, everyone is encouraged to always try and nicely wrap up their "GitFlow" in context of a hot fix.  It's quite simple actually; I did all of this just from the GitHub website in my browser.  Just had to wait for the tests (for `dev` pull request) to complete; but I could simply push the button when done (that was even from my phone, if I recall correctly).  Even small merge conflicts can now be resolved via an editor in the website it appears.

Issue #32, opened 2014-05-08
Move to C++11


Lestropie:
Just been reading up on this myself; bits of it look like coding nightmares, other bits look super useful. Disadvantage (apart from the work in going through and updating relevant code) is that HPCs may not compile it, but static compiles could be used instead.

Lestropie:
... Well that was kind of anticlimactic.

PR #1483, opened 2018-11-09
mrcalc: added 'pi' as special keyword.

maxpietsch:
Adds mathematical constants pi and e to mrcalc.


maxpietsch:
The only controversial bit I can see about the changes are defining `e` using the decimal [representation](https://github.com/MRtrix3/mrtrix3/pull/1483/commits/a9b7f9a592c6fd6ac954266ba50fb1490ccc7879#diff-48a4a0d15b5a030512e1e6910a7ea1abR35) instead of the `<cmath>`'s built-in `M_e`. This is to be consistent with `Math::pi`. Are we happy to merge this into dev?

thijsdhollander:
All good; happy about the consistence.  Maybe add to the documentation that "e" is Euler's constant. (e.g. between brackets or something)

jdtournier:
Looks good to me too. The reason we don't use the math.h `MATH_*` macros any more is that they're not part of C++11 - see [this post](https://github.com/MRtrix3/mrtrix3/issues/32#issuecomment-52288266). 

I'll merge.

jdtournier:
Oops, sorry @thijsdhollander, forgot about your suggestion for adding to the docs. I don't personally see a need for clarifying what `e` refers to, I think it should be fairly obvious to anyone who needs to use it... Feel free to re-open and add if you feel the need.

thijsdhollander:
No worries at all; it's a minor detail.  It'd certainly be clear to those that need it, but may otherwise confuse a few others since it's a single letter/character.  No need to re-open just for that though.  I may add it at some point if/when working on some docs or something, if I remember it then.  All good for now.

PR #1489, opened 2018-11-17
Command example usages in documentation

Lestropie:
As hinted in comment on #1086, came to the realisation that I won't be able to pass CI testing in #1449 whilst providing an example usage in the help page of `foreach` until the functionality described in #1086 was implemented. So I just knuckled down and did it.

If anyone can think of any additional command example usages that may be informative to provide, would suggest adding them here, as it will probably be less likely to go to the effort of writing them in isolation once this PR is done.


thijsdhollander:
I might change a few things about the `dwi2fod` examples; some are quite exotic, or their text might confuse users or lead them to think they're using another CSD algorithm.

thijsdhollander:
Anyone else still has any problems with this one?  If not, let's hit the button.  We can add further examples later on, now that the feature is functional.

Lestropie:
I have one thing left to do.

Lestropie:
Any Mac-heads provide any insight into the [TravisCI error](https://travis-ci.org/MRtrix3/mrtrix3/jobs/477633712)? Not finding anything online.

```
Error: No such keg: /usr/local/Cellar/qt
The command "if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
      brew install qt5
      brew link --force qt5
  fi
  " failed and exited with 1 during .
```

jdtournier:
No idea. Maybe @maxpietsch will have some insights here. Just in case it's useful, the preceding warning might have some bearing on the problem:
```
Error: HOMEBREW_LOGS was not exported!
Please don't worry, you likely hit a bug auto-updating from an old version.
Rerun your command, everything is up-to-date and fine now.
Error: No such keg: /usr/local/Cellar/qt
The command "if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
      brew install qt5
      brew link --force qt5
  fi
  " failed and exited with 1 during .
```

Issue #1087, opened 2017-08-10
Python: Location of algorithm files


Lestropie:
With the FHS re-arrangement of the repository, we now have the following situation:

`bin/5ttgen`
`lib/mrtrix3/app.py` (and others)
`lib/mrtrix3/_5ttgen/fsl.py`
`lib/mrtrix3/_5ttgen/freesurfer.py`

That placement was maybe not super-critically thought out with the FHS changes. I'm now not sold on it: The files in `lib/mrtrix3/_5ttgen/` aren't really a part of the "library", since they're by design intended to be used by a single script only. Additionally, if someone were to write a standalone script against the *MRtrix3* Python libraries (which they can, and I myself do), they would *not* be able to make use of the `algorithm` module, since the corresponding module files need to go into *MRtrix3*'s `lib/mrtrix3/` directory, regardless of where the executable script resides.

They clearly can't reside directly in `bin/`, since they're not executable in and of themselves, and that could lead to naming conflicts. Likewise they don't really belong in `core/`, since a packaged installation needs to contain these files yet that whole directory doesn't exist in the case of packaged installs.

I'm wondering if they should instead reside in sub-directories of `bin/`? One sub-directory per module-based script as is currently done, with the name of the sub-directory being the executable script's name, somehow modified to prevent naming conflicts. So we'd have e.g. `bin/_5ttgen/` and `bin/_dwi2response/`. The `algorithm` module would then find them based on the real path (i.e. follow symlinks) of the executed script, rather than the location of the library files.

thijsdhollander:
> That placement was maybe not super-critically thought out with the FHS changes. I'm now not sold on it

Yes, I fully agree with this.  But the problem is then indeed what makes more sense (I'm not sure myself...).  I understand the original reasons for having the "main" scripts (i.e. not talking about the algos here) in `bin/`, but I have to admit this still jars a bit for me.  Moving them there has somewhat contributed to the issue voiced here now as well.  I've got a similar feeling that the main scripts (e.g. `dwi2response`) should be sticking closer to their algorithms.  But where I'm undecided is whether that should be both in `bin/` (as you suggest; I wouldn't be against this in fact), or somewhere else.  And if somewhere else, where that should be then of course...

[...]

PR #1492, opened 2018-11-20
Move Python script algorithms to src/

Lestropie:
In addition to moving the code for these script algorithms out of the `lib/` directory, highlighting that they are in fact not a part of the script library but are non-executable "source files", this change also references these source files from the location of the executed binary rather than the module file: this means that external projects can make use of the algorithm module.

I never liked the location of these script algorithm files ever since converting the repo to FHS. This I think makes more sense.


jdtournier:
I'm not sure I get the motivation behind these changes... What's wrong with the current location, other than you don't like it? The current location is the most appropriate, as far as I understand current conventions on Unix packaging - see [this post](https://github.com/MRtrix3/mrtrix3/issues/1087#issuecomment-321578248) and associated links therein to reacquaint yourself with all the discussions we had at the time. [This post](https://github.com/MRtrix3/mrtrix3/issues/864#issuecomment-270337588) in particular deals with the `src/` folder issue (at the end). 

So basically, unless there is a _technical_ reason why this is necessary, I don't see why you'd need to go against the FHS. The layout we have now is very consistent with what you'd find in most packages. Take a look at the file listing for packages on your system, you'll find that 'plugins' typically reside in `lib/<package>/` .

Lestropie:
My concern with being in `lib/` is that there may be a perception that these components are somehow immutable, whereas in reality I'd very much encourage anyone to attempt creation of new algorithms, and `src/` maybe has a connotation of being less hard-wired and more conducive to introducing new files without "affecting the library". But if this is strongly against FHS (leaning on your expertise here), that's fine.

But I would like to keep the change whereby the location of these files is determined relative to the executable file, rather than the library module file: this permits use of the `algorithm` module in external *MRtrix3* projects. This would rely on predictable positioning of e.g. the `lib/mrtrix3/dwi2response/*.py` files relative to `bin/dwi2response`, as opposed to relative to `lib/mrtrix3/`. Is it possible for that to cause any issues in specific installations? I could potentially query both relative paths if necessary.

thijsdhollander:
I'm happy to keep it as is as well, to be honest.  I had other ideas about this initially, but at some point we had the discussion, came to a conclusion and went with it.  I simply don't want to change this all over again.  There's already more than enough changes (again) coming up; this becomes a burden to maintain each time.
In my collaborations, for users building their own scripts for certain tasks, if they don't rely on intricate image manipulations, I've started advising them against using our framework, at least for the time being (until things settle a bit more).  I get that we love tinkering with this, but users aren't involved in all these developments, and with each release they're being thrown off their game.  It's already become quite an investment to "learn" our scripting library to begin with (and if you're only going to need it for a single task, the question becomes whether it's worth that investment/time to learn it), but if each time the rules of the game change, and more and more rules are being put into place, it's not going to help.

Ahem, absolutely sorry for my little slightly off-topic rant here.  But well, all meant constructively for sure: we've got to keep an eye on the ever-evolving complexity we're creating, and probably keep on wondering (and challenging our ideas on this) whether it's necessary, or what purpose we're creating certain rules and structures for.

On topic again: to be very honest, I don't think either location gives most of the users either impression or causes a certain perception.  The algorithms are the algorithms, and if someone wants to add another file in the directory they are stored in, surely their first guess would be that the "master" script for those algorithms is going to pick it up, right?  Given the growing complexity of creating a simple script or algorithm, most users would probably do it "by example" anyway: simply copy an existing algorithm, strip the functional code out of it and replace it with their own (probably also "by example").

jdtournier:
> But I would like to keep the change whereby the location of these files is determined relative to the executable file, rather than the library module file: this permits use of the algorithm module in external MRtrix3 projects. 

OK, I see your point now. 

> This would rely on predictable positioning of e.g. the lib/mrtrix3/dwi2response/*.py files relative to bin/dwi2response, as opposed to relative to lib/mrtrix3/. Is it possible for that to cause any issues in specific installations?

Not for our standard stuff, no. At least not as far as I can see. But to make this useful for external apps, the location to be queried would need to be specified in the app itself right? So presumably we'd need to provide a convenience function to set that? In which case, you could have a default location (relative to the `lib/mrtrix3` folder as currently), and allow apps to _add_ locations to what in effect becomes a plugin search path. Is this more or less what you have in mind...?

Lestropie:
> But to make this useful for external apps, the location to be queried would need to be specified in the app itself right?

Not if:

> ... rely on predictable positioning of e.g. the `lib/mrtrix3/dwi2response/*.py` files relative to `bin/dwi2response`

I.e. If the external app conforms to the *MRtrix3* external project directory structure, you can have an executable at `~/module/bin/mrapp` and algorithm files at `~/module/lib/mrtrix3/mrapp/*.py`.

Issue #364, opened 2015-09-28
New command - mrgrid


draffelt:
As discussed originally in Gitlab, we decided that mrresize should be incorporated into mrtransform (#15). Even though this does not strictly transform the image, it was decided that it should be ok to have the mrresize options (-scale, -voxel, -size) as a part of the "regridding options" in mrtransform. 

However, given that we also have mrcrop and mrpad, I'm wondering if it makes more sense to have a single `mrgrid` command. We could then have separate option groups for crop, pad, and resize.

This would also mean that we avoid having to hide the mrresize options to mrtransform (which is already a fairly large command). 

What do you think?

thijsdhollander:
Yes, I think this is a great idea.  Now that I started thinking about it, `mrresize` was actually a bit deceiving: it sounds like it will (allow to) scale the actual images, i.e., as in actually transforming them.  A command named `mrgrid` to basically just map an image to a new grid will avoid that confusion.

Maybe within the context of such a command, it might be worth to "generalize" some cropping and padding options, i.e., turning them into more unified "field of view" options (if that makes sense).

jdtournier:
Happy enough with all of the above - although I'm still generally of the opinion that these operations are all sufficiently similar to justify being part of a single command. I can see that there is a subtle distinction: one set of operations doesn't modify the position of the underlying object in real space or the axes used for the grid (`mrgrid`), the other set does (`mrtransform`). It's just a matter of whether that distinction is sufficient to warrant a different command. 

I'll go through my arguments again, and you can happily ignore me if you still disagree, happy to go with the majority vote on that one. 

First off, I would expect that all of the various modes of operation offered by `mrgrid` will also need to be available within `mrtransform`, so that invoking `mrtransform` using an identity transform and appropriate options should allow you to do everything that `mrgrid` does. Maybe that might require tweaking an existing image (and maybe that's what `mrgrid` might allow you to do) to pass to `mrtransform`'s `-template` option, but the point is it should all be feasible out of the box. So the only functionality to add to `mrtransform` would be options to somehow modify the template image provided - the actual regridding code would be more or less unmodified.

[...]

PR #1494, opened 2018-11-21
mrpad: fix header translation of y and z axes

maxpietsch:
Previously the header translation was erroneously or not modified if the padding in y- and z-direction differs from that in the x-direction.


jdtournier:
It's a minor issue, but a clear bug, so worth fixing now. Note that it'll most likely be redundant in `dev` by the time #364 gets finalised.

PR #1497, opened 2018-11-23
DICOM: increase resilience to errors and empty fields

jdtournier:
A bunch of simple changes to prevent crashes when reading slightly malformed DICOM files with field present but left empty.


jdtournier:
Changes are best appreciated by [hiding whitespace changes](https://github.com/MRtrix3/mrtrix3/pull/1497/files?utf8=%E2%9C%93&diff=split&w=1) - it's not as big a changeset as the regular diff suggests...

Issue #1495, opened 2018-11-22
Screen capture translate type "camera" is broken


thijsdhollander:
People trying to do some surgical planning here, and coming across issues with the screen capture tool.  

Goal: moving a coronal slice from back to front of the brain; but they want to choose a non-voxel-grid-aligned (and also non-scanner-axis-aligned) coronal plane.

Practice: essentially, they set up their (e.g. anatomical) volume by switching to sagittal view, and then choose the desired angle via ctrl+right-click (i.e. rotate in plane sagittally); i.e. they choose which coronal slices they'll be seeing. Then they switch back to coronal view, scroll to the back of the brain, and want to use the screen capture tool to move through the brain to the front, while capturing stuff.  To that end, a number is put in the second (y-axis) box of the translate axes in the screen capture tool.  But they can't get the job done with any of the translate types.  They asked me, so I naturally assumed the "camera" type is the one supposed to be doing this job (and essentially show the same slices you'd be seeing if you where manually scrolling through at this point), but the coronal slice, while moving through the brain as desired, also moves up/down and/or left/right in the viewer.

Temporary solution I hacked them together (as a manual process): using the scanner option instead, and adding number in the z-axis and potentially even x-axis boxes to compensate.  Essentially: inherently constructing the camera axis manually as a linear combination of the scanner axes.  I advised the scanner axes to use for the purpose of this "hack" for the time being, as the camera ones behave very unpredictably (so to put it simply; we don't trust the camera option at this point :slightly_smiling_face:).

I've got the feeling a wrong combination of (scanner or voxel) axes is used maybe somewhere in the code to compute the camera axis... or something (ignore this if it sounds like nonsense)?  Not sure... 

@maxpietsch @jdtournier , also not sure if this might have to do anything (at all) with the linked axes issue you guys were recently looking into (?); but bringing it up just in case...

jdtournier:
Confirmed. I'll look into it when I have a minute. 

Nothing to do with the linked axis issue though...

jdtournier:
Closed with #1498

PR #1498, opened 2018-11-23
MRView: fix translation in camera coordinates in screen capture tool

jdtournier:
Suggested fix for #1495. Note that this maps the x,y,z axes of the camera to the display's left->right, bottom->top, far->near. So to move forwards through the data requires a negative displacement along z. 

Suggest merge to `master` as clearly a bug fix, and unlikely to have unsavoury side effects.


thijsdhollander:
Same problem here as with #1499.  I've not been keeping up with all the action that happened late on Friday and/or over the weekend; so it's a bit of a blur for me at the moment. :confused::question:

jdtournier:
OK, I cleaned this up by re-committing the relevant changes _without_ changing the test_data commit... Requires a forced-pushed, but hopefully it'll make for a cleaner history...

jdtournier:
In case anyone's wondering: remaining issues relate to the new version of `pylint`. Fixes are in #1504. I'll wait until I get confirmation that they're OK from @Lestropie, and once merged  to `master`, we can bring these changes into this PR (and #1499).

PR #1503, opened 2018-11-25
dwiextract: Fix extraction based on total readout time

Lestropie:
Fix to selection of DWI volumes where there are volumes with the same phase encoding direction but different readout times.

Thanks to @treanus for report, solution & test.


jdtournier:
Sorry for dropping the ball on that one - forgot all about it...

treanus:
No worries ☺️

PR #1500, opened 2018-11-23
merge hotfix of mrpad and mrpad test data

maxpietsch:

PR #1504, opened 2018-11-25
mrtrix3.image.statistic(): Fix parameter naming

Lestropie:
`app.error()` call contained a reference to the function being invoked, rather than the parameter that was passed to the function.

Not hugely consequential given the script is throwing an error at that point anyway; but nevertheless preferable for a more appropriate error message than "`TypeError: must be str, not function`".

Observed on [forum](http://community.mrtrix.org/t/dwi2response-dhollander-script-meet-error/2054).

### DIFF
--- a/lib/mrtrix3/_5ttgen/gif.py
+++ b/lib/mrtrix3/_5ttgen/gif.py
@@ -18,15 +18,13 @@ def checkGIFinput(image_path):
 
 
 def getInputs(): #pylint: disable=unused-variable
-  import os, shutil #pylint: disable=unused-variable
   from mrtrix3 import app, path, run
   checkGIFinput(path.fromUser(app.args.input, True))
   run.command('mrconvert ' + path.fromUser(app.args.input, True) + ' ' + path.toTemp('input.mif', True))
 
 
 def execute(): #pylint: disable=unused-variable
-  import os, sys #pylint: disable=unused-variable
-  from mrtrix3 import app, path, run #pylint: disable=unused-variable
+  from mrtrix3 import app, run #pylint: disable=unused-variable
 
   # Generate the images related to each tissue
   run.command('mrconvert input.mif -coord 3 1 CSF.mif')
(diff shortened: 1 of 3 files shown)


jdtournier:
Same issue here as on #1500, but with added goodness:
```
## testing "lib/mrtrix3/_5ttgen/gif.py"...
 [ ERROR ]
************* Module mrtrix3._5ttgen.gif
lib/mrtrix3/_5ttgen/gif.py:21:2: W0611: Unused import os (unused-import)
lib/mrtrix3/_5ttgen/gif.py:21:2: W0611: Unused import os (unused-import)
lib/mrtrix3/_5ttgen/gif.py:28:2: W0611: Unused import os (unused-import)
lib/mrtrix3/_5ttgen/gif.py:29:2: W0611: Unused app imported from mrtrix3 (unused-import)
lib/mrtrix3/_5ttgen/gif.py:28:2: W0611: Unused import os (unused-import)
```
Given that this is all based on `master`, all parent commits passed testing, and none of the commits since have touched this file, I'm stumped. The only explanation as far as I can tell is that `pylint` has suddenly decided to become stricter...

In any case, I'll try to fix this one too.

jdtournier:
I note a stack of errors in the TravisCI logs at the `pip3 install pylint` stage. Not sure whether the version of `pylint` that is eventually installed is the right one. I note `pylint` 2.2 just came out - tests run clean locally on my system if I use the previous 2.1.1 version, but not 2.2 (on top of that, the tests for `labelsgmfix` cause a frank crash of `pylint`...). I've just added a line to `run_pylint` to output the version information so we can see what is running exactly (though I note I'll need to fix that, it's not quite right...).

jdtournier:
OK, this looks like it works...

@Lestropie: please review these changes and make sure you're happy. There's changes to your GIF algorithm, and to the `run_pylint` script - both of which are yours. Since this is going to `master`, I think it's best you give it the thumbs-up (I can't add you as a reviewer since you submitted the PR originally...).

Lestropie:
> The only explanation as far as I can tell is that pylint has suddenly decided to become stricter

On cursory inspection perhaps the new `pylint` emits `unused-import` instead of `unused-variable`, the latter of which was explicitly silenced on the relevant lines (which is acceptable but should only be used where absolutely necessary).

> There's changes to your GIF algorithm ...

[Not mine](https://github.com/MRtrix3/mrtrix3/pull/1249).

PR #1509, opened 2018-12-08
dwipreproc: Compatibility with FSL6.0.0

### LINKED ISSUE #1493: dwipreproc explicit volume recombination
https://github.com/MRtrix3/mrtrix3/blob/16090b5b4ccdad9652bada2473e904ecfc64c047/bin/dwipreproc#L996

translates to

```
mrcalc field_map.nii.gz 0.123 -mult - | mrfilter - gradient - | mrconvert - field_deriv_pe_1.mif -coord 3 1 -axes 0,1,2
```
causing

```
mrcalc: [100%] uncompressing image "field_map.nii.gz"
mrcalc: [100%] computing: (field_map.nii.gz * 0.123)
mrfilter: [100%] applying smoothing prior to calculating gradient
mrfilter: [100%] applying gradient filter to image -
mrconvert: [ERROR] omitted axis "4" has dimension greater than 1
```
when performing explicit volume recombination

field_map.nii.gz is 4D containing 6 volumes
after mrfilter this becomes 5D containing 3 x 6 volumes
subsequent mrconvert then fails when trying to omit a non-singleton dimension

Lestropie:
My guess is that somewhere in the `topup` code, when provided with the `--fout` option to output the estimated field, it generates a deep copy of the input *b*=0 series, and then overwrites the contents of the first volume with the estimated field. However because they erroneously don't reduce the template image to 3 axes when generating said copy, the output "field map" image contains copies of all but the first input *b*=0 volume.

Would expect to be fixed reasonably quickly at their end, but need to catch and handle here nonetheless.

Closes #1493.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ install:
   - |
     if [[ "${test}" == "build" || "${test}" == "run" ]]; then
         export EIGEN_CFLAGS=-I`pwd`/../eigen;
-        (cd ..; hg clone https://bitbucket.org/eigen/eigen/; cd eigen; hg update 3.3);
+        ( cd ..; git clone https://github.com/eigenteam/eigen-git-mirror.git eigen; cd eigen; git checkout branches/3.3 )
     fi
 before_script:
   #######################################################################################################################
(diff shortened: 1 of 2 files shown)


jdtournier:
Ok, might be time to use [Eigen's new official git repo](https://github.com/eigenteam/eigen-git-mirror) and ditch mercurial...?

Issue #1514, opened 2018-12-13
version info on non-git builds & platforms with SSE support only?


yarikoptic:
Debian supports (way too ;)) many non x32 architectures, but builds there fail:
https://buildd.debian.org/status/package.php?p=mrtrix3&suite=sid
and quite often simply due to demand of the following header file for SSE support (available only for i386 and amd64)
```
In file included from cmd/labelconvert.cpp:16:
./core/command.h:20:10: fatal error: xmmintrin.h: No such file or directory
 #include <xmmintrin.h>
          ^~~~~~~~~~~~~
compilation terminated.
(  1/505) [CC] tmp/cmd/labelco
```

before we go ahead and just blindly restrict architectures, I wondered to ask -- do you see it feasible to support non-SSE ones?  I could outline potential benefits if you are interested ;)

jdtournier:
No problem, this one is not actually used for production code... Have a look at #1515, see if that fixes it for you.

jdtournier:
> I could outline potential benefits if you are interested ;)

No need, but you've caught my interest now... What are we talking about?

jdtournier:
One more thing: I notice that many of these builds are failing due to failed dependencies. I assume this is related to Qt? If so, you may still be able to offer a command-line only build: just run `./configure -nogui`, and it'll skip the Qt tests and skip building `mrview` and `shview`.

jdtournier:
And one more while I'm at it: it looks like the install runs from a download of the source code, rather than a git clone. The issue there is that the version information that gets baked into the executables is derived from `git describe`. So these executables will show up as 'version unknown' in their help page and `--version` output, which is not ideal. Is there any way to use a git clone here...?

yarikoptic:
> Is there any way to use a git clone here...?

not directly.  But since I will be preparing the source tarball out of the git repo, we could generate a file with versioning information which could be picked up during build.  Many projects have such a practice:
```shell
$> grep export-subst */.gitattributes 
cctools/.gitattributes:configure export-subst
dipy/.gitattributes:dipy/COMMIT_INFO.txt export-subst
nipy/.gitattributes:nipy/COMMIT_INFO.txt export-subst
nipype/.gitattributes:nipype/COMMIT_INFO.txt export-subst
nipype-master/.gitattributes:nipype/COMMIT_INFO.txt export-subst
pandas/.gitattributes:pandas/_version.py export-subst
[...]

PR #1515, opened 2018-12-13
only include xmmintrin.h header if actually used

jdtournier:
This otherwise prevents compilation on non x86 platforms (see #1514)

This is a harmless change, no reason not to merge to `master`.


jdtournier:
@maxpietsch: I assume this is OK with you too...?

PR #1516, opened 2018-12-14
build: update version information before checking presence of config

jdtournier:
This is to allow version information to be updated easily for packaging purposes (see #1514).

Pushing this for discussion purposes - no urgency to merge as yet.

This probably needs to be tested fairly thoroughly before we go ahead with it. Seems to work on my system, but you never know...


Lestropie:
@jdtournier How necessary is it that this go to `master`?

jdtournier:
OK, it looks like all this machinery is already on `dev` (#1593). Let's close this, there's no real need to for it on `master`.

Issue #771, opened 2016-09-02
dwidenoise: Don't skip edge voxels


Lestropie:
At the outer edges of the volume, where the kernel window centred at the voxel of interest is partly outside the image FoV, currently those voxels are skipped during denoising.

Instead, define a 'fattening' window that extends further along those axes that are not constrained by the image FoV, in order to provide a comparable total number of voxels within the 'fattened' kernel to that of the standard cubic kernel. This ensures that the kernel window is still unique for each voxel in the image.

Lestropie:
Had a change of thought process on how to do this, figured I'd write it down.

Ultimately one wants a set of nearby voxels for which the noise level is comparable. The cubic sliding window is just one way of doing this; but how best to deal with voxels near the edge of the FoV is ambiguous (could be done, but there's a few alternatives). A more precise expression of "nearby" voxels would be a *spherical* kernel, of an appropriate radius to capture an appropriate number of voxels within the kernel. This has two advantages:

- It intrinsically works better with anisotropic voxels, since expansion of the voxel selection could be done in scanner space rather than being a fixed number of voxels per spatial axis.

- When near the image FoV edge, simply continue expanding the selection radius until an appropriate number of image voxels has been included in the kernel.

Lestropie:
Third benefit of above proposal: The size of the kernel can easily be modulated so as to avoid the region in which the method becomes numerically unstable.

No issue with going with the solution already in #1525; just wanted this to be documented on the off chance that there are further relevant discussions in the future.

PR #1525, opened 2018-12-20
Fix issues with MP-PCA denoising

dchristiaens:
These changes address two major issues with `dwidenoise`:

1. **MP matrix ratio**: This relates to a long standing issue, reported [here](http://community.mrtrix.org/t/dwi-denoising/1221), [here](http://community.mrtrix.org/t/denoising-residuals/1715), and [here](http://community.mrtrix.org/t/dwidenoise-correct-use/586/4), where we found that MP-PCA denoising performs poorly when the no. DWIs (M) is close to the total patch size (N). We also found that cases where M<N and M>N can lead to very different noise maps. We have discussed this issue with Jelle and agreed that the asymmetry stems from the definition of gamma = (M-P)/N in the paper, where P is the number of signal components. In recent work by a colleague at King's ([paper](https://arxiv.org/abs/1812.05954); Fig.4) we also found this definition of gamma to cause the biased noise estimates in near-square matrices, and proposed an alternative definition gamma = (M-P)/(N-P) that leads to more precise estimates. I have now also made this change in the `dwidenoise` code, pending Jelle's approval.

2. **Edge voxel handling**: As reported in #771, the original implementation skipped voxels at the edge of the image, potentially leading to different noise characteristics near the image boundaries when the mask extends to n=(patch size - 1)/2 voxels from the image bounding box. This is now addressed by using a non-centred patch in these voxels. The noise level is then effectively constant in n+1 voxels next to the image boundary, and the SVD spectrum is truncated at this noise level in all affected voxels.

Given that both of these are major changes to the output users can expect, I think this should be merged into a next release and properly documented in the changelog. I will draft in due time.

This PR is provisional while we discuss with NYU; don't merge yet.


thijsdhollander:
> Given that both of these are major changes to the output users can expect, I think this should be merged into a next release and properly documented in the changelog. I will draft in due time.

:+1:  All good.  Yes, definitely via `dev` due to the changed output.

dchristiaens:
Now that Lucilio's paper is out, I have done some work on `dwidenoise` to bring these changes to release. These updates add a capability to revert to the noise level estimator as used in Veraart et al. (2016), implemented with an option `-estimator` that is set by default to the improved formula introduced in Cordero-Grande et al. (2019).

There is one vital caveat though: the "legacy" option `-estimator Exp1` will match the current `master` branch *if and only if* the patch size (N) exceeds the number of DWIs (M) in the dataset (notwithstanding the changes in edge handling). This corresponds to the case evaluated in Jelle's paper. Part of the problem that triggered our search for a new estimator was that the formula's used for cases M < N and M > N in current master are incompatible. In other words, there was an asymmetry in the way both cases were treated, for instance leading to a different degree of denoising on matrices X and X^T in testing. I consider this a bug, and think it is unnecessarily complicated to maintain this behaviour since users who wish to have full "legacy" behaviour can achieve so through the `git` history. Therefore, the recent updates allow users to choose the noise level estimator used in Jelle's paper, rather than Lucilio's, but will not revert to the exact same output as before. This is thus a major update to the command, best merged in as part of the next release and documented clearly on the changelog.

For the same reason, we will also need to update the test data.

dchristiaens:
Complex input data is now supported transparently. Input images with complex datatype will be processed as such. Their output will also be complex. Real input images, on the other hand, are processed as real so there is no performance penalty. It only took me three days to get this working... 🙄

The testing data has been updated. If Travis agrees this should be ready to merge.

dchristiaens:
Updated user docs too. Merging and closing this PR.

PR #1526, opened 2018-12-20
re-formulate and multi-thread amp2response

jdtournier:
On the back of #1406, I've had a look at the maths, and I think this works...

Some timings:
```
$ time bin/amp2response dwi.mif sf.mif ev.mif res.txt -force -noconstraint
amp2response: Shell b=0: Response function [1153.03821] solved via ordinary least-squares from 100000 voxels
amp2response: Shell b=2984: Response function [   141.870285   -25.4580612     7.9886179   -2.04643297   0.499316067 -0.0580862164] solved via ordinary least-squares from 100000 voxels

real    0m0.516s
user    0m5.885s
sys     0m0.010s
```
compared to the current version:
```
$ time amp2response dwi.mif sf.mif ev.mif res.txt -force -noconstraint
amp2response: Shell b=0: Response function [ 1153.03821 ] from average of 100000 voxels
amp2response: Shell b=2984: Response function [   141.870285   -25.4580612     7.9886179   -2.04643297   0.499316067 -0.0580862164] solved via ordinary least-squares from 100000 voxels

real    0m4.963s
user    0m4.261s
sys     0m0.693s
```


Lestropie:
Can confirm identical result in a case where constraints are enforced.

cmd/amp2response.cpp
```diff
@@ -150,12 +152,13 @@ class Accumulator {
     ~Accumulator ()
     {
       // accumulate results from all threads:
+      std::lock_guard<std::mutex> lock (S.mutex);
```
jdtournier: no mutex required. Construction and destruction of functors is guaranteed to happen in main thread. Only the [`execute()` method](https://github.com/MRtrix3/mrtrix3/blob/dev/core/algo/threaded_loop.h#L343-L347) is invoked in each thread.

jdtournier:
> Can confirm identical result in a case where constraints are enforced.

Thanks for confirming - that's the bit I didn't get the chance to validate.

>> amp2response: Shell b=0: Response function [1153.03821] solved via ordinary least-squares from 100000 voxels
> 
> Just wondering why the wording changed from "average" to OLS? It's of course the same thing (which is probably why I'm wondering about the re-wording).

I got rid of that code branch since it seemed unnecessary to me - the OLS branch achieved the same thing anyway, so I made the _l_=0 case run through that too. I can see that it affects the message - hadn't thought of that. Happy to revert or to modify the message for _l_=0 cases - up to you. Personally, I don't see that it makes much difference. I was actually thinking these messages could be simplified, I'd rather see an output like:
```
amp2respone [100%]: estimating responses using constrained least-squares from 100000 voxels
amp2response: response function at b=0: [1153.03821] 
amp2response: response function at b=2984: [   141.870285   -25.4580612     7.9886179   -2.04643297   0.499316067 -0.0580862164]
```
The number of voxels is always the same across all shells, so only needs to be reported once. Same for the fitting strategy. I don't think the fact the _l_=0 shells are estimated via OLS (or averaging) is worth reporting, since monotonicity is guaranteed, and negativity should never be an issue. So the constrained solver should always give the same answer as the OLS anyway. Up to you guys how exact you want to be here.

thijsdhollander:
> I got rid of that code branch since it seemed unnecessary to me - the OLS branch achieved the same thing anyway, so I made the l=0 case run through that too. I can see that it affects the message - hadn't thought of that. Happy to revert or to modify the message for l=0 cases - up to you. Personally, I don't see that it makes much difference. I was actually thinking these messages could be simplified, I'd rather see an output like:

It's been a little while, but I think I'd initially overlooked that the code branch itself had effectively been removed.  So no worries at all then.  That said, I like your proposed simplification, even just because it makes the output shorter; agreed the information on the solver and number of voxels is unnecessarily repeated.

I now realise this can theoretically change the output though, should isotropic (l=0) shells average to a negative amplitude; but that's indeed unlikely for the most common use cases.  And if it would happen, using OLS (or not) consistently for any shell (regardless of l=..) is probably even better.

jdtournier:
OK, I reckon this one is good to go too. I just tweaked the terminal output as I'd suggested above, this is what it produces now:
```
amp2response: estimating response function using constrained least-squares from 163 voxels
amp2response:   b=0.5: [3765.95532]
amp2response:   b=700: [ 2197.87231 -610.310364   89.502243 -8.81993484  7.35450888  3.93733931]
amp2response:   b=1200: [ 1698.30029 -709.048096  166.638474 -30.9259472  5.12017012 -1.11742842]
amp2response:   b=2800: [ 1040.42493 -644.057495  277.310516  -86.348526  19.8303623 -4.77585554]
```

Feel free to merge if happy.

thijsdhollander:
:+1:

PR #1530, opened 2019-01-09
dwinormalise & dwiintensitynorm: Refactor

### LINKED ISSUE #1064: dwinormalise & dwiintensitynorm as script algorithms
Following comment [here](http://community.mrtrix.org/t/preproc-and-dwiintensitynorm-dwinormalise/10450).

The difference between these two is certainly not clear from the command names. Moreover, the internals of `dwinormalise` are quite simple. Apart from the `-percentile` option (which is probably never used, but could nevertheless theoretically be added to `mrstats`), it's probably possible to embed the functionality of this entire command into the `dwiintensitynorm` script, and remove `dwinormalise`. All that is required for each subject is to get the *b*=0 volumes, calculate the median within the mask, and multiply the whole DWI series by the ratio of the target value to that median.

Lestropie:
Decided to attack this given the scope of other changes going on in #1449.

- Remove binary command `dwinormalise`.
- Remove Python script `dwiintensitynorm`.
- Add new Python script `dwinormalise`, which utilises the algorithm module:
  - Algorithm 'individual' duplicates functionality of former binary `dwinormalise`.
  - Algorithm 'group' duplicates functionality of former script `dwiintensitynorm`.

Closes #1064.


docs/reference/commands/tcksift2.rst
```diff
@@ -79,7 +79,7 @@ Options for controlling the SIFT2 optimisation algorithm
 
 -  **-max_coeff_step step** maximum change to a streamline's weighting coefficient in a single iteration (default: 1)
 
--  **-min_cf_decrease frac** minimum decrease in the cost function (as a fraction of the initial value) that must occur each iteration for the algorithm to continue (default: 2.5e-05)
+-  **-min_cf_decrease frac** minimum decrease in the cost function (as a fraction of the initial value) that must occur each iteration for the algorithm to continue (default: 2.5e-005)
```
thijsdhollander: Not really on topic of the pull request, but I've noticed this coming up more often here and there: the precision or number of digits here changing (back and forth, depending on who generated the .rst).  Just wondering: is this something we can control better, or is this dependent on some software version that's out of our hands?

thijsdhollander:
In general, I'm entirely ok with it.  I haven't tested, but all functionality seems to remain in place, just under a different name/algo/call; so all good.  Of course don't forget to modify the documentation.  Off the top of my head, I'm guessing this is only referred to in the old single-tissue FBA pipeline...?

Lestropie:
> ... the precision or number of digits here changing (back and forth, depending on who generated the .rst)

Windows systems generate three digits for exponents in scientific notation, compared to two on Linux. I'm constantly `git checkout`-ing a couple of documentation pages whenever I work from home in order to avoid exactly that, because it'll lead to failure of the `sphinx` CI tests. Windows also reverses the sorted order of `tcksift` and `tcksift2`. Haven't actually looked into preventing it from happening though.

> I haven't tested ...

I'll be adding a bunch of Python script tests in a separate PR to `dev` ahead of #1449. So because this PR is going to the `script_changes` branch, those new tests will first be derived based on current `dev` code, and then modified accordingly (obviously with manual verification that the results are the same) in order to resolve #1449.

> Of course don't forget to modify the documentation.

Yep, I'll get to that.

Lestropie:
If there's no other opinions regarding this refactoring, I'll merge it into `script_changes` ahead of other things in preparation for #1449.

PR #1531, opened 2019-01-11
DICOM: Variable parameters

Lestropie:
Playing with a project involving T2 relaxometry, and it was annoying me that I couldn't just see the echo times using `mrinfo`. This just concatenates unique values across frames for echo time, flip angle & repetition time.

Technically I think this should also work if such values vary not just between volumes but also between slices (as long as such is reflected in the DICOM headers).


bjeurissen:
That's great!

PR #1535, opened 2019-01-13
TravisCI: Try to fix Qt installation

Lestropie:
Issue raised in #1489.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ install:
   ##################################################
   - |
     if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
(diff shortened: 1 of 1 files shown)


Lestropie:
Fix proposed in #1489 works around the initial error, but results in an attempt to build Qt from source, which inevitably leads to a timeout (due to no terminal output for 10 minutes, but would likely exceed the total build time restriction otherwise).

Lestropie:
Appears to be fixed by explicitly calling `brew update` prior to Qt5 installation as per [this comment](https://github.com/Homebrew/brew/issues/5513#issuecomment-452772638).

PR #1540, opened 2019-01-29
NIfTI: default back to sform in case of mismatch

jdtournier:
See discussion at http://community.mrtrix.org/t/inconsistent-qform-sform-on-nifti-image/1663/9
Carries on from #1212 

We should also discuss further about the possibility of zeroing the qform on write if it doesn't match the sform (should only happen for non-rigid transforms). 

Also should add FAQ section in the docs about this.


Lestropie:
> We should also discuss further about the possibility of zeroing the qform on write if it doesn't match the sform (should only happen for non-rigid transforms).

Not sure it needs discussion; I think it's pretty clear that this is the best behaviour.

jdtournier:
OK, I reckon this one is good to go, if someone finds the time to have a look at it. Might need to double-check that I haven't borked anything up with all the changes to the NIfTI handling... Are there any known NIfTI validation sets that can be used for this...?

Lestropie:
> Are there any known NIfTI validation sets that can be used for this...?

Not that I'm already aware of. We could derive something but we'd probably need some sort of cross-checking with `fslhd` and `mri_info`.

jdtournier:
OK, this isn't as straightforward as I'd like... The handling of what happens to the voxel sizes when you have a non-orthonormal transform is inconsistent between NIfTI and mif. This is going require a bit more thought...

jdtournier:
OK, just to document the proposed solution, there's a few issues at play here...

Regarding NIfTI handling, there is a potential conflict between the voxel spacings specified in the `pixdim` NIfTI fields and the implicit scaling in the `sform`. This detects and warns about any discrepancies, and in the event of discrepancy, we use the `sform` version (commit 0be8f59b). It's also worth mentioning that we assume the spacing corresponds to the length of the cosine vectors (the first 3 columns of the `sform`), since these correspond to the distance from voxel centre to the adjacent voxel centre along the corresponding axis. 

There were also issues in our handling of the `mif` format: the transform is stored as-is, with no attempt at verifying that the cosine vectors are unit length (which is what the format implicitly assumes). So this PR extends the `sanitise_transform()` function (which is called on `Header` open/create) to check whether the cosine vectors are unit length, and if not, normalise them and update the voxel spacings accordingly (commit 8dc12a07). 

---

Just to verify that everything works as expected, here's a few tests. First, to check image writing, I transformed a T1 image `in.nii` using `mrtransform` with an affine transform `trans.txt` containing significant shear & scaling components:
```
2 0.5 0 0 
0 2 0 0 
0 0 2 0
0 0 0 1
```
using this command:
```
mrtransform -linear trans.txt in.nii <out>
```
where `<out>` was one of:
- `out_master.nii` : NIfTI produced using `mrtransform` built from `master` branch
- `out_master.mif` : equivalent `.mif` image
- `out_proposed.nii` / `out_proposed.mif` : as above using this PR

Then, to check image reading, I call `mrinfo` on all of these files, first using the version from the `master` branch, then using the version in the current PR (the version from `dev` produces identical results to `master`). I'm calling `mrinfo -spacing -transform` to focus on what we're looking at there. 

To summarise: 
- the outputs produced using the current `master` (or `dev`) branch are incorrect, but the information is actually still there if we read the data correctly. 
- The output produced using the proposed changes is correct and self-consistent, and can be read correctly using the `master` (or `dev`) branch. 
- The `master` (or `dev`) branch can't read its own data back correctly, and doesn't interpret the NIfTI and `mif` in the same way. 
- The proposed changes allow all the data produced using any version in either format to be read correctly and consistently (same, correct answer in all cases). 

---

### using `master` version of `mrinfo`:

- for outputs produced either `master` (or dev`): 
  - the voxel size in the `.mif` is the original voxel size, ignoring the scaling in the transformation. But the transformation does contain the scaling within it. 
  - the voxel in the NIfTI is also the original voxel size, but on top of that the transformation no longer contains the scaling...
- the outputs produced using the proposed changes are both correct and consistent with each other

**out_master.mif:**
```
1 1 1 nan
    0.504156686938913    -0.106954100835559    0.0034431868673037     -31.3473205566406
   -0.017965381670192     0.499677139761029 -3.31065931610214e-05     -48.0264434814453
 -0.00343388173681405 -9.03344968521086e-05     0.499988200156504     -88.2809066772461
                    0                     0                     0                     1
```
**out_master.nii:**
```
mrinfo: [WARNING] voxel spacings inconsistent between NIFTI s-form and header field pixdim
mrinfo: [WARNING] qform and sform are inconsistent in NIfTI image "out_master.nii" - using sform
1 1 1 nan
    0.999342543647121    -0.209305344527182   0.00688637298147955     -31.3473205566406
  -0.0356110911549963     0.977850316511019 -6.62131787833158e-05     -48.0264434814453
 -0.00680666171139882 -0.000176781384037188     0.999976286460322     -88.2809066772461
                    0                     0                     0                     1
```
**out_proposed.mif:**
```
0.50448800000000005 0.51099600000000001 0.5 nan
    0.999342543611934    -0.209305354983506    0.0068863729505582     -31.3473205566406
  -0.0356110921076096      0.97785031427288 -6.62131787833299e-05     -48.0264434814453
 -0.00680666189348019 -0.000176781383632564     0.999976286460534     -88.2809066772461
                    0                     0                     0                     1
```
**out_proposed.nii:**
```
0.50448834896087646 0.51099550724029541 0.50000005960464478 nan
    0.999342543647121    -0.209305344527182   0.00688637298147955     -31.3473205566406
  -0.0356110911549963     0.977850316511019 -6.62131787833158e-05     -48.0264434814453
 -0.00680666171139882 -0.000176781384037188     0.999976286460322     -88.2809066772461
                    0                     0                     0                     1
```

---

### using proposed version of `mrinfo`:

All data interpreted correctly and consistently: 

**out_master.mif:**
```
0.5044883660378694 0.51099553016208188 0.50000005692758676 nan
    0.999342543611934    -0.209305354983505   0.00688637295055821     -31.3473205566406
  -0.0356110921076095     0.977850314272881 -6.62131787833298e-05     -48.0264434814453
 -0.00680666189348019 -0.000176781383632564     0.999976286460534     -88.2809066772461
                    0                     0                     0                     1
```
**out_master.nii:**
```
mrinfo: [WARNING] voxel spacings inconsistent between NIFTI s-form and header field pixdim
mrinfo: [WARNING] qform and sform are inconsistent in NIfTI image "out_master.nii" - using sform
0.50448838781286764 0.51099554093762312 0.50000005505147915 nan
    0.999342543647121    -0.209305344527182   0.00688637298147955     -31.3473205566406
  -0.0356110911549963     0.977850316511019 -6.62131787833158e-05     -48.0264434814453
 -0.00680666171139882 -0.000176781384037188     0.999976286460322     -88.2809066772461
                    0                     0                     0                     1
```
**out_proposed.mif:**
```
0.50448800000000005 0.51099600000000001 0.5 nan
    0.999342543611934    -0.209305354983506    0.0068863729505582     -31.3473205566406
  -0.0356110921076096      0.97785031427288 -6.62131787833299e-05     -48.0264434814453
 -0.00680666189348019 -0.000176781383632564     0.999976286460534     -88.2809066772461
                    0                     0                     0                     1
```
**out_proposed.nii:**
```
0.50448834896087646 0.51099550724029541 0.50000005960464478 nan
    0.999342543647121    -0.209305344527182   0.00688637298147955     -31.3473205566406
  -0.0356110911549963     0.977850316511019 -6.62131787833158e-05     -48.0264434814453
 -0.00680666171139882 -0.000176781384037188     0.999976286460322     -88.2809066772461
                    0                     0                     0                     1
```

core/header.cpp
```diff
@@ -632,6 +632,23 @@ namespace MR
...
+    // check that cosine vectors are unit length (to some precision):
+    bool rescale_cosine_vectors = false;
+    for (size_t i = 0; i < 3; ++i) {
+      auto length = transform().matrix().col(i).head<3>().norm();
+      if (abs (length-1.0) > 1.0e-6)
+        rescale_cosine_vectors = true;
+    }
+
+    // if unit length, rescale and modify voxel sizes accordingly:
+    if (rescale_cosine_vectors) {
```
Lestropie, line 645: Might be worth having an `INFO` or `DEBUG` statement in here.

Lestropie:
Thanks for documenting, and for putting the effort into getting it all correct.

Are you confident about the new Analyse image handling? The code path has changed quite a bit but it's not covered in the CI tests.

core/header.cpp
jdtournier: Good call. Done with 3ea5dd9

jdtournier:
> Are you confident about the new Analyse image handling? The code path has changed quite a bit but it's not covered in the CI tests.

Pretty confident. The Analyze handling was all done using the NIfTI functionality anyway, if you [look at the code](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/formats/analyse.cpp)...

jdtournier:
All good - apart from:

> - Remove ".img" from NIfTI2 permissible suffix list.

No, that's still allowed, [according to the NIfTI-2 specs](https://nifti.nimh.nih.gov/nifti-2):

> - will support the existing file formats and naming: a .nii single-file, a .hdr/.img file-pair and their gzipped versions

I'll revert that bit...

jdtournier:
Just looking at the Analyze handling a bit more closely, I think there's a few more changes to make...

We're fine on read - [the code will detect whether the input in Analyze or NIfTI](https://github.com/MRtrix3/mrtrix3/blob/ba06ce1bfb27b8f5a8f927e872d54dddedbccb2c/core/file/nifti_utils.cpp#L92-L100) and act accordingly. 

On write, there is some ambiguity when creating a `.img` output as to whether it should be in Analyze or NIfTI format. So far, the code has produced Analyze output when a `.img` output is requested (that's what the boolean flag does [here for Analyze](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/formats/analyse.cpp#L60) and [here for NIfTI](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/formats/nifti1.cpp#L59)). You'll note the [current NIfTI output handler only accepts `.nii` suffixes](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/formats/nifti1.cpp#L52). 

Is it time to write NIfTI for `.img` outputs too...? With the code as it is in the branch, that's what it will do. Maybe we could set that up as a config option?

jdtournier:
Actually, I've just noticed it's worse than that... The code currently suggests that we'll write an Analyze output for `.nii`, and a NIfTI-1 for `.img`... I'll fix that now!

Lestropie:
> Is it time to write NIfTI for .img outputs too...?

Seems to be what's happening prior to these changes also; so calling it an "Analyze" handler is a bit misleading, it's more like a `.img` handler. 

> Maybe we could set that up as a config option?

Given the above, I don't think it'd be worth the effort to set up the capability to write genuine Analyze. If there were a desire for it we'd have heard about it by now.

jdtournier:
> Seems to be what's happening prior to these changes also; so calling it an "Analyze" handler is a bit misleading, it's more like a .img handler.

OK, there's a bit more to it than that. The NIfTI-1 standard is supposed to be backwards-compatible with Analyze, basically because most packages out there ignore half of the header anyway. So they added important information in the fields that were basically unused in practice. We produce a NIfTI-1 compliant output when writing Analyze, but what happens with Analyze is that we reset the strides to pure axial, and also potentially flip the left-right axis depending on the `Analyse.LeftToRight` config option. I think these are the only differences. 

But the point remains: should we treat `.img` as NIfTI, and forgo the modifications we've been making so far? Personally, I think there may still be selected applications out there that might rely on old-school Analyze data, so it might be good to at least provide the option to output more Analyze-like NIfTIs. 

Any objections to me adding a config file option for this...?

jdtournier:
OK, I've just added a config file option `IMGOutputsAnalyze` to enable more Analyze-friendly NIfTI output - if I can call it that. It's off by default, which means we'd produce straight NIfTI for both `.nii` and `.img` images. 

I've also removed the AnalyzeAVW description from the format string for NIfTI1 - it's always NIfTI compliant, the only thing is it should also accept old-school Analyze files at input. But I'm not sure it's worth reporting that..

jdtournier:
OK, I think I'm happy with this now. I'll merge this shortly unless there's any last-minute objections?

Lestropie:
Concern with the image format string was that if someone runs `mrinfo` on an Analyze image, but the format is reported as "NIfTI", they may think that there's some kind of bug making *MRtrix3* fail to recognise the format properly. Probably too hypothetical though, and Analyze images should be pretty bloody rare these days. 

If you're happy with your double-checking of the Analyze handling then go for it.

jdtournier:
OK, I think that last commit should finalise things. Basically:

- With this, we can _only_ write NIfTI output - not Analyze (but then we've not produced actual Analyze images for a very long time). The option to write Analyze-friendly output is gone. 
- To produce Analyze-friendly output, we can use `mrconvert -stride -1,2,3`. This matches the output on `master` - as long as the fix in #2014 is applied... :blush:
- The `AnalyzeLeftToRight` config option is honoured only on read for _actual_ Analyze images (not NIfTI). 
- I've tidied up the NIfTI API a touch as well while I was at it, but nothing major.
- I've also fixed up missing documentation and an incorrect specification for the `NIfTIAutoSaveJSON` config file option

PR #1541, opened 2019-01-30
PNG support & mrcolour command

### LINKED ISSUE #4: New binary: mrslice2image
In some cases it is useful to output a slice of an image volume to an image file at native resolution, rather than using the `mrview` screenshot tool. This can also simplify ensuring that the intensity scaling between different images is consistent.

Desired features:
- Either receive a pre-cropped image from mrconvert, or user provides -coord options to specify where they want the slice(s) to be
- If input is 4D and the 4th axis has dimension 3, output as RGB colour image
- Option to manually specify minimum / maximum image intensity (useful for outputting multiple images that should have a common colour bar, even if they don't all have the same dynamic range). Otherwise the intensity range is determined from the image data.
- Output as .tga; this won't require any code dependencies, and should be supported by more software than .pgm & .ppm
- Apply colour bar mapping to image on output
- Options to flip / swap axes (maybe a comma-separated list providing x,y axes for output image as +/- input axis indices)

Lestropie:
Even though it's not strictly necessary for RC4, I was not far from finishing it off, so we'll see if we can get it included, and I can hopefully get `libpng` out of my head to go focus on other requirements.

I regularly use a clumsy > 5-year-old command for exporting slice data directly to 2D image file in my own figure generation. The approach implemented here makes it far more convenient, supports [numbered multi-file](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#multi-file-numbered-image-support), and particularly when combined with `mrcolour` means that e.g. quantitative images generated during the course of an experiment can be exported with choice of applied colour map & intensity scaling directly to PNG file(s) at the native resolution of the underlying image; even if ssh'ing to computing hardware.

The included PNG CI tests are only simple, but I have tested read & write of [PngSuite](http://www.schaik.com/pngsuite/), and the only differences are the variable gamma images. Though obviously these test data are all only 2D (plus colour channels).

There are some peculiarities that may necessitate some documentation... I don't have a proper sense of how intuitive these might be for others to figure out. Independent testing welcomed.

Some more details in commit message of 5afcfc2.

Closes #4.

While this could be used to justify closing #8 based on its title, there may nevertheless still be merit in pursuing what's described in the description of that issue, i.e. implementing format & I/O handling to be able to interpret matrix data in text files as images.

Maybe one day this will also contribute toward some kind of image summary compilation QA akin to FSL's `slicesdir`, which has been raised on the forum a couple of times but doesn't yet have its own issue listing on GitHub.


Lestropie:
Quick demonstration of how this might be used for e.g. QA compilations:

```
mrconvert fa.mif -coord 2 15:15:45 Axial[].png &&  mrcat Axial* Axial.png -axis 0
mrconvert fa.mif -coord 1 17:17:51 -axes 0,2,1 Coronal[].png && mrcat Coronal* Coronal.png -axis 0
mrconvert fa.mif -coord 0 14:14:42 -axes 1,2,0 Sagittal[].png && mrcat Sagittal* Sagittal.png -axis 0
```

axial

coronal

sagittal

PR #1545, opened 2019-02-04
Update dwi2response dhollander

thijsdhollander:
Updates to `dwi2response dhollander` algorithm and user interface (_not_ the arguments/options, only what information appears on the terminal as feedback for the user).

@Lestropie : I've created this pull request against the `script_changes` branch for convenience.  It's entirely up to date with `script_changes` and thus uses the new API throughout.  I've also manually checked that those API changes don't break my intended behaviour (as I had finished most of these changes before the new API, and later on merged the API in), similar to how I believe you've manually checked outputs for other scripts before/after `script_changes`'s impact.

Don't merge yet; I still have to have a final check, and update some documentation.


thijsdhollander:
Alright, all checked and done.  I'll merge this into `script_changes` tomorrow morning (Tuesday morning AEDT).  Rob indicated to still have the `script_changes` PR open until next video call as well; just in case.

PR #1546, opened 2019-02-04
Docs: Combine binaries & scripts into single list

### LINKED ISSUE #1468: merge list of executable and scripts

Lestropie:
Closes #1468.

Will post link to resulting list page once RTD refreshes its list of GitHub branches so that I can activate it.


Lestropie:
[Preview](https://mrtrix3-dev.readthedocs.io/en/docs_combine_commands/reference/commands_list.html)

maxpietsch:
Is it on purpose that `dwishellmath` and `foreach` are not listed?

Lestropie:
`dwishellmath` has always been absent from the docs because it doesn't use the Python libraries. Same with `foreach`; but that will get included once this change gets resolved with #1449. Existing discussion in #692.

Issue #1184, opened 2017-10-31
Delimiter in matrix save


Lestropie:
Partly inspired by [this post](http://community.mrtrix.org/t/the-output-of-tck2connectome/345/16).

Functions `save_matrix()` and `save_vector()` could select the column delimiter based on the file extension of the output path. So for instance, if the output path is a `.csv`, use a comma rather than a space. Similar for `.tsv`. Thoughts?

maxpietsch:
The only problem I can see is that users might not expect the separator format to change if they forget to add the file extension. Also, probably due to the silly use of comma as decimal separator, the German version of Excel uses semicolons as delimiters for `.csv` files if I am not mistaken. So again, there might be a mismatch of expected separators.

Lestropie:
Part of the problem is that programs like Excel / Matlab set their expected delimiters for file import based on the file extension; so it wouldn't surprise me if space-separated text files and comma-separated .csv files import exactly as expected, whereas other combinations end up with import results like that shown in the forum post. So was hoping to address that, at least for most common use cases.

But I failed to think of localisation at the time; that's likely the bigger problem. Some recommend encasing all values within double-quotes, but it's kind of ugly. It's possible to query the [decimal point separator](http://www.cplusplus.com/reference/locale/numpunct/decimal_point/), but the locale-specific list separator doesn't appear to be defined anywhere. Could infer "," if "." is used for decimals, and ";" if "," is used for decimals; but it might be too much hidden magic dust...

Lestropie:
Closed by #1549.

PR #1549, opened 2019-02-04
Matrix save: Delimiter based on file extension

Lestropie:
Addresses #1184 if wanted.

Needs testing on locale with comma as decimal point separator, which I couldn't get working on my Windows system.


thijsdhollander:
> Needs testing on locale with comma as decimal point separator, which I couldn't get working on my Windows system.

@maxpietsch @bjeurissen @dchristiaens ... maybe?

Lestropie:
In retrospect, I don't think there's any need to check the locale of the output stream, as since #1323 we [force the C locale in `App::init()`](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/core/app.cpp#L1156-L1157).

Lestropie:
Converging Python matrix import code to behave equivalently to C++ wasn't quite as easy as I expected: `split()` doesn't accept multiple delimiters.

-----

@maxpietsch: The modified `save_transform()` & `load_transform()` that include a centre position are on reflection incongruent with #1603. See post-`dev`-merge code [here](https://github.com/MRtrix3/mrtrix3/blob/a4f6fffaedf314e2019a09ef7c2f5f2fd9672da1/core/math/math.h#L281-L305)

What I think would make more sense would be:

- For save:

    - Construct a `KeyValues` instance;

    - Insert into it an entry with key `centre` and value being the text string of the `Vector3`;

    - Pass that to the existing `save_transform()`.

- For load, have a wrapper function that:

    - Calls `load_transform()`;

    - Uses the existing `File::KeyValue::Reader` (which would require modification to support a comment symbol to strip from the start of each relevant line) to populate a `KeyValues` instance;

    - Validates the presence of the key "`centre`" & converts the value string to a `Vector3`;

    - Returns the transform and centre.

lib/mrtrix3/matrix.py
```diff
@@ -106,7 +107,10 @@ def decode(line):
         line = regex_comments.split(line, maxsplit=1)[0]
       line = line.strip()
       if line:
-        data.append([dtype(a) for a in line.split(delimiter) if a])
+        if len(delimiter) == 1:
+          data.append([dtype(a) for a in line.split(delimiter) if a])
+        else:
+          data.append([dtype(a) for a in [''.join(g) for k, g in itertools.groupby(line, lambda c : c in delimiter) if not k ]])
```
jdtournier: 🤯

core/math/math.h
```diff
@@ -19,13 +19,15 @@
 
 #include <cmath>
 #include <cstdlib>
+#include <locale>
```
jdtournier: Where is this required...?

lib/mrtrix3/matrix.py
Lestropie: Rectangle indeed.

core/math/math.h
Lestropie: Vestigial from previous attempts to detect whether or not it was safe to change the delimiter based on the decimal point symbol in use.

Issue #1042, opened 2017-07-07
build script: remove outdated executables if corresponding .cpp file no longer exists.


jdtournier:
Following from #1036: if a command is removed, but was built from a previous version, it will still be in the `bin/` folder, and presumably in the PATH, after a standard `git pull && ./build`, leading to potential issues with users not noticing important changes.

PR #1550, opened 2019-02-05
Build remove outdated bin

maxpietsch:
Addresses #1042: remove binary files in the target bin folder if no corresponding `.cpp` file can be found. The check whether a file is binary should be conservative but `check_encoding` makes it a rather expensive operation and could be relaxed if need be.

Removal is triggered for every build, irrespective of the build target.

### DIFF
--- a/build
+++ b/build
@@ -246,7 +246,32 @@ def get_real_name (path):
   return path
 
 
+def _get_expected_bin_filenames (exe_suffix, cmd_directory=cmd_dir):
+  binary_apps = set()
+  for entry in os.listdir(cmd_directory):
+    if entry.endswith(cpp_suffix):
+      binary_apps.add (entry[:-len(cpp_suffix)] + exe_suffix)
+  return binary_apps
 
+
+def list_expected_bin_files (exe_suffix, cmd_directory=cmd_dir, bin_directory=bin_dir):
+  bin_files = []
+  binary_apps = _get_expected_bin_filenames(exe_suffix, cmd_directory=cmd_directory)
+  for app in sorted(binary_apps):
+    if os.path.isfile(os.path.join(bin_directory, app)):
+      bin_files.append(os.path.join(bin_directory, app))
+  return bin_files
+
+
+def list_unexpected_bin_files (exe_suffix, cmd_directory=cmd_dir, bin_directory=bin_dir):
+  unexpected = []
+  if os.path.isdir(bin_directory):
+    binary_apps = _get_expected_bin_filenames(exe_suffix, cmd_directory=cmd_directory)
+    for filename in sorted(os.listdir(bin_directory)):
+      filepath = os.path.normpath(os.path.join(bin_directory, filename))
+      if os.path.isfile(filepath) and filename not in binary_apps and not is_likely_text(filepath):
+        unexpected.append(filepath)
+  return unexpected
 
 
 def list_untracked_bin_files (directory = '.'):
@@ -257,8 +282,8 @@ def list_untracked_bin_files (directory = '.'):
   return []
 
 
-def has_shebang(filename, nbytes=1024):
-  """try to detect if file has a shebang
+def is_likely_text(filename, nbytes=1024):
+  """ detects if file has a shebang or is utf-8 encoded, empty file is deemed non-text
   :arg filename
   :arg nbytes sample size to analyse
   """
@@ -270,15 +295,16 @@ def has_shebang(filename, nbytes=1024):
     log('could not read file ' + filename + '\n' + str(e) + '\n')
 
   if chunk:
-    line = chunk.split(b'\n', 1)[0]
-    try:
-      line = str(line.decode('utf-8'))
-    except UnicodeDecodeError:
-      return False
-    line = line.strip()
-    if len(line) > 2 and line[0:2] == '#!':
-      log('has_shebang: ' + filename + ' line: ' + line + '\n')
+    if chunk.startswith(b'\x23\x21'):  # starts with shebang '#!'
+      log('is_likely_text: ' + filename + ': starts with shebang \n')
       return True
+
+    try:
+      chunk.decode('utf-8')
+      log('is_likely_text: ' + filename + 'utf-8 encoded \n')
+      return True
+    except UnicodeDecodeError:
+      pass
   return False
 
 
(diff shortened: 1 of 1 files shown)


maxpietsch:
Feel free to merge if you're happy with it.

maxpietsch:
@jdtournier suggested checking for the presence of a shebang line in the file Instead of detecting whether a file is binary. This approach deletes everything that does not have a corresponding .cpp file, is not decodable as UTF-8, or does not start with `!#`.

Lestropie:
> @jdtournier suggested checking for the presence of a shebang line in the file Instead of detecting whether a file is binary. This approach deletes everything that does not have a corresponding .cpp file, is not decodable as UTF-8, or does not start with `!#`.

[Sounds familiar](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/lib/mrtrix3/run.py#L366-L412) :sweat_smile: :+1:  

Only risk is that if a bash script were added to `bin/` that didn't contain a shebang, it'd get yeeted; it would be pretty bad practise, but still... Would it be safer (given we're talking irreversible deletion of files) to use detection of a shebang as confirmation to definitely keep a file, but not explicitly require the presence of such to avoid deletion?

thijsdhollander:
> Would it be safer (given we're talking irreversible deletion of files) to use detection of a shebang as confirmation to definitely keep a file, but not explicitly require the presence of such to avoid deletion?

Yeah, I was thinking along the same lines as well.  I think I'd be more comfortable with that.

jdtournier:
> [Sounds familiar](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/lib/mrtrix3/run.py#L366-L412) 😅 👍

I knew that rang a bell... 🔔 

Otherwise, I'd tend to agree with @Lestropie & @thijsdhollander on the potential for scripts to get clobbered because the developer forgot to include the shebang line, and hadn't committed his changes yet. Note that my previous suggestion of relying on `git` would suffer from the same problem. 

As far as I can tell, the current approach would remove a script that's missing a shebang? Even if it was valid utf8? Maybe it would be better to:

- first check for shebang in first 2 characters, and keep if present
- check first 1024 characters for valid utf8, and keep if valid
- assume binary otherwise, check for corresponding `cpp` in `cmd/` folder, and keep if found
- otherwise remove file

Other idea might be to move 'removed' files to a dedicated folder (say `invalid_files` or something) and provide a clear warning that this has happened and where to find the file? This would guarantee no loss of data, and effectively take out any outdated binaries. If we go along with this idea, I reckon we don't even need the valid utf8 check - any scripts missing a shebang will need to be amended, but at least they won't disappear...

Also: we should use this approach in `./build clean` too - basically would need to replace the [list_untracked_bin_files() function](https://github.com/MRtrix3/mrtrix3/blob/master/build#L232-L238).

What do you reckon?

jdtournier:
> Other idea might be to move 'removed' files to a dedicated folder (say `invalid_files` or something)

On that note, maybe `purged_files` would be better...?

maxpietsch:
Agreed, moving instead of removing is a better solution. Any preference on `purged_files/`, `tmp/purged_files/` or `bin/purged_files/`? I think I prefer `bin/purged_files/` as that would be the place where I'd look for a missing binary / script.

> As far as I can tell, the current approach would remove a script that's missing a shebang? Even if it was valid utf8?

Yes.

> I reckon we don't even need the valid utf8 check - any scripts missing a shebang will need to be amended

This would require handling the byte order mark (see [here](https://en.wikipedia.org/wiki/Shebang_(Unix)#Magic_number)). Any reason not to decode as utf8? If so, it might be better to fall back to binary detection (8b5d8d0) which also does not have the problem of removing scripts / text files that might miss a shebang. 

> Also: we should use this approach in `./build clean` too

Not sure I understand, are you referring to moving objects instead of deleting or replacing `git ls-files` with detecting binary files via absence of shebang or utf8 encoding? The latter would change the behaviour as currenty `./build clean` also deletes scripts (valid shebang or not) that are not in the git tree.

jdtournier:
> This would require handling the byte order mark

I didn't know about that, good catch. However, we don't use it, and I've just tried inserting it at the beginning of the `build` script with this line:
```
$ sed -i '1s/^/\xef\xbb\xbf/' build
```
verified that it inserts the right character (0xEF 0xBB 0xBF) using `hexdump`:
```
$ hexdump -n 8 -x build
0000000    bbef    23bf    2f21    7375
0000008
```
looks like it's reading pairs of bytes back to front, but it checks out: `vim` and `less` open the file without problem, and don't show the BOM, and `vim` shows that it's interpreted the BOM as expected (`:set fileencodings` returns `ucs-bom,utf-8,default,latin1`). 

But `bash` won't run the file (at least not on my MSYS2 / Win10 laptop):
```
$ ./build
./build: line 1: ﻿#!/usr/bin/env: No such file or directory
./build: line 3: usage_string: command not found
./build: line 87: import: command not found
./build: line 90: syntax error near unexpected token `and'
./build: line 90: `if sys.executable[0].isalpha() and sys.executable[1] == ':':'
```

So to ensure compatibility at least with MSYS2, it looks like we _can't_ use the BOM. Besides, from that same wikipedia page: 

> Some authorities recommend against using the byte order mark in POSIX (Unix-like) scripts,[14] for this reason and for wider interoperability and philosophical concerns. Additionally, a byte order mark is not necessary in UTF-8, as that encoding does not have endianness issues; it serves only to identify the encoding as UTF-8.

--- 

Regarding the `./build clean` issue: currently, it does indeed rely on `git ls-files` and removes anything not under version control. This means there is also scope for data loss if `./build clean` is invoked with a non-versioned script in `bin/`. It also can't work on standalone, non-git setups. So I reckon we can simply use the same approach as here in the clean as well, and also move non-shebang files to the purged folder (whatever name we want to give it). 

What do you reckon?

maxpietsch:
This version now moves files out of `bin` into `bin/purged_files` if any of the following are true, checked in that order:
    - does not match an expected name (using .cpp files in `cmd_dir` and `exe_suffix`)
    - the file is empty
    - `b'\x23\x21'` is at the beginning of the file
    - the first 1024 bytes can be decoded as utf8

This binary file detection requires knowing the `exe_suffix` so I moved the config file parsing up before the `clean` part of the code.
Would you change `store_current_build` as well? Then we could get rid of `list_untracked_bin_files` and replace it with `list_unexpected_bin_files` but that requires moving config parsing before config handling.

Lestropie:
Open question: If one runs `./build clean`, and `bin/purged_files/` already exists, currently the script will move any *additional* new unknown files it finds in `bin/` into `bin/purged_files/`. Given the user would already have received a prior warning regarding the moving of unknown files into that location *before* they run `./build clean`, should the pre-existing contents of that directory be wiped at that point, and then an empty directory constructed if necessary?

Lestropie:
Bumping and adding "question" tag; I tested & was happy to merge save for the question in the prior [comment](https://github.com/MRtrix3/mrtrix3/pull/1550#issuecomment-516368404).

maxpietsch:
No strong preference but I'd suggest keeping the contents of `bin/purged_files/`, it feels unexpected that `./build clean` would be deleting files previously moved because they are not part of MRtrix3. Other than taking up disk space I see no harm in keeping them in that folder.

Issue #907, opened 2017-02-09
mrcat not appending dw_scheme


bjeurissen:
Was testing the 0.3.16 branch and one of my scripts fell over on concatenating two images.

The initial images were as follows:

```$ mrinfo ${tmp}/img1.mif ${tmp}/img2.mif
************************************************
Image:               "/tmp/tmp.nDfv3FugGJ/img1.mif"
************************************************
  Dimensions:        100 x 100 x 58 x 3
  Voxel size:        2.4 x 2.4 x 2.4 x ?
  Data strides:      [ 1 2 3 4 ]
  Format:            MRtrix
  Data type:         unsigned 16 bit integer (little endian)
  Intensity scaling: offset = 0, multiplier = 1
  Transform:                    1           0           0      -122.3
                                0           1           0      -67.68
                                0           0           1      -69.52
  DW_scheme:         0,0,0,0
                     0,0,0,0
                     0,0,0,0
************************************************
Image:               "/tmp/tmp.nDfv3FugGJ/img2.mif"
************************************************
  Dimensions:        100 x 100 x 58 x 3
  Voxel size:        2.4 x 2.4 x 2.4 x ?
  Data strides:      [ 1 2 3 4 ]
  Format:            MRtrix
  Data type:         unsigned 16 bit integer (little endian)
  Intensity scaling: offset = 0, multiplier = 1
  Transform:                    1           0           0      -122.3
                                0           1           0      -67.68
                                0           0           1      -69.52
  DW_scheme:         0,0,0,0
                     0,0,0,0
                     0,0,0,0
```

And the resulting concatenate failed to append both dw_schemes:

```
$ mrcat -axis 3 ${tmp}/img1.mif ${tmp}/img2.mif - | mrinfo -
mrcat: [100%] concatenating "/tmp/tmp.nDfv3FugGJ/img1.mif"
mrcat: [100%] concatenating "/tmp/tmp.nDfv3FugGJ/img2.mif"
************************************************
Image:               "/tmp/mrtrix-tmp-QOgfFD.mif"
************************************************
  Dimensions:        100 x 100 x 58 x 6
  Voxel size:        2.4 x 2.4 x 2.4 x ?
  Data strides:      [ 1 2 3 4 ]
  Format:            Internal pipe
  Data type:         unsigned 16 bit integer (little endian)
  Intensity scaling: offset = 0, multiplier = 1
  Transform:                    1           0           0      -122.3
                                0           1           0      -67.68
[...]

PR #1555, opened 2019-02-07
4D schemes preserve on [] notation

### LINKED ISSUE #1150: mrconvert with [] in output does not split gradients
```
mrconvert dwi.mif volume_[].mif
```
results in ``volume_[].mif`` files containing the full gradient table, instead of just a single row of the gradient table.

This can lead to further problems down the road, e.g.:
```
mrcat -axis 3 volume_*.mif all_volumes.mif
```
will now contain a gradient table with n^2 rows, where the original gradient table had n rows.

Lestropie:
Closes #1150.

Seems to work for `dw_scheme`; will test with phase encoding tables in a moment.

Some details that may need to be discussed:

- Had to move `src/dwi/gradient.*`  to `core/dwi/`; otherwise the shared library wouldn't compile. Not sure if this is something you want to address separately, but given how deep in the back-end the handling of the DW scheme has to be it probably makes sense.

- On split using `[]` notation, the full DW scheme will be copied to `prior_dw_scheme` as per #774.

- Comments are also handled slightly differently: On loading using `[]`, contents of "`comments`" will be the concatenation of *unique* entries.

- Where key-value contents vary between two images loaded using `[]`, I instead write the value "`variable`", which is something I also do in `tckedit` when merging multiple track files. Doing a concatenation may not make sense for arbitrary key-value pairs.

- Commands such as `mrmath` & `mrcalc` that may operate on multiple input images (and no one of those images can reasonably claim "precedence" as the "primary" input image on which the output image key-values should be based) could possibly do with a similar operation. It's not a *unique* operation though: what's being augmented in this PR is strictly a *concatenation* (and the function has been renamed as such), whereas for `mrmath` / `mrcalc` it's more like a "merge" (which was the old name of this function). But the latter could reasonably be added as a `Header` member function.


Lestropie:
#907 is also strongly related. The case of `mrcat` concatenating specifically across volumes should behave similarly to `[]` notation along the volume axis; but for other axes it should behave more like a "merge" rather than a "concatenate".

Lestropie:
OK, that should give consistent behaviour in header key-value management (not only the 4D schemes) across `[]` notation, `mrcalc`, `mrcat` and `mrmath`. Even erased a bunch of code that'd been put in those commands at various times to deal with 4D scheme data.

PR #1556, opened 2019-02-07
Prevent duplicated ROI entries in track files

### LINKED ISSUE #1368: Include and exclude ROIs are duplicated in tck header
While preparing the tractography tutorial, I noticed that ``tckinfo`` reports all include and exclude ROIs twice. This is not the case for the seed ROI.

The issue is not with ``tckinfo`` itself as the raw header also has every include and exclude ROI duplicated.

The issue happens regardless of whether the ROI is an image or a sphere.

```
$ tckgen ms_wm.mif -seed_image mask.mif -include mask.mif -exclude -30,20,20,20 test.tck
tckgen: [100%]    10589 seeds,     7142 streamlines,     5000 selected
$ tckinfo test.tck 
***********************************
  Tracks file: "test.tck"
    count:                5000
    downsample_factor:    3
    fod_power:            0.25
    init_threshold:       0.0500000007
    lmax:                 8
    max_angle:            45
    max_dist:             200
    max_num_seeds:        5000000
    max_num_tracks:       5000
    max_seed_attempts:    1000
    max_trials:           1000
    method:               iFOD2
    min_dist:             10
    mrtrix_version:       3.0_RC3-27-ga1b87312
    output_step_size:     1
    rk4:                  0
    samples_per_step:     4
    sh_precomputed:       1
    source:               ms_wm.mif
    step_size:            1
    stop_on_all_include:  0
    threshold:            0.0500000007
    timestamp:            1528710640.7401590347
    total_count:          7142
    unidirectional:       0
    ROI:                  exclude -30,20,20,20
    ROI:                  exclude -30,20,20,20
    ROI:                  include mask.mif
    ROI:                  include mask.mif
    ROI:                  seed mask.mif
```

Lestropie:
While the fix for #1368 was trivial, there was some slightly confusing code lying around, which I decided to somewhat address while I was in there.

- The `Properties` class contains `ROISet` instances for include / exclude / mask, but also contains a `std::multimap<std::string, std::string>` called "`roi`". I've renamed this to "`prior_rois`", to hopefully highlight the fact that the former classes are the ROIs that are in use during code execution, whereas the latter are classification & specifiers that have been used prior & now been read from the header of an *existing* track file that those ROIs were involved in the creation of.

- Member function `Properties::load_ROIs()` was not called from anywhere; this is old code from back before the creation of `tckedit`, when the `tckgen` code would put ROIs from the command-line into the "`roi`" class and *then* load them (also contributing to the confusion in the first point). This is now handled explicitly in function `Tractography::load_rois()`, which *doesn't* go through the `multimap`. This is necessary in the design because `tckedit` may read ROIs from the header of an input track file, but not actually want to load or utilise those ROIs.

Closes #1368.


Lestropie:
Yeah, I saw those too; just forgot to go back to them :-/

Issue #263, opened 2015-05-31
Prevent deep multi-threading


Lestropie:
Many low-level operations are now multi-threaded. However if some high-level functionality is multi-threaded, and itself calls multi-threaded low-level operations, one could end up with e.g. 64 threads running instead of 8, which could cause slowdown.

It may be better to set an atomic flag to indicate that a multi-threaded process is running; and if another function requests number_of_threads(), it should return 0 and instead run in a strictly single-threaded manner, until the flag is released.

Lestropie:
In a similar vein to this:

Some functors contain a progress bar, yet may be called within a loop which itself has a progress bar. A static atomic flag could be added to the progress bar by which a particular instantiation 'claims' the terminal output, and other progress bars do not appear.

Lestropie:
On reflection, these could be toggled via [`static MR::Thread::__Backend backend`](https://ark.intel.com/products/64583/Intel-Xeon-Processor-E5-2680-20M-Cache-2_70-GHz-8_00-GTs-Intel-QPI).

jdtournier:
I assume that link should have been [this one](https://github.com/MRtrix3/mrtrix3/blob/dev/core/thread.h#L89)...?

And yes, I agree this might be a good way to handle things. We could add a static method in there to query [the number of registered threads](https://github.com/MRtrix3/mrtrix3/blob/dev/core/thread.h#L87), and maybe compare that to the[ number that it's supposed to use](https://github.com/MRtrix3/mrtrix3/blob/dev/core/thread.h#L243)?

Lestropie:
:rofl: Clipboard fail...

I probably didn't think about this enough: One thread within a multi-threading queue still needs to obtain terminal "privileges" for constructing a progress bar. So this mechanism in particular would still need its own dedicated atomic flag: Couldn't rely on one thread "claiming" the terminal before any other threads are even spawned. This should be handled by creating a ProgressBar within the constructor of any multi-threading class. But for preventing execution of nested multi-threading, simply testing if `backend` is non-zero should suffice (as far as I can tell).

I kept bouncing back to this while thinking about #984. But that's fundamentally a destructor throwing issue, not a multi-threading issue; keep seeing similarities where there are none... :-/

Lestropie:
[...]

PR #1563, opened 2019-02-12
progressbar: disable line wrapping while updating terminal

jdtournier:
Prompted by discussion in #263...

This means the terminal doesn't get clogged up with the same line repeated over and over again when the progressbar line width exceeds the terminal's width...


thijsdhollander:
Probably won't have time soon to check this; added all devs to review for better chances of success.

jdtournier:
Merging this since it's a trivial change. We can revert if it proves problematic.

PR #1583, opened 2019-03-19
mrtransform: default to input transform

jdtournier:
Addresses this issue:
http://community.mrtrix.org/t/mrtransform-changes-strides-but-also-qform/2353


jdtournier:
Actually, I don't think this is quite right... I'll can this and try again.

Issue #1578, opened 2019-03-12
mrview: Invisible toolbar on load


Lestropie:
If, when initially running `mrview`, the input image is sufficiently large that a progress bar appears, then the `mrview` toolbar does not get drawn initially; it appears as a blank white strip. Everything then appears once the mouse is clicked anywhere on the toolbar.

Observed on a Windows machine; don't know whether or not it is specific to such.

jdtournier:
Not sure whether this is still an issue, but for a while mrview was displayed without a toolbar at all using Qt5 on my Windows 10 laptop. This was only resolved recently with the Qt update to version 5.12. Do you reckon this might be fixed with an update...?

Lestropie:
I'm pretty sure that my personal laptop (where I observed this) is on Qt4 due to #594. But I'll endeavour to test with different versions.

Issue #378, opened 2015-10-09
mrview -overlay.load causes qt: recursive repaint warning


maxpietsch:
Starting mrview with -overlay.load causes the warning 
`QWidget::repaint: Recursive repaint detected`
in the terminal. This is probably a thread related issue. No warning is produced if the overlay is loaded from the GUI.

Lestropie:
Couldn't reproduce on either branch on my laptop; someone else will have to have a go.

jdtournier:
This was on Ubuntu 14.04, Qt 5.2.1 (I think). I'm pretty sure the issue only occurs when the overlay image is large enough that it triggers the progressbar to display (i.e. it takes more than a second to load).

maxpietsch:
All correct. I get no message if I down-sample the image and also no message if I convert the image to a compressed nifti. Here is the file (dimensions: 561 x 647 x 481): 
<https://drive.google.com/file/d/0B4LjOWoWoZPZTW04eG5KWDN0X0U/view>

On Os X (Yosemite 10.10.5 mac book air, QT  5.4.1) mrview crashes with the error message "Abort trap 6" for both uncompressed and compressed but not for down-sampled overlay. That might be a different issue though (384MB graphics card RAM).

jdtournier:
@maxpietsch, when you have a minute, can you confirm whether this is still an issue...?

maxpietsch:
I still get the same error message on ubuntu. When mrview is run with `-debug` option the menu and toolbar are not drawn correctly until I click on them. 

![2016-05-26t08_12_05 288598 0100_4xiewb](https://cloud.githubusercontent.com/assets/10046944/15566567/a3a9189e-2319-11e6-88e9-d915498f4a94.png)


Here is the start of the debug output:

`mrview: [INFO] GL renderer:  GeForce GTX 760/PCIe/SSE2
mrview: [INFO] GL version:   3.3.0 NVIDIA 352.63
mrview: [INFO] GL vendor:    NVIDIA Corporation
mrview: [DEBUG] loading font into OpenGL texture...
mrview: [DEBUG] compiling OpenGL vertex shader:
#version 330 core
layout(location = 0) in vec2 pos;
layout(location = 1) in vec2 font_pos;
uniform float scale_x;
uniform float scale_y;
out vec2 tex_coord;
void main () {
  gl_Position = vec4 (pos[0]*scale_x-1.0, pos[1]*scale_y-1.0, 0.0, 1.0);
[...]

PR #1586, opened 2019-03-22
Progressbar updates in main thread

### LINKED ISSUE #256: mrview: Creating ProgressBar from another thread
Whenever I create a ProgressBar within my connectome toolbar, I get the following in the terminal:

```
QObject::startTimer: Timers can only be used with threads started with QThread
QObject::killTimer: Timers cannot be stopped from another thread
```

Not critical, but kind of annoying; sometimes if a user selects some option that's going to take some time to load, I want to show a progressbar so they know the program hasn't hung.

jdtournier:
Took quite a bit of experimentation, but I reckon this may finally fix #256 and #378 (and probably others besides). Commit a87abc3 is the relevant one here. I'll need all MRView users who've reported issues when the progressbar is shown to give this a spin and see whether the issue persists...


Lestropie:
Does *not* fix #1578.

Lestropie:
Extracted some unrelated code and moved to #1663.

The actual technique used here, I'd need to spend a fair bit of time on to wrap my head around.

I suppose that as long as it does not *regress* for GUI / non-GUI on any of the three OS's, it should be safe to merge, even if it fails to exhaustively solve all issues?

It also looks like it may only fix the issue if `ThreadedLoop` is utilised; so for any GUI tool that uses custom multi-threading code with a progress bar, `ProgressBar::run_update_thread()` won't be utilised?

jdtournier:
Yes, as far as I can tell, there's no regression. And yes, it would only solve issues when using our own ThreadedLoop class - hopefully there won't be any other instances of this going on within MRView... 

Most of the changes are actually cosmetic, even though the diff looks bigger than it needs to. As mentioned earlier, the bulk of the logic is contained within a87abc3: basically set up condition variable that can be notified in the main thread, whose job is then to update the progressbar or exit if all worker threads have actually completed.

PR #1589, opened 2019-04-02
Tractography::Writer: Preserve order

### LINKED ISSUE #502: Tractography::Writer: Preserve order

### LINKED ISSUE #1576: tcknormalise (tcktransform?) randomly changes order of streamlines when multi-threading
When using multi-threading to warp a .tck file with tcknormalise, the order of the streamlines is not preserved. On the contrary, when using single-thread, the order is preserved. In some cases this may be a problem, e.g. when dealing with labelled streamlines such as during segmentation. As evidence of this behavior, I attach two plots showing the distribution of the distance of corresponding streamlines before and after the warp with (i) single thread and (ii) multi-thread. With single thread, the mean is plausibly centered around 4 mm, instead with multi-threading it is centered around 60 mm, meaning there is no more correspondence between streamlines after the transformation. 

hist_single_thread

hist_multi_threads

This problem may affect the new tcktransform too.

Could you please fix the issue?

Lestropie:
Closes #502.

Closes #1576.

Listing as draft PR in order to show progress.

However something that chipped away at me during the course of this is the magnitude of overlap between the classes used for streamline vertex read/write vs. track scalar read/write. While they do currently derive from common base classes, there's a lot of functionality common between them that is instead duplicated between the two specialisations; and that fraction of overlap is accentuated by the changes made here. So I'm quietly wondering if more functionality should be shifted to the base classes, with the separation in functionality between 1 / 3 values per vertex handled via eg. template specialisations.

Also have not yet decided whether to shift the per-streamline weights to set/get functions rather than a public member variable, as has happened here for the streamline index.

FYI The way I've disambiguated between an empty streamline (used to indicate that the streamline shouldn't be written) and a streamline with no vertices that should nevertheless be written (e.g. `tcktransform` but the entire streamline is outside the FoV) is, by assigning to any streamline that should not be written to file, a *non-finite weight* (even if SIFT2 weights are not provided / derived, the weight is nevertheless an embedded member variable of the Streamline class).

### DIFF
--- a/cmd/connectome2tck.cpp
+++ b/cmd/connectome2tck.cpp
@@ -279,11 +279,11 @@ void run ()
       std::mutex mutex;
       ProgressBar progress ("generating exemplars for connectome", count);
       if (assignments_pairs.size()) {
-        auto loader = [&] (Tractography::Connectome::Streamline_nodepair& out) { if (!reader (out)) return false; out.set_nodes (assignments_pairs[out.index]); return true; };
+        auto loader = [&] (Tractography::Connectome::Streamline_nodepair& out) { if (!reader (out)) return false; out.set_nodes (assignments_pairs[out.get_index()]); return true; };
         auto worker = [&] (const Tractography::Connectome::Streamline_nodepair& in) { generator (in); std::lock_guard<std::mutex> lock (mutex); ++progress; return true; };
         Thread::run_queue (loader, Thread::batch (Tractography::Connectome::Streamline_nodepair()), Thread::multi (worker));
       } else {
-        auto loader = [&] (Tractography::Connectome::Streamline_nodelist& out) { if (!reader (out)) return false; out.set_nodes (assignments_lists[out.index]); return true; };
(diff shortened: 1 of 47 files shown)


Lestropie:
Closed in favour of #1939 and #1988.

PR #1590, opened 2019-04-03
Update python signal handling

### LINKED ISSUE #1457: Setting preexec_fn to call os.setpgrp() breaks scripts calling matlab
On linux, setting `preexec_fn` to call `os.setpgrp()` in `run.py` caused all my scripts that call matlab to break.

Calling matlab directly using `subprocess` (without `preexec_fn`) would fix the scripts themselves, but when calling those scripts from yet another script using `run.command` would fail again.

Removing the `preexec_fn` option ensured that all scripts directly or indirectly calling matlab, work fine.

Calling matlab with `-singleCompThread` and `-nojvm` also did not help.

Lestropie:
Attempting to resolve the issue of signal handling in the Python libraries ahead of merging #1449.

The approach here is driven by [this page](https://stefan.sofa-rockers.org/2013/08/15/handling-sub-process-hierarchies-python-linux-os-x/) which describes some of the issues. I've erased a lot of relevant code and started from scratch, and things seem to be working from what I've tried, and shouldn't have the race conditions that the old approach had. It also doesn't involve use of `os.setpgrp()`, which should mean it fixes #1457 (nevertheless @bjeurissen it's worth testing calling Matlab functions).

One change here is that I switched `foreach` to use the `multiprocessing` module rather than `threading`. This is because when using `foreach` the individual jobs may be quite long, but there's no robust way in Python to terminate a thread; even more so when it's not the Python interpreter carrying the processing load. So instead, each working "thread" in `foreach` now gets its own *process*, which can then be terminated on receiving a signal, e.g. propagation of SIGINT. One little downside here is that if you Ctrl-C `foreach` when it's running multi-threaded, you get duplicates of the signal handler message, as the custom signal handler is invoked separately for each process. It also renders my gymnastics in making `run.command()` thread-safe somewhat redundant... but at least now with the new `foreach` you can kill the script with a single Ctrl-C regardless of whether or not it's running multi-"threaded". Tagging @jdtournier in case I've overlooked an alternative solution here.

Ideally I'd like to test this in a genuine Windows Python environment before merging; think I had this set up at home previously.


Lestropie:
In native Windows, the file handlers for stdout / stderr of subprocesses remain open and prevent os.unlink() from erasing the temporary files. It will be necessary to keep track of the file IDs in addition to the file names while subprocesses are being run so that they can later be explicitly closed prior to file deletion.

Lestropie:
Okay, 0eba2357 seems to solve the Windows issue. Still awaiting testing with calling MatLab functions and/or more general comments.

Lestropie:
> please check my fix in 5ef5b84 is correct.

I swear I fixed that :confounded:

jdtournier:
OK, given that all remaining test failures are due to pre-existing issues in the `script_changes` branch, I'd say we're good to merge...

PR #1593, opened 2019-04-12
Improve build versioning when git not present

jdtournier:
This allows version information to be added to the repo itself, and it'll be used in the event that `git` isn't present or the code was downloaded as an archive - addresses #1514. The version string must be set to the corresponding tag. When `git` can be used, the `build` script will check that the version information stored in the repo matches the version string produced by git (`git describe --abbrev=0`), and fail if it doesn't match. This means when a new version is to be released, the version string will need to be updated and then that commit tagged with the same version string before pushing with tags. 

This will need a bit of finessing when merging to `master`. Mostly likely strategy will be to force a fast-forward merge of a manual non-fast-forward merge - if that makes sense to anyone...

### DIFF
--- a/build
+++ b/build
@@ -1042,65 +1042,116 @@ def build_next ():
 
 
 
-
-def get_git_version (folder):
+def version_from_git (folder):
   try:
-    return get_git_version.version
+    log ('fetching version information from git in folder "' + folder + '": ')
+    with open (os.devnull) as devnull:
+      tag = subprocess.check_output ([ 'git', 'describe', '--abbrev=0' ], cwd=folder, stderr=devnull).decode(errors='ignore').strip()
+      commit = subprocess.check_output ([ 'git', 'describe', '--abbrev=8', '--dirty', '--always' ], cwd=folder, stderr=devnull).decode(errors='ignore').strip()
+    log (tag + ', ' + commit + '\n')
+    return [ tag, commit ]
+  except (OSError, subprocess.CalledProcessError):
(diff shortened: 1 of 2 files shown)


build
```diff
@@ -1042,65 +1042,116 @@ def build_next ():
...
-          hard_coded_version = line.split('MRTRIX_BASE_VERSION')[1].strip().strip('"')
-          log (hard_coded_version + '\n')
-    if not len (hard_coded_version):
-      raise NameError
-  except ( OSError, NameError ):
-    fail ("ERROR: failed to read version information\n")
+    log ('  ')
+    version = version_from_git (mrtrix_dir[-1])
+    if version[0] != tag:
+        fail ('ERROR: version stored in "' + os.path.join (lib_dir, 'version.h') + '" does not match git tag name\n')
```
maxpietsch: This trips up run_tests when updating from an older version as testing/src/project_version.h does not seem to be cleared before running the tests. The version string in that file was 3.0_RC2-931-g3688ac7a. Is this a one-off?

jdtournier: Yes, that file is not under version control, and used to be generated by the `build` script. This new version will no longer update that file, but if present, check that it matches the nearest tag (I think...). So it should just be a matter of deleting that file...

thijsdhollander: > This trips up run_tests when updating from an older version as testing/src/project_version.h does not seem to be cleared before running the tests.

Yep, I ran into the same issue yesterday.

PR #1594, opened 2019-04-12
holding branch for changes to be pushed to dev

jdtournier:
This is to avoid the need to wait for CI to complete before updating other pull requests. I'm using the `dev_pending` branch to merge all the other pull requests into, and getting the CI to run on this one single pull request...


thijsdhollander:
Ok, all good, but I suggest we merge this then whenever such a batch of pull requests ends up in here and the tests are successful.  Things should only end up here if their own pull request is finalised, including discussions, etc...

Merging this to `dev` as frequently as possible (i.e. completed tests) helps for things to get the proper further "`dev`-testing" that the `dev` branch is meant for; otherwise this risks to become "yet another `dev` layer"...

So I've taken the initiative to merge this batch: all pull requests were approved, and the tests ended up passing now too.

thijsdhollander:
...in practice: I think once new commits arrive in this branch that this very pull request can be reused (maybe after consciously re-opening it).  But I might be wrong.  If the latter, it only takes a new pull request to be opened for the same branch of course.

jdtournier:
Yep, that's the idea. I don't find the time to review and merge pull requests very often, and the CI testing makes it very difficult to process multiple requests in quick succession. So I'm thinking we could protect the `dev_pending` branch to force code reviews but no CI, allowing quick merges, and the CI could run for the combined merge to `dev` in its own time. It'll typically only raise glitches that can be fixed up without affecting the validity of the code reviews, so I don't think it's a problem. But having the ability to decouple the reviewing and merge process from the CI is hopefully going to help speed things up...

thijsdhollander:
Agreed with everything except

> So I'm thinking we could protect the dev_pending branch to force code reviews 

since we consciously didn't force code reviews for merge to `dev` before either.  That'll slow down the process in other ways.  Only `master` should be "protected" with code reviews.  Anything that lands on `dev` can always still be reviewed ("just in case") before it all lands on `master`.

It'd be weird to protect `dev_pending`, since it's already 2 whole steps away from `master`.

jdtournier:
Sorry, I meant to request code reviews for merging, but allow merging anyway as currently done on `dev` - no problem there.

Also, if merging to `dev_pending`, I'd rather the code reviews were done before the merge to `dev_pending` branch since the changes would be smaller in scope at that point. By the time that branch might be considered for merging to `dev`, it might already have accumulated a number of pull requests, which makes reviewing a lot more challenging - and reduces the chances of someone finding the time and motivation to look at it meaningfully. 

In any case, I just thought this strategy might provide a way to quickly approve a few pull requests and merge them all together in a batch for a single round of CI testing. Once that branch passes its CI, it should always be considered good to merge, I think. 

But I don't think that should change the current workflow. I still expect most of us to create our pull request with `dev` as the destination... If one of us wants to merge a few pull requests quickly, then this option is available - but either way, pull requests should be reviewed and approved before merging to `dev` or `dev_pending`.

thijsdhollander:
> Sorry, I meant to request code reviews for merging, but allow merging anyway as currently done on dev - no problem there.

Ok, yep, I understand now; looks like we're on the same page. :+1:

Issue #1485, opened 2018-11-12
mrtransform: disable implicit FOD reorientation


bjeurissen:
`mrtransform` currently performs FOD reorientation by default if the number of volumes indicates that the series can be interpreted as a series of SH coefficients. This is quite a dangerous practice as it could silently mess up pipelines, e.g. if input data happens to have a number of gradient orientations that corresponds to a number that can be interpreted as a series of SH coefficients. So I feel it would be a safer option to disable FOD reorientation by default.

Lestropie:
Just to add to this: I suspect FOD reorientation was turned on by default at the time because this was perceived at the time to be the most common usage. However with the current FBA pipeline reorientation actually has to be turned off explicitly at the transformation stage, as this is instead deferred to `fixelreorient`. So having this off by default and instead activated explicitly within `population_template` would actually be simpler usage for users.

maxpietsch:
Agreed, it makes sense to change this behaviour. I am just concerned that changing this might silently trip off scripts that implicit relied on FOD reorientation. This is a source of error that is hard to detect. The safest approach would be to rais an error if the number of volumes indicates an SH file and the orientation option is not provided.

thijsdhollander:
It let this idea sink a bit in: initially, I was feeling quite opposed to it, because of the initial problems and confusion this could again unleash upon users of the FBA pipeline.  The biggest problems (every time) being that, once this release would go out, the online docs get updated, but that doesn't mean users update their install; yet they still use the (new) documentation they find online.
So in this case, that would've meant: in the online docs, the `-noreorient` option disappears, causing copy-pasting users to not specify it any more.  But if they're still using a previous version, that would mean they do end up reorienting.  ...and then another time at the fixel stage.  So 2 reorientations, which is obviously wrong.

[...]

Issue #1484, opened 2018-11-12
mrtransform: modulate non-FOD data sets


bjeurissen:
At the moment, the `-modulate` option can only be used for FODs. Performing modulation of scalar images (e.g. for VBA) requires users to get the determinants of the Jacobians of the deformation field and then applying those in the correct way to the image intensities, leaving a lot of room for error. It would be very useful just to support the `-modulate` option for both FODs and (series of) scalar images. This would require an explicit option to signal that the dataset should be interpreted as FODs in addition to a generic `-modulate` option.

bjeurissen:
Any suggestions on how to approach this? Currently the modulation seems to be completely tied to the reorientation. Any ideas @maxpietsch?

maxpietsch:
Sorry, I am not familiar with the modulation code. Interface-wise: Modulation means different things depending on whether one refers to fibre ODFs (scale so that fibre density is constant across fibre bundle) or scalar images (scale with local volume change). I think the two are equivalent for spherical ODFs but I am not sure if it would be a good idea to choose which to perform, depending on the `-reorient` option. We could make it explicit via `-modulate fc` and `-modulate jac` though?

Lestropie:
> I think the two are equivalent for spherical ODFs ...

I'd have thought it *wouldn't* be: E.g. If you have compression in all 3 spatial dimensions, a scalar image modulated by the full Jacobian is going to incorporate all 3 factors, whereas a spherical ODF using "FOD modulation" would effectively "regress out" the expansion along the direction of the aPSF independently for each, resulting in the latter version not being scaled to the same magnitude as the former. So the more explicit version may be necessary.

jdtournier:
> If you have compression in all 3 spatial dimensions, a scalar image modulated by the full Jacobian is going to incorporate all 3 factors, whereas a spherical ODF using "FOD modulation" would effectively "regress out" the expansion along the direction of the aPSF independently for each, resulting in the latter version not being scaled to the same magnitude as the former. 

I'm not sure, I have a feeling that ought to give (approximately) the right answer...?

But more to the point, if it's using the full aPSF backend to do the modulation on a scalar image, rather than simply modulating by the determinant of the Jacobian (I hope I've got that right...?), that alone is reason enough to change it...

maxpietsch:
[...]

PR #1598, opened 2019-04-17
mrtransform Jacobian modulation and explicit reorientation

maxpietsch:
This addresses #1484 and #1485:

- added option for modulation by the determinant of the Jacobian of the warp of by the determinant of the linear transformation.
- makes FOD reorientation a mandatory option if the number of volumes indicates a SH image with lmax >= 2

The new option-logic follows what we discussed a while back [here](https://github.com/MRtrix3/mrtrix3/issues/1485#issuecomment-460703161):
```
# sh.mif: lmax>=2, scalar.mif: number of volumes not in [6, 15, 28, 45, ...]
mrtransform sh.mif -reorient_fod bool ...  #  reorientation explicit
mrtransform sh.mif  ...  #  error: missing option -reorient_fod
mrtransform scalar.mif ...  #  fine
mrtransform scalar.mif -reorient_fod true ...  #  error

mrtransform sh.mif -reorient_fod true -modulate fod ... #  reorientation and aPSF modulation
mrtransform sh.mif -reorient_fod true -modulate jac ... #  reorientation and jacobian modulation, warning
mrtransform sh.mif -reorient_fod false -modulate jac ... #  jacobian modulation, warning
```

I implemented Jac modulation in the warp adapter (and filter) as this makes it independent of the dimensionality of the image and required the least changes. Alternatively, one could implement Jacobian modulation as a second step after warping which could be faster for images with more than 3 dimensions.

Feel free to break it!


maxpietsch:
Is everyone happy to merge this PR? I would like to include changes in the workshop if possible.

bjeurissen:
Should

```
mrtransform sh.mif -reorient_fod false -modulate jac
```

not just perform jacobian modulation of the raw intensities _without warning_? 

And do

```
mrtransform volumes.mif -modulate jac
```

and 

```
mrtransform scalar.mif -modulate jac
```
also work?

maxpietsch:
Yes, Jacobian modulation works with any number of volumes.

Technically it's perfectly fine to perform jacobian modulation on SH images but I'd warn users that they might want to use FOD modulation instead to prevent the use of Jacobian modulation in the FBA pipeline. The warning is:

```
mrtransform sh.mif -warp warp.mif -modulate jac warped_j.mif -reorient_fod yes
mrtransform: [WARNING] Jacobian modulation performed with FOD reorientation. Did you mean FOD modulation?
```

Lestropie:
Not crucial for merge, but something I've thought about in the context of other commands: it should be possible to add to the list of tests verification that the command in fact fails appropriately given a certain combination of input data & command-line options.

-----

> mrtransform: [WARNING] Jacobian modulation performed with FOD reorientation. Did you mean FOD modulation?

Presumably part of the design of this wording is that if the number of volumes is appropriate for an SH image, but FOD reorientation is explicitly disabled, then Jacobian modulation is entirely reasonable (it might just be that the number of volumes happens to correspond to an SH series by chance); whereas if the input data are to be interpreted as FOD data (as inferred from the request for FOD reorientation), then Jacobian modulation doesn't make a lot of sense. But I wonder if the current message is a bit too brief for users to disentangle such.

Maybe something like:

"Input image being interpreted as FOD data (user requested FOD reorientation); FOD-based modulation would be more appropriate for such data than the requested Jacobian modulation"

jdtournier:
Happy to merge, but maybe you'd like to add the more explicit error message suggested by @Lestropie first...?

maxpietsch:
> Presumably part of the design of this wording is that if the number of volumes is appropriate for an SH image, but FOD reorientation is explicitly disabled, then Jacobian modulation is entirely reasonable (it might just be that the number of volumes happens to correspond to an SH series by chance); whereas if the input data are to be interpreted as FOD data (as inferred from the request for FOD reorientation), then Jacobian modulation doesn't make a lot of sense.

Yes, I've modified the warning message as suggested. The logic behind it is [here](https://github.com/MRtrix3/mrtrix3/blob/ebb953e6e264d242790e1923419e0b200363b678/cmd/mrtransform.cpp#L426).

PR #1603, opened 2019-04-20
Write command string and software version to more outputs

Lestropie:
May need some careful consideration about which files should or should not have the relevant comment added at the head. Chance of gotchas: e.g. `dwi2response tax` and `dwi2response tournier` had naive code for reading response function files that failed to skip the new comment line. I've disabled writing the header line for `bvec` / `bval` outputs and `eddy` config / index outputs; I'm wondering if similar should be done for saving specifically transformations to file in case other softwares fail to parse such?


thijsdhollander:
I'm not a big fan of automatically adding any comments to directions and/or response function text files. This effectively changes the format, and will break backwards compatibility for a lot of peoples' scripts; even those written outside of MRtrix.  The benefit of those text files, to me, is that they're simple and clean.

thijsdhollander:
Travis CI is very insightful here, with `population_template` encountering a:

```
population_template:
                     could not convert string to float: #
```

While this occurrence is of course easily fixed, it shows what we're in for if we add these comments to those text files; so not a big fan of these changes.

Lestropie:
Was slightly concerned about the response functions similarly to the transformations. For something larger like a connectome matrix people are probably more likely to rely on a library function to import data, which one would expect to be compatible with comments, but it's still not guaranteed. Conversely there are areas where writing the software version in a comment has been replaced with the full command string & version, which should have no concomitant effect at all on a parser. The question is where that line gets drawn. We may want to be more on the conservative side than where this is currently, this just needs to be decided for individual outputs; but conversely it's not an absurd requirement for scripts to ignore comments in text files, whether our own or others.

thijsdhollander:
> For something larger like a connectome matrix people are probably more likely to rely on a library function to import data, which one would expect to be compatible with comments, but it's still not guaranteed.

Yes, that makes sense. Although talking about connectome matrices specifically, I suspect a lot of people use the ones MRtrix outputs as input to "non-MRtrix" software too.

> Conversely there are areas where writing the software version in a comment has been replaced with the full command string & version, which should have no concomitant effect at all on a parser.

Yep, I'm not concerned about those. These are also relatively MRtrix-specific (as far as that's a thing).

> The question is where that line gets drawn. We may want to be more on the conservative side than where this is currently, this just needs to be decided for individual outputs;

So well, in general, I'd err on the conservative side. I wouldn't introduce comments for most files where they weren't already present. I'm feeling particularly conservative about the more "simple" files, with information in them that's easy to interpret, and thus have a higher chance of users using other software or own scripts to do exactly that. Things that come to mind are the response function files and generally anything to do with directions (both which I mentioned above), but then Travis CI highlighted transformation matrices as well. So those 3 are some that I'd certainly not like to introduce comments in, but I might be overlooking other ones for sure.

> but conversely it's not an absurd requirement for scripts to ignore comments in text files, whether our own or others.

While I don't disagree entirely, it's simply not necessarily the reality, I reckon. As the changes as well as Travis CI pointed out, our own "previous" MRtrix version doesn't comply here. This itself, I think, is going to cause inconvenience popping up here and there.

In general, I'd really only add info on MRtrix version and/or commands to 100% truly MRtrix-specific file types, such as `.mif` and `.tck`.

maxpietsch:
I like the idea of adding the comments in plain text files but I'd draw the line at not including the command string if it breaks backward compatibility with our tools, at least to a certain version, as that is harder to fix for users and might cause frustration when simultaneously working with different versions (HCP and local installation). As far as I can tell, we should be able to support reading text matrices with comments as far back as [4 years](https://github.com/MRtrix3/mrtrix3/blame/5408098ca7c6d24e00553dfde9f39e147e28711c/core/math/math.h#L154) via the c++ functions.

numpy's `loadtxt` ignores `#` by default, not sure about other software, but I'd expect it to be a non-issue or easy fix for user-written scripts. The `population_template` CI test failure is easy to fix and would not cause backwards compatibility issues as it is self-consistent within the script.

`average_response` would not be backwards-compatible.

So, I'd lean towards writing command strings to transformation files but leave response function files unchanged for now.  I'd also suggest moving the matrix parser from `population_template` into the python script library and also use it in `responsemean` to allow adding comments in the future.

Lestropie:
> As far as I can tell, we should be able to support reading text matrices with comments as far back as 4 years via the c++ functions.

Yep, it's only Python code that was ever going to be problematic here.

> I'd also suggest moving the matrix parser from `population_template` into the python script library and also use it in `responsemean` to allow adding comments in the future.

`dwi2response tax` and `tournier` also read response function matrix files. A library function for such is probably a good idea given we've tried to avoid having `numpy` as a dependency. Where to put it though...?

-----

For transparency, the current proposed behaviour is:

- Command & version string is a `std::string` in `App` namespace, so easy to utilise in any location deemed relevant.

- `.tck` files get `command_history`.

- For `save_matrix()` / `save_transform()` / `save_vector()`, the behaviour is now to *omit* this comment string, *unless* the programmer specifically asks for it. So existing outputs won't change, but if there's some future capability that utilises one of these functions but provenance is desired, the precedent can be set for that particular type of output.

- Text file outputs that don't call one of these three functions (and hence don't fall under the "simple text numerical data" definition) do get the comment string added.

Personally I'd rather have this included in as many files as possible, but I do appreciate the ramifications of doing so... Do we think that it would be "safe" to include this line by default for specifically transformation data, but omit it for generic matrx / vector data?

jdtournier:
Just my 2 cents:

I'm all in favour of adding comments in these text files, not just for command and version history, but any other purpose that developers reckon might be useful. One of them I had in mind was for instance to store the b-value of each shell as a comment at the end of each line in the response function. It shouldn't interfere with current operation if comments are stripped out, but would require special handling if the import code actually wanted access to that information. 

If we can support it on the Python side with a generic access function, we should do so. The issue of maintaining compatibility with previous version isn't too big a deal, I don't think - it'll only affect users who want to share their results with others whose installations are older: quite a rare use case, and easily dealt with on a case-by-case basis. What is more potentially problematic is interaction with other packages or custom scripts. To handle that case, I reckon we could simply provide a one-liner that can be run to remove comments, e.g. (from [StackOverflow](https://unix.stackexchange.com/a/157619/166790)):
```
# note this edits the file in-place
# remove the `-i` option to preserve the original and print to stdout. 
sed -i '/^[[:blank:]]*#/d;s/#.*//' your_file
```
With this one-liner, I reckon it would be relatively easy for others to adjust their scripts to maintain compatibility - what do you reckon?

maxpietsch:
> A library function for such is probably a good idea given we've tried to avoid having `numpy` as a dependency. Where to put it though...?

> Do we think that it would be "safe" to include this line by default for specifically transformation data

I'd default to adding the strings to all text files unless it causes downstream user frustration with little gain. That being said I am not aware of any python code that parses transformation files other than population_template, which is self-consistent. So, I'd say we add the comments.

I'd just add the following to `app.utils` (untested, inspired by numpy's `loadtxt`):

```python
def load_txt(filename, dtype=float, delimiter=" ", comments='#', encoding='latin1', errors='ignore'):
    import re
    def decode(line):
        if type(line) is bytes:
            line = line.decode(encoding, errors=errors)
        return line

    if comments:
        regex_comments = re.compile('|'.join(comments))

    def split_line(line):
        line = decode(line)
        if comments:
            line = regex_comments.split(line, maxsplit=1)[0]
        return line.rstrip().split(delimiter)

    with open(filename, 'rb') as fh:
        return [[dtype(a) for a in split_line(l)] for l in fh.readlines()]


def load_key_val(filename, comments='#', encoding='latin1', errors='ignore'):
    import re
    def decode(line):
        if type(line) is bytes:
            line = line.decode(encoding, errors=errors)
        return line

    if comments:
        regex_comments = re.compile('|'.join(comments))

    res = {}
    with open(filename, 'rb') as fh:
        for line in fh.readlines():
            line = decode(line)
            if comments:
                line = regex_comments.split(line, maxsplit=1)[0]
            if len(line) < 2:
                continue
            name, var = line.rstrip().partition(":")[::2]
            if name in res.keys():
                res[name].append(var.split())
            else:
                res[name] = var.split()
    return res
```


> it'll only affect users who want to share their results with others whose installations are older

And anyone with heterogeneous computing environments such as HCP clusters.

Lestropie:
> One of them I had in mind was for instance to store the b-value of each shell as a comment at the end of each line in the response function.

I've encountered this issue myself: it would really be preferable to be verifying equivalence of b-values when e.g. computing a mean response function. I'd maybe do a comma-separated list in one comment line rather than a comment at the end of each line though; would make more sense when updating the parsing to handle keyvals in comments separately to the data itself.

> With this one-liner, I reckon it would be relatively easy for others to adjust their scripts to maintain compatibility - what do you reckon?

If people require instruction to maintain compatibility with their own scripts, I'd personally give instructions for updating their scripts to skip comments / empty lines in text files, rather than suggesting modification of the data itself. But each to their own. Could add an FAQ page explaining the concepts (#1394).

-----

Ended up putting in a bit of time; 47421c4 reverts default behaviour based on consensus and adds capability to add more than just the command string to the header; 5c3b787 adds a new Python module for achieving similar functionality there. Not yet tested, have committed the first versions to compile / pass `pylint`; but committing them makes it a little easier for me to do testing.

core/math/math.h
```diff
@@ -240,29 +246,40 @@ namespace MR
   }
 
   //! write the transform \a M to file
-  inline void save_transform (const transform_type& M, const std::string& filename, const bool write_cmdstring = false)
+  inline void save_transform (const transform_type& M, const std::string& filename, const vector<std::string>& header_lines)
```
jdtournier: I reckon I'd be in favour of merging both aspects. This modification forces the user to format the command history string as soon as they want to add any comments, which I think is going to lead to inconsistencies down the track. Why not have a declaration with both the `vector<string>` and `bool` arguments, like this:
```
void save_matrix (const MatrixType& M, const std::string& filename, const vector<std::string> header_lines = vector<std::string>(), const bool write_cmdstring = true)
```
Note this assumes that we do want to default to inserting the command history every time - which I'd be happy to support. 

Another potential modification would be to add the version string in there along with the command history, if that bool is true. Thoughts?

Lestropie: Part of the reasoning there was that if it's just the one line in isolation, you can just add the command string as a comment and that's it; whereas if you're wanting to have more than that, there's a decent chance that you may want the comments in some kind of format (e.g. key-value pairs), and therefore would do it all manually.

I'm happy to go with that solution, but I think the command string would need to be prepended with "`# command_history:`"...?

> Another potential modification would be to add the version string in there along with the command history

Already done: the variable "command_string" has the command, double-space, then the software version in brackets, just as has always be done in `command_history`. So maybe "`command_string`" isn't the ideal name?

Lestropie: See 5eec075.

PR #1605, opened 2019-04-21
New script: mrtrix_cleanup

Lestropie:
Keep getting annoyed coming across old scratch directories in various locations ever since 5b7af2ab / #785 moved the default location from `/tmp/` to the working directory. This finds and deletes them.

@jdtournier: Gist instead?


Lestropie:
I'm finding this useful myself, so think users would find the same. If there's no contention or better suggestions around the command name, I'll merge to `dev`.

jdtournier:
Just had a look, I reckon this is a good idea. Happy for it to go into the main repo. 

Two thoughts:

- does this clean out temporary piped images (stored in `/tmp`) by default? I had a quick look through, and it didn't look like it... Yet I reckon leftover temporary piped images are the most annoying, because they reside in a non-obvious location.

- It's probably worth adding a warning, at least in the docs, that it's not a good idea to run this while _MRtrix3_ jobs are running...

Lestropie:
> does this clean out temporary piped images (stored in /tmp) by default?

No, I chose to omit that, from memory at least in part because:

> it's not a good idea to run this while MRtrix3 jobs are running

on a multi-user system. If user permissions would mean that other users' piped images couldn't be deleted, and hence this issue would only arise if running as super-user, that would be adequate protection I suppose. 

The difficulty then though becomes the interface and resulting search space: when only the path corresponding to the singular input filesystem path is searched, the script's operation is fairly unambiguous; whereas if other paths were to be added "automatically" (e.g. `TmpFileDir` / `/tmp/` / `ScriptScratchDir` / `cwd`), and maybe the manual specification of a search path were optional instead of compulsory, it might not be immediately clear what will be searched and what will not depending on invocation.

Maybe the most sensible would be:

- Input path provided: Search that path *only*.

- Input path *not* provided: Search `TmpFileDir` (`/tmp/` if not specified), `ScriptScratchDir` (if specified), and `cwd`.

Wouldn't be against that if preferred by others.

> I reckon leftover temporary piped images are the most annoying, because they reside in a non-obvious location.

If anything those are actually vastly easier as one could historically run "`rm -f /tmp/mrtrix-tmp-*`"; it's the Python script scratch directories that get scattered all over the filesystem.

Plus, one could now just do "`mrtrix_cleanup /tmp/`".

-----

Another thought I had while writing this: If config file entries `TmpFilePrefix` / `ScriptScratchDir` / `ScriptScratchPrefix` or environment variables `MRTRIX_TMPFILE_DIR` / `MRTRIX_TMPFILE_PREFIX` are set, this script may need to alter its search accordingly depending on implementation...

PR #1309, opened 2018-04-25
Correct documentation of response function estimation

dchristiaens:
With the upcoming release of `3.0_RC3`, several changes have been made to the documentation of response function estimation that need to be discussed and agreed upon before we can merge. Naturally, anything discussed here also applies to the dangling `doctest` branch. #1244

First and foremost, the user guides of MT-CSD ([master](http://mrtrix.readthedocs.io/en/latest/constrained_spherical_deconvolution/multi_tissue_csd.html) / [doctest](http://mrtrix.readthedocs.io/en/doctest/constrained_spherical_deconvolution/multi_shell_multi_tissue_csd.html)) and MT-GT ([master](http://mrtrix.readthedocs.io/en/latest/quantitative_structural_connectivity/global_tractography.html) / [doctest](http://mrtrix.readthedocs.io/en/doctest/quantitative_structural_connectivity/global_tractography.html)) have recently been modified, changing the recommended response function estimation methods in these pages from `msmt_5tt` to `dhollander`. Whilst I agree that these sections deserve an update now that more than one option is available, I object against the switch to `dhollander` in these pages. In my view, the MT-CSD and MT-GT pages should be consistent with what was used in their respective papers. Since we already have a separate page on response function estimation, there is no point repeating anything else on this topic here, or make a recommendation for what to use. Instead, I would prefer to see a section that says: here is a page about various options; and this is the pipeline that was used if you wish to replicate the original paper.

My proposed changes in 32913dc do exactly that: both sections now start by linking to the page on response function estimation, and then outline the steps used in the respective papers. See [this page](https://github.com/MRtrix3/mrtrix3/blob/11c68f968a63070ed11581239d3c66a2f6222fe0/docs/quantitative_structural_connectivity/global_tractography.rst) for rendered output; you will notice that this is heavily based on the original MT-CSD user guide.

[...]

PR #1612, opened 2019-04-24
mtnormalise: added reference

maxpietsch:
added the ISMRM abstract as reference http://archive.ismrm.org/2017/3541.html


thijsdhollander:
> I have to admit I'm a bit unimpressed with this addition.
> ...
> ......
> ...........
> ... I think it comes across pretty badly, unfortunately.

Please, don't lose sleep over 2 words in a commit in a feature branch that wasn't even finished yet, because it was contributed on a public holiday between other work for this software, and I only have so much time to work on public holidays (or institute holidays, like last Tuesday).

It would be nice if you weren't so quick to judge a single commit in a feature branch that wasn't even elaborated upon yet. I would be really thankful for that. If you felt inclined to flag some uncertainties on your part about this, surely there are more careful and kinder ways (and a one-line email is even less effort, I'd argue); but just waiting a little bit might've also done the job. Again, I'd be really thankful for that. Thanks.

My intentions here actually had nothing to do with ownership (I could've also told you that if you had simply asked of course, or even waited just a little bit). I just want to be entirely clear about the substantial differences in the operation of the algorithm; this is documentation after all. Having designed this algorithm, I know it rather well, so I prefer to document it clearly. If the reference is added, then elaboration is needed about the differences; otherwise the message the reference sends about the way in which this command operates would be confusing, or even misleading to some. My latest commit has further formalised this, and taken out the bit that got you fired up to some degree.

jdtournier:
So you make add a statement that you don't deny was maybe a bit unfair to other contributors, but I'm the bad guy for pointing this out? Sure.

I'll ignore the blatant gaslighting, and stick to the statement itself. I still don't see much point in it. It doesn't add to the ability of users to use the command, and the previous description already documented the fact that it performs in the log domain. You could add outlier rejection and balancing to that paragraph if you want, but I don't think it warrants a whole new paragraph. If someone did go to the trouble of digging up that reference, they'd quickly figure out where the differences are from the existing description. I think the command help page should provide all the information needed to understand what it's for and how to use it, but not a lot more. As I suggested, if you want to go into the details of the implementation, the best place would be an article or technical report that can be referenced in the command. And to be honest, it would be a good idea to do this regardless.

thijsdhollander:
> So you make add a statement that you don't deny was maybe a bit unfair to other contributors, but I'm the bad guy for pointing this out? Sure.

It was an incomplete commit on a holiday. I'm only asking not to judge it immediately and wait for context. Of course I didn't deny it, I was still working on it; it's changed accordingly in the mean time. Why is this even a discussion about "bad guys"? :confounded: 

> I'll ignore the blatant gaslighting,

I don't know how much more I can emphasise this was not my intention; but again: this was not my intention. I really don't know how else I can make this more explicitly clear.

> I still don't see much point in it. It doesn't add to the ability of users to use the command, and the previous description already documented the fact that it performs in the log domain. You could add outlier rejection and balancing to that paragraph if you want, but I don't think it warrants a whole new paragraph. If someone did go to the trouble of digging up that reference, they'd quickly figure out where the differences are from the existing description. I think the command help page should provide all the information needed to understand what it's for and how to use it, but not a lot more.

I have had users asking me, implying that it wasn't clear enough. That was before the reference was added (obviously), but they'd wondered essentially what the link was between `mtnormalise` and the algorithm described in the abstract; around the time `mtnormalise` was introduced. The "log-domain" mention didn't seem to get it across on its own. If a single paragraph can clarify this, I don't see how that could be harmful in any reasonable way. We have commands with longer descriptions.

-----

Apart from the single paragraph, this is what really has me worried about the copyright of contributions that cost so much time and work. I just want to document it clearly in place with a single short paragraph, because that's all the time I can allocate to it efficiently for now. The amount of pressure to have a tiny bit of documentation, that could really clarify or at least highlight the essential difference for people who wonder about it, thrown out; the amount of resulting tensions with "bad guy" discussions erupting, further lost time needing to "justify" that small bit of documentation, etc... jeez... :disappointed:

Lestropie:
> It would be nice if you weren't so quick to judge a single commit in a feature branch that wasn't even elaborated upon yet. I would be really thankful for that. If you felt inclined to flag some uncertainties on your part about this, surely there are more careful and kinder ways (and a one-line email is even less effort, I'd argue); but just waiting a little bit might've also done the job. Again, I'd be really thankful for that. Thanks.

What you are asking for is special treatment. @jdtournier's communications here are in no way inconsistent with the posts others - yourself included - have historically made on GitHub questioning or criticising the contributions of others. I certainly don't receive this privilege myself: my myriad parallel developments are persistently picked apart here in the public domain.

> I'm only asking not to judge it immediately and wait for context. 

The original text was entirely consistent with both your prior position on the politics surrounding this command (#1034, #1036) and your proclivity for redundant self-attribution (e.g. #1309, [#1545](https://github.com/MRtrix3/mrtrix3/pull/1545/files#diff-60d7b231493a39269204158b943bd330R100)); so there would be every reason for any rational observer to expect no further modification of the text in the absence of confrontation.

> I was still working on it

Given it's a single paragraph of text, I'd consider intentionally spreading changes across multiple commits to be rent-seeking behaviour.

Moreover, given the political tensions surrounding this command and its relative attributions, to make a change to the documentation that was more self-aggrandising than what you intended the final result to be, but to nevertheless publicly commit that change with no corresponding communication, would be an exceptionally poor calculation on your part, since the ensuing backlash would have been entirely predictable.

> Why is this even a discussion about "bad guys"?

Because this has become a recurring motif, with highlighting of unacceptable modifications being met with passive-aggressive pontification / attempts to smear such criticisms as irrational / denial of intent / virtue signalling / straw-manning / exhaustive excuse listing (which can lead to self-contradiction). I doubt I'm the only one both seeing this same pattern repeat itself and growing tired of such.

> I have had users asking me, implying that it wasn't clear enough. That was before the reference was added (obviously), but they'd wondered essentially what the link was between mtnormalise and the algorithm described in the abstract; around the time mtnormalise was introduced. If a single paragraph can clarify this, I don't see how that could be harmful in any reasonable way.

The criticism is very clearly not directed principally at this aspect. A single concise description of the aspects of the command not described in the corresponding reference is entirely rational IMO. The line gets drawn when you belittle the contributions of a colleague in the process of self-promotion. That signals an unhealthy collaborative philosophy both to the rest of the team, and to any other passive observers.

jdtournier:
OK, I've modified the description to something I'd be comfortable with. I think it says everything it needs to say, I hope this will settle the issue...

thijsdhollander:
Ok, fresh start. Let's please stay on topic.

I was entirely ok with the command/documentation before this pull request, and still am entirely ok with that version.  Adding the reference yet without further documentation along with it is problematic for me, since key parts of the implementation and the reference's (serving as documentation) description of algorithms differ. These matter for the results (as presented to all and accepted when `mtnormalise` was introduced), which is why I care about not creating a misunderstanding. This is the reason I either want to retain `mtnormalise` as it was and currently still is on `master` (which has passed all of the many stages of our reviewing systems), or alternatively add the reference but not without documentation that I can be comfortable with as well.
Rejecting the pull request is the simplest solution, but I want to commit constructively towards trying to find another common version that we can all be comfortable with.

We're now at @jdtournier's most recent proposal, which is represented in the squashed latest commit (hence why it appears as a commit that I authored, but the last commit in the squash is @jdtournier's). The relevant file sits [here](https://github.com/MRtrix3/mrtrix3/blob/4a24d2d63e244e8751356a7095f0d709ff45f9e3/cmd/mtnormalise.cpp).

I have now carefully considered @jdtournier's proposal, and I am currently not comfortable with it. For a new version of the documentation which includes the reference to work for me, I'd like to be 100% clear what bit in the description mentions the differences. The word "now" is not clear enough in that regard for me to be comfortable with the description. I will propose a minimal change which would result in a version that I can be comfortable with. This can then be re-evaluated.

thijsdhollander:
The new proposal has been committed. I have even removed a mention of "improved" from the previous proposal. It is now a very formal and neutral statement, in the location which @jdtournier prefers it to sit, relative to the remainder of the description. Feel free to re-evaluate when you have the opportunity, and we can assess whether convergence is possible.

jdtournier:
OK, I'm happy with the current statement. Glad we could reach agreement. I'll merge now.

PR #1614, opened 2019-04-29
Script testing

Lestropie:
Given the manual testing only of #1449, the increasing scope of #1603, and the possibility of various other changes in `dev` having undetected detrimental effects, I would prefer to have this up and running before `3.0_RC4` is tagged.

These are not intended to be run as part of CI; rather, they are a secondary set of tests that should be run manually to protect against regression whenever changes are made to either executable scripts or the Python library itself.

The data currently reside in a second repository "`MRtrix3/script_test_data`", so as not to unnecessarily increase the size of the test data to be downloaded during CI, which is then accessed via a second submodule path. Tests are executed via `./run_tests -scripts`, and will require prior installation of other softwares (e.g. FSL, ANTS, FreeSurfer) to complete successfully.

Listing as draft PR; I've not yet ironed out all kinks, but I've at least brought it up to date with `dev` and addressed the major divergences. 

Current observations:

- BET gives a slightly different outcome between FSL5 and FSL6, which has a concomitant effect on the output of `5ttgen fsl` since it crops the FoV based on the brain mask.

- `population_template` seems to be not perfectly reproducible if not run on identical software / hardware; whether this is due to differences in hardware / OS / *MRtrix3* versions I have not yet determined. Could do something like dilation / erosion of intersection masks to permit some degree of error in overlap, but obviously that will then limit sensitivity in detecting actual regressions in the future. Or could just make sure that the script runs but not test the output against existing data.

- #1555 has introduced an issue with concatenation of volumes in the presence of phase encoding BIDS fields, which is currently preventing some `dwipreproc` tests from passing.


Lestropie:
Boosted from draft to full PR; looking to have this merged to `dev` in order to validate Python changes, both existing and any incoming, prior to tag update. Open to comments on repository arrangement / test interface, and any additional tests.

Regarding the earlier points:

- Test data is provided based on both FSL5 and FSL6, and generated data must match one of these.

- I never fully isolated the source of the `population_template` discrepancy; I believe that initially my test data were generated in a Docker container running on Windows, and that did not match data generated on any other system, but I no longer have that Windows system. So I have simply re-generated the corresponding test data; if anybody ever encounters a failure of these tests on a particular system we can investigate further then.

- Residual failing tests should have been resolved by the merge of #1618.

Lestropie:
While I have resolved most issues in bringing this branch up to date and using it to test some other recent changes to the Python API in `dev`, there has been considerable change over time in the results of `population_template` (over and above the issue previously described). Compared to the most recent version of the `population_template` target data I generated, what's currently being produced on `dev` code has a template location that's translated by around a full voxel out on 2 of 3 axes, despite the voxel grid on which each is defined being identical.

For this PR I've disabled the checking of `population_template` output against existing data (the commands will only be run to ensure successful completion). But the irreproducibility of `population_template` does I think warrant its own investigation.

Lestropie:
```
[rob@rsmithmetabox mrtrix3]$ ./build; ./run_tests -scripts
logging to "testing.log"
fetching test data... OK
building testing commands... OK
running "5ttgen"... 8 of 8 passed
running "dwi2response"... 40 of 40 passed
running "dwibiascorrect"... 8 of 8 passed
running "dwicat"... 4 of 4 passed
running "dwigradcheck"... 4 of 4 passed
running "dwinormalise"... 3 of 3 passed
running "dwipreproc"... 15 of 15 passed
running "dwishellmath"... 3 of 3 passed
running "foreach"... 3 of 3 passed
running "labelsgmfix"... 4 of 4 passed
running "population_template"... 13 of 13 passed
running "responsemean"... 2 of 2 passed
[rob@rsmithmetabox mrtrix3]$ mrconvert -version
== mrconvert 3.0_RC3-852-gcaf08fc4 ==
```

Lestropie:
@jdtournier Thoughts on file / directory naming here?

I don't particularly like having `testing/data/`, `testing/script_data/`, `testing/tests/`, `testing/script_tests/`, and then `./run_tests` and `./run_tests -scripts`. It feels too much like the script testing has been tacked on; which is the case, but it feels unnatural nevertheless.

I think I'd rather have something like: `testing/tests_binaries/`, `testing/data_binaries/`, `testing/tests_scripts/`, `testing/data_scripts/`, and then `./run_tests binaries` and `./run_tests scripts`.

jdtournier:
As we discussed earlier, how about:
```
testing/
 ├── binaries/
 │   ├── cmd/
 │   ├── data/
 │   ├── src/
 │   └── tests/
 └── scripts/
     ├── data/
     └── tests/
```

And we can then modify the `run_tests` to take an argument (`binaries` or `scripts`) and run whatever tests are in that particular sub-folder. We would then have the potential to have more categories should we ever need them... (unlikely at this stage, I'll admit).

Lestropie:
I'd consider modifying that to:
```
testing/
 ├── binaries/
 │   ├── data/
 │   └── tests/
 ├── cmd/
 ├── scripts/
 │   ├── data/
 │   └── tests/
 ├── src/
 └── tests/
```
, since some of the `testing_diff_*` commands are utilised in testing of script outputs. Means that root directory will classify as an *MRtrix3* project rather than `binaries/`. Execution of unit tests could be managed via entries in `/tests/` rather than `/binaries/tests/`? There's also the prospect of future Python-based unit tests, which would go in `/bin/`.

jdtournier:
I had similar thoughts, so quite happy to go along with your suggestion.

Lestropie:
@jdtournier: With the re-arrangement in e2ea118, if you could adequately anonymise the data, you could quite easily add your private DICOM import test suite as an additional data module & set of tests.

Lestropie:
Merging in an attempt to get things progressing. Still some results to look into, but diagnosing them may be made easier by getting this stuff into `dev`.

PR #1617, opened 2019-05-03
Sh2amp enhancements

jdtournier:
Adds multi-shell and multi-tissue capability to `shconv` and `sh2amp`. 

Briefly, this extends SH storage to allow storing SH along the 5th dimension of an image, one per shell. `shconv` now allows multiple ODF/response pairs (much like `dwi2fod` and `mtnormalise` do), and allows multi-shell responses (in which case it'll produce a 5D output). `sh2amp` can now accepts directions files in spherical or Cartesian coordinates, full-blown DW encoding files, or an image with a DW encoding in its header (required to handle multi-shell inputs). 

This allows forward-modelling of all our spherical deconvolution operations, e.g. for MSMT-CSD:
```
shconv wm.mif wm.txt gm.mif gm.txt csf.mif csf.txt - | sh2amp - dwi_orig.mif dwi_predicted.mif
```

Also in there are a few changes to docs handling, and minor updates to the tests.


jdtournier:
OK, quite a few tweaks to the docs - triggered by how bad the description of the SH coefficients looked... 
Preview on [readthedocs](https://mrtrix3-dev.readthedocs.io/en/sh2amp_enhancements/index.html), e.g. [shconv](https://mrtrix3-dev.readthedocs.io/en/sh2amp_enhancements/reference/commands/shconv.html) and [sh2amp](https://mrtrix3-dev.readthedocs.io/en/sh2amp_enhancements/reference/commands/sh2amp.html).

Lestropie:
- Might be worth adding a reduced-FoV multi-shell DW image to the main test data repo for functionalities like these.

- Changes are substantial enough in both commands I think to throw in secondary authorship.

jdtournier:
Tests are already in there - see d1ba865.

Hadn't noticed the authorship. I reckon I might have earned it here... :Grin:

thijsdhollander:
I love the overall idea (both the new functionality, as well as the "convention" for 5D multi-SH-shells). Won't have time soon (aka before ISMRM) to do any testing though, but I suppose you've done quite some testing to make sure it all works, right?
Let's maybe briefly chat/confirm on Tuesday during the meeting to make sure we're all good and understanding it all correctly; that'll be more efficient rather than typing up long texts here.

One thing, now that I think of it (and so I don't forget); I'm sure it's probably clear from the code (but again, sadly I don't have the time to look thoroughly into it), but: when you provide a full gradient table (b-values and all) to the new `sh2amp`:

1. I suppose it writes that gradient table also to the output header, so it becomes a "stand-alone" and fully self-consistent "dMRI dataset", right? If not (yet), I think it should definitely do that; that'd be really handy.

2. And while I typed up point one (and made it into points 1 and 2), just realised: there must be some definition then of how the SH "shells" from the 5D input are linked to b-values, correct? I'm really assuming stuff now all the way since point 1, but I'm guessing the "definition" here is then just via increasing b-values?  I.e. if 4 "shells" in the SH file and 4 b-values in the gradient table, the first "SH shell" goes with the lowest b-value, and so on? Another definition I could imagine (not per se prefer or anything though) would be the order of first appearance in the gradient table, which could be different from increasing b-value order of course. And then finally, it has to be documented clearly enough of course (if it isn't already and I'm overlooking it right now). Well, just bringing this up since it's something to consider (not sure what my opinion would be here; will have to sleep over it for a night or two, I guess). Let's go over that on Tuesday or something; it's probably simpler than I'm imagining now.

(and the authorship thing is all good of course)

thijsdhollander:
Wrt the above, I've come round (rather quickly after posting the above) to preferring assuming ordering with respect to increasing b-value. Still haven't found time to look into the code, documentation, or anything really; but I'm more or less assuming you probably went with that as a convention as well... it's the most consistent with a few other tools (e.g. response function estimation and such).

thijsdhollander:
Since this alters the documentation, and introduces a new sort of "convention" wrt how SH coefficients are stored (i.e. the addition of how the 5th dimension is used in this context), it might be worthwhile or more efficient to mash this up with #1635 (or somehow combine / coordinate / ... the effort).

thijsdhollander:
I just coincidentally also remembered http://community.mrtrix.org/t/csd-on-standard-dwi-data/2499/5 : a current mistake in the `lmax` bit of the `dwi2fod` docs (it's essentially "lagging behind": it used to be correct, but isn't anymore since a long while). Might be worthwhile to patch that up at the same time. Essentially, this sentence:

> If omitted, the command will use the highest possible lmax given the diffusion gradient table, up to a maximum of 8.

...should become:

> If omitted, the command will use the highest possible lmax given the ***lmax of the response function(s)***, up to a maximum of 8.

Issue #1443, opened 2018-08-29
mrcat DWI images with gradient scheme on axis 4 does not work


maxpietsch:
Currently `mrcat` concatenates the dw_schemes and throws an error as the output has only half the number of volumes as the created dw_scheme suggests. 
`mrcat: [ERROR] Number of volumes in image "dwi.mif" (300) does not match that in phase encoding table (600)`.

We probably should check whether both gradient schemes are identical and use only one, strip it otherwise?

maxpietsch:
```
$ mrinfo dwi.mif
************************************************
Image:               "dwi.mif"
************************************************
  Dimensions:        100 x 100 x 64 x 300
  Voxel size:        1.51515 x 1.51515 x 1.5 x 1
  Data strides:      [ -2 3 4 1 ]
  Format:            MRtrix
  Data type:         32 bit float (little endian)
  Intensity scaling: offset = 0, multiplier = 1
  Transform:                    1          -0           0      -78.41
                               -0           1           0      -54.45
                               -0          -0           1      -39.96
  dw_scheme:         0.019866,0.411523,0.911183,0
  [300 entries]      -0.385352,-0.305637,0.870684,2600
                     ...
                     -0.305057,-0.445414,-0.841752,2600
                     -0.19586,0.857924,0.47498,1000
  mrtrix_version:    3.0_RC3-266-gebbabc8d-dirty
  pe_scheme:         1,0,0,0.1
  [300 entries]      -1,0,0,0.1
                     ...
                     0,1,0,0.1
                     0,-1,0,0.1
```
Concatenate twice:
```
 $ ~/mrtrix3_dev [dev]$ mrcat dwi.mif dwi.mif -axis 4 - | mrinfo -
mrcat: [ERROR] Number of volumes in image "dwi.mif" (300) does not match that in phase encoding table (600)
```

thrice:
```
$ mrcat dwi.mif dwi.mif dwi.mif -axis 4 - | mrinfo -
mrcat: [ERROR] Number of volumes in image "dwi.mif" (300) does not match that in phase encoding table (900)
```

`mrcat` works when the pe_scheme is removed in one of the files but then the output does not have a `pe_scheme` header entry:
```
mrcat dwi.mif dwi.mif $(mrconvert dwi.mif -clear_property pe_scheme - ) -axis 4 - | mrinfo -
```

thijsdhollander:
[...]

PR #1618, opened 2019-05-04
Modifications to 4D scheme handling in mrcat and [] notation

Lestropie:
This leads on from #1555, and was pursued in trying to resolve #1614 but has been branched away from that as a separate issue.

A number of scripts were failing to complete after #1555, complaining of incorrect phase encoding tables. But it'll take a couple of steps for me to explain why...

1. As described back in #747, in order to provide support for BIDS, if all volumes within a 4D image have the same phase encoding direction & total readout time, then this information will be written to the image header via BIDS fields "`PhaseEncodingDirection`" and "`TotalReadoutTime`". However for various functionaities it was necessary to support in *MRtrix3* image series with different phase encoding per volume; in these cases, this information is written to field "`pe_scheme`".

2. Whenever concatenation of volumes occurs, this is achieved by converting the header key-value information to numerical matrix format (regardless of the format in which such data are actually stored), concatenating the matrices from different files, and then saving back to the image header (wherein the test for phase encoding contrast is re-applied in order to determine the format in which the resulting data will be stored in the header).

3. When the BIDS fields are present in the header, the equivalent phase encoding table is generated by producing one row of the table based on those two fields, and then duplicating based on the number of volumes in the header.

4. If either the dimensionality / number of volumes in an image header is changed, or the number of rows in a DW / PE table is changed, this leads to an intrinsic discordance between two different information sources within the one header. This may or may not be flagged, depending on the precise order of modification operations / presence of sanity checking.

5. When it comes to both `mrcat` and use of the square-bracket image path notation, turns out it's quite regular for points 3 and 4 to butt heads. For instance, the number of volumes in the resulting header would be set, and *then* an attempt would be made to concatenate the phase encoding table; but when generating the table for the concatenation of the first two input files, the first of these two would be given a number of rows corresponding to the number of volumes in the *output image*.

I had to start & cancel a couple of different possible solutions before I ended up with acc9bd2. The heavy-lifting of both `mrcat`, and loading of multiple image files using the `[]` notation, has been moved to a new function `Header::concatenate()`, which constructs a new `Header` based on the concatenation of the contents of a `vector<Header>` along a specified axis. This resolves discrepancies between header key-values in a similar manner to that done previously, but also deals with diffusion gradient table and phase encoding table data directly within the process of header concatenation such that the mismatch described in point 4 can never eventuate. It also implements checking of DW / PE table validity whenever writing to a header, which probably had to be omitted previously (at least in part) because of some of the issues described above.

The behaviour is best demonstrated using the new `mrcat` tests:

```
$ mrinfo voxel1.mih
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: i-
TotalReadoutTime: 0.1
dw_scheme: -1,0,0,1000

$ mrinfo voxel2.mif
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: i
TotalReadoutTime: 0.1
dw_scheme: 1,0,0,1000

$ mrinfo voxel3.mif
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: j-
TotalReadoutTime: 0.2
dw_scheme: 0,-1,0,1000

$ mrinfo voxel4.mif
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: j
TotalReadoutTime: 0.2
dw_scheme: 0,1,0,1000

$ mrinfo voxel5.mif
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: k-
TotalReadoutTime: 0.3
dw_scheme: 0,0,-1,1000

$ mrinfo voxel6.mif
Dimensions: 1 x 1 x 1
PhaseEncodingDirection: k
TotalReadoutTime: 0.3
dw_scheme: 0,0,1,1000

$ mrcat voxel*.mih -axis 0 - | mrinfo -
Dimensions: 6 x 1 x 1
*empty: different PE / DW directions within the one volume*

$ mrcat voxel*.mih -axis 3 - | mrinfo -
Dimensions: 1 x 1 x 1 x 6
dw_scheme: -1,0,0,1000
           1,0,0,1000
           0,-1,0,1000
           0,1,0,1000
           0,0,-1,1000
           0,0,1,1000
pe_scheme: -1,0,0,0.1
           1,0,0,0.1
           0,-1,0,0.2
           0,1,0,0.2
           0,0,-1,0.3
           0,0,1,0.3

$ mrcat voxel1.mih voxel1.mih voxel1.mih voxel1.mih voxel1.mih voxel1.mih -axis 0 - | mrinfo -
Dimensions: 6 x 1 x 1
PhaseEncodingDirection: i-
TotalReadoutTime: 0.1
dw_scheme: -1,0,0,1000

$ mrcat voxel1.mih voxel1.mih voxel1.mih voxel1.mih voxel1.mih voxel1.mih -axis 3 - | mrinfo -
Dimensions: 1 x 1 x 1 x 6
PhaseEncodingDirection: i-
TotalReadoutTime: 0.1
dw_scheme: -1,0,0,1000
           -1,0,0,1000
           -1,0,0,1000
           -1,0,0,1000
           -1,0,0,1000
           -1,0,0,1000

```

-----

What I have omitted from this PR is the issue of using multiple square-brackets within a single image path, as described in the [documentation](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#multi-file-numbered-image-support):

- Conceptually, what this actually is (for image load) is a *recursive concatenation*. I have constructed the code logic to be faithful to this as much as possible, but this particular capability is not yet tested for the reasons below.

- As far as I can tell, this is currently not working, even in `master`. E.g.:
```
$ mrinfo dwi.mif 
Dimensions: 6 x 8 x 9 x 68

$ mrcat dwi.mif dwi.mif -axis 4 dwix2.mif
$ mrinfo dwix2.mif
Dimensions: 6 x 8 x 9 x 68 x 2

$ mrconvert dwix2.mif tmp-[]-[].mif
[ERROR] output file "tmp-00-00.mif" already exists

$ ls
tmp-00-00.mif   tmp-01-00.mif   tmp-02-00.mif   ...   tmp-67-00.mif
tmp-00-01.mif   tmp-01-01.mif   tmp-02-01.mif   ...   tmp-67-01.mif
tmp-00-02.mif   tmp-01-02.mif   tmp-02-02.mif   ...   tmp-67-02.mif
     ...             ...             ...        ...        ...
tmp-00-67.mif   tmp-01-67.mif   tmp-02-67.mif   ...   tmp-67-67.mif
```

  **Edit**: This specific point should be resolved in #1619.

- I've *not* implemented support for 5D schemes, as discussed in #1443; partly because of the issue above being more pertinent to resolve, and partly because I don't personally have a need for such and therefore can't justify putting time into it right now. The changes within this PR should however make it much easier to control the behaviour of such; under concatenation at least. It would also centralise the location for if handling of arbitrary 4D scheme key-values were desired, which was raised somewhere. Having a mirror function to `Header::concatenate()` responsible for separating a single input `Header` out into one or more `Header`s along a specific axis - either based on a set of desired indices as in `mrconvert -coord`, or one `Header` per output file on use of square-bracket notation - would similarly centralise the inverse operation, along with the specialised handling of header key-values when appropriate to maintain concordance between header dimensions and key-value contents.


jdtournier:
Darn, I'll have to wrap my head around all this again, it's going to take a while... 
I guess the main thing is that all tests still run OK, so that's the main thing. I'll try to review this later tonight.

core/header.cpp
```diff
@@ -192,35 +160,73 @@ namespace MR
...
+          while (axis < H.ndim() && H.stride(axis))
+            ++axis;
+          loopindex2axis.push_back (axis);
         }
+
+        // Reimplemented support for [] notation using recursive function calls
+        // Note that the very first image header has already been opened before this function is
+        //   invoked for the first time; "vector<Header>& this_data" is used to propagate this
+        //   data to deeper layers when necessary
+        std::function<Header(vector<Header>&, const size_t, size_t&)>
```
jdtournier: auto?

core/header.cpp
```diff
@@ -703,4 +709,131 @@ namespace MR
...
+
+    auto concat_scheme = [] (Eigen::MatrixXd& existing, const Eigen::MatrixXd& extra)
+    {
+      if (!existing.rows())
+        return;
+      if (!extra.rows() || (extra.cols() != existing.cols())) {
+        existing.resize (0, 0);
+        return;
+      }
+      existing.conservativeResize (existing.rows() + extra.rows(), existing.cols());
```
jdtournier: will likely need to cater for the possibility of different number of columns across the different inputs, and resize to the smaller of the two...

core/header.h
```diff
@@ -377,8 +377,8 @@ namespace MR
 
 
 
-
-
+  // Can't be a static member function due to memory alignment requirements of vector<>
```
jdtournier: This is an odd one... What's the issue here, exactly?

jdtournier:
OK, a few comments, but nothing that scream out at me here. But then I'm really struggling to form a full picture of what's going on here. Ultimately, I reckon the main criterion should be that everything checks out on testing as many common and edge cases as is reasonable. You've already got a few in the tests, maybe we can add a few more to cover our backs? Have you for instance tried handling explicit number sequences, e.g. `input-[1:3].mif` or `input-[0:2:6].mif`, etc? As long as all the main scenarios check out OK, I'm happy to sign off on this...

core/header.h
Lestropie: If I try to make that function a `static` member of `Header`, I get a compiler error, because it receives a `vector<Header>`, `vector<Header>` requires definition of aligned memory allocation functions as per `core/types.h`, but such functions can't be defined if the definition of class `Header` is not yet complete and hence its memory requirements are not yet confirmed.

Lestropie:
> You've already got a few in the tests, maybe we can add a few more to cover our backs? Have you for instance tried handling explicit number sequences, e.g. input-[1:3].mif or input-[0:2:6].mif, etc?

It passed all the existing `mrcat` tests; admittedly this doesn't include non-unity step though.

In thinking about additional tests here, I suppose that technically concatenation along `SliceEncodingDirection` should result in concatenation of the contents of `SliceTiming`...

core/header.cpp
Lestropie: Resizing only makes sense if it's guaranteed that a scheme with more columns will strictly be a superset of one with less columns. Not sure this would be the case; e.g. if storing the full *b*-tensor in `dw_scheme` I'd expect 6 values for the tensor then the *b*-value.

Lestropie:
Took a bit of stuffing around, but that works.

I couldn't get the `std::function` to return `Header` by move: whatever I tried, it would either copy-construct (which loses the IO handlers), or the calling function would start referencing data that had been popped off the stack (eeek). You're welcome to give it a go if you want, but just having a `Header&` function argument catch the result of the `std::function` call works.

There's also some extra tests.

One thing that caught me out here is that `mrconvert -coord 3 0:2:end` works, but `tmp-[0:2:end]` does *not*. Don't know if that's worth the effort implementing, since it might require a bit of effort to modify `File::NameParser` to achieve such.

jdtournier:
> I couldn't get the std::function to return Header by move: whatever I tried, it would either copy-construct (which loses the IO handlers), or the calling function would start referencing data that had been popped off the stack (eeek). 

I assume you're talking about [this line](https://github.com/MRtrix3/mrtrix3/pull/1618/files#diff-7499bbba61a7436a4e24dba406ac37cdR176)? In which case I really wouldn't worry about it, what you've got works, and it's an internal lambda anyway...

> One thing that caught me out here is that mrconvert -coord 3 0:2:end works, but tmp-[0:2:end] does not. Don't know if that's worth the effort implementing, since it might require a bit of effort to modify File::NameParser to achieve such.

OK, it would be _possible_ to support this, but pretty hard, I have to admit... I don't think we should worry about it at this stage. 

> It passed all the existing mrcat tests; admittedly this doesn't include non-unity step though.

Good point. Together with your additional tests, I reckon that's enough. Let's merge, if there's remaining issues, we can fix them when they crop up.

PR #1622, opened 2019-05-23
Documentation to Discourse wiki

### LINKED ISSUE #1394: Need easy to edit, user-editable FAQ section

Lestropie:
I made it partway though this a while back, but I think this is everything now.

- I made the executive decision and renamed the "`frequently_asked_questions`" category on the community forum to "`wiki`". I think that's more accurate, and will be how we refer to it in discussion anyways.

- I went to the extreme and moved *everything* from the "Troubleshooting" section. There is instead just a very small "Troubleshooting" page within the "Tips and tricks" section, directing people to the community forum. As far as content that perhaps *shouldn't* be moved, first cabs off the rank are two areas that are linked to from elsewhere in the docs:

   - In the FBA pipelines, what to do when CFE runs out of memory (could maybe be substituted with a small piece of inline text);

   - Remote display management for linux / HPC install (but this is one thing that may change over time, so would be preferable to be in the Wiki for that reason).

   Open to continuation of discussion from #1394 regarding what should / shouldn't be moved.

- Also finally added a `README.md` page with instructions for compiling the documentation.

Closes #1394.


Lestropie:
[Docs link](https://mrtrix3-dev.readthedocs.io/en/docs_to_discord/)

Lestropie:
Opening as full PR and closing #1394 in the hope of attracting thoughts on what stays in the inline docs vs. moving to the wiki prior to merge some time soon.

jdtournier:
Had a quick look, I'm happy enough with this. :+1:

Issue #1428, opened 2018-08-15
New script: dwicat


Lestropie:
(name suggestions welcome)

As raised in multiple locations:

A convenience script for concatenating multiple DWI series where the intensity scaling of image reconstruction may have varied due to recalibration. This would involve:

- Extracting the *b*=0 volumes from each series.
(Easiest would be to require .mifs at input so that gradient tables are embedded)

- Using `mrhistmatch scale` or `mrstats -output median` to determine the global intensity fluctuations between acquisitions.

- Concatenate all data into a single 4D image, applying appropriate multiplicative factors.

It should also check for variations in TE and TR between the input series, and warn where appropriate.

Lestropie:
Bump

jdtournier:
Sounds good, agree on the principle, and will definitely come in handy. I'd prefer the simpler name of `dwicat` personally. We could also make the intensity normalisation conditional on detecting a sufficiently large difference in scaling, along with a clear report about this? Also, I'd advocate using the raw DICOMs as input - though mifs would be fine too, of course.

thijsdhollander:
Yep, 100% agreed with all of @jdtournier's comments.

Lestropie:
> Also, I'd advocate using the raw DICOMs as input - though mifs would be fine too, of course.

Any risk of interaction with `dwidenoise`?

Lestropie:
Closed by #1626.

PR #1626, opened 2019-06-08
New script dwicat

jvohryzek:
Hey Robert,
Could you get it approved by some other developer. 

Best
Lena & Jakub

### DIFF
--- a/bin/dwicat
+++ b/bin/dwicat
@@ -63,13 +63,9 @@ def execute(): #pylint: disable=unused-variable
     run.command('mrconvert ' + path.from_user(app.ARGS.mask) + ' ' + path.to_scratch('mask.mif') + ' -datatype bit')
   app.goto_scratch_dir()
 
-  # extract mean b=0 volume for each image series
+  # extract b=0 volumes within each input series
   for index in range(0, num_inputs):
-    run.command('dwiextract ' + str(index) + 'in.mif - -bzero | ' + \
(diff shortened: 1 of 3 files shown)


Lestropie:
Closes #1428.

Lestropie:
Manually setting average *b*=0 intensity would be pretty easy. What I actually wanted to do here, but couldn't convince myself that my math was correct, was to intensity-match all protocols to all other protocols, producing a scaling factor *matrix*, and then scale each protocol individually in such a way as to converge their intensities toward a common target, rather than arbitrarily using the first scan as the "target" intensity.

treanus:
Hi developpers,

I did a git checkout dev and tried dwicat.

`dwicat part1.mif part2.mif full.mif`

Get the error:

```
[ERROR] mrhistmatch scale 1b0.mif 0b0.mif 1rescaledb0.mif (dwicat:84)
dwicat: [ERROR] Information from failed command:
dwicat:
        mrhistmatch: [ERROR] mrhistmatch currently only works on 3D images
```

jdtournier:
The script relies on updates to the supporting executables, which need to be recompiled. You need to run `./build` after checking out the `dev` branch to get it to work.

treanus:
> The script relies on updates to the supporting executables, which need to be recompiled. You need to run `./build` after checking out the `dev` branch to get it to work.

Works fine now.

@jdtournier Thx

PR #1627, opened 2019-06-08
mrhistmatch: Enable operation on 4D images

Lestropie:
Small tweak to `mrhistmatch` to enable operation on 4D images.

By merging this prior to #1626, that script will be able to operate using all *b*=0 data from each image series, rather than having to collapse them into a mean *b*=0 image per series. Therefore will merge this PR first, then make the corresponding change in #1626.


Lestropie:
Closed in favour of #1628.

PR #1632, opened 2019-06-26
permute_axes adapter: fix handling of additional axes

jdtournier:
This allows the correct interpretation of negative axis indices in the specification of the new axes, which is required to insert additional axes (e.g. going from dimensions [ X Y Z 3 ] to [ X Y Z 1 3 ]). This behaviour is required and documented for mrconvert's -axes option, which currently doesn't work as stated (fixed with this change).


Lestropie:
Would benefit from a test case I think: it's the sort of thing that isn't obvious from the code itself, and so could easily be re-broken if future changes are made for some other purpose in the corresponding code.

jdtournier:
Was thinking the same... I'll push one up later today.

jdtournier:
Ok, tests added, should be good to go. Just need someone to give it the thumbs up and merge...

PR #1635, opened 2019-06-29
SH description: Revise

Lestropie:
Was utilising the spherical harmonic basis description for something else, and found a bit of an internal inconsistency:

> ... it has conjugate symmetry, i.e. Y(l,-m) = Y(l,m)* (where * denotes the complex conjugate).

> volume 1: l = 2, m = -2 (imaginary part of m=2 SH) ;

So after saying that coefficients for negative values of *m* can be omitted from the basis, negative values of *m* are then used in the description of the basis, but to instead refer to the imaginary components of the coefficients for positive values of *m*. :clown_face:


core/math/SH.cpp
```diff
@@ -36,11 +36,13 @@ namespace MR
         "Each volume in the output image corresponds to a different spherical harmonic "
         "component. Each volume will correspond to the following: \n"
         "volume 0: l = 0, m = 0 ; \n"
-        "volume 1: l = 2, m = -2 (imaginary part of m=2 SH) ; \n"
```
thijsdhollander: I think you may have misunderstood the intention of @jdtournier's initial wording here. The l and m information indicates the parameterisation, i.e. just like the volume number (as an alternative parameterisation) does as well.  The bit between brackets is merely a description of the definition for said parameters.  E.g., for l=2 and m=-2, the definition of the function implies that you have to compute the imaginary part of the l=2 and m=2 SH.  The confusion might stem from the different use of m as a parameter: the first m (in this example -2) is a parameter of the "modified" SH basis, i.e. the one we define here in; whereas the second m is a parameter of the "standard" or "unmodified" SH that is used "within" the definition of the "modified" SH basis.  My wording here is on purpose a bit informal, but I suppose you get the point.

Just as a more formal example, I think one of the first ones (or at least one of the most often referred to) is Maxime's early PhD paper where he used symmetrical SH to model ADCs: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.20948 .  See equation 11 in there. It might not be exactly the basis used in mrtrix, but the general format of defining it is the same.  The original version of the documentation here defines the basis in function of volume number, similar to how Maxime's paper defines it in function of parameter "j".  The "j" parameter (or volume index) itself is defined in function of l and m (see the sentence before equation 11 in the paper). Finally there is equation 11 itself, which explains how this is then actually evaluated, and part of that is an "unmodified" SH evaluation; but for some values of m you have to take the real part, and for others the imaginary part; and this choice indeed depends on the sign of m.

So long story short: the original documentation here is very much in line to the way in which this is usually formally defined in the literature.  But the brevity of the documentation might lead to confusion.  I would therefore personally argue the brevity is the problem; but rather than making individual command documentation excessively long, it might be wise to keep it short on purpose, and include a reference to another place (this might e.g. be a documentation page), where there is enough space / flexibility / words to describe this in detail in a way that doesn't allow for confusion.  My final 2 cents would be that most users don't need this in detail in the command description; it's a lot of lines that they probably want to scroll past each time again just to look up the name of an option of said command, in my experience the most common use of the inline docs for many frequent users.  Users that require this description, are probably at least already trying to port information between software packages, or are implementing stuff on their own.  They would probably look into more detailed docs anyway, rather than going by a command description.

jdtournier:
Yep, you're both right here. There are two inconsistent definitions in use at the same time. This is at the very least confusing (unless you already know a fair bit about the SH basis). I like the suggestion of moving the description into the main documentation and clarifying it properly, which will de-clutter the app's help page while providing a more accurate and detailed description. 

As to the description itself, there's still the glaring issue that it describes the non-normalised basis... I'll have a go at clarifying when I find the time.

jdtournier:
OK, I've had a stab at updating the description of the SH as used in MRtrix3 in the main docs, so that we can refer to it in the help pages for individual commands. 

I've used the original orthonormal SH page for this, which was where the change of basis was described. I've renamed the page just to 'Spherical Harmonics', and left the change of basis discussion at the end. The page isn't finished yet (still need to discuss zonal harmonics as used to store responses), but it would be nice to get some feedback and a few other pairs of eyes on this to make sure I've not made any egregious mistakes so far... [Please take a look at the preview here](https://mrtrix3-dev.readthedocs.io/en/sh_description/concepts/spherical_harmonics.html) and let me know if you spot anything that needs amending!

jdtournier:
OK, I reckon this is one is good to go. Please [check the docs preview](https://mrtrix3-dev.readthedocs.io/en/sh_description/concepts/spherical_harmonics.html) and give it the thumbs up/down.

Lestropie:
There's still a separate "concepts" docs page on *l<sub>max</sub>*; want to merge that with this one?

jdtournier:
Good point. We probably should look into doing that...

Lestropie:
In "Storage conventions", I think I'd rather still have a table showing which *l* & *m* are encoded within each image volume for the first ~ 7 volumes, rather than relying entirely on text & equations.

Lestropie:
- Associated Legendre polynomials are not defined.

- Should a note be added upfront regarding the sqrt(2) multipliers in the m!=0 functions yielding an orthonormal basis, rather than only mentioning it later in the "differences with previous versions" section?

jdtournier:
:+1: I like those changes. Made a tiny edit on the wording in d0cdb77, but nothing major.

> There's still a separate "concepts" docs page on lmax; want to merge that with this one?

I had a look into this, and I don't think it would be appropriate. It reads like a set of practical recommendations, rather than the much more theoretical nature of the SH page itself. I think it's best left separate, but I have added a link from the *l*<sub>max</sub> page pointing to the SH page itself. 

Assuming you're happy with this, I'm happy to merge.

Issue #738, opened 2016-07-25
mrthreshold optimal threshold with -mask option bug


thijsdhollander:
When running `mrthreshold` with the `-mask` option, the idea is that the algorithm only uses the data inside the mask to compute the optimal threshold.  That works ok (so no need changing), but there's a bug in initialising the golden section search.  The latter would need to be initialised with the minimum and maximum values, but for this step it accidentally uses the whole image (rather than just the data in the provided `-mask`).  That becomes problematic when the minimum or the maximum of the whole image is (much) lower or higher than the actual minimum or maximum of the data in the mask's region.

What basically needs fixing is the `min_max` function, that would need to offer a mask option.  @draffelt and I took a brief look at it, somewhere at the end of a long day several weeks ago; but we stuffed up somewhere.  What we had done is https://github.com/MRtrix3/mrtrix3/commit/c2fabb087c5269cff5f6b59e57221797871762a9 in the `optimal_threshold_tweaks` branch back then, but we must've made a mistake somewhere, as it introduces random other bugs here and there.

I don't have the time to look into it, sadly... so just putting it up here now that I just remembered it again...

jdtournier:
Discussed in #1481 - closing.

Issue #205, opened 2015-04-02
mrthreshold: replace -abs with -above and -below


draffelt:
Sometimes it would be handy to compute a mask based on a range of values. I realise this can be done currently with multiple steps, however it would be nice to have it in one command. 

If both options are input we would need to compare the magnitude of the -above and -below to determine if an || or && operation should be used. 

The only issue is that it might break a few scripts...being such a fundamental command.  Any objections?

jdtournier:
Could that not be done using:

```ShellSession
mrcalc in.mif <above> -gt in.mif <below> -lt -mult in.mif -mult out.mif
```
??

Not too pretty admittedly, but the main feature of `mrtheshold` is the automatic threshold determination. If you're using predefined thresholds, you might as well use `mrcalc`...

Incidentally, this would be much easier with a more natural syntax like what I suggested in issue #27. This would then look like (assuming thresholds are 0 & 1):

```ShellSession
mrcalc '(a>0)*(a<1)*a' in.mif out.mif
```

Might be thinking about doing something about this?

Lestropie:
We discussed it a bit earlier today, tend to think the feature of `mrthreshold` is _applying_ thresholds, doesn't necessarily matter how those thresholds are determined, whether automatic or manual. If it's a regular enough operation I can see the `mrcalc` solution being a bit clunky, especially for those not versed with stack-based syntax unless we do #27. But I'm not pushing strongly either way.

draffelt:
I agree it's not that hard for most technical users to figure out, however it might not be that obvious for some RA users (with either syntax). 

I agree with Rob in that if someone needs to threshold an image, the first place they would look is mrthreshold. In most cases I use it with the -abs option (and don't use mrcalc -lt). The automatic stuff is just a bonus as far as I see it. Is there any reason why we can't have the ability to do it in either command?

jdtournier:
Yes, I agree it doesn't have to be either/or. If you find you use the command enough for this type of operation (or at least you would if it could do it), then by all means go for it, it shouldn't take much to implement. I agree it's probably the first place people will look... 

And I like the idea of `-above ` and `-below ` options rather than `-abs`, it's more explicit that way. Not sure how many scripts use `mrthreshold` with the `-abs ` option currently, but I wouldn't worry about this at the moment - we haven't officially released yet...

thijsdhollander:
[...]

PR #1636, opened 2019-07-03
mrthreshold: Rewrite

### LINKED ISSUE #1481: mrthreshold behaviour with exotic uses
Has some degree of overlap with #205 and #738.

Currently the input to the `-percentile` option is immediately converted to the equivalent behaviour as if the `-top` or `-bottom` option had been used (which select the N top-valued or bottom-valued voxels respectively). However it does this conversion based on the `voxel_count()` function, which assumes:

1. There are no non-finite values in the input image.

2. The `-ignorezero` option has not also been used.

3. The `-mask` option has not also been used (it seems the intention is that it only applies to calculation of the optimal threshold, but it's not inconceivable that a user may try to use this option elsewhere, e.g. calculating a percentile within a provided mask).

There's also the presence of the `-toppercent` and `-bottompercent` options in addition to the `-percentile` option, which are quite clearly intended to behave differently given the code branching: I believe that one is intended to calculate a singular threshold from inside the mask that is then applied to the entire image, whereas the other intends to only highlight voxels within the provided mask. But that's not really reflected in the documentation.

I can also see other areas of code where the behaviour in the presence of non-finite input data may not be as expected.

Given this, the discussion in #205, and the bug report in #738, I think the structure & interface of `mrthreshold` is in need of close scrutiny, with closer attention paid to the potential confounds listed above. Not going to put the time into fixing bugs before determining whether or not the actual command-line interface is to be changed, since that could plausibly have an influence on the target code structure, as well as whether there's a consensus on the intended behaviour for all possible usages.

Lestropie:
First attempt at restructuring following relevant discussion.

Code is drastically changed; the three command-line option groups defining the interface are fairly reflective of the code structure. Main pragmatic differences are:

- `-bottompercent` and `-toppercent` are removed since `-percentile` covers that functionality;

- The `-bottom` option will find the threshold that will result in removal of some number of voxels; i.e. if one wants to *retain* some bottom number of voxels, then `-invert` also needs to be provided *explicitly*. This is kind of the final step for entirely disentangling the `-invert` functionality from all other command-line options.

There are some finer-grained less-relevant differences:

- `-percentile` doesn't quite match the old behaviour perfectly (so tests currently fail), but I think this implementation is more mathematically precise;

- `-top` / `-bottom` will determine a threshold that will result in that number of voxels, rather than selecting that number of voxels; if there are multiple voxels with an identical value at the location where the threshold is being applied, this is identified explicitly and a warning is issued.

Have not yet addressed #738.

Have not done anything explicitly about 4D input images.

Single-threshold-per-invocation approach chosen in favour of #205.

**Edit**: Closes #1481.


Lestropie:
Flagged as ready for review.

-  Have added support for 4D inputs: will either determine a single threshold across all data (`-allvolumes` option), or treat each 3D volume individually. If `-mask` is used, this can be either 3D or 4D.

-  #738 should be fixed.

-  A longer description regarding the change to the `-top` / `bottom` behaviour.

   In the previous implementation, the command would sort the voxels by intensity, and select voxels from the start / end of this sorted list until the desired number of voxels was selected. This was an issue if there were multiple voxels in the image with identical values, as the command would somewhat arbitrarily select some number of those identically-valued voxels until the desired number of voxels was selected, arbitrarily omitting the other voxels despite having identical values.

   In this new implementation, the command seeks a threshold value that will result in the requested number of voxels being selected / omitted. That threshold is then applied, which results in the selection / omission of some number of voxels. If this resulting number differs from the number requested by the user, a warning is issued.

thijsdhollander:
Hey Rob,

Thanks for the rewrite of the command and the fix to the min-max initialisation.  One constructive comment though, about this:

> The -bottom option will find the threshold that will result in removal of some number of voxels; i.e. if one wants to retain some bottom number of voxels, then -invert also needs to be provided explicitly. This is kind of the final step for entirely disentangling the -invert functionality from all other command-line options.

After having given this a lot of thought, I do think this is a very confusing change though.  I've asked this to a few collaborators who use this very option, and they're unable to see the logic in the new definition of this particular option.  I see where you're probably coming from (I suppose the idea that a threshold by default is retaining things above the threshold value), but it jars with what's expected if you read the name of the option, especially in the presence of the "equivalent" `-top` option.  To put it like this maybe: you'd expect `-top` and `-bottom`, by virtue of they're "paired" names, to have the same description apart from the word "top" or "bottom".  With the proposed change, the description also swaps an "include" for an "exclude", which is the non-intuitive bit, I reckon.  Furthermore, in practice you'd kinda think the default behaviour gives you (or aims for, with the other new change) the number of voxels you're asking for, rather than excluding them for the one but not the other option.  The `-invert` would then swap the include for the exclude behaviour.

Apart from this in and of itself, I think it's also maybe quite dangerous to change the behaviour of such a basic functionality to now by default mean the opposite... But in any case, I'd think intuitively it makes slightly more sense to have both of them mean "to include the top/bottom number of voxels".  Finally, I reckon that would also be the more prevalent use case (e.g. select the top something FA voxels vs selecting e.g. the bottom something ADC values, etc...).  Just my 2 cents of course.  Trying to get some reviewing going. (I'd thought about this a few weeks ago, but forgot to write this down back then)

Lestropie:
The names "top" and "bottom" can't be fully disentangled from "invert"; so maybe it's the names of those options that is in fact the source of the problem. E.g. If they were "include" / "exclude", or "select" / "omit", or something similar, it might be clearer, and it then wouldn't be altering the operation of a pre-existing command-line option.

jdtournier:
I agree with @thijsdhollander here, the `-top` and `-bottom` options really should preserve their existing behaviour. I understand that this conflicts with the logic behind your option grouping, with the clear separation between threshold determination and application of said threshold, but maybe we can resolve this fairly easily. How about: 

#### Threshold determination

- `-abs value`
- `-percentile value`
- **`-rank N`** or `-nth N` or `-voxel N` or ... (up for discussion)

#### Modifiers for threshold determination

- `-allvolumes`
- `-ignorezero`
- `-mask image`
- **`-decreasing`** : rank voxels in order of decreasing intensity. This should affects `-percentile` and `-rank` (or whatever we end up calling it).

#### Application of threshold

- `-invert`
- `-nan`

#### Convenience options

- **`-top N`** : produce mask consisting of N highest intensity voxels (equivalent to `-decreasing -rank N`)
- **`-bottom N`** : produce mask consisting of N lowest intensity voxels (equivalent to `-rank N -invert`)


This should hopefully be a minor modification of what you have, preserve the logic of your option grouping and the way the application works, but also provide continuity with existing behaviour and the convenience of the `-top` and `-bottom` options. 

I'd also keep option group titles short - the proposed ones are way too long. Not sure what I've got is necessarily final, but you get the idea... 

What d'ya think?

Lestropie:
> - `-top N` : produce mask consisting of N highest intensity voxels (equivalent to `-decreasing -rank N`)
> - `-bottom N` : produce mask consisting of N lowest intensity voxels (equivalent to `-rank N -invert`)

These are both wrong:

- `-top N` would be no different to `-rank N` (but can just be an alias of such for backwards compatibility);

- `-bottom N` would be equivalent to `-rank N -decreasing -invert`, assuming that `-decreasing` is intended to affect the determination but not application of the threshold.

... unless you are interpreting `N` as being "the index of the sorted array used to determine the threshold", which I'm pretty sure would be counter-intuitive for non-programmers; most would I think expect "`-rank N`" to just select `N` voxels for them, and the requisite help text to convince them otherwise would likely be ugly.

The problem is that `-decreasing` and `-invert` are confusing to disentangle, even though they are in fact separate. E.g. You can reverse the sort operation such that the threshold "isolates" the bottom 10 voxels instead of the top 10 voxels, but application of the threshold will then result in selection of all but those bottom 10 voxels; you still need the `-invert` option to select voxels below rather than above the threshold. Same with `-percentile 90 -decreasing` being equivalent to `-percentile 10`; the former does *not* invoke `-invert` despite one's instincts.

Renaming `-invert` may assist here, in order to prevent `-decreasing -invert` feeling like a double negative. E.g. `-below` to select voxels below the determined threshold rather than above it. Or potentially expand even further and have `-operator {lt,le,ge,gt}` to better deal with equalities?

-----

What I think I'm converging towards in this writing process is:

1. Replace `-invert` with `-operator` as above;

2. Change `-bottom` to result in *selection* rather than *omission* of bottom-valued voxels (as is previous behaviour); but state in the help text of `-bottom` that it additionally implicitly changes the default operator to `<=` unless otherwise specified.

This would shift the onus slightly for disentanglement of concepts from the interface back to the code, but shouldn't be too difficult.

Lestropie:
Okay, that works; and tests passed by effectively reverting the tests and leaving the test data unchanged.

thijsdhollander:
Hmm, I think *generally* @jdtournier's proposal actually went quite in the right direction; but just a few bits weren't maybe 100% thought through in the initial description of it.  There was indeed an (accidental, obviously) mistake with the *exact* description of `top` and `bottom` in function of the other things; `top` obviously, but `bottom` is indeed a bit more subtle due to the exact behaviour for values that match the threshold.  But that's in and of itself a good showcase of the value of `top` and `bottom` indeed: they're simple, uncomplicated for most general use cases, and that's a good thing.  The idea of putting those in convenience category exactly reflects that.

The `-rank N` or `-voxel N` (another alternative for something along those lines: `-count N`) was also not a bad idea, I reckon.  And yes, fully aware that that one then is equivalent to `-top`, but that's ok I think.  Not only is it preferable to retain backwards compatibility for such a "low level" type of working horse command, `-top` and `-bottom` are a very convenient and easy to understand pair of options that have the benefit that their names are (have always been for many users!) perfectly self-explanatory; that's been my actual experience with those users.

Similarly for `-invert`: it's quite simple and clear what it does, it's been there before, it worked well for users before.  If it works, don't fix it, I reckon.  But yes, with the new categories, care needs to be taken where to put it and how to describe it, for people who read the whole documentation and are thus "dealing" with those categories.

Finally, the (what I think is the intended) idea of `-decreasing` didn't seem bad to me as well, but again some extra consideration on where to put it (category wise) and how to describe it is indeed needed; and that wasn't 100% done yet in that proposal maybe.

Now, on the other hand, the `-operator` proposal is fair as well, I think; although I'd (subjectively) argue it's a bit more cryptic though, and it does require you to then go and check the "operators", again see them strongly in the logic of the (internal!) logic and operation of the command, and then put it all together to sometimes achieve something that, I'd think, might or should have been more simple.  Not against it though; the definitions are quite clear for sure.

But in the end with the current most recent changes, I'd also still keep `-invert` in though, for several reasons: it's still pretty clear what it does; it'll remain a shorter and more obvious way to do exactly that rather than achieve it via a particular `-operator` for many people; and finally of course backwards compatibility.  There's no strong reason I can see to really need or want to get rid of it, even together with the latest changes and proposal.

Lestropie:
The issue is that `-bottom` is not precisely equivalent to using `-top` with the threshold based on the (`voxel_count - N`)th voxel in the sorted list and then also using `-invert`: the particular voxel that is utilised in determining that threshold needs to be retained despite containing a value equivalent to the threshold (which would normally be "accepted" by an intrinsic `>=` operation, but would then be "inverted"). Prior to c398151 there was a latent boolean "`equal_as_above`" that accounted for this, by essentially making use of `-bottom` do a kind of internal switch between `>=` and `>`; but even then that code was kinda clunky due to the ambiguity of the definition of "above" in the presence of the use of `-invert`. The `-decreasing` suggestion has its own problems, in terms of knowing precisely how the command will behave right where values are equal-to or not-quite-equal-to the threshold, the relationship between `-decreasing` and `-invert` / definition of the "`-rank`" option, and what `-bottom -invert` would mean. The `-operator` solution solves all ambiguities.

jdtournier:
OK, there's a bit more subtlety to the issue than my suggestion gave credit for. No matter, I like the latest set of suggestions, they do convey all the functionality in a way that's least open to misinterpretation, and they maintain backward compatibility with the `-top` and `-bottom` options. Happy to go with that.

cmd/mrthreshold.cpp
```diff
@@ -14,222 +14,442 @@
...
+Image<bool> get_mask (const Image<value_type>& in)
+{
+  Image<bool> mask;
+  auto opt = get_options ("mask");
   if (opt.size()) {
-    percentile = opt[0][0];
-    ++nopt;
+    mask = Image<bool>::open (opt[0][0]);
+    if (mask.ndim() > in.ndim())
+      throw Exception ("Cannot use mask image with more axes than input image");
```
jdtournier: This won't allow e.g. using a 4D mask with a single volume (effectively 3D) with a genuine 3D image. Is this intentional? Maybe we can relax this a bit?

cmd/mrthreshold.cpp
```diff
@@ -14,222 +14,442 @@
...
+                             const size_t max_axis,
+                             const bool ignore_zero)
+{
+  vector<value_type> data;
+  data.reserve (voxel_count (in, 0, max_axis));
+  if (mask.valid()) {
+    Adapter::Replicate<Image<bool>> mask_replicate (mask, in);
+    if (ignore_zero) {
+      for (auto l = Loop(in, 0, max_axis) (in, mask_replicate); l; ++l) {
+        if (mask_replicate.value() && in.value() != 0.0f)
```
jdtournier: Any reason we don't check if value is finite when running with `-ignore_zero`? Even if this is interpreted as using zeros to indicate NaN, if there are genuine NaNs in the image, there's no way they can be handled appropriately here, right?

cmd/mrthreshold.cpp
Lestropie: 05c736e permits that case.

cmd/mrthreshold.cpp
Lestropie: Good catch; see 05c736e.

Also revised to use `!std::isnan()` rather than `std::isfinite()`, since technically infinities can be sorted.

Lestropie:
Would `-comparison` be more user-friendly than `-operator`?

jdtournier:
> Would -comparison be more user-friendly than -operator?

Happy with either, personally, though `-comparison` is probably a tiny bit more easily interpreted.

thijsdhollander:
> though -comparison is probably a tiny bit more easily interpreted

Yes, I tend to agree with this one.

> maintain backward compatibility with the -top and -bottom options

It'd be nice to retain backward compatibility with `-invert` as well though.

thijsdhollander:
:+1: Yeah, the docs do read better with "comparison"; so makes sense.  I'm pretty sure the command is complete now: I've brainstormed for a minute or two and can't see anything obvious missing in terms of thresholding capabilities.

PR #1637, opened 2019-07-09
mrmetric: fix cross correlation metric looping

maxpietsch:
fixes the cross-correlation metric, updated mrmetric's command description.
reported and discussed [here]( http://community.mrtrix.org/t/cross-correlation-and-mrmetric/2648/9)

### DIFF
--- a/cmd/mrmetric.cpp
+++ b/cmd/mrmetric.cpp
@@ -172,7 +172,7 @@ void usage ()
   SYNOPSIS = "Computes a dissimilarity metric between two images";
 
   DESCRIPTION
-  + "Currently only the mean squared difference is implemented.";
+  + "Currently only the mean squared difference is fully implemented.";
 
   ARGUMENTS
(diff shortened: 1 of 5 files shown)


maxpietsch:
There is something wrong with the macos GUI build: https://travis-ci.org/MRtrix3/mrtrix3/jobs/556216368

jdtournier:
Guess the version of Qt used for the macOS build is a bit more up to date than on the Linux side... Should be an easy fix if you're happy to give it a shot?

maxpietsch:
Yes, QFontMetrics::[width](https://doc.qt.io/qt-5/qfontmetrics-obsolete.html#width) is now obsolete but QFontMetrics::[horizontalAdvance](https://doc.qt.io/qt-5/qfontmetrics.html#horizontalAdvance) is only available in qt5.11. I added macros to check the QT version for backwards compatibility.

PR #1638, opened 2019-07-11
mrview fix volume group navigation

### LINKED ISSUE #1369: mrview: view tool syncs volume and volume groups
Once the view tool has been opened, volumes and volume groups are linked. shift + left or right changes  both dimensions. Appears to not happen with RC2.

jdtournier:
Fixes #1369

### DIFF
--- a/cmd/mrmetric.cpp
+++ b/cmd/mrmetric.cpp
@@ -173,7 +173,7 @@ void usage ()
   SYNOPSIS = "Computes a dissimilarity metric between two images";
 
   DESCRIPTION
-  + "Currently only the mean squared difference is implemented.";
+  + "Currently only the mean squared difference is fully implemented.";
 
   ARGUMENTS
   + Argument ("image1", "the first input image.").type_image_in ()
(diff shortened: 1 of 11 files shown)


thijsdhollander:
Not 100% sure it was introduced by these changes, but even if not, it's closely related; so just mentioning it here to not unnecessarily create a new issue:

There's a scenario that crashes out into a segmentation fault, as well as one other (possibly) related oddity in the "view options" tool.  2 scenarios to reproduce:

1. The oddity: open a DWI dataset (or maybe any multi-volume image; no volume _groups_ though, only 4D); open the "view options" tool; go to the next volume either by keyboard (right arrow) or using the "Index:" field in the GUI; observe the "Group:" field change from 0 to 99 (even though it can't be interacted with).

2. The segfault: open a DWI dataset as in the other scenario; now _first_ go to the next volume by keyboard (right arrow); only then open the "view options" tool (unlike the above scenario, the "Group:" field will still read 0 here); now either hit the right arrow again or use the "Index:" field in the GUI to go to another (next) volume; this leads at that point to a crash / segmentation fault.

The above using current `dev`, i.e. with the above changes included.  I haven't checked whether this particular issue was already there before though.

thijsdhollander:
...so just to "generalise" the segfault scenario: this can be reproduced by navigating (e.g. left/right arrow buttons) any number of volumes before opening the "view options" tool, then opening the "view options" tool, then navigating to another volume; i.e. a scenario that can easily happen (open a dataset, inspect a bit by browsing volumes, and at that point it's almost "bound to happen" if you work with the view options later on).  I just checked, and it also still reproduces if the view tool is closed in the process; so browse some volumes, open and optionally close the view tool, try to navigate to another volume and segfault.

Issue #1580, opened 2019-03-13
Statistical inference: Add strong FWE control


Lestropie:
Currently, when multiple hypotheses are tested in a single execution of a statistical inference command (either using the `-negative` option currently, or more generally in #1543), or when the command is executed multiple times to address multiple hypotheses, each hypothesis has its own null distribution to which the default permutation is compared. This provides only weak FWE control, in that the probability of a false positive finding in any particular experiment is e.g. 0.05, but the more hypotheses one tests, the probability of finding *any* false positive *across* the set of hypotheses tested exceeds the intended *p*-value threshold.

A command-line option to activate strong FWE control would construct a *single* null distribution for all hypotheses, by sampling the maximal enhanced statistic from any location in the template from any hypothesis.

This could still be used even if different hypotheses require different input data (e.g. FD / FC / FDC), by concatenating all of the input data and using appropriate zero-filling in the design matrix to essentially "separate" the multiple hypotheses into blocks.

It would also likely only behave as intended if adequate non-stationarity correction were performed. Without this, one hypothesis may dominate contributions toward the null distribution, making it increasingly unlikely to be capable of reporting a significant effect for any other hypothesis where the raw numbers are smaller in magnitude (for whatever reason).

Lestropie:
Closed by 3f629e8; to be merged in #1639 followed by #1543.

Issue #1068, opened 2017-07-22
fsl.py (5ttgen) does not pick up combined first segmentation file in fsl v5.09


AlistairPerry:
I'm running `5ttgen fsl` on fsl version 5.09 (with eddy patch http://fsl.fmrib.ox.ac.uk/fsldownloads/patches/eddy-patch-fsl-5.0.9/centos6/eddy_openmp).

I'm running into the following error:

```
File "/software/mrtrix3/mrtrix3-20170707/lib/mrtrix3/_5ttgen/fsl.py", line 181, in execute
    app.error('FSL FIRST has failed; not all structures were segmented successfully (check ' + path.toTemp('first.logs', False) + ')')
NameError: global name 'path' is not defined
```

`fsl.py` on line 174 looks for file `first_all_none_firstseg.nii.gz` in the temp 5tt directory.

With the above `fsl `version, however, the output filename of the combined first segmentation is `first_all_fast_firstseg.nii.gz`

Would it be more appropriate for this line to look for filenames matching the string *firstseg.nii.gz?

Lestropie:
Hey Alistair, thanks for the report.

> `NameError: global name 'path' is not defined`

That's a missing `from mrtrix3 import path` call; it should have given you a nicely-formatted terminal message...

> With the above fslversion, however, the output filename of the combined first segmentation is `first_all_fast_firstseg.nii.gz`

OK, various thoughts:

- I suspect different FSL versions are using different default settings for the boundary correction (see [here](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FIRST/UserGuide), `run_first_all` "output" bullet). Seems there is a manual control over what is done here using the `-m` option; however they claim their default is "auto", which uses different settings for different structures, yet this is not being reflected in the name of this combined image...

- I'm pretty sure the `run_first_all` script has an error in it: It seems to be setting the name of the output combined image based on the boundary correction method of *the last structure in the list*, rather than the overall setting (see "`methname=$btype ;`"; should be based on "`$bmeth`")... So explicitly using `-m auto` is unlikely to solve the issue.

- The boundary correction method *appears* to be image-based, and does not manipulate the surface mesh representations in any way. Therefore, given `5ttgen fsl` ignores FIRST's image outputs and only accesses the mesh files, using "`run_first_all -m none`" may be the easiest solution to this particular issue, and will avoid some unnecessary computations.

[...]

PR #1639, opened 2019-07-16
Statistical inference: Use Z-statistic

Lestropie:
Extension of #1543, but listing on its own in order to be better able to describe the change and link to such.

@maxpietsch provided me with [this link](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;eb9b9693.1712) in relation to an issue I've had surrounding per-fixel per-subject design matrix exclusion. Initial tests suggest that it may not have fully resolved that specific issue, but I think it's a worthwhile change to be making more broadly.

With this change, both t-tests and F-tests are immediately converted to Z-statistics; it is these Z-statistics that are then fed as the input to the statistical enhancement algorithm.

This provides a number of benefits:

- Empirical behaviour of statistical enhancement algorithms should be consistent across different cohort sizes.

- Statistic becomes pivotal across hypotheses, which is necessary for strong FWE control (#1580) in the absence of non-stationarity correction.

- For F-tests:

    - Any elements with p>=0.5 (z<=0.0) do not contribute to statistical enhancement.

    - Distribution of statistical measure is no different to that of a t-statistic.

    This is a much more elegant solution to the issue described in #1068, and also results in much faster CFE execution for F-tests due to decreased cluster size at small values of *h* (and doesn't require the awkward gymnastics of internally storing `sqrt(F)` which is what I did in #1068).

- Statistic remains pivotal even when the rank of the design matrix varies between those elements tested (e.g. if subject data are differentially cropped due to FoV & subject motion, such that valid diffusion data are not available for all fixels for all subjects). This *should* improve stationarity of statistical inference when per-subject per-fixel exclusion is used, though my initial tests suggest it wasn't a panacea to the particular issue I was observing.

Change has minimal effect on execution time: lookup tables for direct t->z and F->z conversion are constructed on demand based on the relevant full analytic transformations (with some mathematical trickery to avoid floating-point precision issues), and Hermite interpolation is performed on this table. Total lookup-based transformation precision is approx. that of single-precision floating-point, i.e. 6 significant figures.

Enhancement parameters should not I think need to be modified: the experiments used to determine those parameters could simply be re-interpreted as having been done based on simulated-z rather than simulated-t effects.

There's a unit test added in there specifically for the inverse error function calculations; I also have a standalone command that has been used to verify the operation of the lookup tables, which I might try to coax into a unit test.


Lestropie:
One catch found here: `Eigen::betainc()` is not available in Eigen 3.2, only in Eigen 3.3. So either the minimum installation requirements will need to be bumped, or a custom implementation of that function will be required, or an environment variable will need to be used to toggle use of the Z-transformation depending on detection of the availability of that function at config time.

Lestropie:
7820da1 adds support for Eigen 3.2 (should probably test both Eigen versions as part of CI...). Currently the Eigen function will be used if available, and this alternative used otherwise (the outcomes of F->Z transformations, which incorporates this calculation, are identical to at least 6 significant figures).

jdtournier:
Alternatively, we could simply add Eigen as a git submodule, now that there's an [officially maintained git repo](https://github.com/eigenteam/eigen-git-mirror)...?

Lestropie:
While that's definitely an option, I think it's a packaging / distribution / installation question that extends well beyond the scope of this PR, so probably needs to be had separately. I kinda need to merge this into `stats_enhancements` so that I can resolve with #1662.

PR #1642, opened 2019-07-17
mrview: fix bug introduced in #1638

jdtournier:
This should fix #1640...

PR #1643, opened 2019-07-18
Resolve with latest TravisCI updates

Lestropie:
Cherry-picking commits from #1642.

Given TravisCI failures will preclude any bug fixes from being merged to `master`, I think the preferable approach in such a circumstance is:

1. Get CI working on `master`;

2. Merge latest updates on `master` into `dev`;

3. Pull requests merging into `dev` remain focused on the issue at hand and don't get conflated with CI experimentation / fixes.

4. Add a "`CI`" label in GitHub so that issues / PRs specific to continuous integration are flagged as such, and e.g. can be ignored when generating changelogs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
       - llvm-toolchain-xenial-8
     packages:
       - g++-9
-      - clang-8.0
+      - clang-8
       - zlib1g-dev
       - libqt4-opengl-dev
       - python3
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ install:
     fi
   - |
     if [[ "${test}" == "build" || "${test}" == "run" ]]; then
-        export EIGEN_CFLAGS="-isystem `pwd`/../eigen";
+        export EIGEN_CFLAGS="-idirafter `pwd`/../eigen";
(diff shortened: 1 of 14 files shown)


Lestropie:
bba9b51 works on GCC9 on my machine. There was a particular issue [here](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/src/gui/opengl/shader.h#L140-L145), which looks suspiciously like a move assignment rather than a copy assignment; it's possible that the implicit copy constructor of that class might lead to unwanted behaviour somewhere like [here](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/src/gui/mrview/mode/ortho.h#L37), as multiple instances of the `Program` class could contain the same index value, and any one of them could call `gl::DeleteProgram()` on it, invalidating the underlying data of the other instances.

Basically the changes are to prevent copy-construction or copy-assignment of that `Program` class, and to use `NOMEMALIGN` to make sure that the `std::vector` specialisation isn't invoked for duplicates of the `Projection` class (as definition of a specialisation of `std::vector` requires that the underlying class has a copy constructor, and therefore you can't have the class or any of its derivatives forbidding the copy constructor). None of those classes have underlying Eigen members so I don't think it should cause any issues.

Lestropie:
I can't review since I listed the PR; need somebody else to click the OK button.

Issue #842, opened 2016-11-30
detect attempts to overwrite input data


jdtournier:
As discussed on [this community forum thread](http://community.mrtrix.org/t/unexpected-mrtransform-behaviour/585?u=jdtournier).

I don't think there's any simple portable way of querying whether a file we've just opened read-write was already being accessed read-only - that would be the ideal scenario. But I do think we can keep track of this ourselves relatively easily, since 99% of file access will be via the [`File::MMap`](https://github.com/MRtrix3/mrtrix3/blob/master/lib/file/mmap.h#L31) class - at least for images. That uses a standard C `open()` call, which returns an integer file descriptor. I _think_ this call is supposed to return the same fd when opening the same file again. So if we keep a list of file descriptors that have previously been opened read-only, we can check when opening a file read-write whether we had already opened it read-only. We could also check whether we had already opened it read-write, but that's less likely to be a problem.

This should sort out _most_ issues with trying to write to one of our inputs, although there will still need to be some changes to some of the image formats that open the file using C++ IO streams (e.g. the MRtrix formats), since it's basically impossible to retrieve the associated fd from that - at least not in a portable way (see [this page](http://www.ginac.de/~kreckel/fileno/) to get an idea of how complicated that might turn out to be...). The simplest thing to do in these cases will be to enforce that _all_ file access is done via standard C calls that operate on file descriptors (not even `FILE*` pointers)... 

An alternative approach which might actually work more seamlessly is to rely on the device ID & inode number for the file as retrieved via the `stat()` call - performed in the `File::MMap` constructor. That combination _should_ be unique to that specific file (have to check what happens with hard or soft links), so all we'd need to do is keep track of those entries that have already been opened read-only - something that should be achievable even if using C++ streams. But we'd still need to catch any uses of `std::ifstream` and make sure they record these bits of information... 

So not trivial, but potentially doable...

jdtournier:
Worth checking whether using a `MMAP_PRIVATE` mapping for the read-only input file fixes the issue - at least on platforms where `mmap()` is available. Will need to check whether Windows/MSYS2 support equivalent behaviour.

jdtournier:
OK, I've had a look into this. Turns out the `MAP_PRIVATE` isn't going to work, since according to the [mmap() man page](http://man7.org/linux/man-pages/man2/mmap.2.html): 

> MAP_PRIVATE
[...]

PR #1646, opened 2019-07-18
Allow command inputs to be overwritten safely

jdtournier:
Addresses #842. 

Turns out it's a simple matter of deleting the file rather than opening with `O_TRUNC`. If the file is already opened, the kernel ensures it remains usable by processes that hold its file descriptor.

This may not work on Windows since it tends not to allow deletion of open files (I've yet to test this). But while this doesn't allow writing to one of the inputs, it should at least catch and prevent users from doing so inadvertently, so that should still be a good thing...

### DIFF
--- a/core/file/utils.h
+++ b/core/file/utils.h
@@ -119,6 +119,7 @@ namespace MR
         return __tmpfile_prefix;
       }
 
+
       /* Config file options listed here so that they can be scraped by
        * generate_user_docs.sh and added to the list of config file options in
        * the documentation without modifying the script to read from the scripts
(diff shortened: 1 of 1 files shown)


maxpietsch:
It works on ubuntu and macOS but has the side-effect of converting symlinks and hard links to new inodes. It might be worth warning about that or in the future [parse](https://en.cppreference.com/w/cpp/filesystem/status) the link once we switch to c++17?

Lestropie:
> This may not work on Windows since it tends not to allow deletion of open files (I've yet to test this)

Correct:

`mrconvert: [ERROR] error deleting file "duplicate.mif": Permission denied`

Issue #1513, opened 2018-12-12
ENH: mrview overlay: allow volume selection for >3D images


maxpietsch:
Currently, the volume field is greyed out if a 4D image is loaded. Also, it would be good to add support for selecting the volume group.

PR #1649, opened 2019-07-22
mrview: overlay n dimensional data

maxpietsch:
I ported the changes in 33dc1403b869f8c0de0eca32 to add support for n-dimensional images in the overlay tool in mrview (fixes issue #1513).

Compared to the previous behaviour, the volume index spin boxes are removed when multiple images are selected as they do not apply to all images. I kept the bounding box though as removing that can change the FOV due to resizing of the overlay tool which I find distracting especially for screenshots. In place of the empty Volume indices box a text is shown indicating the reason why volumes can not be changed.

10D image selected:
image

3D image selected:
image

Multi-selection:
image

### DIFF
--- a/src/gui/dialog/progress.cpp
+++ b/src/gui/dialog/progress.cpp
@@ -42,7 +42,7 @@ namespace MR
             p.data = new Timer;
           }
           else if (reinterpret_cast<Timer*>(p.data)->elapsed() > 1.0) {
-            Context::Grab context;
+            GL::Context::Grab context;
             if (!progress_dialog) {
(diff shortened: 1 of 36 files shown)


jdtournier:
OK, I've gone and reworked a lot of the OpenGL context check routines to try to hunt down a failed context assert when closing overlay images (apparently this was already an issue on `dev`). It's a big changeset, but basically I'm replacing macros with inline functions, and adding a couple more checks for the main OpenGL-related classes. Seems to fix the issue at any rate...

PR #1658, opened 2019-07-26
tckgen: Recursive back-tracking

Lestropie:
I just remembered that this code was sitting here from almost 2 years ago; creating a draft PR to remind myself to revisit and test, see whether I can get out an objective measurable change (good or bad).

I've never been content with the ACT back-tracking algorithm. For a given erroneous termination point, it just kind of slowly increments the length of back-tracking, trying to produce a track that's longer than it was when the erroneous termination was first encountered. But there's a floating free parameter in there (number of attempts before incrementing the back-tracking length), and the truncated streamlines can encounter their own issues when trying to traverse the difficult area. The code is also not particularly elegant.

What this change did is re-implement back-tracking as a *recursive algorithm*. So even those streamlines that are the result of truncation by the back-tracking mechanism may themselves undergo back-tracking (and so on and so on). The code is actually kinda cool.

From memory a thread could every now and then noticeably spend time on a single streamline, likely due to the size of the stack exploding. But it might theoretically do better in tracking particularly difficult regions.


Lestropie:
Have tested this a little on up-to-date data & code, including the SH calculation fixes & revised FOD amplitude cutoff. Experimental results unfortunately don't really give a strong indication in any particular direction. In general the results are very similar for both backtracking algorithms; one could maybe make the case that with recursive back-tracking, the SIFT model residuals are *slightly* smoother, i.e. large single-voxel errors can be slightly blurred out.

For dynamic seeding:
- About 10% slower for same number of streamlines
- Total streamlines density (sum of lengths) about the same
- Cost function both before and after SIFT about the same
- Number of streamlines after SIFT about the same
- Fraction of seeds producing accepted streamlines fractionally higher

For GMWMI seeding:
- About 10% faster for same number of streamlines
- About 20% less streamlines density
- Cost function before SIFT fractionally lower, after SIFT about the same
- Number of streamlines after SIFT increased by 10%, but total streamlines density decreased by 10%
- Fraction of seeds producing accepted streamlines increased by 12%

-----

Pro for the recursive algorithm: It (basically) removes the free parameter in the current implementation, being the number of propagation failures at a given length before the truncation length is increased; for the recursive algorithm this is logically set to 1, and in fact the command can basically freeze if this exceeds 1.

Potential con: If back-tracking were to be re-implemented for bootstrap-based probabilistic algorithms, this would necessitate storing bootstrapped model data per recursive call, rather than per thread. With the current implementation, one would (theoretically) just truncate the streamline, remove any bootstrapped model data corresponding to the truncated segment only, and re-generate data for those voxels whenever re-intersected, repeatedly deleting and re-generating such data as the streamline continuously truncates and re-tracks. For recursive back-tracking, *every* recursive call would need to *explicitly duplicate* all bootstrapped model data, which may be slow.

-----

While I really like the elegance of this solution, I'm struggling to derive a solid empirical justification for the change. It might shine better in difficult targeted tracking cases; it'd probably also do better on really difficult phantoms, but don't particularly care about that. Opinions welcome: I'll swallow discarding it if I have to.

PR #1659, opened 2019-07-30
mrcalc: Updates to complex data handling

Lestropie:
Mostly needed to add the `-polar` option to support some local data manipulation; but tweaked a few other things while I was in there (description in 0cadc52).

### DIFF
--- a/cmd/mrcalc.cpp
+++ b/cmd/mrcalc.cpp
@@ -34,13 +34,13 @@
   + OptionGroup (TITLE)
 
 #define UNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
-  + Option (#OPTION, DESCRIPTION).allow_multiple()
+  + Option (#OPTION, FEEDBACK " : " DESCRIPTION).allow_multiple()
 
 #define BINARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
-  + Option (#OPTION, DESCRIPTION).allow_multiple()
+  + Option (#OPTION, FEEDBACK " : " DESCRIPTION).allow_multiple()
 
 #define TERNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
-  + Option (#OPTION, DESCRIPTION).allow_multiple()
+  + Option (#OPTION, FEEDBACK " : " DESCRIPTION).allow_multiple()
 
 # elif SECTION == 2 // code section
 
--- a/cmd/mrcalc.cpp
+++ b/cmd/mrcalc.cpp
@@ -14,6 +14,158 @@
  * For more details, see http://www.mrtrix.org/.
  */
 
+
+
+
+/**********************************************************************
+  CONVENIENCE MACROS:
+ **********************************************************************/
+
+#ifdef SECTION
+
+#undef SECTION_TITLE
+#undef UNARY_OP
+#undef BINARY_OP
+#undef TERNARY_OP
+
+# if SECTION == 1 // usage section
+
+#define SECTION_TITLE(TITLE) \
+  + OptionGroup (TITLE)
+
+#define UNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  + Option (#OPTION, DESCRIPTION).allow_multiple()
+
+#define BINARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  + Option (#OPTION, DESCRIPTION).allow_multiple()
+
+#define TERNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  + Option (#OPTION, DESCRIPTION).allow_multiple()
+
+# elif SECTION == 2 // code section
+
+#define SECTION_TITLE(TITLE)
+
+#define UNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  class Op_##OPTION : public OpUnary { NOMEMALIGN \
+    public: \
+      Op_##OPTION () : OpUnary (FEEDBACK, FLAGS & COMPLEX_MAPS_TO_REAL, FLAGS & REAL_MAPS_TO_COMPLEX) { } \
+      complex_type R (real_type v) const REAL_OPERATION \
+      complex_type Z (complex_type v) const COMPLEX_OPERATION \
+  };
+
+#define BINARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  class Op_##OPTION : public OpBinary { NOMEMALIGN \
+    public: \
+      Op_##OPTION () : OpBinary (FEEDBACK, FLAGS & COMPLEX_MAPS_TO_REAL, FLAGS & REAL_MAPS_TO_COMPLEX) { } \
+      complex_type R (real_type a, real_type b) const REAL_OPERATION \
+      complex_type Z (complex_type a, complex_type b) const COMPLEX_OPERATION \
+  };
+
+#define TERNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
+  class Op_##OPTION : public OpTernary { NOMEMALIGN \
+    public: \
+      Op_##OPTION () : OpTernary (FEEDBACK, FLAGS & COMPLEX_MAPS_TO_REAL, FLAGS & REAL_MAPS_TO_COMPLEX) { } \
+      complex_type R (real_type a, real_type b, real_type c) const REAL_OPERATION \
+      complex_type Z (complex_type a, complex_type b, complex_type c) const COMPLEX_OPERATION \
+  };
+
+# elif SECTION == 3 // parsing section
+
+#define SECTION_TITLE(TITLE)
+
+#define UNARY_OP(OPTION,FEEDBACK,FLAGS,DESCRIPTION,REAL_OPERATION,COMPLEX_OPERATION) \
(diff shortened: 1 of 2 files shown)


jdtournier:
Ok, I've also tweaked a few things... 

Main change is centralising the definition of operators using a bit of macro magic. Each operator is now defined entirely on one line, including its usage, code and parsing. This should reduce the barrier to adding new operators as needed. 

Second change is to insert the *feedback template string* (for want of a better term) into the option description itself. This serves two purposes: it mirrors the output shown to the user when the command is running; and makes it clear how many operands each operator will use. The reason this matters is that the current option grouping was based purely on the number of operands consumed (i.e. unary, binary, ternary), which makes any finer subgrouping problematic. With this simple addition (prompted by @dchristiaens), we can group on a different, more appropriate basis - which brings me to the final change:

Operators are now grouped by functionality, which I think makes the help page much easier to navigate, and hopefully better addresses the concerns that prompted the addition of option groups for complex numbers.

Still a bit of work to do on it, but pushed it anyway to get feedback on whether there is support for this approach...

Lestropie:
I still find the presence of `-phase` but absence of `-magnitude` in the complex data section really counter-intuitive and off-putting :man_shrugging:

For a non-programming user, the fact that the absolute value of a real number and the magnitude of a complex number are both achieved by calling `std::abs()` is totally irrelevant. At which point the only similarity between "`real_data.mif -abs`" and "`complex_data.mif -magnitude`" is that they have the same symbol in the equation printout ("`|  |`"). So recognising that the magnitude of a complex number can be obtained using `-abs` may take other users longer than it took me, even after adding the word "magnitude" to the option string of the `-abs` option.

Would having `-abs` operate on real data only, and `-magnitude` operate on complex data only, really be that much of a loss?

thijsdhollander:
> Operators are now grouped by functionality, which I think makes the help page much easier to navigate, 

Yes, that is useful.

> I still find the presence of -phase but absence of -magnitude in the complex data section really counter-intuitive and off-putting

Yep, I agree.  I haven't read the whole discussion or checked all code or changes, but slightly like with the `mrmath` / `mrstats` discussion, I have the feeling I'm missing the point of the issue; as in: I'm not seeing any issues to begin with... real numbers are complex numbers (with zero complex part) too, right?  So all that applies to complex numbers, naturally works on real numbers too.  So yeah, in this case, the magnitude becomes the thing also commonly known as the absolute value for real numbers specifically.  That, in my mind at least, provides the straightforward answer to what seems to be the issue (2 cents starts here): both are valid and existing operators, so provide them both in their sensible categories.  `-abs` in one of those more general categories (wherever you think it belongs) among some other stuff that works on real numbers, and `-magnitude` in some category dedicated to complex numbers (if that's how you like to categorise).  I'd just have them do (behave) the same thing then (yeah, gasp, duplicate functionality).  Or alternatively, along the lines of what Rob suggests: have `-abs` operate on real data only.  I'm not sure it's that important though that `-magnitude` operates on "complex data only", in the sense that the weird bit here is in fact "complex data only", as real data is complex data too (in the sense of where my argument starts somewhere earlier on in this post).  I mean, it's not "wrong" to compute the magnitude of a real number, right?  It's just a synonym of the absolute value, for the case of a real number...?

Can't be careful enough, so all the above is just 2 cents, and maybe I missed the issue entirely (?).  If it happens to be a case of the latter, please just ignore and gently walk away from the post.

Lestropie:
Rather than of `-abs` for real only and `-magnitude` for complex only, they could instead literally be synonyms: one `Operator` class, but two different option strings spawning an instance of such.

thijsdhollander:
> Rather than of -abs for real only and -magnitude for complex only, they could instead literally be synonyms: one Operator class, but two different option strings spawning an instance of such.

Yep, 100% agreed.  Not only the most elegant, but also an entirely "correct" solution, I'd argue.  Having the 2 options then each sit in their own "contextually sensible" category will just help with discoverability.

jdtournier:
Glad there's agreement that the new section layout is better. I'll carry on with that and get I finished off.

Regarding the addition of `-magnitude`: I don't have enormously strong feelings about it, but I personally don't like the idea much. I'd be happy enough for it to be a synonym, and documented as such, but I'd still consider it to be entirely redundant. Main reason is that of the current range of operators, only `-round`, `-ceil`, `-floor`, the comparison operators (`-gt`, `-lt`, `-ge`, `-le`), and `-min` & `-max` _don't_ handle complex data. All other operators are defined for real and complex inputs. We're not arguing to have additional operators for `-neg`, `-sqrt`, or any others. The only reason that I can see why it makes sense to add the `-magnitude` option specifically in the complex number section is purely because `-phase` exists in that section and its natural counterpart `-abs` (or `-magnitude`) is conspicuous by its absence from the same section. But I think we can trivially counter that by documenting the availability of the `-abs` option in the description of the `-phase` operator, and clarifying that `-abs` handles both real and complex inputs (which you'll note is already the case with the latest set of changes).

More generally, `abs` is widely used in different contexts to handle both real and complex numbers (MATLAB, C++, python, ...), and is I think the natural and logical choice. Yes, we can make life a touch easier for users not familiar with how programming languages handle these things, but it feels like overkill when all we need is a bit of additional documentation.

As always, just my 2 cents.

Lestropie:
> But I think we can trivially counter that by documenting the availability of the `-abs` option in the description of the `-phase` operator

That would do it actually. Describing that `-abs` works on complex data in the `-abs` help text gets kind of lost; whereas doing so in the `-phase` help text would be vastly less likely to be missed.

jdtournier:
> That would do it actually. 

Glad to hear it. Otherwise it would have been difficult to implement cleanly given the macro-based approach I've just introduced... :sweat_smile:

jdtournier:
OK, that last commit fixes a few issues that were already present. Some of these change behaviour, which warrants some discussion:

- `-if`operator: complex version used `a.real() ? b : c` - modified to `is_true(a) ? b : c`, in line with logical operators introduced in this PR (the `is_true()` function returns `z.real() || z.imag()`).

-  `-replace` operator: real version was `((a==b) || (std::isnan(a) && std::isnan(b))) ? c : a`, while the complex was `(a==b) ? c : a`. I get the idea that if both `a` & `b` are NaNs, they could be considered the same, but it's not strictly speaking correct, right? I've changed them both to the simpler complex version, but it does mean that non-finite numbers will never be replaced using this approach. Not sure what others make of that... 

- `-isnan`, `-isinf`, `-finite`: simplified tests from e.g. `std::isnan(v) != 0` to a simple `std::isnan(v)`, since [the standard](https://en.cppreference.com/w/cpp/numeric/math/isnan) shows them as returning a `bool` anyway. 

- `-finite`: complex version was `std::isfinite (v.real()) || std::isfinite (v.imag())`, which would return true if only one of the real or imaginary component was non-finite. I'm guessing that's not the expected behaviour, so changed to `std::isfinite (v.real()) && std::isfinite (v.imag())`

I'm assuming everyone will be OK with this, but thought I'd document this here in case there is disagreement.

Lestropie:
RE `-replace`: The primary purpose for implementing that operator was in fact for manipulating NaNs. Even if NaN != NaN in IEEE spec, there's utility in saying "change all instances of this value to this other value", which isn't stringently tied to a definition of "equals". That's not to say that the same couldn't be achieved using other operations, this was simply seen as the most convenient option at the time, and I'm not sure if there's any tangible benefit to preventing such (other than spec conformance).

Agree with the rest.

jdtournier:
Yes, I'd expected this would be the one to be discussed. One specific use case I can think of where allowing NaN == NaN might be problematic is if given two images `a.mif` and `b.mif`, both masked with NaNs outside the brain, and running the operation:
```
mrcalc a.mif b.mif 100 -replace out.mif
```
This would immediately replace the whole region outside the mask (originally filled with NaN) with the value 100, whereas the user would presumably have intended that operation to apply purely within the mask. It's admittedly a rare use case, though.

Conversely, if the intention is explicitly to replace NaNs with zeros (or any other value), then this can be achieved with:
```
mrcalc a.mif -isnan 0 a.mif -if out.mif
```
I agree it's not as clean as
```
mrcalc a.mif nan 0 -replace out.mif
```
which is presumably the use case this is designed to deal with. I'd be happy enough to match NaNs as happened before, as long as it's documented and consistent between real and complex. As you say, if we admit that the primary purpose of that operator is to replace NaNs, this would make the operator a bit useless...

Lestropie:
> One specific use case I can think of ...

Yes, I'd always thought of it in terms of an image and 2 values; giving 2 images will indeed send it skewiff. I'm content with the behaviour being reverted to standard; I've not documented my handling of such data anywhere so there's nothing that would need to be revised in that respect.

> As you say, if we admit that the primary purpose of that operator is to replace NaNs, this would make the operator a bit useless...

It does now have usage outside of NaN manipulation I think; not in the repo, but I'm sure in a script somewhere I use it to change parcellation indices one at a time.

jdtournier:
Oh well, latest commit re-instates NaN matching in the `-replace` option, documents it, and makes the handling consistent between real & complex. I'm happy enough to keep it like that, but not fussed either way.

thijsdhollander:
> -isnan, -isinf, -finite: simplified tests from e.g. std::isnan(v) != 0 to a simple std::isnan(v), since the standard shows them as returning a bool anyway.

This doesn't change behaviour, right?

jdtournier:
> This doesn't change behaviour, right?

No, this is purely stylistic, really.

jdtournier:
Need to make a decision about whether `(NaN == Nan)` returns true in the `-replace` operator - this is the current behaviour with that last commit. If we're all happy with that, I reckon this is good to merge?

maxpietsch:
> (NaN == Nan) returns true
Shiver. 

Would make sense to only apply this rule if one of the nans is a value rather than an image? Would solve both issues and be true to the meaning of `replace` but keep the mathematical convention of nans not being equal.

In the context of registration and transformation nan-masking images make sense to mark invalid data (`warpconvert`, `population_template` in `mrreg` branch).

Lestropie:
Sticking to spec across the board makes most sense I think. A couple of operations will be less intuitive for users, so they'll just have to be documented and explained clearly.

> In the context of registration and transformation nan-masking images make sense to mark invalid data

Yep; I've been tempted to add NaN-masking to the recommended FBA pipeline in #1543, even if not exploiting the per-fixel subject exclusion capability, precisely for this reason.

jdtournier:
> Would make sense to only apply this rule if one of the nans is an argument value rather than an image value?

Not a bad idea in principle, but would require quite a bit of finessing to implement in practice...

> Sticking to spec across the board makes most sense I think. 

OK, happy with that. I'll revert that last commit and we can merge.

jdtournier:
OK, hadn't noticed @Lestropie has already force-pushed the revert... 
I'll merge now then.

PR #1665, opened 2019-08-02
linear transformation from sets of scanner coordinates

maxpietsch:
* fixed transformcalc align_vertices_rigid
* added align_vertices_rigid_scale
* modified output of `mrview --> view tool --> copy` to match format of plain text matrices


maxpietsch:
Merging as it fixes previously broken code only affecting `transformcalc` and presumably uncontroversial unification of numerical output format.

PR #1672, opened 2019-08-09
dwicat: Fix for updated Python API

Lestropie:
Error in resolving divergence between #1626 and #1449.


Lestropie:
Correction: Divergence was with #1603.

PR #1682, opened 2019-08-20
Smarter default patch size for denoising

dchristiaens:
With these changes, the patch size in `dwidenoise` is by default selected as the smallest cubic extent at least as large as the number of DWI volumes in the data. So, 

- datasets with <=27 volumes will now be processed with `-extent 3x3x3`
- datasets with <=125 volumes will now be processed with `-extent 5x5x5`
- datasets with <=343 volumes will now be processed with `-extent 7x7x7`
- etc.

This helps ensure that the data redundancy is exploited to its fullest, and avoids implausible sudden jumps in the estimated noise level maps when the patch size gets too small. This will obviously affect default run time for HCP-like datasets, but after discussion with @maxpietsch I feel that this is a cost worth paying. In any case, users can always override the option if run time becomes prohibitive.

### DIFF
--- a/cmd/dwidenoise.cpp
+++ b/cmd/dwidenoise.cpp
@@ -20,8 +20,6 @@
 #include <Eigen/Eigenvalues>
 
 
-#define DEFAULT_SIZE 5
-
 using namespace MR;
 using namespace App;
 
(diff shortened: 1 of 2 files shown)


Lestropie:
If it's highly preferable for the method to have very close to the same number of voxels as volumes, would it be better to go all-in on the radial kernel idea? You could get a very close match, and even expand the radius just enough at the FoV edge to encapsulate a comparable number of voxels rather than relying on wrap-around. You could also keep the kernel approximately spherical in the presence of anisotropic voxel sizes.

dchristiaens:
> would it be better to go all-in on the radial kernel idea?

I suppose it would be. I doubt I'll have time to work on that one anytime soon though... If you want to give it a go, feel free. Otherwise, I suggest we merge this PR, perhaps under the header of "slightly" smarter default patch size...

jdtournier:
> rather than relying on wrap-around.

I hope that's not what we're doing? I thought this was now relying on a non-centered patch, but still all on the same side of the image...


Otherwise, happy to merge this as-is. We can always follow up with the radial patch idea in due course.

dchristiaens:
> > rather than relying on wrap-around.
> 
> I hope that's not what we're doing? I thought this was now relying on a non-centered patch, but still all on the same side of the image...

Correct, we are currently using a non-centred, but cubic, patch that aligns with the edge of the image.

> Otherwise, happy to merge this as-is. We can always follow up with the radial patch idea in due course.

Thanks.

PR #1690, opened 2019-08-22
mrinfo: transform doc

maxpietsch:
changes the description of `-transform` from  "the voxel to image transformation" to " the transformation from image coordinates [mm] to scanner / real world coordinates [mm]".


jdtournier:
😊

PR #1692, opened 2019-08-23
Run time improvements for nonlinear registration

dchristiaens:
These two changes are the result of a discussion with @maxpietsch yesterday. Pairwise nonlinear registration was very slow on his system, especially when using fODF reorientation. I had a look at the code and suggest two run time improvements:

1. A faster implementation of the reorientation step. The SH basis evaluation was/is currently done looping over all delta-functions. Some of the work can be avoided by evaluating the basis across the full set of (cartesian) directions straight away. This is implemented in 4b8e3ba and yields a 10% speed improvement. The results are exactly the same (zero residuals in warps and in output images).
```
on dev:
$ time mrregister -type nonlinear odf_sub-1.mif odf_sub-2.mif -mask1 mask1.mif -mask2 mask2.mif -transformed_midway test01.mif test02.mif -nl_warp warp01.mif warp02.mif
... 
1675.12s user 112.60s system 1837% cpu 1:37.30 total

with 4b8e3ba:
$ time mrregister -type nonlinear odf_sub-1.mif odf_sub-2.mif -mask1 mask1.mif -mask2 mask2.mif -transformed_midway test11.mif test12.mif -nl_warp warp11.mif warp12.mif
...
1453.35s user 120.21s system 1799% cpu 1:27.42 total
```

2. Taking warp field inversion out of the main loop. The displacement field in currently inverted upon every Demons iteration, but this is not needed for any of the intermediate calculations. Taking this calculation out of the loop, to compute it once at the end, saves an additional 7% on the run time (implemented in ce0bf28). With this change, there is a *tiny* (< 0.002 mm) change in the output warp fields; the output transformed fODFs are identical. @maxpietsch reported that this doesn;t work for him with multi-level Lmax during registration, but I can't replicate this behaviour.  
```
with ce0bf28:
$ time mrregister -type nonlinear odf_sub-1.mif odf_sub-2.mif -mask1 mask1.mif -mask2 mask2.mif -transformed_midway test21.mif test22.mif -nl_warp warp21.mif warp22.mif
...
1374.30s user 116.81s system 1837% cpu 1:21.16 total
```

While the total speed improvement is admittedly small (at least, I was hoping for more), a similar test took 8 minutes on Max's system and 11 minutes on the last year's workshop laptop, so those total 16% can matter...

Working on this has also made it clear to me that the SH basis implementation, and the associated Legendre functions, can benefit from refactoring. I'm happy to have a go at this at some point, though it's probably better dealt with in a separate PR.


jdtournier:
Sounds good! Slightly worried about this though:

> @maxpietsch reported that this doesn;t work for him with multi-level Lmax during registration, but I can't replicate this behaviour.

Can we get to the bottom of this before merging? 

Otherwise: 

> Working on this has also made it clear to me that the SH basis implementation, and the associated Legendre functions, can benefit from refactoring. I'm happy to have a go at this at some point, though it's probably better dealt with in a separate PR.

More than happy with that if you feel it'll make a substantial difference. :+1:

thijsdhollander:
> Taking warp field inversion out of the main loop. The displacement field in currently inverted upon every Demons iteration, but this is not needed for any of the intermediate calculations.

This *is* important: the inverse is not computed from scratch each iteration, *but uses the previous iteration's inverse as initialisation*. This gradual updating of the inverse is a mechanism to effectively obtain it in a robust way at the end of the algorithm. Without this mechanism, computing the inverse of a "strong" warp "at once" is notoriously problematic: just using the gradient descent approach that's typically used (and we use here too) will get you stuck in local optima for certain specific "types" of warps. The most simple artificial example of such a problematic warp I can think of of the top of my head is (think 2D now, for simplicity) a whirlpool shaped warp that twists at least 180 degrees (maybe less is sufficient too to trigger the problem; would need to draw that on paper to figure out). I think the typical circle-to-C registration example might have it too, at least if the final warp at least registers the circle fully to the C shape. In 3D, there's even far more kinds of cases of warps where local optima would be encountered if attempting to compute an inverse "at once".

The challenge of inverting a warp is essentially one of figuring out a possible diffeomorphic integration path from start (no warp) to end (warp), and then inverting that integration path. But the moment you get a "twist" in the warp (I'm still talking about a diffeomorphic warp here, that is invertible; by "twist" I mean a whirlpool type of localised rotation), the most naive interpolation path between no-warp and the twist-warp takes you via non-diffeomorphic warps. The implication of that is local optima in the gradient descent inversion approach.

The only solution is to only invert very small warps (with only small limited displacements). Each update step in a registration process is such a limited displacement field, luckily. So the update step can far more safely be inverted, and this can then be composed with the "current" complete inverted warp at this point in the algorithm, and then once more gradient descent optimised to be entirely accurate again.

Or to put it once more differently: the path the registration process takes is a trajectory through the space of diffeomorphisms, so this trajectory itself can be employed to "guide" the computation of the inverse through this space from start to end, without encountering bumps along the way.

> @maxpietsch reported that this doesn;t work for him with multi-level Lmax during registration, but I can't replicate this behaviour.

This might just have been a coincidence of one of those more challenging warps being the final result in his experiment at that point. For smoother warps, the chances of encountering this problem are indeed smaller; so the problem might only occasionally pop up, or more often in scenarios where less regularisation is used. In any case, you don't want to risk it.

By the way, this reminds me (I think about it every so often and then forget to mention it again): you might want to stick a big fat warning label on `warpinvert` for this very reason. If you rely on it to invert "any" warp (that's perfectly diffeomorphic), you're going to get caught out every once in a while for certain kinds of warps. When it's wrong (essentially a local optimum), it can be *very* wrong, depending on the warp.

thijsdhollander:
Just to add: when it doesn't work for Max, I suppose you mean it outputs a different inverse before and after the relevant commit?  If that's the case, you can actually assess the inaccuracy of the inverse: composing a warp with its inverse should give a reasonably accurate identity transform/warp.  If it doesn't, you know it's wrong.  If it's wrong, I'm 100% sure the problem is a local optimum.

dchristiaens:
> This _is_ important: the inverse is not computed from scratch each iteration, _but uses the previous iteration's inverse as initialisation_. This gradual updating of the inverse is a mechanism to effectively obtain it in a robust way at the end of the algorithm.

Yes, I know. I don't see a significant difference in subject-to-subject registration though (residuals < 0.002 mm as I said above). Of course we can come up with whirlpools and other artificial examples, but these are usually the cases where nonlinear registration struggles or fails anyway. On a 3-D to 3-D registration, updating the inverse field takes up about 50% of the run time!

The proper way to address this is to move to log-Demons registration, where the inverse is available for free. I'll give that a go at some point.



> Just to add: when it doesn't work for Max, I suppose you mean it outputs a different inverse before and after the relevant commit?

We were testing on Max's machine the night before he left on vacation and we probably screwed something up in the rush. I had another go with an awake brain the day after so I think's it's fine, but to be safe asked him to test this commit when he gets back.

> Sounds good! Slightly worried about this though:
> 
> > @maxpietsch reported that this doesn;t work for him with multi-level Lmax during registration, but I can't replicate this behaviour.
> 
> Can we get to the bottom of this before merging?

Yes, happy to wait.

thijsdhollander:
> Yes, I know. I don't see a significant difference in subject-to-subject registration though (residuals < 0.002 mm as I said above). Of course we can come up with whirlpools and other artificial examples, but these are usually the cases where nonlinear registration struggles or fails anyway.

Ok, we're on the same page about what I explained; was just using a lot of words to make sure we were indeed on the same page. The artificial example was only meant for that purpose; it's the most "visual" thing I can think of to make it clear, but it wasn't needed since we're aware about the issue in any case.

That said, this problem happening, i.e. the inverse ended up being inconsistent due to the class of warps where inverting ends up in local optima with gradient descent, is not all that rare though. There's a reason there's large volumes of literature around preserving inverse consistency, in medical imaging and more specifically brain imaging; and similarly that several packages out there (ITK and the like) even offer different algorithms to invert a warp, where none of them works across the board (i.e. converges on the true optimum), as far as I'm aware.  The circle-to-C example is overly extreme, but it's not dissimilar to patterns happening around the cortex, even in registrations between relatively "normal" subjects. Where we might encounter it less, is with most FOD-based registrations, but only coincidentally because of a few reasons: often (but not always) very low spatial resolution in dMRI, more focus on a very smooth warp that merely matches the core WM "skeleton", the FODs also constraining that a bit more, etc... however, with larger resections, tumour mass effects, large and lobe-specific atrophy, these warps do come about. You can question whether registration can then entirely accurately register such cases; that's a valid question, but it's entirely independent from preserving inverse consistency when you output 2 warps that claim to indeed be each other's inverse (up to precision of the discrete voxel grid at least).  Also, it's quite possible that users would be using the registration code e.g. for inter-subject T1w image (or similar quality) registration or template building.  Even with the current limitation to SSD, that's doable (and if normalised CC is introduced, that'd open it up to even wider ranges of applications).

> On a 3-D to 3-D registration, updating the inverse field takes up about 50% of the run time!

Well, the inverse also kind of is half of the output you're getting. I'd happily pay even more to have inverse consistency guaranteed; if you want to rely on that inverse warp, you wouldn't want it to be entirely wrong every once in a while.  You could make it an optional output though, or could even have a kind of "`-fast`" switch where you warn that you're throwing inverse consistency "potentially" out of the window.  Or additionally output some metric on accuracy of inverse consistency (often computed as the average difference between identity transform and composition of forward and inverse warp); but that might be hard to interpret for most users.

> The proper way to address this is to move to log-Demons registration, where the inverse is available for free. I'll give that a go at some point.

That's one way to preserve inverse consistency indeed; and should for sure be lots faster of course. It'd change the current output / behaviour though (but strictly for the better, I'd argue).

> We were testing on Max's machine the night before he left on vacation and we probably screwed something up in the rush. I had another go with an awake brain the day after so I think's it's fine, but to be safe I asked him to test this commit when he gets back.

> EDIT: No that is not what I meant. Max ran into a run time error that aborted registration.

Ok, was likely not an actual case of such a warp then. Other than the inverse consistency, there's not much else in the commit that could be wrong or a problem or similar; it's a pretty straightforward difference: just computed the inverse "at once" at the end instead of updating it each iteration.

maxpietsch:
With this code, I can't reproduce the error I saw (runtime Exception due to dimension mismatch if I remember correctly). I already undid the changes on my machine and I can't see how I managed to get that error from modifying `nonlinear.h`...

With these changes, on unmasked 2.5mm adult data, I get a speedup from 60s to 20s (3 run average, no linear registration, 10 iterations per resolution stage, without writing output files or warps but otherwise default parameters).

On my two test subjects, I get differences in the inverse warps of up to 0.0057mm. Probably neglectable but, I'd test on more challenging data (HCP inter-subject, Alzheimer's data, or longitudinal neonatal data) what worst-case effect we'd expect due to replacing the iterative warp inversion. If inverse consistency is significantly violated in some cases, I'd opt for keeping the default to inverting on every iteration (or every n-th iteration), changeable via command-line argument.

thijsdhollander:
> With this code, I can't reproduce the error I saw (runtime Exception due to dimension mismatch if I remember correctly). I already undid the changes on my machine and I can't see how I managed to get that error from modifying nonlinear.h...

Looks like that one was just some confusion back then I reckon. Can probably be safely dismissed.

> On my two test subjects, I get differences in the inverse warps of up to 0.0057mm.

Yeah, when you get a small difference like that, it's merely a precision thing; that's not an issue for sure. These aren't the kinds of cases where it got stuck in a local optimum.

> I'd test on more challenging data (HCP inter-subject, Alzheimer's data, or longitudinal neonatal data) what worst-case effect we'd expect due to replacing the iterative warp inversion. If inverse consistency is significantly violated in some cases, I'd opt for keeping the default to inverting on every iteration (or every n-th iteration), changeable via command-line argument.

When it does get violated, then it's not a matter of "how significant": it's not a precision issue then, but a discrete jump from the optimum it should reach to one that's simply wrong. Due to how the gradient descent is set up to compute an inverse, the wrong bit will be "local" in the warp as well. In any case I'd highly recommend having the default to update the inverse at each iteration (note, this is not "inverting on every iteration" in a certain interpretation of those words at least). An option to change it should probably come with a strong warning. You could also have an option to simply not output the inverse (and then you don't have to compute it at all), if the user reckons they won't need it. Still safer than outputting an inverse that might not actually be an inverse if you happen to be unlucky.

PR #1693, opened 2019-08-24
Fixel-fixel connectivity matrix format

Lestropie:
I had a go at polishing off the stats code using the basic initial fixel-fixel connectivity matrix format (text file with one line per fixel, comma-separated fixel index - connectivity value pairs separated by commas per row), but was never happy with the performance of it. In particular if the data are on network storage it can stall completely. So I think that I want to finish off the .mif-based fixel-fixel connectivity matrix format before merging #1543.

From memory, what I had already working:

- `fixelconnectivity` command to generate and save matrix.

- `fixelfilter smooth` reading and using that data to smooth fixel data.

What needs to be done:

- Get `fixelfilter connect` working using new format. **EDIT**: Done.

- Strip from `fixelcfestats` ability to receive track file as input rather than pre-calculated connectivity matrix (code becoming too difficult to manage). **EDIT**: Done.

- Get `fixelcfestats` working using new matrix format. In particular the CFE enhancer will need to change quite a bit. **EDIT**: Done.


Lestropie:
Note to anybody out there previously accessing the "`fixel_matrix_format`" branch (likely mostly workshop attendees): The relevant functionalities should now be accessed via the "`stats_enhancements`" branch. Subsequent changes to how this code is accesssed will be announced in #1543.

Issue #1697, opened 2019-08-27
General issue template not in correct folder?


thijsdhollander:
@maxpietsch, are you sure you've put the general.md issue template in the correct folder on `master`?  When I created this very issue, I was only given the choice between bug report and feature request.

PR #1702, opened 2019-09-02
moved github general issue template #1697

maxpietsch:
The added `General` button makes the small message `Don’t see your issue here? Open a regular issue.` somewhat redundant but much more visible. Addresses #1697.

image


thijsdhollander:
> The added General button makes the small message Don’t see your issue here? Open a regular issue. somewhat redundant but much more visible.

No, that's always there regardless. The templates are always in addition to a "regular" issue. Well, they're all "regular" issues of course, the template just sets it up... with a template; for convenience.

thijsdhollander:
Oh, now I see; your "general" template is basically a no-template template. So well, in line with the above, you can actually remove it if you wish: the "Open a regular issue" link will still be there.

PR #1703, opened 2019-09-03
Progressbar fix quiet

### LINKED ISSUE #1700: progress bar shown despite log level 0
In nonlinear registration code, I now get lots of progress messages somewhere introduced between e14c89cfc8 (current dev) and 326f11edd2fbcf3d5d827f991196380a89b46a4a (dev 22 August). Could be related to #1586 but haven't looked into it.

example to reproduce (note `LogLevelLatch level (0)`):
```
#include "command.h" // key header for building an executable command against MRtrix3
#include "image.h" // defines class for accessing image data
#include "algo/threaded_loop.h" // tools for multi-threaded looping over images
using namespace MR;
using namespace App;

// this function sets up the command-line interface for the command
void usage()
{
  // details that must be defined for all commands
  AUTHOR = "W. Heath Robinson (h.robinson@implausible.com)";
  SYNOPSIS = "Compute voxel-wise geometric mean across volumes";

  // these are compulsory command-line inputs that must always be provided by the user
  ARGUMENTS
    + Argument ("input", "an input image").type_image_in()
    + Argument ("output", "the output image").type_image_out();

  // these optional inputs can be either provided or omitted by the user
  OPTIONS
    + Option ("invalid", "value if any intensity is non-positive (default = NaN)")
      // this option requires that a value be specified alongside it when used
      + Argument ("value").type_float();

  // example citation for the command docs
  REFERENCES
    + "Robinson, W.H. A treatise on geometric averaging. Contrived Science Weekly, 2019.";

  // manually set the copyright notice if the default is inappropriate
  COPYRIGHT = "Copyright (c) 2019 Heath Robinson Labs.";
}

// this is the functor class defining the operation to be applied
class GeometricMean
{
  public:
    GeometricMean (const float value) : invalid (value) { }
    // the operation to be performed independently for each voxel
    void operator() (Image<float>& in, Image<float>& out) {
      // grab voxel intensities across volumes as a vector:
      values = in.row(3);
      if (values.minCoeff() > 0.0)
        out.value() = std::exp (values.array().log().mean());
      else
        out.value() = invalid;
    }
    Eigen::VectorXd values;
    const float invalid;
};

// this function contains the primary operation of the command
void run ()
{
  // access the input image and check validity
  auto in = Image<float>::open (argument[0]);
  if (in.ndim() != 4)
    throw Exception ("expected 4D input image");

  // copy the input image's header, and modify to suit
  Header header = in;
  header.ndim() = 3; // <= remove 4th dimension
  header.datatype() = DataType::Float32; // <= use floating-point output

  // create the output image, using the modified header as a template
  auto out = Image<float>::create (argument[1], header);

  // create an instance of the functor with the value optionally provided by the user
  GeometricMean functor (get_option_value ("invalid", NaN));
  // performs the operation using multiple threads while displaying a progress bar
  {
    LogLevelLatch level (0);
    ThreadedLoop ("computing geometric mean", in, 0, 3).run (functor, in, out);
  }
  // output image write to disk is finalised automatically when it goes out of scope
}
```

output:
```progress in.mif out.mif -force
progress: [WARNING] existing output files will be overwritten
progress: [100%] computing geometric mean
```

jdtournier:
Closes #1700 

I think this should do it.... The first commit here is the actual fix (silly mistake on my part). The second commit cleans up the implementation a bit (better encapsulation, but really mostly cosmetic). The last commit actively prevents a ProgressBar from displaying if another is already active - which should mean there is no need to use a loglevel latch to keep nested ProgressBars quiet...

### DIFF
--- a/core/progressbar.cpp
+++ b/core/progressbar.cpp
@@ -155,6 +155,7 @@ namespace MR
   std::condition_variable ProgressBar::notifier;
   std::mutex ProgressBar::mutex;
   void* ProgressBar::data = nullptr;
+  bool ProgressBar::progressbar_active = false;
 
   ProgressBar::SwitchToMultiThreaded::SwitchToMultiThreaded () {
     ProgressBar::previous_display_func = ProgressBar::display_func;
--- a/core/progressbar.cpp
+++ b/core/progressbar.cpp
@@ -51,23 +51,23 @@ namespace MR
     void display_func_terminal (const ProgressBar& p)
     {
       __need_newline = true;
-      if (p.multiplier)
+      if (p.show_percent())
         __print_stderr (printf (WRAP_OFF_CODE "\r%s: [%3" PRI_SIZET "%%] %s%s" CLEAR_LINE_CODE WRAP_ON_CODE,
-              App::NAME.c_str(), p.value, p.text.c_str(), p.ellipsis.c_str()));
+              App::NAME.c_str(), p.value(), p.text().c_str(), p.ellipsis().c_str()));
       else
         __print_stderr (printf (WRAP_OFF_CODE "\r%s: [%s] %s%s" CLEAR_LINE_CODE WRAP_ON_CODE,
-              App::NAME.c_str(), busy[p.value%6], p.text.c_str(), p.ellipsis.c_str()));
+              App::NAME.c_str(), busy[p.value()%6], p.text().c_str(), p.ellipsis().c_str()));
     }
 
 
     void done_func_terminal (const ProgressBar& p)
     {
-      if (p.multiplier)
+      if (p.show_percent())
         __print_stderr (printf ("\r%s: [100%%] %s" CLEAR_LINE_CODE "\n",
-              App::NAME.c_str(), p.text.c_str()));
+              App::NAME.c_str(), p.text().c_str()));
       else
         __print_stderr (printf ("\r%s: [done] %s" CLEAR_LINE_CODE "\n",
-              App::NAME.c_str(), p.text.c_str()));
+              App::NAME.c_str(), p.text().c_str()));
       __need_newline = false;
     }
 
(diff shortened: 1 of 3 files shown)


Lestropie:
946f1ce echoes discussion in #263. Historically a `static bool` would not have been adequate; I couldn't even get the appropriate behaviour using a `std::atomic_flag` due to the race condition involved. But perhaps after #1586 it's only within-thread nested `ProgressBar`s that may remain an issue. I'll see if I can find the test command I wrote at the time.

maxpietsch:
Fixes the `-quiet` and loglevel latch issues in `mrregister`.

jdtournier:
> Historically a static bool would not have been adequate; I couldn't even get the appropriate behaviour using a std::atomic_flag due to the race condition involved.

OK, I guess there's quite a few edge cases to consider here. I'd only considered the case of nested ProgressBars, or at least ProgressBars instantiated sequentially within a single thread. The current safeguard should cater for _most_ cases since even in a multi-threading context, the classes handling the tasks are instantiated within the main thread prior to launching the worker threads, so as long as things happen in the right order, the first to claim the ProgressBar gets to display it. 

I should also point out that the LogLevel latch is still effective in silencing unwanted progress messages if required, so that's always an option.

Where things might fall over is in a context where the top-level multi-threaded operation does not itself instantiate a ProgressBar at all, but invokes multiple tasks that do instantiate their own ProgressBar. In that case, the terminal won't have been claimed first, and we end up with a race condition to claim it. But even in theory, if this was to work we would end up with multiple threads each competing to update the terminal with presumably the same progress message and different percentages - which I reckon is a situation we should be avoiding. 

One additional safeguard we could put in there is to add a check at [this line](https://github.com/MRtrix3/mrtrix3/blob/946f1cef3e8a5c82122d2b7af8302d69527643a4/core/progressbar.h#L187) to ensure that the [thread ID](https://en.cppreference.com/w/cpp/thread/get_id) is that of the main thread. We'd have the original main thread ID stored in a global static variable within the [App::init() call](https://github.com/MRtrix3/mrtrix3/blob/946f1cef3e8a5c82122d2b7af8302d69527643a4/core/app.cpp#L1212), and then simply check that `std::this_thread::get_id() == ::MR::App::main_thread_ID` when setting the `show` variable. This would effectively prevent any ProgressBar instantiated outside the main thread from displaying anything or interfering with another ProgressBar. This might prove a bit too restrictive, but I'm not sure, I can't think of instances where a ProgressBar that we actively want to show is instantiated outside the main thread. But I reckon it might prove effective in dealing with any edge cases: it's automatically thread-safe (regardless of any potential race conditions on the `progressbar_active` bool variable I've just introduced), would immediately silence any unwanted progressbars in separate threads, and _probably_ not require any modifications to existing code. The only worry I have is with Thread::Queues where the source or sink instantiates the progressbar within its execute() call, rather than declaring the progressbar as a member. Worth considering...?

jdtournier:
I suggest we merge this PR to ensure more or less correct behaviour on `dev`, and carry on discussing the issue of how to better manage claiming the terminal in a multi-threaded context in #263 (it's still open...). Things should work fine at the moment since any instance where this would otherwise be problematic is currently dealt with using the LogLevel latch approach - what we're talking about beyond that is coming up with a strategy that won't requite such explicit handling, and that can wait...

jdtournier:
That last commit fixes issues that @maxpietsch pointed out to me. _Should_ be good now...

PR #1707, opened 2019-09-06
Progressbar restrict construction to main thread

jdtournier:
As per discussion on #1703 and #263. 

Still need to figure out whether there are instances where this is going to be problematic - hence this remains a draft for now.


jdtournier:
OK, I guess this isn't going to be looked at unless I un-draft it... Everyone happy to merge and see what happens...?

PR #1711, opened 2019-09-09
foreach fixes

maxpietsch:
- change to PRE: now removes .gz if present then splits off extension instead of truncating after the first '.'
- fix to UNI: if the common suffix is empty (common for directories without common name suffix) the output was emtpy, now it defaults everything after the common prefix


Lestropie:
> change to PRE: now removes .gz if present then splits off extension instead of truncating after the first '.'

Yeah multi-period filenames are bloody annoying for trying to extract extensions. The alternative would be `.split('.')[0]` :nauseated_face:

PR #1715, opened 2019-09-10
Fix sh precomputer again

jdtournier:
Seems the SH precomputer was still wrong, even after #1206...

The fix we did then made it work for the `Math::SH::get_peak()` call, but screwed up the `PrecomputedAL::value()` call... This is because the √2 factor had originally been absorbed in the precomputed Associated Legendre polynomials, which wasn't taken into account in the `get_peak()` call, but was taken into account by the `value()` call. This PR now ensures the precomputed Associated Legendre polynomials still _don't_ have the additional √2, so that they are interchangeable with the non-precomputed ones, but fixes the `value()` call so that that factor is correctly taken into account there. 

On the funny side, it looks like there had actually been no problem with the tracking until we decided to 'fix' it with #1206... 🤦‍♂ 

I've added a small test command to check for regression to make sure we don't bork this yet again in future. In the meantime, this is likely to change the tracking behaviour yet again...

### DIFF
--- a/core/math/SH.h
+++ b/core/math/SH.h
@@ -434,7 +434,7 @@ namespace MR
                 ValueType c = c0 * cp - s0 * sp;
                 ValueType s = s0 * cp + c0 * sp;
                 for (int l = ( (m&1) ? m+1 : m); l <= lmax; l+=2)
-                  v += get (f,l,m) * (c * val[index (l,m)] + s * val[index (l,-m)]);
+                  v += get (f,l,m) * Math::sqrt2 * (c * val[index (l,m)] + s * val[index (l,-m)]);
                 c0 = c;
                 s0 = s;
               }
--- a/testing/cmd/testing_sh_precomputer.cpp
+++ b/testing/cmd/testing_sh_precomputer.cpp
@@ -1,0 +1,58 @@
+/* Copyright (c) 2008-2019 the MRtrix3 contributors.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Covered Software is provided under this License on an "as is"
+ * basis, without warranty of any kind, either expressed, implied, or
+ * statutory, including, without limitation, warranties that the
+ * Covered Software is free of defects, merchantable, fit for a
+ * particular purpose or non-infringing.
+ * See the Mozilla Public License v. 2.0 for more details.
+ *
+ * For more details, see http://www.mrtrix.org/.
+ */
+
+#include <sstream>
+
+#include "command.h"
+#include "math/SH.h"
+
+
+using namespace MR;
+using namespace App;
+
+
+void usage ()
+{
+  AUTHOR = "J-Donald Tournier (jdtournier@gmail.com)";
+
+  SYNOPSIS = "Test the accuracy of the spherical harmonic precomputer";
+
+  REQUIRES_AT_LEAST_ONE_ARGUMENT = false;
+}
+
+using value_type = float;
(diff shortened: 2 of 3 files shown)


thijsdhollander:
Just to confirm whether I understand it correctly:

* `fod2fixel` is unaffected; i.e. the `-peaks` option was fixed with #1206; has been correct since; is not affected by this PR.  Also `fod2fixel -afd` was never affected, not before nor after #1206, so also not by this PR.
* `sh2peaks` was never affected, not before nor after #1206, so also not by this PR. However, might want to double-check what's exactly the case for the `sh2peaks -fast` option.
* `tckgen` is affected; i.e. was in fact correct *before* #1206 (?); has been "broken" since; so is affected by this PR. This affects the amplitudes as evaluated during `tckgen`, so basically as well whether a given amplitude is above/below `-cutoff`.

So in conclusion of all the above, my understanding is that this PR only affects `tckgen`, correct?

If `tckgen` was fine before #1206, and given that the "fix" back then was orientationally dependent (i.e. on the inclination angle of spherical polar coordinates); this means this PR also causes an orientationally dependent change (again), right?

This might very well catapult the sensible `-cutoff` value or range back again too; but wrt that one, I think it's at this point more clever to first fix this, wrap up the `dwi2response dhollander` business and take into account `mtnormalise`.  And maybe consider first whether there's anything else that's going to affect the actual amplitudes `-cutoff` is typically applied to.  And then evaluate what's a sensible value for `-cutoff`.

dchristiaens:
Great catch @jdtournier !

jdtournier:
OK, just to clarify things (although it might complicate matter further):

- `fod2fixel -peaks` was affected before #1206, fixed by #1206, and remains unaffected here.
- `fod2fixel -afd` remains unaffected as far as I can tell (doesn't rely on SH precomputer)
- `sh2peaks` was unaffected before #1206, the `-fast` option was added in #1206 to allow use (and testing) of the SH precomputer, and remains unaffected here (checks out in testing). 
- `tckgen` with iFOD1 or iFOD2 was unaffected before #1206, messed up after #1206 with this bug, and fixed with this PR.
- `tckgen` with SD_STREAM must have been affected before #1206 (although I've not checked that), since it uses the SH precomputer to get the peaks - the exact piece of code that was problematic with `fod2fixel -peaks`. So it must have been fixed with #1206, and should still be fine now. 

So it's a bit of a messy story...

On a different note, I've double-checked that `tckgen` now produces the correct output, it that the output is strictly identical with and without `-noprecompute` (the default is to use the SH precomputer, which is what triggered the bug). I'll add further tests for that shortly.

thijsdhollander:
Thanks for taking the time to clarify; it's appreciated and useful.

> fod2fixel -peaks was affected before #1206, fixed by #1206, and remains unaffected here.
> fod2fixel -afd remains unaffected as far as I can tell (doesn't rely on SH precomputer)
> sh2peaks was unaffected before #1206, the -fast option was added in #1206 to allow use (and testing) of the SH precomputer, and remains unaffected here (checks out in testing).

Ok, good to see that clarified.  Makes sense; so those are all fine and unaffected.

> tckgen with iFOD1 or iFOD2 was unaffected before #1206, messed up after #1206 with this bug, and fixed with this PR.

> tckgen with SD_STREAM must have been affected before #1206 (although I've not checked that), since it uses the SH precomputer to get the peaks - the exact piece of code that was problematic with fod2fixel -peaks. So it must have been fixed with #1206, and should still be fine now.

Hmm, ok, indeed messier than I anticipated; at least it's contained to `tckgen` only, even to the probabilistic algorithms only.

Other than the `dwi2response` discussion though, here `mtnormalise` can't play the role of equaliser, since this change's effect happens after any such steps.  Once all of these, and possibly the angle constraint for SD_STREAM too, wrap up, it'll definitely be a good idea to re-evaluate `-cutoff` defaults again.

PR #1720, opened 2019-09-12
MRView tractogram tool: fix bug in lines render when thresholding

jdtournier:
Crashes out on ATI hardware when displaying a TSF-thresholded tractogram and switching to lines render mode. This seems to fix it...


jdtournier:
Anyone willing to review this one?

PR #1723, opened 2019-09-17
JSON export: Fix PE scheme

Lestropie:
Use of `auto` keyword resulted in null phase encoding direction in instances where saving a NIfTI image requires permutation / flipping of axes; this would lead to an `Exception` being thrown in `Axes::dir2id()` function.


Lestropie:
Note: Echoes 79edb809, but needs to go to `master`.

PR #1732, opened 2019-09-23
core/app: fix abbreviated option parsing

maxpietsch:
fix abbreviated option parsing with identically named options 
example in mrfilter `-std` gets mapped to the first `-stdev` option.
Prior to these changes `mrfilter -std` fails with:
`[ERROR] several matches possible for option "-std": "-stdev", "-stdev"`


Lestropie:
This seems to be an issue on `master` also; the fix should therefore I think be directed there.

PR #1735, opened 2019-09-25
Fixes to slice timing handling

Lestropie:
It seems a bit of a mess has accumulated with respect to JSON data and the handling thereof, which is now having downstream effects at BIDS-Apps/MRtrix3_connectome#59.

In addition to this fix for `master`, further modifications will be required on `dev` to get JSON data handling working properly:

- The behaviour of `mrinfo -json_keyval` / `mrinfo -json_all` and `mrconvert -json_export` differs; the latter handles rotation of phase encoding / slice encoding directions while the former do not.

- While `mrinfo -json_all` appropriately writes basic header information in their native data format (e.g. numerical matrix for header transform), the contents of `keyval()` are all written as strings. `mrinfo` was intended to convert numerical matrix data appropriately, but that seems to not be working right now. It's also the case that simple things like integer values may be better written to JSON as integer values rather than the string representations thereof.


jdtournier:
I have to admit I've not had a chance to get my head around these issues to comment meaningfully. I assume this fix is necessary to fix issues on `master`. But presumably this doesn't fix #1771...?

Lestropie:
> But presumably this doesn't fix #1771...?

Trouble is that there's three different PRs (#1735, #1738, #1771),  all dealing with a common underlying fundamental issue but at different levels and with different scopes. Since #1771 is not mine I don't 100% understand exactly what the overlap is there, and there's also the issue of `master` vs. `dev` changes. So I'll see how far I can get here in disentangling them.

- This PR is specifically about handling of slice timing data; in particular for `dwipreproc` to be able to handle JSON data that were *not* generated by *MRtrix3*. On current `master`, the slice timing information extracted from DICOM is not written in the appropriate format, but `dwipreproc` was programmed to read the data from this inappropriate format; hence the incompatibility with other DICOM converters. So this change does two things:

  - Fix specifically the saving of slice timing information from DICOM into the file header, so that it is then formatted correctly when written to JSON.

  - Make the import of slice timing information in `dwipreproc` as robust as possible, so that it is maximally likely to be able to correctly import such information no matter what software or version of such it was generated from.

- #1738 is intended to be a more all-encompassing "fix" of JSON import and export. It's not addressing any specific observation of any particular erroneous behaviour, but trying to make the import / export "behave as expected" in all scenarios.

- #1771 is where I get slightly stuck:

  - It's not clear whether the issue reported there is over and above what was rectified in #1738. I did quite a bit of testing for that one, so would be slightly surprised if there was still an outstanding issue after that changeset.

  - If it is purely a matter of 1D array data being transposed, then I don't think there is any issue with collapsing into a single header entry. But I would want test data to run against #1738 to confirm such, since I already had to muck about quite a bit there with how string header data and JSON import / export interact.

  - I do think there is an issue with using comma delimiters to write such data to the header at import stage. Comma-delimited slice timing information in the header gets exported to JSON as a list of lists, not a 1D numerical data array.

I suspect we need to have some exemplar data, on which we can test both `dev` and `master`, in order to determine whether or not there are further changes that need to be pushed to `master` to fix outright bad behaviour, as well as whether #1771 provides a necessary change to `dev` over and above #1738.

Lestropie:
Aborting; given magnitude of issues with JSON data handling, will not be resolving such on `master`, instead relying on imminent tag update that will include changes in #1843.

PR #1736, opened 2019-09-25
dwipreproc: Permit rigid-body between DWI and SE-EPI

Lestropie:
Better support in `dwipreproc` for a very specific type of image acquisition protocol. The only change is to prevent an unnecessary image re-gridding of the SE-EPI data onto the DWI voxel grid.

While an API function is modified to add an optional input argument, the default value of this parameter means that behaviour and interface are unchanged elsewhere. I will resolve the change manually with `dev`, since I would prefer to more consistently use key-value function arguments in the updated API.


Lestropie:
Also: See relevant discussion in BIDS-Apps/MRtrix3_connectome#61.

PR #1740, opened 2019-09-25
mrclusterstats: Make sure template mask is 3D

Lestropie:
Command would execute for an erroneous 4D mask image, and would take an inordinate amount of time to execute.


cmd/mrclusterstats.cpp
```diff
@@ -159,6 +159,8 @@ void run() {
     throw Exception ("the number of contrasts does not equal the number of columns in the design matrix");
 
   auto mask_header = Header::open (argument[3]);
+  if (!(mask_header.ndim() == 3 || (mask_header.ndim() == 4 && mask_header.size(3) == 1)))
```
jdtournier: This won't catch 5D images... Maybe we need something a bit more formal, e.g. a `check_effective_dimensionality (HeaderType, int)` function in the `core/image_helpers.h` header...? This might look something like:
```
Template <class HeaderType>
void check_effective_dimensionality (const HeaderType& H, size_t ndim) 
{
  if (H.ndim() < ndim)
    throw Exception ("image \"" + H.name() + "\" is not " + str(ndim) + "D"); 
  for (size_t n = ndim; n < H.ndim(); ++n)
    if (H.size(n) != 1)
      throw Exception ("image \"" + H.name() + "\" is not " + str(ndim) + "D"); 
}
```
What do you reckon?

Lestropie: That'd work; I just don't know how important it is to not produce an error when a user attempts to use a 5D image as a mask for voxel stats... If you wanted to add a function like this, there'd be a lot of other areas in the repo where it would be applicable, so it would warrant its own PR.

jdtournier: True enough. Might merge this and create another one on `dev` for this idea...

jdtournier: True enough. Might merge this and create another one on `dev` for this idea...

Issue #1729, opened 2019-09-19
tckgen: Revert max angle behaviour


Lestropie:
In #1507, more accurate quantification of streamline curvature was necessary in order to derive the necessary parameters related to streamline length calculation. As part of this process however, the determination of default maximum curvature angles was also reverted to reflect a target minimum radius of curvature of one voxel, rather than a curvature of 90 degrees along a streamline arc length equivalent to one voxel. Furthermore, it appears that while the intent was for this new calculation to apply to first-order methods only, it is nevertheless being applied to higher-order methods as well.

Targets:

- Revert `dev` such that behaviour of tracking algorithms is identical to that prior to #1507, whilst retaining the length quantification aspects of #1507.

- Generate second pull request proposing alteration to the default tracking algorithm maximum curvature angles based on a target radius of curvature derivation. Given iFOD2 would be excluded from this, and other algorithms are being changed by #1630, it would likely only affect iFOD1.

- For both of the above, need to re-check the precise behaviour of RK4.

PR #1747, opened 2019-10-01
tckgen: Change default step size for RK4

Lestropie:
First proposed change to default tracking parameters aspart of #1729.

First-order algorithms use a very small default step size in order to mitigate curvature overshoot. An alternative solution to curvature overshoot is to use the 4th-order Runge-Kutta method. Therefore, when the latter is activated using the `-rk4` option in `tckgen`, there is no longer a good justification to keep the default step size very small. However in the current code, the same default step size of 1/10th of a voxel is used regardless of the use of RK4.

Initially I was going to make the default step size with RK4 1/2 of a voxel, mimicking the default for iFOD2. However I then considered that for default iFOD2 parameters, there are in fact 3 vertices per 1/2-voxel "step" that are used for sampling 5TT image / ROIs more densely. If the default step size for these first-order algorithms with RK4 were to be 1/2 the size of a diffusion image voxel, there may be a risk of "jumping" features that are defined at a smaller spatial scale. I therefore went for a compromise of 1/4 of a voxel size.

This brings run times with RK4 closer to that without. The fraction of streamlines retained vs. rejected is generally comparable between no RK4 / RK4 with 1/10th voxel step / RK4 with 1/4 voxel step. As far as results are concerned it's difficult to judge the tractograms as being any different.

(`iFOD1` and `sd_stream` outputs are looking particularly woeful however; the latter is probably related to #1630, may need to look into the former).

If this change is accepted it may alter the considerations within #1630.

There is also a further question in here with respect to the interaction between RK4 and the maximum angle. [Internally](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/src/dwi/tractography/tracking/exec.h#L486-L520) RK4 involves taking two fibre orientation estimates at half a step size away from the current vertex. Currently, no maximum angle constraint is applied at those two points (with the exception of `ifod1`, which needs its rejection sampling constrained); the maximum angle constraint is only applied based on the fibre orientation estimate at the new vertex. Question is: Should a restriction of half / all of the maximum angle constraint be applied to those two interior calculation points?

### DIFF
--- a/docs/reference/commands/tckgen.rst
+++ b/docs/reference/commands/tckgen.rst
@@ -53,7 +53,7 @@ Streamlines tractography options
 
 -  **-select number** set the desired number of streamlines to be selected by tckgen, after all selection criteria have been applied (i.e. inclusion/exclusion ROIs, min/max length, etc). tckgen will keep seeding streamlines until this number of streamlines have been selected, or the maximum allowed number of seeds has been exceeded (see -seeds option). By default, 5000 streamlines are to be selected. Set to zero to disable, which will result in streamlines being seeded until the number specified by -seeds has been reached.
 
--  **-step size** set the step size of the algorithm in mm (default is 0.1 x voxelsize; for iFOD2: 0.5 x voxelsize).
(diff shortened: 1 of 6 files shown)


Lestropie:
Superseded by #1833.

Issue #1746, opened 2019-09-30
New assertion failure in Eigen


Lestropie:
The following error has arisen in CI testing for `dev`. I have been unable to reproduce it locally despite checking out the very same Git branch of Eigen as is used in TravisCI.

```
# command: mkdir -p tmp-population_template-in/ && ln moving.mif.gz tmp-population_template-in/s1.mif.gz && ln template.mif.gz tmp-population_template-in/s2.mif.gz && population_template tmp-population_template-in/ tmp-population_template-out1.mif.gz -type rigid_nonlinear -rigid_scale 0.2,0.3 -rigid_niter 20,20 -nl_scale 0.2,0.3 -nl_niter 1,1 -scratch tmp-population_template-out1 -force [ ERROR ]
population_template: Generating a population-average template from 2 input images
population_template: [WARNING] no masks input. Use input masks to reduce computation time and improve robustness
mrinfo: /home/travis/build/MRtrix3/mrtrix3/../eigen/Eigen/src/Core/Transpose.h:378: static void Eigen::internal::checkTransposeAliasing_impl<Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Transpose<Eigen::Matrix<double, -1, -1, 0, -1, -1> >, true>::run(const Derived &, const OtherDerived &) [Derived = Eigen::Matrix<double, -1, -1, 0, -1, -1>, OtherDerived = Eigen::Transpose<Eigen::Matrix<double, -1, -1, 0, -1, -1> >, MightHaveTransposeAliasing = true]: Assertion `(!check_transpose_aliasing_run_time_selector <typename Derived::Scalar,blas_traits<Derived>::IsTransposed,OtherDerived> ::run(extract_data(dst), other)) && "aliasing detected during transposition, use transposeInPlace() " "or evaluate the rhs into a temporary using .eval()"' failed.
```

Lestropie:
Fixed by 5940137.

PR #1748, opened 2019-10-01
[DONOTMERGE] Activate AppVeyor debug

Lestropie:
Being used to investigate #1746.

PR is not to be merged; just activating a build so that I can RDP to the AppVeyor VM. Still can't reproduce the Eigen assertion failure on TravisCI / AppVeyor locally, so will try this instead.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,16 +5,20 @@ configuration: Release
 clone_folder: c:\msys64\mrtrix3
 shallow_clone: false
 
+environment:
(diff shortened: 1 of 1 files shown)


maxpietsch:
possibly related: http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1646

Lestropie:
Fixed; mistake was in #1738, was under the mistaken impression that I had seen a different PR to `dev` with the same issue.

PR #1754, opened 2019-10-03
MRView: report 3D texture size

jdtournier:
Turns out the max texture size isn't the same for 2D and 3D textures... We basically only use 3D textures in `mrview`, so that's what should be reported in the OpenGL dialog. Could have saved me a lot of time debugging a non-issue...

### DIFF
--- a/src/gui/dialog/opengl.cpp
+++ b/src/gui/dialog/opengl.cpp
@@ -61,8 +61,8 @@ namespace MR
         root->appendChild (new TreeItem ("VSync", format.swapInterval() ? "on" : "off", root));
         root->appendChild (new TreeItem ("Multisample anti-aliasing", format.samples() ? str(format.samples()).c_str() : "off", root));
 
-        gl::GetIntegerv (gl::MAX_TEXTURE_SIZE, &i);
-        root->appendChild (new TreeItem ("Maximum texture size", str (i), root));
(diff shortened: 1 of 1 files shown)


Lestropie:
> We basically only use 3D textures in `mrview` ...

Do you mean that `mrview` is the only place where we use 3D textures, or that within `mrview` only 3D textures are used? Because I would have thought the latter were nowhere near the truth. Maybe one is more likely to encounter an issue with size limits with 3D textures? Would have thought it would be better to simply report both.

jdtournier:
> Do you mean that mrview is the only place where we use 3D textures, or that within mrview only 3D textures are used? 

Both, really... In mrview, all image data is handled via 3D textures - even 2D slices. We could try to change the handling to only resort to 3D textures when we really need to, but it'll complicate the code somewhat. We only use 2D textures for non-imaging things like fonts, colourmaps, etc. 

> Maybe one is more likely to encounter an issue with size limits with 3D textures? 

Definitely. Typical 2D texture size is limited to 16384, but 3D texture size to 2048 (at least that's what it is on my system). 

> Would have thought it would be better to simply report both.

No point unless we actually use 2D textures to render 2D slices - which we currently don't... I'm not against the idea though, it's trivial to add. I'll get that done now.

PR #1763, opened 2019-10-08
connectomestats: Rename NBSE to TFNBS

Lestropie:
As raised on the [forum](https://community.mrtrix.org/t/group-connectivity-analysis/2333/3). We no longer have the authority to attempt to change the name of a published method, given the duration of time since its publication, our generally pessimistic outlook on its performance, and the subtlety of our criticism of the existing name; so we'll be sticking to this name from now on.

### DIFF
--- a/cmd/connectomestats.cpp
+++ b/cmd/connectomestats.cpp
@@ -50,6 +50,18 @@ void usage ()
 
   SYNOPSIS = "Connectome group-wise statistics at the edge level using non-parametric permutation testing";
 
+  DESCRIPTION
+  + "For the TFNBS algorithm, default parameters for statistical enhancement "
+    "have been set based on the work in: \n"
+    "Vinokur, L.; Zalesky, A.; Raffelt, D.; Smith, R.E. & Connelly, A. A Novel Threshold-Free Network-Based Statistics Method: Demonstration using Simulated Pathology. OHBM, 2015, 4144; \n"
+    "and: \n"
+    "Vinokur, L.; Zalesky, A.; Raffelt, D.; Smith, R.E. & Connelly, A. A novel threshold-free network-based statistical method: Demonstration and parameter optimisation using in vivo simulated pathology. In Proc ISMRM, 2015, 2846. \n"
+    "Note however that not only was the optimisation of these parameters not "
+    "very precise, but the outcomes of statistical inference (for both this "
+    "algorithm and the NBS method) can vary markedly for even small changes to "
+    "enhancement parameters. Therefore the specificity of results obtained using "
+    "either of these methods should be interpreted with caution.";
+
 
   ARGUMENTS
   + Argument ("input", "a text file listing the file names of the input connectomes").type_file_in ()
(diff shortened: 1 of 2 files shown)


thijsdhollander:
Not entirely sure, but isn't it a bit strange to have both

* If using the TFNBS algorithm: Baggio, H.C.; Abos, A.; Segura, B.; Campabadal, A.; Garcia-Diaz, A.; Uribe, C.; Compta, Y.; Marti, M.J.; Valldeoriola, F.; Junque, C. Statistical inference in brain graphs using threshold-free network-based statistics.HBM, 2018, 39, 2289-2302

* If using the TFNBS algorithm: Vinokur, L.; Zalesky, A.; Raffelt, D.; Smith, R.E. & Connelly, A. A Novel Threshold-Free Network-Based Statistics Method: Demonstration using Simulated Pathology. OHBM, 2015, 4144

...?  I get the intention (I think), and don't per se disagree, but it might be quite strange to most people who don't really have the full context available to them.  To be clear: we definitely need the name change to TFNBS, and the first relevant reference.  But I'm not sure whether it's beneficial to *anyone* to have the OHBM reference; that includes (each for their own unique reasons) the authors of the HBM paper, those of the OHBM abstract and finally users in general (and those users in the role of authors of their own work, where they'd then be citing both these references).  Might be worthwhile to consult with the authors of the OHBM abstract whether they'd like to see the latter reference pop up in statements along the likes of "...using the TFNBS algorithm (....et al.)" where more citations might be perceived as throwing more support behind the method.

Not going to add opinions here myself on what should or should not be done, but I thought it'd be useful to bring these elements up should they be of value in any way.

jdtournier:
Happy with the name change. Also agree with @thijsdhollander here, that the current reference list is not great. I'm happy to keep both though, as long as they're under the same heading...

Lestropie:
Still pretty salty that they published an abstract at OHBM a year after ours, describing exactly the same method, as though ours never existed. I wouldn't be that upset if publications cited the Baggio paper and not our abstract, but I like the idea of at least having that fact documented there.

Alternatively, one could consider that the Baggio paper fails to define default enhancement parameters, whereas that was one of the key goals of Lea's work. So if a user were to utilise that algorithm in *MRtrix3* specifically *with the default parameters*, that outcome would have been informed by Lea's work. Maybe if that condition were added to the reference text it would be sufficient contextualisation?

thijsdhollander:
Ok, I'm suddenly realising I might have caused some confusion: whenever above I said "OHBM abstract", I definitely referred to Lea's abstract (if that wasn't clear).  I'm not arguing either way, and generally I'm actually more for inclusion of references (with relevant context if needed).  What I think I tried to bring up, among a few things, is that maybe Lea and all or some co-authors on *her* abstract *might* not be keen to eventually see citations in the literature reading: *"...using the TFNBS algorithm (Baggio et al. 2018, Vinokur et al. 2015)"*, given current insights in the performance of the method.
So when I said *"Might be worthwhile to consult with the authors of the OHBM abstract whether they'd like to see....."*, I'm referring to the authors on Lea's abstract from back then (not Baggio et al.).

So I'm in no way being critical of the inclusion of the reference because of reasons of proper credit or anything; the credit is deserved for sure, but the authors might not actually want to get that kind of credit in that kind of context.  But to be clear: that fully depends on what those authors think about this.  I'm just bringing it up as something that might be an element to consider.

PR #1766, opened 2019-10-11
mtnormalise: alternative implementation

jdtournier:
This follows on from, and hopefully replaces #1756. 

I had a go at tweaking @JRosnarho's previous changes  to `mtnormalise` to see if I could squeeze a bit more performance out of it. I thought maybe all the interactions with images might actually be a performance bottleneck, so decided to rewrite the implementation using Eigen primitives only, avoiding interactions with images as much as possible (basically only at read and write times). Seems to have made quite a bit of difference, even though multi-threading is only used at read & write time - the meat of the algorithm is all single-threaded.

On the high-res dataset typical of the ones I'd like to use this with.

- Current implementation: 1m30s
- @JRosnarho's changes (#1756): 50s
- this pull request: 18s

Runtime on the BATMAN dataset is about 0.7s (about 4s with current implementation). 

There is a minor difference with the previous implementation, due entirely (as far as I can tell) to a slight difference in the way outliers are handled. In the previous implementation, outliers were removed from the mask, and not considered again (as far as I can tell). In this implementation, all voxels in the original mask are considered at every iteration - so voxels that might have been rejected initially may be re-included at a later stage. This is mostly because it keeps the estimation of the upper & lower quartiles simpler. I figured this is unlikely to cause much of a problem, unless the number of outliers becomes comparable to a quarter of the number of voxels in the mask, which would indicate a pretty bad mask, really... On my high-res data, the difference between the final masks produced by both approaches is only 16 voxels out of 1.8M (and 23 out of 92k for the BATMAN data).

I could try to match the handling of the outliers exactly, but I'm not sure it's worth it. Depends on whether there was a strong rationale behind the current strategy (maybe there's some weird datasets where this is a real problem...?).

---

Another question mark is whether there is genuine need to fully converge the balancing factor estimation & outlier rejection before estimating the field again. On my high-res data, the algorithm still hasn't quite converged after 15 full iterations (with max 8 balancing/outlier iterations per field update). Using 20 iterations with a max of 3 balancing/outlier rejection iterations per update gives faster runtimes, but gets much closer to convergence (on the BATMAN data, it's pretty much fully converged within 10-11 iterations regardless). 

Thoughts welcome, as always.


thijsdhollander:
> I had a go at tweaking @JRosnarho's previous changes to mtnormalise to see if I could squeeze a bit more performance out of it. I thought maybe all the interactions with images might actually be a performance bottleneck, so decided to rewrite the implementation using Eigen primitives only, avoiding interactions with images as much as possible (basically only at read and write times). Seems to have made quite a bit of difference, even though multi-threading is only used at read & write time - the meat of the algorithm is all single-threaded.

:+1: 

> In the previous implementation, outliers were removed from the mask, and not considered again (as far as I can tell).

No, that's not correct.  First things first; the intended design is absolutely to "reconsider" outliers, or to put it differently always re-estimate the full set of outliers again at each iteration (even each iteration of the "inner" loop where the tissue scaling factors are optimised).  As the other parameters' (tissue scaling factors and the normalisation/"bias" field) estimated values improve, it makes sense that the outliers can in turn be better estimated.  Also, if they weren't re-estimated "fresh" at each iteration, the algorithm wouldn't come as close (as it does now) to having the property that, if it converged sufficiently, and you feed the outputs of `mtnormalise` to `mtnormalise` again, the second time round the normalisation field is very or close to entirely spatially constant: because then the second new run of `mtnormalise` *would* re-estimate outliers fresh at the start, but at that point indeed re-include voxels that were in the final outlier set in the first run of `mtnormalise`, and thus the second run could easily optimise towards a new different optimum.

But more to the point: to make sure that the original implementation does indeed follow this intended design, I just double-checked it.  This can be relatively easily done by running with `-debug`, as it then opens an `mrview` showing the current mask (i.e. where outliers are removed) at each iteration (even inner loop).  I just did that, and saved the first 20 masks for a run with default parameters on a dataset; so well into the 3rd iteration of the outer loop.  Initially, most of the time more voxels are added to the set of outliers, but at almost all iterations (of the inner loop) there's also voxels that re-enter the mask.  So that confirms outliers are considered again at each and every iteration for sure.

Trying to think long and hard back to some of the details of the original design, I remember @rtabbara and myself had to check quite a few internal definitions of "the mask" though.  At the very start of the algorithm, it checks explicitly for anything non-physical that isn't compatible with estimating a multiplicative factor to end up as a positive given number.  So basically, any non-finite values are kicked out for sure.  Also voxels where all tissue signal amounts sum to zero or less are kicked out "forever".  I think that's it though.  So you've got:

1. the mask that the user feeds in via `-mask`
2. from that mask, non-physical voxels are kicked out (non-finite and sum to zero or less) "forever"; this then provides what I'll informally call "the basic mask"
3. "the basic mask" is always "reconsidered"; at each iteration, outliers are estimated and scale factors are estimated from "the basic mask minus outliers", but the next iteration, the outliers are freshly re-estimated from "the basic mask", yet now working with updated scale factors of course
4. the outer loop isn't much different: basically, the normalisation field is estimated given current outliers and current scale factors, and the next bunch of inner loops work with the updates normalisation field (and each time estimate outliers from "the basic mask' fresh again, and so on).

The only other thing about the outlier estimation is that it initially (before the first iteration; you could call it "initialisation" in a way) estimates those with a more generous threshold once.  I explained that at some point here on GitHub: there's an annoying balance to be found between dealing with huge outliers *or* dealing with very un-balanced tissues.  We wanted to cater to both extreme scenarios, but each of those makes the estimation of the other initially unsafe (outlier rejection would detect a whole "tissue" area as outliers with extremely unbalanced tissues; but tissue balancing would severely trip up in case of extreme outliers, and then effectively result in extremely unbalanced tissue factors).  So trying to find some sensible balance that caters for a wide range of each possible extreme scenario, we ended up to start with outlier rejection but with a more generous threshold.  The choice of starting with outlier rejection is considering that outliers (even extreme) would still be a more common "possible" scenario than hugely unbalanced tissues (the latter would almost always have to be an inappropriate multi-tissue model; you could reason the users is a bit to blame for that, in some ways).  But then starting with outlier rejection anyway due to the latter argument, still protecting ourselves a bit against unbalanced tissues, comes in the form of a more generous outlier threshold initially.  We found that in all possible scenarios and datasets we had access to, and even actively messing with some of those to make them worse / more challenging, this strategy was able to safely "ease" the algorithm into a workable situation.  Once it's in such a workable situation with a good / reasonable outlier guess and reasonably suitable tissue balance factors for that particular dataset (even if unbalanced), then it's good.  It's just about making sure it "sets off in the right direction"; if that makes sense.

> In this implementation, all voxels in the original mask are considered at every iteration - so voxels that might have been rejected initially may be re-included at a later stage.

So that's good, but it was also like that before.  If the output is different though, my initial worry would be that the difference lies at step 2 in the above list: those non-physical cases (non-finite and zero or lower summed voxels) should be excluded at the start, but also never ever be considered again.

Finally, the other risky thing I can remember is the domain you estimate outliers in... maybe?  That should also be the log-domain (well, everything that's estimated is basically estimated in the log-domain across the board).

> I could try to match the handling of the outliers exactly, but I'm not sure it's worth it. Depends on whether there was a strong rationale behind the current strategy (maybe there's some weird datasets where this is a real problem...?).

So in conclusion, I think you should've "exactly" matched it, at least as to the bits you've described...?  I know during development, at some point (because @rtabbara and I hadn't discussed that particular bit into sufficient detail at that very point, I think) our implementation was actually not re-considering outliers.  We stumbled upon that fact because there are indeed weird datasets or cases where *that* would be a problem. :slightly_smiling_face:   Especially when pushing the algorithm to be robust for the aforementioned challenging scenarios, you might initially "briefly" have quite a bit too many outliers.  But that should rectify itself when the algorithm sets of in the right direction.

So as far as I can tell, the way you intended it to work in this implementation, should actually have ended up matching what it was before as well.  If there's a difference somewhere in outcome, it would probably be due to small numerical imprecisions, or maybe due to some of those other "details" I mentioned above?  But as I mentioned, some of those other details are quite crucial; and also of the nature that you'd only become aware of it in more exceptional cases (where it might then fail potentially "quite drastically").

> Another question mark is whether there is genuine need to fully converge the balancing factor estimation & outlier rejection before estimating the field again.

No, not absolutely fully *per se*.  But it should be quite good though, otherwise you'll end up effectively "more inefficiently" over the outer loop iterations.  This is particularly the case because some tissues typically appear in distinct-ish spatial areas; i.e. CSF in a big bunch of voxels in the middle (the ventricles), lots of GM towards the outside (the cortex), and particular WM features, including outliers, again in coherent specific areas (CST and bits of CC are often outliers due to high values, etc...).  This matters because the bias field is spatially smooth.  So if the balance factors and outliers are "way off", then the bias field due to this particular spatial layout can really be quite far off at that point as well.  So in some ways, there is "a lot to gain" by making sure that outliers and balance factors are "as good as they can be" whenever estimating the bias field.

Other than that, in a way it makes the algorithm a bit better defined to describe (depending on how you think about this of course): you can say at a higher level of abstraction that at each outer loop iteration, it estimates 2 things in turn: the bias field, and the "tissue-factors-plus-outliers".  Each one given the latest estimation of the other.  Then going down into the abstract level of the latter, again 2 things are optimised given the latest estimation of the other (and the "fixed" estimation of the bias field, for one series of inner loops).  I'd agree though that this bit isn't a serious argument per se (it's just a bit more neat, I'd say).  But the *previous* paragraph does matter though!

> On my high-res data, the algorithm still hasn't quite converged after 15 full iterations (with max 8 balancing/outlier iterations per field update). Using 20 iterations with a max of 3 balancing/outlier rejection iterations per update gives faster runtimes, but gets much closer to convergence (on the BATMAN data, it's pretty much fully converged within 10-11 iterations regardless).

Yeah, sure.  When we chose those default values for number of inner and outer loop iterations, this was a bit of a trade off between everything we saw in all the data we had access to (this did vary quite broadly though).  Indeed, it's not strictly needed for the inner loop to converge.  That's also why, well, it's actually not implemented to require that in practice.  For the inner loop, it's easy enough to check true convergence by checking that the masks match exactly.  Yet, there's a cap that the user can set and we chose a default value of 7 for it.  So that's fine.  We noticed that the inner loops go toward convergence initially relatively quickly, but it can depend on the size of the data.  It'll definitely take more "inner" iterations to truly converge when you have more voxels (i.e. higher spatial resolution).  But as you indeed say, that becomes a bit nonsensical then, because the algorithm as a whole ends up "converging" faster if you *don't* let it converge all the way in the inner loop and move to runs of the outer loop more quickly.  So the point ends up being: there's some balance in terms of optimal running, *but* more crucially: that balance depends on lots of properties having to do with your data, and your particular setting (it'll again be different if there's more or less outliers, if there's only 2 tissue types, if any of the tissues are more or less presence (e.g. atrophy, etc...)).  We've seen that, because we tested across some of those things varying between datasets we used.

So the point ends up being here, we've determined some defaults that are ok, and furthermore are for a lot of data "on the safe side".  The "on the safe side" means: rather run a bit longer but get a bit of a better solution.  We reasoned that's fine, because the algorithm isn't that "heavy" and typically doesn't take long to run on typical resolution datasets out of the box.  When for those datasets, you choose to upsample the data, it's likely because you're doing a group analysis (e.g. FBA), which has far more "heavy" steps in any case, and you'll probably automate a lot of these steps and let it run.  If you have data that is out of the box high resolution, you'll likely be used to longer running times for some of the other steps in any typical preprocessing pipeline.  And finally, if you have exceptionally high resolution data, or data that is in any other way "exceptional", you'll probably want to be actively on top of setting some parameters yourself and not go with the defaults.

...and as I'm hoping to now end up this post, I'm realising we didn't allow the user to set the maximum number of inner loop iterations themselves (it defaults to 7, but it's also fixed at 7).  Ahh, ok, I now realise why you brought up the inner loop and its convergence.  Ok.  I'm not going to edit all the above to correct that; but I'll just pick one quote of myself above:

> Yet, there's a cap that the user can set and we chose a default value of 7 for it.

So that's not completely correct: the user doesn't have access to this yet.  Your scenario does show that there might be reasons for informed users to set this to optimise their automated pipelines to run even faster.  So it makes sense to provide the option.  I note that at the moment (e.g. on `dev` and `master`), there's only `-niter` for the outer loop (default at 15).  So maybe an `-inner_niter` (default at 7) can be offered as well.

thijsdhollander:
Small extra:

> (on the BATMAN data, it's pretty much fully converged within 10-11 iterations regardless)

Yeah, so that's what I mentioned above in terms of 

> So the point ends up being here, we've determined some defaults that are ok, and furthermore are for a lot of data "on the safe side". The "on the safe side" means: rather run a bit longer but get a bit of a better solution.

The safe side being that for a lot of data, the current 7 inner and 15 outer loops almost always get you to virtually full convergence for a lot of "common" (e.g. spatial resolution) data.

Note that we can detect convergence for the inner loop *exactly*, because a discrete thing (the set of outliers) is alternated with a non-discrete thing (tissue balance factors).  Once the discrete thing doesn't change over 2 subsequent iterations, the non-discrete thing also won't.  This is (yet) another reason as to why the algorithm was organised the way it is.  Even when you get in a run up to a point where the inner loop each time only runs 1 (inner) iteration for subsequent outer loops, and you'd conclude that the whole algorithm is has then converged, you might in fact be technically wrong: those balance factors and the bias field do still change, but probably very very very little (the outlier set not changing being evidence of little change happening to the rest).  But note that you might suddenly get to a point where these small changes do result in a voxel being classified differently as to outlier status (be it becoming outlier or no longer being outlier), and that may again cause the rest to change a little.

In any case, all of that matters very little if at least the number of *outer* loops is sufficiently large to converge sufficiently.  The default of 15 is safe for a wide range of data; and you might indeed have good reasons to "manually" set it to a larger value for exceptionally high spatial resolution data.  The default of 7 max. for the inner loop is a generous "safe" default value.  Lots of scenarios that are well enough behaved might be fine with less than those 7; but I'd rather have the default be safe.  We've seen some weird outliers here over the years in some cases. :man_shrugging: 

But all that just to also support again that an option that allows the user to override that 7 max. value for inner loops is certainly a sensible thing to offer.

jdtournier:
OK, glad to hear I'd misunderstood how the outlier rejection worked in the current version - all good. So the only reason for the difference then is the subtle difference in how the quartiles are computed. In the current version, only voxels in the current mask are used to compute the upper & lower quartiles. In this proposal, all voxels in the mask are included in that calculation (it makes things a fair bit easier). My feeling is that as long as the numbers of outliers isn't too huge, this should affect the overall performance in any noticeable way - and this is more or less what I've observed so far. 

Otherwise, pretty much everything is (or should be) strictly identical, including clipping values to non-negative values, the initial rejection of extreme outliers, etc. 

I'll add an option to tweak the number of iterations for the inner loop, I agree it would be good to have.

thijsdhollander:
> In the current version, only voxels in the current mask are used to compute the upper & lower quartiles. In this proposal, all voxels in the mask are included in that calculation (it makes things a fair bit easier). My feeling is that as long as the numbers of outliers isn't too huge, this shouldn't affect the overall performance in any noticeable way - and this is more or less what I've observed so far.

Ah ok, another subtlety I hadn't appreciated at first.  But I agree; I can't imagine a scenario where this would be a problem.  And if it were, the algorithm would likely be in trouble from the start anyway, so such a scenario would simply be too extreme a priori.

This then does explain that the result might not be *exactly* the same as before; but not in a problematic way, so should be all good.

PR #1771, opened 2019-10-17
JSON: import simple array as a single header entry

jdtournier:
Apparently converting data via `dcm2niix` and converting to `mif` with a JSON import results in the slice timings being represented as multiple separate header entries, rather than the single entry with a comma-separated list that a direct `mrconvert` call produces. This means a call to `dwipreproc` fails when used with slice to vol, with error message:
```
dwipreproc: No slice encoding direction information present; assuming third axis corresponds to slices
  File "/home/finn/Software/mrtrix3/bin/dwipreproc", line 233, in <module>
    app.error('Cannot use slice timing information in image header for slice-to-volume correction: Number of entries (' + len(slice_timing) + ') does not match number of slices (' + dwi_header.size()[2] + ')')
TypeError: cannot concatenate 'str' and 'int' objects
```
Note there's another fix here for that python error...

This pull request modifies the handling of the import of JSON arrays, so that:

- if the array contains **no** sub-arrays, it'll concatenate into a single comma-separated list and insert the information as a single header entry;
- if the array contains **only** sub-arrays, it'll concatenate each sub-array into a comma-separated list and insert the information as one header entry per sub-array
- if it contains a mixture of simple elements and arrays, it'll throw an Exception

That last behaviour might need to be revised depending on whether there are any instances where we would reasonably be expected to handle a mix of elements and arrays. Alternatives are:

- issue a warning and ignore
- issue a warning and insert as one entry per element or sub-array. 

But I'll admit I don't know enough about BIDS to make an informed decision here. I just expect that if we ever encounter a BIDS entry that consists of such a mix, we'll have trouble interpreting it correctly anyway...

### DIFF
--- a/bin/dwipreproc
+++ b/bin/dwipreproc
@@ -230,7 +230,7 @@ if eddy_mporder:
     slice_timing = dwi_header.keyval()['SliceTiming'][0]
     app.var(slice_timing)
     if len(slice_timing) != dwi_num_slices:
-      app.error('Cannot use slice timing information in image header for slice-to-volume correction: Number of entries (' + len(slice_timing) + ') does not match number of slices (' + dwi_header.size()[2] + ')')
+      app.error('Cannot use slice timing information in image header for slice-to-volume correction: Number of entries (' + str(len(slice_timing)) + ') does not match number of slices (' + str(dwi_header.size()[2]) + ')')
 
 # Use new features of dirstat to query the quality of the diffusion acquisition scheme
 # Need to know the mean b-value in each shell, and the asymmetry value of each shell
--- a/core/file/json_utils.cpp
+++ b/core/file/json_utils.cpp
@@ -54,24 +54,33 @@ namespace MR
             H.keyval().insert (std::make_pair (i.key(), str<int>(i.value())));
           } else if (i->is_number_float()) {
             H.keyval().insert (std::make_pair (i.key(), str<float>(i.value())));
+          } else if (i->is_string()) {
+            H.keyval().insert (std::make_pair (i.key(), str (i.value())));
           } else if (i->is_array()) {
-            vector<std::string> s;
-            for (auto j = i->cbegin(); j != i->cend(); ++j) {
-              if (j->is_array()) {
+            size_t num_array = 0;
+            for (const auto j : *i)
+              if (j.is_array())
+                ++num_array;
+
+            if (num_array == 0) {
+              vector<std::string> line;
+              for (const auto k : *i)
+                line.push_back (str(k));
+              H.keyval().insert (std::make_pair (i.key(), join (line, ",")));
+            }
+            else if (num_array == i->size()) {
+              vector<std::string> s;
+              for (const auto j : *i) {
                 vector<std::string> line;
-                for (auto k : *j)
+                for (const auto k : j)
                   line.push_back (str(k));
                 s.push_back (join(line, ","));
-              } else {
-                s.push_back (str(*j));
               }
+              H.keyval().insert (std::make_pair (i.key(), join(s, "\n")));
             }
-            H.keyval().insert (std::make_pair (i.key(), join(s, "\n")));
-          } else if (i->is_string()) {
-            const std::string s = i.value();
-            H.keyval().insert (std::make_pair (i.key(), s));
+            else
+              throw Exception ("JSON entry contains mixture of elements and arrays");
           }
-
         }
 
         if (!Header::do_not_realign_transform) {


Lestropie:
> if the array contains no sub-arrays, it'll concatenate into a single **comma-separated** list and insert the information as a single header entry;
> if the array contains only sub-arrays, it'll concatenate each sub-array into a **comma-separated** list

With #1735 (specifically [here](https://github.com/MRtrix3/mrtrix3/pull/1735/commits/2ff43df55d7ab3c1ded817333c286abcb05fa20f#diff-bed1352315cfa1a544ec2f1d9fd643a2)) and #1738 I found that in order to have the JSON export operate as intended for slice timing information extracted from DICOM, such array data needed to be saved as *space-separated* rather than comma-separated in the header. If it's comma-separated, the JSON `.dump()` function generates a list of lists with one entry (demonstration in #1738). What does the JSON data output by `dcm2niix` look like with respect to #1738?

jdtournier:
This is all based on on-going discussions with [Finn Lennartsson](https://community.mrtrix.org/u/finn/summary) - might be worth involving him in the discussion, but I don't think he has a GitHub account... 

In any case, this is what `dcm2niix` produces for that entry:
```
"SliceTiming": [
0,
2.7625,
1.745,
0.7275,
3.49,
2.4725,
1.455,
0.4375,
3.2,
2.1825,
1.165,
0.145,
2.9075,
1.89,
0.8725,
3.635,
2.6175,
1.6,
0.5825,
3.345,
2.3275,
1.31,
0.2925,
3.055,
2.035,
1.0175,
0,
2.7625,
1.745,
0.7275,
3.49,
2.4725,
1.455,
0.4375,
3.2,
2.1825,
1.165,
0.145,
2.9075,
1.89,
0.8725,
3.635,
2.6175,
1.6,
0.5825,
3.345,
2.3275,
1.31,
0.2925,
3.055,
2.035,
1.0175 ],
```

And on the current `master` branch, this gets inserted as one value per `SliceTiming` entry using `-json_import`:
```
$ mrinfo -property SliceTiming dwi_raw.mif.gz
0
2.7625
1.745
0.7275
3.49
2.4725
1.455
...
```

Using the direct DICOM import all within _MRtrix3_, this ends up as a single comma-separated list of values in a single entry:
```
$ mrinfo -property SliceTiming dwi_raw.mif.gz
0,2.76250005,1.745,0.727500021,3.49000001,2.47250009,1.45500004,0.4375,3.20000005,2.18249989,1.16499996,0.144999996,2.90750003,1.88999999,0.872500002,3.63499999,2.61750007,1.60000002,0.582499981,3.34500003,2.3275001,1.30999994,0.292499989,3.05500007,2.03500009,1.01750004,0,2.76250005,1.745,0.727500021,3.49000001,2.47250009,1.45500004,0.4375,3.20000005,2.18249989,1.16499996,0.144999996,2.90750003,1.88999999,0.872500002,3.63499999,2.61750007,1.60000002,0.582499981,3.34500003,2.3275001,1.30999994,0.292499989,3.05500007,2.03500009,1.01750004
```

I guess #1735 and #1738 will definitely interact here, but I'm unsure about the details...

Lestropie:
I suspect there's a number of gremlins arising from the fact that I hurriedly got the JSON handling working well enough to be able to produce and process data that looked like BIDS, but wasn't cross-checked for compatibility with other packages.

#1735 is a fairly gross deviation from what the data should have been, for which I put a hack into `dwipreproc` to compensate.

#1738 better demonstrates the fundamental issue: JSON supports multiple data types and hierarchical data, whereas we internally store any non-compulsory header data as a `std::map<std::string, std::string>`. We therefore need to carefully consider how that conversion should take place in both directions. #1738 does a decent amount of gymnastics to make this work for export in what I think is a sensible way. This issue shows that the import stage needs more careful consideration as well more generally.

In terms of this specific case, one might ask: Given these data contain one entry per volume, and other header entries containing one entry per volume (e.g. `dw_scheme`) have one header entry (/ row) per volume, should the same be done for `SliceTiming`, even if it's only a single scalar value per volume? I tend to think that collapsing such data into a single row makes sense (in #1735 I put a [bunch of logic gymnastics in place in `dwipreproc`](https://github.com/MRtrix3/mrtrix3/pull/1735/files#diff-9a2510352b82dd6f73530274c31a7e1f) just in case the data come in in unexpected formats), but that's nevertheless an explicit decision. But the reason I made that first comment is that if you make these changes in order for JSON *import* of `SliceTiming` to work as intended, but comma-delimit the string data in the header, then upon JSON *export*, `SliceTiming` will come out as a list of lists of floats rather than a list of floats.

Ideally we should have a unit test that imports a pre-prepared JSON file (with as many nasties as we wish to support) to an empty header, exports it, then raw diffs the JSON contents.

Lestropie:
Aborting; given magnitude of issues with JSON data handling, will not be resolving such on `master`, instead relying on imminent tag update that will include changes in #1843.

PR #1788, opened 2019-10-24
mrinfo -imagepath option

maxpietsch:
Added an option that prints the filename as the first output:
```
mrinfo f/* -datatype -imagepath
f/1.mif
Float32LE
f/2.mif
UInt8
...
```

Could be extended/modified to add more functionality to dereference symbolic links or output the absolute path once we go to C++17.


jdtournier:
I like the idea, but I'd favour a less clunky name for the option, e.g. `-path` or `-name`.

FWIW, `-name` is what the `find` Unix command uses for a similar purpose...

maxpietsch:
4586cce changes the option name to `-name` and modifies the description header to `Image name:          "1.mif"` instead of `Image     :          "1.mif"` to make the option easier to guess.

```
$mrinfo 1.mif
Image name:          "1.mif"
  Dimensions:        50 x 50 x 32 x 45
  Voxel size:        3.0303 x 3.0303 x 3 x 1
  Data strides:      [ -2 3 4 1 ]
  Format:            MRtrix
  Data type:         32 bit float (little endian)
  Transform:                    1          -0           0      -77.66
                               -0           1           0       -53.7
                               -0          -0           1      -39.21

$ mrinfo ?.mif -name -size
0.mif
50 50 32 1
1.mif
50 50 32 45
```

cmd/mrinfo.cpp
```diff
@@ -274,6 +276,7 @@ void run ()
     else if (export_grad || check_option_group (GradImportOptions) || dwgrad || shell_bvalues || shell_sizes)
       DWI::set_DW_scheme (header, DWI::get_valid_DW_scheme (header, true));
 
+    if (name)       std::cout << argument[i] << "\n";
```
Lestropie: Note that in some instances this will give a different result to `header.name()`. Would the latter make more sense, i.e. as with other options, providing the header data "as *MRtrix3* sees it"?

maxpietsch: Not sure what the instances are in which there is a difference between argument and name but `header.name()` is now probably more in line with the option name.

jdtournier: The main reason `header.name()` might differ from the argument given is with DICOM data: in this case, the `header.name()` is formed from the DICOM headers. 

Personally, I'm happy either way, though.

Lestropie: I think was actually referring to number sequence usage; looks like it comes out the same though.

PR #1797, opened 2019-11-10
mrview: fixel --> vector field and ROI opacity

maxpietsch:
extracting some mrview-related changes from the `mrreg` branch that I found sensible:

- added config option to change _fixels_ into _vector_ field (origin at voxel centre), allow greated dynamic range of fixel length
- ROI alpha defaults to 0.5 instead of 1. config file option added to change the default.

### DIFF
--- a/docs/reference/config_file_options.rst
+++ b/docs/reference/config_file_options.rst
@@ -357,6 +357,12 @@ List of MRtrix3 configuration file options
 
      The factor by which the ODF overlay is scaled.
 
+.. option:: MRViewRoiAlpha
+
+    *default: 0.5*
+
+     The default alpha of a ROI overlay.
+
 .. option:: MRViewRotateModifierKey
 
     *default: ctrl*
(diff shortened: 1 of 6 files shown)


Lestropie:
I don't find "fixels are vectors" to be an intuitive name for that. It's more like whether or not fixels are unidirectional. Would be preferable to have that as a check box if possible.

maxpietsch:
Voila
Screenshot 2019-11-11 14 58 33

Screenshot 2019-11-11 14 58 43

Would be good if someone more versed with the intricacies of mrview double checked my implementation, in particular [displayable.h](https://github.com/MRtrix3/mrtrix3/pull/1797/commits/cbcb065d85604cb7a8b488bb5af3d5595828a7a0#diff-21ae34a81917bb6d480413271aa3af2cR44).

jdtournier:
Superseded by #1812 - closing.

PR #1798, opened 2019-11-10
adding mrregister / population_template features

maxpietsch:
I stripped unfinished features out of the `mrreg` branch (see #1575) to bring nan-masking, robust initialisation and multi-contrast-, weighted- and leave-one-out-registration into `dev`. 

This PR contains nearly all registration-specific features of #1575 with the exception of the splitting of `mrregister` into multiple binaries and related code which remains in the `mrreg` branch.

Notable changes besides pure registration code are the command line parsing accepting paired inputs (+ separated) and a new filter in `mrfilter` that produces images with clipped intensities using z-scores which can be useful for initial alignment in `population_template`. 

I am not sure if this should be part of the 3.0 release given the large number of changes. I've been using the multi-contrast registration for a while now but broader testing will likely uncover rough edges / bugs in the user interfaces.


jdtournier:
Personally, I think you should go ahead and merge. This is a massive changeset, and there's no way any of us are going to have the resources to inspect this properly. The only serious testing this is going to get is in the community... 

And if you are introducing large changes, now is the time, before we push out a major release. Yes, there's likely to be issues that'll crop up down the line. But that's invariably the case. Users understand this.

This does however mean that you have to fairly confident what you're adding is sensible and robust - we'll be directing all relevant bug reports and support queries your way... :stuck_out_tongue_winking_eye:

bin/population_template
```diff
@@ -33,48 +28,58 @@ DEFAULT_NL_SCALES = [0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0
...
-  linoptions.add_argument('-affine_lmax', help='Specify the lmax used for affine registration for each scale factor, in the form of a list of integers (default: %s). The list must be the same length as the linear_scale factor list' % ','.join([str(x) for x in DEFAULT_AFFINE_LMAX]))
-  linoptions.add_argument('-affine_niter', help='Specify the number of registration iterations used within each level before updating the template, in the form of a list of integers (default:500 for each scale). This must be a single number or a list of same length as the linear_scale factor list')
+  linoptions.add_argument('-linear_estimator', help='Choose estimator for intensity difference metric. Valid choices are: l1 (least absolute: |x|), l2 (ordinary least squares), lp (least powers: |x|^1.2), Default: None (no robust estimator used)')
+  linoptions.add_argument('-rigid_scale', help='Specifiy the multi-resolution pyramid used to build the rigid template, in the form of a list of scale factors (default: %s). This and affine_scale implicitly  define the number of template levels' % ','.join([str(x) for x in DEFAULT_RIGID_SCALES]))
```
Lestropie: *Specify

bin/population_template
```diff
@@ -84,168 +89,613 @@ except NameError:
...
+          res[name] = var.split()
+    return res
 
   good = True
   run.command('transformcalc ' + transformation + ' decompose ' + transformation + 'decomp')
-  if not os.path.isfile(transformation + 'decomp'): # does not exist if run with -continue option
+  if not os.path.isfile(transformation + 'decomp'):  # does not exist if run with -continue option
     app.console(transformation + 'decomp not found. skipping check')
     return True
-  data = utils.load_keyval(transformation + 'decomp')
```
Lestropie: What's the rationale for `load_key_value()` rather than `utils.load_keyval()`? It seems to be doing an almost identical function, except that the version here is skipping lines commencing with "`#`" (which is where key-values go in text files). If there's some unmet need, would rather address it in the library function.

bin/population_template
```diff
@@ -255,38 +705,55 @@ def execute(): #pylint: disable=unused-variable
...
 
   if app.ARGS.linear_transformations_dir:
     if not dolinear:
       raise MRtrixError("linear option set when no linear registration is performed")
     app.ARGS.linear_transformations_dir = relpath(app.ARGS.linear_transformations_dir)
     app.check_output_path(app.ARGS.linear_transformations_dir)
 
+  # automatically detect SH series in each contrast
+  do_fod_registration = False  # in any contrast
+  CON.n_volumes = []
```
Lestropie: Capitalisation is intended to discriminate between variable vs. (effective) constants, not global vs. local

cmd/mrregister.cpp
```diff
@@ -107,24 +118,49 @@ void usage ()
...
 }
 
 using value_type = double;
 
+void run () {
 
+  vector<Header> input1, input2;
+  const size_t n_images = 1 + (argument.size() - 2 ) / 3;
+  { // parse arguments and load input headers
+    if (MR::abs(float (n_images) - (1.0 + float (argument.size() - 2 ) / 3.0)) > 1.e-6) {
```
Lestropie: This would make more sense as a modulo operation

cmd/mrregister.cpp
```diff
@@ -164,72 +200,138 @@ void run ()
...
+    if (nvols1 != nvols2)
+      throw Exception ("input images do not have the same number of volumes: " + str(input2[i].name()) + " and " + str(input1[i].name()));
+
+    // set do_reorientation and image_lmax
+    if (nvols1 == 1) { // 3D or one volume
+      mc_params[i].do_reorientation = false;
+      mc_params[i].image_lmax = 0;
+      CONSOLE ("3D input pair "+input1[i].name()+", "+input2[i].name());
+    } else { // more than one volume
+      value_type val = (std::sqrt (float (1 + 8 * nvols1)) - 3.0) / 4.0;
```
Lestropie: Similar to previous: an `SH::LforN()` followed by a modulo would make more sense when reading the code

cmd/mrregister.cpp
```diff
@@ -846,77 +1027,154 @@ void run ()
...
+      {
+        // LogLevelLatch log_level (0);
+        Image<value_type> im1_image = Image<value_type>::open (input1[idx].name());
+
+        Header transformed_header (input2[idx]);
+        transformed_header.datatype() = DataType::from_command_line (DataType::Float32);
+        Image<value_type> im1_transformed = Image<value_type>::create (im1_transformed_paths[idx], transformed_header);
+
+        const size_t nvols = im1_image.ndim() == 3 ? 1 : im1_image.size(3);
+        const value_type val = (std::sqrt (float (1 + 8 * nvols)) - 3.0) / 4.0;
```
Lestropie: Similar to previous

src/registration/metric/demons4D.h
```diff
@@ -29,26 +30,40 @@ namespace MR
...
+                     const Im1ImageType& im1_image, const Im2ImageType& im2_image, const Im1MaskType im1_mask, const Im2MaskType im2_mask,
+                     const vector<MultiContrastSetting>* contrast_settings = nullptr) :
                        global_cost (global_energy),
                        global_voxel_count (global_voxel_count),
                        thread_cost (0.0),
                        thread_voxel_count (0),
                        mutex (new std::mutex),
                        normaliser (0.0),
-                       robustness_parameter (-1.e12),
+                       robustness_parameter (1.e-12), // MP: used to be -1.e12
```
Lestropie: Does this require greater investigation / tracking? I.e. was this a bug that requires reporting?

maxpietsch: I guess Dave put this parameter in the code to prevent floating-point rounding errors in the cost function value to affect the convergence check. I am not sure if it makes a difference in practice and if so I'd expect the code to terminate earlier at roughly the same solution (relevant [code](https://github.com/MRtrix3/mrtrix3/blob/b9c0114afbed727655cb59b80458fdb0987debf8/src/registration/metric/demons.h#L106-L123)).

bin/population_template
maxpietsch: The reason that this does not use `utils.load_keyval` is that this code is 3 years older (0f85864c542908e7330c0e1e). It's still there because the output format of `transformcalc decompose` is not the key-value format. We might want to change that?
The current format is:
```text
# transformcalc trafo.txt decompose trafo.dec  (version=3.0_RC3-1182-gcbcb065d)
scaling: 1 0.999999999999999 0.999999999999999
shear: 3.05311331771918e-16 -8.32667268468867e-17 1.66533453693773e-16
angles: 0.443338396922925 0.0206736092750134 -0.0770429056771817
angle_axis: 0.449556495888172 0.983870564432754 0.0831931410457268 -0.158359128962367
translation: 0.114196353312504 -1.08436236954077 -0.922431985054211
R: 0.996820605461134 0.0769502650670704 0.0206721366628588
R: -0.0606847655421982 0.901327603695001 -0.428865841550147
R: -0.051633707587086 0.426247824068974 0.903131636428088
S: 1 3.05311331771918e-16 -8.32667268468867e-17
S: 1.87350135405495e-16 0.999999999999999 1.66533453693773e-16
S: -2.70616862252382e-16 1.11022302462516e-16 0.999999999999999
jacobian_det: 1
```

cmd/mrregister.cpp
maxpietsch: I can change line 133 to `if (n_images - 1) * 3 + 2 != argument.size()` if you don't like the floating-point operations but not sure how to use the modulo to make it more intuitive? The arguments are paired input images (subject i and j), with contrast pairs separated by '+': 
For instance: `['i1', 'j1', '+', 'i2', 'j2', '+', 'i3', 'j3']`.

bin/population_template
maxpietsch: Never mind, I misread the code of `utils.load_keyval`. Change is in a21cebd07788

Lestropie: Actually that was me forgetting about the nature of the output of `transformcalc decompose`: `utils.load_keyval()` is supposed to only load key-value pairs from commented lines. Maybe if you were to provide kwarg "`comments=''`", and then explicitly remove any commented lines yourself, it might work? Or the internals of `utils.load_keyval()` could be tweaked to make this work?

I don't think the output of `transformcalc decompose` should be changed, since those key-value pairs are the principal output of the command rather than sidecar information, so it makes sense to not be commented.

cmd/mrregister.cpp
Lestropie: For the case of SH data, the logic that makes most sense to me is something like what's used in [`Math::SH::check()`](https://github.com/MRtrix3/mrtrix3/blob/dev/core/math/SH.h#L696-L698): Map number of volumes to *lmax*, then check that *lmax* maps to the number of volumes. We could put such functions in a module if it'd be useful. What you've got here is the same functionality, just not functionalised.

What I was thinking in terms of a modulo operation basically avoids having to derive the inverse function: take the expression for `n_images`, replace the division with a modulo, and make sure it's zero.

bin/population_template
maxpietsch: Seems to be a feature / bug in `utils.load_keyval` :P
```
utils.load_keyval('trafo.dec')
Out[2]:
{'scaling': ['1', '0.999999999999999', '0.999999999999999'],
 'shear': ['3.05311331771918e-16',
  '-8.32667268468867e-17',
  '1.66533453693773e-16'],
 'angles': ['0.443338396922925', '0.0206736092750134', '-0.0770429056771817'],
 'angle_axis': ['0.449556495888172',
  '0.983870564432754',
  '0.0831931410457268',
  '-0.158359128962367'],
 'translation': ['0.114196353312504',
  '-1.08436236954077',
  '-0.922431985054211'],
 'R': ['0.996820605461134',
  '0.0769502650670704',
  '0.0206721366628588',
  ['-0.0606847655421982', '0.901327603695001', '-0.428865841550147'],
  ['-0.051633707587086', '0.426247824068974', '0.903131636428088']],
 'S': ['1',
  '3.05311331771918e-16',
  '-8.32667268468867e-17',
  ['1.87350135405495e-16', '0.999999999999999', '1.66533453693773e-16'],
  ['-2.70616862252382e-16', '1.11022302462516e-16', '0.999999999999999']],
 'jacobian_det': ['1']}
```

cmd/mrregister.cpp
maxpietsch: I think you refer to the number of volumes check. That's basically how I implemented it [link](https://github.com/MRtrix3/mrtrix3/pull/1798/commits/d854e3320a2450c22c7b38f1b1f139671ad94405#diff-2a6687b787e993424fe0d35053e28d72R251).

bin/population_template
Lestropie: It's probably something that I didn't think through enough, since this is the only location where it's used. I probably also wrote it trying to use the same kwargs as `numpy.loadtxt()`, but in this case slightly different behaviour is probably needed: i.e. "lead characters / strings that will result in *not* reading the line" v.s. "lead characters / strings that *will* result in reading the line", and if the second is empty, then any line not prefixed with an entry in the first will be read. That would permit both reading the output of `transformcalc decompose` (which it seems to do currently), and reading key-values from commented headers (which I suspect was my cognitively dissonant intention).

cmd/mrregister.cpp
Lestropie: I was thinking since that logic appears within `population_template`, Python versions of those functions might be handy.

maxpietsch:
FYI, I am currently running regression tests for `population_template` against `dev`, will merge if these pass to satisfactory precision.

bjeurissen:
Started testing ``population_template`` and ran into the issue that it currently does not support 4D inputs with a size of 1 in the fourth dimension.

Currently, this causes fod_reorientation to be activated erroneously. It is easy to get around this error by only allowing fod_reorientation to be activated if the 4th dimension has a size larger than 1, but then the first ``mrregister`` will still fail because it tries to register a 4D (nx,ny,nz,1) image and a 3D (nx,ny,nz) image..., because the template image is plain 3D.

I am not sure what the best strategy is to get around this. Make sure that all intermediate results also remain 4D, fix ``mrregister`` to handle such cases, remove the 4th dimensions of all the inputs, ... ?

bjeurissen:
> I am not sure what the best strategy is to get around this. Make sure that all intermediate results also remain 4D, fix `mrregister` to handle such cases, remove the 4th dimensions of all the inputs, demand that input to population template is either truly 4D or 3D to begin with, not generate such 4D images to begin with, ... ?

I have a feeling that it might be good to have a generic mechanism in place that can be reused in all situations like this and that allows simply treating a pseudo 4D image as if it were a 3D image if doing otherwise would cause problems.

bjeurissen:
During testing, I noticed that common prefixes in the input files get stripped in the output files.

So ``fod/PAT001.mif``with ``PAT`` a common prefix for all files in ``fod`` will result into warps being output e.g. as ``warp/001.mif``, which I think is not desirable. I think it makes more sense to keep the same filenames in the output as in the input.

bjeurissen:
From the ``population_template`` docs, it is currently unclear how the ``-transformed_dir`` option deals with multiple contrasts.

bjeurissen:
I just realized that https://github.com/MRtrix3/mrtrix3/pull/1798/commits/ebb3b5e2c84b64019d023503d3dcdf5e49d964af only keeps the common prefix, but still strips common suffixes, so it does not really address my concern... Why are these pre- and suffixes being stripped? Should we not simply strip the file extensions and leave everything else as is?

maxpietsch:
I strip pre- and postfixes in each contrast independently and match only the non-common part of filenames across contrasts and masks to allow input structures such as
```
odfA
├── contrastA_0_normalised.mif.gz
├── contrastA_1_normalised.mif.gz
└── contrastA_2_normalised.mif.gz
odfB
├── contrastB_0_suffixB.mif.gz
├── contrastB_1_suffixB.mif.gz
└── contrastB_2_suffixB.mif.gz
mask
├── mask_0.mif.gz
├── mask_1.mif.gz
└── mask_2.mif.gz
```

I'd rather not remove that functionality but use the stored contrast-specific pre- and postfixes when writing the output.

bjeurissen:
> I'd rather not remove that functionality but use the stored contrast-specific pre- and postfixes when writing the output.

That makes sense!

bjeurissen:
> > I'd rather not remove that functionality but use the stored contrast-specific pre- and postfixes when writing the output.
> 
> That makes sense!

Thinking about this more, what will happen in the following situation:

```
odfA
├── PAT00.mif
├── PAT01.mif
└── PAT02.mif
odfB
├── PAT00.mif
├── PAT01.mif
└── PAT02.mif
mask
├── PAT00.mif
├── PAT01.mif
└── PAT02.mif
```

which I think makes a lot of sense. It will then have to figure out that those PAT prefixes are not contrast specific?

Note that something like this could already happen if your subject IDs start with common leading numbers, you definitely do not want to lose those.

Given that each contrast has to be provided in its own folder, is it not easier to just demand that every contrast folder contains exactly the same filenames?

bjeurissen:
> I strip pre- and postfixes in each contrast independently and match only the non-common part of filenames across contrasts and masks to allow input structures such as
> 
> ```
> odfA
> ├── contrastA_0_normalised.mif.gz
> ├── contrastA_1_normalised.mif.gz
> └── contrastA_2_normalised.mif.gz
> odfB
> ├── contrastB_0_suffixB.mif.gz
> ├── contrastB_1_suffixB.mif.gz
> └── contrastB_2_suffixB.mif.gz
> mask
> ├── mask_0.mif.gz
> ├── mask_1.mif.gz
> └── mask_2.mif.gz
> ```
> 
> I'd rather not remove that functionality but use the stored contrast-specific pre- and postfixes when writing the output.

I don't see how one can make a sensible decision on the output filenames of e.g. the warps when each folder of input filenames can have different pre and post-fixes.

Even if we can come up with something, the current approach will leave us with multiple folders describing the same subjects, but each with different filenames, which will make any subsequent operations on those files very tricky. E.g.  transforming the subjects in a specific contrast folder using the warps in a warp folder, will become quite a headache, instead of just a simple foreach.

bjeurissen:
After running

```
population_template -mask_dir mask/ -warp_dir warp/ \
contrast1/ contrast1_template.mif \
+ contrast2/ contrast2_template.mif \
+ contrast3/ contrast3_template.mif
```

```
mrtransform contrast1/01.mif -warp_full warp/01.mif contrast1_warped/01.mif -template contrast1_template.mif 
```

produces

```
[ERROR] no linear transform found in initialisation syn warps image header
```

Is this expected?

maxpietsch:
No, that's a bug. The `mrconvert_keyval` in the `run.command` strips the linear transformations from the header, fixed in 1142b13.

maxpietsch:
> I don't see how one can make a sensible decision on the output filenames of e.g. the warps when each folder of input filenames can have different pre and post-fixes.

I'd use the contrast specific common pre- and postfix for output images and cross-contrasts common pre- and postfix for warps and linear transformations. This makes most sense I reckon and preserves the old behaviour.

bjeurissen:
> I'd use the contrast specific common pre- and postfix for output images and cross-contrasts common pre- and postfix for warps and linear transformations. This makes most sense I reckon and preserves the old behaviour.

If you keep track of all of those, that makes sense :)

maxpietsch:
With 4dfb2d2 warps and linear trafos now use the common pre- and postfix across contrasts and subjects -- excluding whitespace and stripped of leading or trailing underscores and `-`.

Input:
```
odfA
├── cA _0_fA 2.nii
├── cA _1_fA 2.nii
└── cA _2_fA 2.nii
odfB
├── cB _0_fB 2.mif.gz
├── cB _1_fB 2.mif.gz
└── cB _2_fB 2.mif.gz
mask
├── mask_0_fM.mif.gz
├── mask_1_fM.mif.gz
└── mask_2_fM.mif.gz
```
output:
```
warp
├── 0_f.mif
├── 1_f.mif
└── 2_f.mif
lin
├── 0_f.txt
├── 1_f.txt
└── 2_f.txt
transformedA
├── cA _0_fA 2.nii
├── cA _1_fA 2.nii
└── cA _2_fA 2.nii
transformedB
├── cB _0_fB 2.mif.gz
├── cB _1_fB 2.mif.gz
└── cB _2_fB 2.mif.gz
```

maxpietsch:
I did some digging for the origin of the differences in `mrregister` results between master and this branch. 

TL/DR: All good as far as I can tell with `mrregister`. The differences after accounting for bugfixes (d168540d90820ce2f7c91d27b04c7efbc8c0c04e, 58728ba51216c0b2d332f8f333ee403ff4c3a7bd) are identical or close to numerical precision.

`population_template` results are expected to be different for instance due to changes in FOV (tighter crop in mrreg_lite2) and changes to the mask handling. Visually results differ on the order of less than a voxel but I can not tell if one version is better. I'd say good to merge but let me know if I should dig into `population_template` as well.

-----

Below are some more details about differences in `mrregister`:

initialisation:

- the average space bugfix d168540d90820ce2f7c91d27b04c7efbc8c0c04e causes a different grid, average space is identical if this bugfix is reverted
- centre of mass is slightly different due to multi-threading (2e-14mm), f9bc2f38fb7bbc6aff068066fae378b8b37dc6dd

initial transformations and average space are identical for initialisation to centres of mass initialisation with `-nthreads 0` and reverted d168540d90820ce2f7c91d27b04c7efbc8c0c04e. All test below use no threading and reverted d168540d90820ce2f7c91d27b04c7efbc8c0c04e.


Scalar linear (affine) registration:
- identical results

Scalar nonlinear registration:
identical results if average space padding is reverted (mrreg_lite2: 1 voxel, master: 0 voxels). This was changed in 58728ba51216c0b2d332f8f333ee403ff4c3a7bd to unify average space padding for nonlinear and linear registration.

FOD, l_max=0:
- linear, master: slightly different cost comparing lmax=0 and registering only the first volumes (relative cost function difference after 1 iteration: 2e-14)
- linear, mrreg_lite2: after 10 iterations, identical cost and transformation compared to master with the scalar input, for both scalar input or lmax=0

FOD, lmax=2:
- linear: 1 iteration has relative cost function differences on the order of 3e-16, 2e-12 after 10 iterations, transformation matrices RMS between master and mrreg_lite2 after 10 iterations: 4e-13
- nonlinear: cost function differences on the order of 1e-18 per iteration for a cost of the order of 8e-5, resulting warps are identical

maxpietsch:
I updated the test data for `population_template` as all tests required updating the data with this PR. The major difference between the old and the new data is that out of field of view areas are now masked with NaNs, not zeros, but also the registration results are slightly (about a voxel or less) different in scanner space (mostly for affine and nonlinear registration). If interested and for reference, the new data is in the [`registration_new_features_1798`](https://github.com/MRtrix3/script_test_data/tree/registration_new_features_1798/population_template) branch, the old data in the [current `dev`](https://github.com/MRtrix3/script_test_data/commit/8fac49f30224d17a47f0aa7d77096989e02e4d63) branch.

maxpietsch:
I have tested single and multi-contrasts code paths with different options, merging as it seems to work or to get more broad testing in dev. The new tests check out on two different machines (macOS and ubuntu), one of the old tests in dev (voxel size test) did fail on my ubuntu machine due to precision.

jdtournier:
:+1: thanks Max!

maxpietsch:
:) There is a somewhat eery silence after pressing the button that merges code that spans 3 years... 

(BTW I am keeping the `mrreg_lite2` branch around on purpose for convenience of re-working the missing features in the `mrreg` branch and for Alena's code.)

PR #1800, opened 2019-11-11
dirorder: Exhaustive search

Lestropie:
Encountered this internal limitation while writing a script for generating acquisition schemes. Details in 3593415.

Execution is still way faster than other related commands, so that's no issue.

Also considered altering the optimisation of order by calculating the SH condition numbers for prospective addition of individual remaining directions; but undecided as to whether or not it's worth the effort yet.

### DIFF
--- a/cmd/dirflip.cpp
+++ b/cmd/dirflip.cpp
@@ -45,7 +45,7 @@ void usage ()
 
 
   OPTIONS
-    + Option ("permutations", "number of permutations to try.")
+    + Option ("permutations", "number of permutations to try (default: " + str(DEFAULT_PERMUTATIONS) + ")")
     +   Argument ("num").type_integer (1)
 
     + Option ("cartesian", "Output the directions in Cartesian coordinates [x y z] instead of [az el].");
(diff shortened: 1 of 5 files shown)


jdtournier:
I don't see any problem with your proposal, but I'm just trying to get my head around the motivation for it. Can I just check what the 'internal limitation' was exactly? I'm struggling to see how this would have a massive impact on the quality of the scheme. Sure, it might give slightly better looking values for the cost function, but then the cost function is itself relatively arbitrary, and since we're talking about relatively random circumstances that will likely also imply serious issues with the quality of the data that has been acquired, I'm not sure what the problem was that you needed to complement this with an exhaustive search... 

Maybe it's the randomness of the first volume that's the issue? I can see that with your proposal, the output is at least deterministic.

Lestropie:
"Internal limitation" was in reference to `dirorder` using a random first volume, not anything external to such. Can be a pretty decent variation in SH condition number upon truncation depending on which volume is selected as the first. If data quality upon truncation were to be too serious to warrant concern, then `dirorder` would lose its purpose; if it's going to exist & be used, might as well get it right.

jdtournier:
OK, all makes sense. Must admit I haven't looked into how much variation there is based on first volume, so if it's substantial, that makes sense. :+1:

> If data quality upon truncation were to be too serious to warrant concern, then dirorder would lose its purpose; if it's going to exist & be used, might as well get it right.

Agree on the principle, but in my opinion this is a case of making sure it's good enough, given the range of likely associated issues. But I guess that's irrelevant if it's cheap to get it right anyway, which given your changes, seems to be case...

PR #1802, opened 2019-11-11
generate_user_docs.sh macOS compatible 

### LINKED ISSUE #543: generate_user_docs.sh doesn't work with BSD sed
Generate_user_doc does not work on default OS X as BSD sed differs from GNU sed. The former does not allow in-place editing without backup file and one cannot use strings that contain escape sequences.

Current sed command: `sed -ie "1i$filename\n===========\n" $filepath/$filename.rst`

The first problem could be fixed by `sed -i ".bak" ...` and then `rm $filepath/$filename.rst.bak`. The second can supposedly be solved with ANSI escape sequences but I can't find a solution for newline and equal signs in the string. So I am stuck with 
`sed -i '.bak' '1i\'$'\n'$filename$'\n' $filepath/$filename.rst` which works except for the missing ===========\n sequence.

Someone with more bash skillz around? Alternative would be to use perl or something else that is more portable.

maxpietsch:
fixes #543

### DIFF
--- a/docs/generate_user_docs.sh
+++ b/docs/generate_user_docs.sh
@@ -84,7 +84,8 @@ for n in `echo "$cmdlist" | sort`; do
   esac
   $cmdpath __print_usage_rst__ > $dirpath/$cmdname.rst
   case $n in *.cpp)
-    sed -ie "1i.. _$cmdname:\n\n$cmdname\n===================\n" $dirpath/$cmdname.rst
+    s='print ".. _'${cmdname}':\n\n'${cmdname}'\n===================\n\n" if $. == 1; $. = 0 if eof'
+    perl -i -pe "$s" $dirpath/$cmdname.rst
   esac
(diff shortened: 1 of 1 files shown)


Lestropie:
I now get this spammed when running `docs/generate_user_docs.sh`:

```
perl: warning: Falling back to the standard locale ("C").
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_AU.UTF-8"
    are supported and installed on your system.
```

Easy enough to fix; just should have tested locally rather than trusting the CI :-/

Lestropie:
> I now get this spammed when running docs/generate_user_docs.sh:

Turns out this was indeed an erroneous configuration on my part, and not just a system "quirk". So I'm not against keeping this as-is given `perl` should be pretty widely supported (there shouldn't be a flurry of other people hiting the same issue).

PR #1803, opened 2019-11-11
check_syntax macOS grep

maxpietsch:
BSD grep does not support all features of gnu-grep such as lookahead which breaks the script (print `usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]...` many times to terminal). 

This change warns the user and instructs to install gnu-grep if `ggrep` (homebrew's version of gnu-grep) can not be found but proceeds anyway as gnu-grep might be in the path anyway. If `ggrep` is available, it uses that instead of the default grep.


Lestropie:
If the script fails entirely when using BSD `grep`, is issuing a message to the user and then attempting to run anyway the right behaviour to go with, especially if it results in printing a lot of junk to the terminal meaning that the message might get missed? Would it be better to abort outright on `darwin*` if `ggrep` is not installed?

maxpietsch:
The problem is that we'd need to detect which version of grep is in the path as ggrep is a symlink to grep in`/usr/local/opt/grep/libexec/gnubin`:
```
% brew install grep

All commands have been installed with the prefix "g".
If you need to use these commands with their normal names, you
can add a "gnubin" directory to your PATH from your bashrc like:
  PATH="/usr/local/opt/grep/libexec/gnubin:$PATH"
=º  /usr/local/Cellar/grep/3.3: 21 files, 880.7KB
```
Also, there [used](https://discourse.brew.sh/t/why-was-with-default-names-removed/4405) to be a flag `with-default-name`. Given that it's a developer tool I didn't invest the time to write a test for particular versions of grep but it would be cleaner indeed.

bjeurissen:
Note that perl (and python 2.7) will be removed from macos soon, leaving only python3 bundled. A more future proof solution might be to write check_syntax entirely in python. That keeps the dependencies down to just python.

jdtournier:
Given that this is not essential for regular users, I'm happy to merge this as-is. We can always revisit this if it becomes problematic. The long-term solution would indeed be a python script, but that will require someone's time to code up and test... This at least tells developers what to do about it if they're on macOS.

maxpietsch:
I modified the script to abort if the OS is `darwin*` and the grep version string contains `BSD` and no `ggrep` executable is found in the path. Otherwise the terminal is swamped with grep output and one might miss the warning to install gnu-grep.

PR #1812, opened 2019-11-20
mrview fixel plot enhancements

jdtournier:
Supersedes #1797 (actually an old branch I'd forgotten to create a pull request for).
Includes the relevant changes from #1797, plus what I had before. 

Main additions:
- option to display fixel/vectors are unidirectional
- add support for vector overlay 5th dimension to track main image 4th dimension (useful for angio cine applications)
- add wrap volume functionality to cycle through volumes (useful for cine applications)
- wider range of min/max values
- fixed bug in threshold handling 
- default transparency in ROI editor is 0.5 rather than 1, and there's a new config file option to set that default.


jdtournier:
A bit of eye candy to show what things look like when using unidirectional display, volume wrapping, and a 5D vector field with 5th index tracking main image 4th index (used in this [fetal cardiac paper](https://www.biorxiv.org/content/10.1101/635797v1)):

fetal_cardiac

maxpietsch:
Looks good, two small niggles:

- The length of the not bi-directional fixels is half that of the bi-directional fixels.
- `track main image volume` could be more descriptive about which dimensions are tracked

jdtournier:
OK, not sure I'm bothered about the length issue. Personally, I reckon it makes sense that its length from the origin is unchanged, even if that means it's half the length... 

Agree that "track main image volume" is not great. Any suggestions about what might be better? "vector volume group tracks main image volume"? Actually, I reckon it's probably best to have a shorter label like "cine mode", and use a tooltip to give a lot more detail about how it works....?

maxpietsch:
Using the correct length matters when one wants to show displacement fields in physical units. I'd rather half the scale factor if you're bothered by too long vectors. 
I can't think of a more descriptive but equally long option caption. So "cine mode" and a more verbose tooltip that also mentions the volume wrapping is probably best to not inflate the toolbar too much.

thijsdhollander:
> Using the correct length matters when one wants to show displacement fields in physical units.

Was thinking the same; this would be very useful actually.

jdtournier:
> Using the correct length matters when one wants to show displacement fields in physical units.

While this sounds appealing, the reality is that [the scaling is tied to the voxel size](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/fixel/base_fixel.cpp#L50) anyway - so it doesn't bear any relationship with physical distances unless you know this equation and the voxel size of your data... 

That said, I'd be happy enough to change this is it was felt to be useful, so that the scaling actually does reflect a 'millimeters per fixel unit length', which you could then set to 1 to ensure 1 to 1 physical mapping. But personally I reckon that should be another issue...

jdtournier:
@maxpietsch : I vote we merge this now, and create a feature request for the physical scaling. I think handling this well will require a bit of thought, it'll be worth mocking up a few use cases to see how we can make that work. But this is in many ways independent of this pull request. Agreed?

maxpietsch:
Agreed, scaling to physical units can be a separate issue. Having a "millimeters per fixel unit length" box would obviate being consistent in how the length of a fixel is defined but is still feels unnecessarily complex handling the uni- vs bidirectional case differently.

jdtournier:
> it still feels unnecessarily complex handling the uni- vs bidirectional case differently.

Well, this is where I'm not convinced. I understand the length of the fixel to be from the origin to the tip, and if it's bi-directional, then you effectively get two of them back-to-back - i.e. the forwards and backwards projections. I think it feels more 'natural' that when switching from bi-directional to uni-directional, the non-relevant half simply drops off. And as a more minor point, I'm not sure I can think of many cases of bi-directional fixels that should be displayed in millimeters...? So I fully get the case for being able to display uni-directional fixels in millimeters, but not so sure it matters for the bi-directional case - unless you have specific examples in mind?

maxpietsch:
> I understand the length of the fixel to be from the origin to the tip

I understand it as vector fields but I guess your interpretation makes sense since we call it fixels (which is a misnomer for anything but actual fixel data).

>  So I fully get the case for being able to display uni-directional fixels in millimeters, but not so sure it matters for the bi-directional case - unless you have specific examples in mind?

Not really other than for displaying uncertainty in unidirectional measures. Let's merge and discuss semantics another time :P

PR #1814, opened 2019-11-21
tckgen: fix out-of-bounds access in downsampler

jdtournier:
This caused crashes in rare cases, due to downsampling tracks with a single vertex.

First raised by @bjeurissen due to consistent crashes in `dwigradcheck` on Catalina, eventually tracked down to the use of downsampling (so nothing to do with macOS or multi-threading), and the issue isolated with a massive overnight 'valgrind' job on Linux... :sweat_smile:

### DIFF
--- a/src/dwi/tractography/resampling/downsampler.cpp
+++ b/src/dwi/tractography/resampling/downsampler.cpp
@@ -26,8 +26,10 @@ namespace MR {
 
         bool Downsampler::operator() (Tracking::GeneratedTrack& tck) const
         {
-          if (ratio <= 1 || tck.size() < 2)
+          if (ratio <= 1)
             return false;
+          if (tck.size() <= 1)
+            return true;
           size_t index_old = ratio;
           if (tck.get_seed_index()) {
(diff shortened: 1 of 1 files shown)


Lestropie:
- Returning `false` would have this functor shut up shop if used in a multi-threading pipe; it might make more sense to return true as long as ratio > 1.

- There's likely other resampling functors that have the same issue w.r.t. assuming a minimum length, and/or returning false when maybe they shouldn't.

- Regardless of those, downsampling a streamline with one vertex should probably preserve that vertex in the output streamline.

bjeurissen:
I can confirm that this fixes the issue! :+1:

jdtournier:
OK, @Lestropie, how about that last commit?

bjeurissen:
Last commit also passes my test.

Lestropie:
I would think this actually needs to go to `master`.

I want to take the opportunity to go through all of the streamline resampling functors and make sure the logic is correct; but I can do that in a separate PR.

jdtournier:
OK, I'll merge this and commit the same change to `master` too.

PR #1815, opened 2019-11-21
mrview: add support for QFileOpenEvent (macOS)

jdtournier:
This adds support for opening images by dragging onto the dash, and by double-clicking in the Finder.

### DIFF
--- a/cmd/mrview.cpp
+++ b/cmd/mrview.cpp
@@ -20,6 +20,7 @@
 #include "memory.h"
 #include "gui/mrview/icons.h"
 #include "gui/mrview/window.h"
+#include "gui/mrview/file_open.h"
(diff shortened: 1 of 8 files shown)


bjeurissen:
Sorry, I compiled only mrview, so I did not face the above CI errors :(.

Apparently, this patch breaks shview and mrcolour, which seem to have missing definitions for virtual members of QApplications, e.g. "event".

bjeurissen:
> Apparently, this patch breaks shview and mrcolour, which seem to have missing definitions for virtual members of QApplication, e.g. "event". I guess this means the same logic that was applied to mrview now also has to be applied to shview? This also seems to suggest that mrcolour is pulling in more code than it should.

Think I fixed it, will push them ASAP.

PR #1821, opened 2019-11-25
Fix dicom handling with stricter string conversion

jdtournier:
Fix errors while reading DICOMs following #1794. 

This strips the strings prior to parsing with `to<type>()`, and fixes the issue as far as DICOM handling is concerned. 

But there is an argument for stripping all strings within the `to<type>()` call. This obviously fixes this issue, but also provides more lenient handling of string parsing in general. There's a good argument for considering strings with trailing whitespace as valid...

### DIFF
--- a/core/file/dicom/csa_entry.h
+++ b/core/file/dicom/csa_entry.h
@@ -97,7 +97,7 @@ namespace MR {
               for (uint32_t m = 0; m < nitems; m++) {
                 uint32_t length = Raw::fetch_LE<uint32_t> (p);
                 if (length)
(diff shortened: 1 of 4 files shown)


Lestropie:
> But there is an argument for stripping all strings within the `to<type>()` call. 

Agreed; 8c77f3f.

jdtournier:
Cool. I'll revert my first two commits though, no need to use those additional `strip()` statements if that's done internally by the `to()` call...

Lestropie:
There may be an argument for a `strip()` call actually happening *earlier*: i.e. not just for conversion to numerical types, but any kind of DICOM extracted strings should throw out any superfluous whitespace.

PR #1822, opened 2019-11-27
dirflip and dirsplit: Fix command-line option parsing

Lestropie:
Details in af51eef.

#1794 has multiple implications, as it turns out...

### DIFF
--- a/cmd/dirflip.cpp
+++ b/cmd/dirflip.cpp
@@ -21,7 +21,7 @@
 #include "thread.h"
 #include "dwi/directions/file.h"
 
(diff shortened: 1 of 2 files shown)


Lestropie:
> Just changed things slightly to use constexpr rather than a macro as was the case before. It's cleaner and better C++11 anyway...

Should we consider modifying *all* instances of such?

jdtournier:
Yep. I tend to when I come across them. But I don't think it's worth chasing them down systematically unless you've got time to spare...

PR #1823, opened 2019-11-28
MR::to<>(): Strip whitespace (DICOM fix)

Lestropie:
Supersedes #1821.

Rather than reverting initial changes, figured it would be cleaner to instead cherry-pick the relevant commit.

Outstanding question:

> There may be an argument for a `strip()` call actually happening *earlier*: i.e. not just for conversion to numerical types, but *any* kind of DICOM extracted strings should throw out any superfluous whitespace.

### DIFF
--- a/core/mrtrix.h
+++ b/core/mrtrix.h
@@ -77,21 +77,6 @@ namespace MR
 
 
 
-  template <class T> inline std::string str (const T& value, int precision = 0)
-  {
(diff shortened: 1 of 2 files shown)


jdtournier:
Looks good. On that outstanding question: [already the case](https://github.com/MRtrix3/mrtrix3/blob/master/core/file/dicom/element.cpp#L387)...

PR #1831, opened 2019-12-04
support foreach */ again to loop over directories

bjeurissen:
foreach used to support looping over just folders, by adding a trailing '/'
Currently this is broken because basename('foldername/') is empty, causing NAME to be empty.
This should fix this.


jdtournier:
Seems fair enough to me...

PR #1833, opened 2019-12-05
Tracking parameter updates

### LINKED ISSUE #930: tckgen: use more appropriate angle for SD_STREAM algorithm

Lestropie:
This will become a central PR for multiple proposed changes to tracking parameters, due to the potential for interplay between them.

PR is to encapsulate:

- #1630: Changing default angles, which the original issue / PR began as altering only deterministic algorithms but is now proposing to make *all* tracking algorithms operate using fixed angle thresholds independently of step size (closes #930);

- #1747: Changing default step size when utilising 4th-order Runge-Kutta to be 1/4 voxel size.

- Re-addressing appropriate cutoffs for various algorithms. This will be a repeat of aspects of #1228, but specifically to address the combination of:

  - Changes to default WM response function estimation in #1545;

  - Fix to the SH precomputer in #1715 due to the erroneous fix in #1206.

Once I've written a script to generate output data for a range of prospective cutoffs for all possible algorithms & configurations, it'll go here for discussion.

-----

Other points that we may want to consider in this context given we're performing "an overhaul of default tracking parameters":

- Add some downsampling by default for first-order algorithms without RK4 to decrease file size.

- Change iFOD2 power from (1/nsamples) to (1/(nsamples-1)). There may be a more appropriate derivation for such, but for the current derivation the arithmetic is nevertheless incorrect.


Lestropie:
Initial image dump; only a coarse parameter set right now, can generate more fine-grained results as we hone in on targets for particular algorithms.

Note that `sh2peaks` wth default parameters was used for FACT algorithm.

[Full-resolution image link](https://www.dropbox.com/s/a6x9ps7g6am4gh6/Thresholds_compilations.zip?dl=0)

## Without ACT

#### FACT

compilation_None_FACT_small

#### iFOD1

compilation_None_iFOD1_small

#### iFOD2

compilation_None_iFOD2_small

#### SD_STREAM

compilation_None_SDSTREAM_small



## With ACT (no back-tracking)

#### FACT

compilation_ACT_FACT_small

#### iFOD1

compilation_ACT_iFOD1_small

#### iFOD2

compilation_ACT_iFOD2_small

#### SD_STREAM

compilation_ACT_SDSTREAM_small

Lestropie:
- For ACT, I've always wondered whether thresholds should simply be disabled outright. I think there's a decent case to be made for that for the deterministic algorithms. iFOD1 and iFOD2 get a bit messy with too low a threshold, though raising the iFOD2 power may mitigate some of that. Even if non-zero it's possible that the default threshold for these should be lower if ACT is used compared to if it is not. ACT already influences the default minimum length, so there is some precedent for it influencing tracking parameters.

- I made `-power` work for iFOD1 in addition to iFOD2, thinking it might make sense for the power when RK4 is used to be 0.25. That produces absolute junk; but I think keeping the `-power` option operational for both iFOD1 and iFOD2 makes sense.

Lestropie:
Similar for targeted tracking.

[Full-size images download](https://www.dropbox.com/s/q90bi9cw4ymwgji/Thresholds_targeted.zip?dl=0)

## Without ACT

#### FACT

None_FACT_small

#### iFOD1

None_iFOD1_small

#### iFOD2

None_iFOD2_small

#### SD_STREAM

None_SDSTREAM_small

## With ACT (no back-tracking)

#### FACT

ACT_FACT_small

#### iFOD1

ACT_iFOD1_small

#### iFOD2

ACT_iFOD2_small

#### SD_STREAM

ACT_SDSTREAM_small

jdtournier:
😱

Thanks!

Issue #778, opened 2016-09-08
MRView: add ability to view spherical ROIs as stored in tck file headers


jdtournier:
As requested [here](http://community.mrtrix.org/t/showing-the-spherical-seed-roi/428?u=jdtournier).

Lestropie:
Bumping as this got raised a couple of times during the workshop. Progress has been made on the `mrview_spherical_rois` branch, but it's not been touched in 6 months. Might take a look and see whether it can be promoted to the release milestone without too much effort.

PR #1836, opened 2019-12-08
mrview: Draw spherical ROIs

Lestropie:
Largely a reminder to myself to try to finish this off and merge.

From memory the outstanding issue was handling within volume render mode.

### DIFF
--- a/src/dwi/tractography/properties.h
+++ b/src/dwi/tractography/properties.h
@@ -131,11 +131,11 @@ namespace MR
 
       inline std::ostream& operator<< (std::ostream& stream, const Properties& P)
       {
-        stream << "seeds: " << P.seeds;
-        stream << "include: " << P.include << ", exclude: " << P.exclude << ", mask: " << P.mask << ", dict: ";
+        stream << "seeds: " << P.seeds << "; ";
+        stream << "include: " << P.include << "; exclude: " << P.exclude << "; mask: " << P.mask << "; dict: ";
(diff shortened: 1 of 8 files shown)


maxpietsch:
From skimming the code it seems this is this tied to spherical include ROIs stored in tractogram headers? 

Is the rendering code flexible enough to write a simple tool that can display (and create,  import, and export) spherical ROIs outside the tractography tool? I am asking because I am looking for a simple way to create labelled ROIs for landmarking from within mrview for a student project. Otherwise, I'd go with using image masks per landmark.

Lestropie:
Currently yes, the goal is just to display whatever is in the track file header.

If you want to generalise and have the ability to visually control positioning of spherical ROIs, and import / export them, then the interface for that becomes a much larger piece of work than just "displaying spherical ROIs" as is done here. The latest Siemens Frontier TDI WIP is quite good in this respect: you can create inclusion & exclusion regions, move and resize them with the mouse, and then hit a button that runs a tractogram through `tckedit` with those ROIs. But to do something like that in `mrview` would be a fair amount of work.

Lestropie:
Had another quick peek at this. Too much yet to be done though:

- Lighting of spherical ROIs in volume render mode not matching lighting of tractography tool (which itself doesn't currently properly match other tools; think this was what triggered me to suggest that non-material lighting settings should be centralised);

- Culling / ordering of triangles in 3D mode when not fully opaque and lighting enabled not working.

Lestropie:
Addresses #778.

PR #1844, opened 2019-12-11
mrview/mrcolour: split off colourmap code from GUI code

jdtournier:
to avoid dependence of mrcolour on Qt

The diff looks huge, but it really only involves moving the `ColourMap::Entry` class and related code into its own header/cpp in `src/`, and renaming the `src/gui/mrview/colourmap.{h,cpp}` to something else (`colourmapper.{h,cpp}` in this case) to avoid problems when trying to include `colourmap.h` (since it looks in the same folder as the current file first, and so we can't include `colourmap.h` in `src/gui/mrview/colourmap.h`). Then I renamed `ColourMap::Renderer` to `ColourMapper` to avoid conflicts between the `MR::ColourMap` namespace and the now-removed `MR::GUI::MRView::ColourMap` namespace. All other changes are minor consequences of this.


Lestropie:
Had intended to do this code move during the course of #1541, but clearly never got around to it.

I did however find the "`ColourMapper`" class name quite ambiguous as to its functionality, so had a bit of a go:

- Renamed to "`MRView::ColourBars`";

- Moved the addition of colourmap entries to a Qt menu to its own file (it was a static function of that class, but had no interaction with it, only with the colour maps themselves).

- There were a couple of instances of this class that were never utilised whatsoever;

- The class had *both* "`size_t current_index`" *and* "`size_t current_colourbar_index"` members :sweat_smile:

jdtournier:
Seems all good - merging.

Issue #1462, opened 2018-10-04
mrview: command line option overlay.intensity does not update the overlay


maxpietsch:
As reported in #1461 by @dnvaughan:
Unfortunately -overlay.intensity just sets the text box entries on the overlay tool tab, and doesn’t update the actual image or the overlay color bar.

jdtournier:
That option is processed [here](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L769L781), and seems ok (good example of blocking signals BTW, in case you're looking for inspiration for #1369). If you look at the slot that the `max_value` is [connected to](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L149), to note that it [applies to the currently selected overlays](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L477). My guess is, at the point where the option is processed, no overlay has been selected.
Not sure why though, the `-overlay.load` processing block calls `add_image()`, which itself [explicitly sets the selection](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L215L223). :confused: 

Can I check whether the display updates properly if you interact with the main window in any way? Is it just a matter of explicitly calling `updateGL()`?

Zach-ER:
If I interact with the window (moving the cursor around, hiding/viewing main image, hiding/viewing overlay) it does not update. The value is in the text box in the 'overlay' panel, though.

Lestropie:
Bumping and elevating milestone based on [forum thread](https://community.mrtrix.org/t/overlay-colourmap-and-intensity-are-not-applied-in-mrview-via-command-line/3115).

jdtournier:
Closing - fix is hopefully in #1850...

PR #1850, opened 2019-12-16
mrview: fix -overlay.intensity

jdtournier:
As discussed in #1462. Would appreciate others giving this a shot. 
Note this fixes the `-overlay.intensity` issue, but is unlikely to fix the `-overlay.colourmap` one discussed in [this forum thread](https://community.mrtrix.org/t/overlay-colourmap-and-intensity-are-not-applied-in-mrview-via-command-line/3115?u=jdtournier). I'll keep digging on that one...


jdtournier:
Actually, the colourmap handling seems to be OK on my system. Not sure why it wouldn't be working for the user. I'll request an update on the forum.

Issue #1845, opened 2019-12-11
run.command() interface & path quoting


Lestropie:
Branching discussion from #1689, specifically [this comment](https://github.com/MRtrix3/mrtrix3/pull/1689#issuecomment-564711729), so as not to blur distinct issues.

> As you can see, the issue is that the argument passed to mrconvert's -copy_properties was quoted with single quotes (presumably by the `shlex.quote()` call), but these are then passed along with the argument itself to the command. I can fix this by adding an additional `.strip ("'")` to this line in `run.py`, so that it strips both single and double quotes.

I have never previously seen single-quotes introduced like that for the purpose of encapsulating filesystem paths. I would need you to trace back and find precisely what the source of it is, so that it can either be modified to be double-quote at the source, or all API filesystem path handling checked to ensure compatibility with either single or double quotes.

> This was kind of discussed in #928, but I do think that things would really be a lot easier if we passed lists of strings to the `run.command()` call rather than trying to tokenise a single command string, with all the gotchas involved in doing this...

Of course it'd be easier from an API management perspective to do so; but it'd be way less convenient when actually writing scripts to run against the API. Constructing command strings to pass to `run.command()` is a lot more intuitive than constructing `argv`s; hence why I've put so much effort into providing that capability for my own sake. So removing that capability would feel like a pretty large regression to me. Unless it becomes unmanageable, I'd rather put in the effort to keep string-based inputs to `run.command()` working. If it's just a matter of a particular version of `shlex.quote()` using singles rather than doubles, it shouldn't be too hard to rectify; much easier than what's already been put in place in this context.

jdtournier:
Yes, I agree it would be quite a bit of effort and a loss in convenience if we had to enforce the use of lists instead of command strings - not to mention throwing out quite a bit of work you've done on this. But it would avoid these types of issues...

In any case, I think I've figured out where it's coming from, and the fix is simple - see c7f91da. I have a feeling this will need to be replicated across all instances where the `mrconvert_keyval` argument is set in `run.command()`... It might be simpler to allow that command to take the original input (not the output of `path.from_user()`), which `run.command()` could then handle appropriately internally. I guess this would mean it couldn't be used for cases where `app.WORKING_DIR` is not set? Or maybe that's not an issue? The only instances I could identify where `mrconvert_keyval` doesn't receive the output of `path.from_user()` are these ones (which I'm not sure I understand, TBH):
```
[...]

PR #1853, opened 2019-12-17
run.command(): Fix mrconvert_keyval usage

Lestropie:
Passing a user-specified path to `run.command()`'s keyword argument `mrconvert_keyval` must not be escaped, as the string is inserted directly as a list element and is not parsed as part of a command string.

Cherry-picked first commit from #1689.

Comments:

- This arose because I failed to appropriately modify usage during the evolution of #1449. In an earlier iteration I was inserting these entries into the `mrconvert` command string, which required path escaping, and then trying to subseequently erase them from the command string printed to the terminal; eventually I realised I could just separate the actual command `argv` from the terminal printout at an earlier point in the function, and have this parameter only influence the former. I simply failed to update the calls to specify `escape=False` to match.

- I'm still struggling to see why this is being portrayed as such a severe issue (#1845). In some contexts, filesystem paths need to be escaped, in others they need to not be; that's why `path.from_user()` has the `escape` parameter. Only possible change in light of this issue would be to revert that parameter to not have a default value, and force it to always be specified. But the challenge of filesystem path handling is no more an existential crisis now than it was before observing this specific mis-use.

- I also still don't see why the question of running `quote()` twice on one string is being elevated. In the commit message of c7f91da it seems to get raised out of nowhere; there's no indication that this is actually happening anywhere, or that it is a legitimate concern for any hypothetical case that would not itself be a mis-use of the API. If you want a warning comment above the `path.from_user()` function indicating to not run it on a string that may have itself already been quote-escaped, that's fine; it just seems to me that this is a non-issue right now.

### DIFF
--- a/bin/dwishellmath
+++ b/bin/dwishellmath
@@ -57,7 +57,7 @@ def execute(): #pylint: disable=unused-variable
     files.append(filename)
   # concatenate to output file
   run.command('mrcat -axis 3 ' + ' '.join(files) + ' out.mif')
-  run.command('mrconvert out.mif ' + path.from_user(app.ARGS.output), mrconvert_keyval=path.from_user(app.ARGS.input), force=app.FORCE_OVERWRITE)
+  run.command('mrconvert out.mif ' + path.from_user(app.ARGS.output), mrconvert_keyval=path.from_user(app.ARGS.input, False), force=app.FORCE_OVERWRITE)
 
 
 # Execute the script
--- a/bin/labelsgmfix
+++ b/bin/labelsgmfix
@@ -164,7 +164,7 @@ def execute(): #pylint: disable=unused-variable
   # Insert the new delineations of all SGM structures in a single call
   # Enforce unsigned integer datatype of output image
   run.command('mrcalc sgm_new_labels.mif 0.5 -gt sgm_new_labels.mif parc.mif -if result.mif -datatype uint32')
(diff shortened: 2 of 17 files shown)


jdtournier:
I'm a bit surprised by the strength of your reaction here, I apologise if this came across as a bigger deal than it is. I think a lot of this is simply that I tend to document what I've tried and my thoughts on the issue so that it's out there and we can refer to it in future - maybe I do that more than I should...

I'm quite happy to just pass the `escape=False` flag through `path.from_user()`, but simply concerned that this was likely to happen again in future - so I thought I'd look into options to avoid this. Can't think of one, so that's that. It's the correct fix as things stand, no debate. As you can see from my various musings on #1845. 

> I also still don't see why the question of running quote() twice on one string is being elevated. 

Because that's kind of how the problem manifested in the first place. It was clear that the path had been quoted but then passed as-is, quotes and all, to the command. As it turns out, no double-quoting was actually taking place (my mistake), but just once when it shouldn't. But the reason I even looked into trying to 'fix' the quoting was because of [this comment](https://github.com/MRtrix3/mrtrix3/issues/1845#issuecomment-565250407):

> I'm surprised that shlex.quote() or pipes.quote() would behave like that. If that's the case, then we may want to make path.quote() a little more advanced, by invoking path.unquote() before runing the library quote() function.

[My last comment](https://github.com/MRtrix3/mrtrix3/issues/1845#issuecomment-566553243) on #1845 is purely in response to this. All I'm saying is I don't think this can be done. Which wouldn't have changed the fact that the current fix (don't quote in the first place) is the correct one. 

So as far as I can tell, we're all in agreement about how to handle this.

PR #1858, opened 2019-12-19
constrained least-squares: support non-zero constraints and equality constraints

jdtournier:
Additions to the constrained least-squares solver in response to requests:

- support for non-zero constraint values - i.e. allow constraints of the form _Ax_ ≥ _t_, where _t_ ≠ 0
- support for equality constraints - i.e. allow constraints of the form _Ax_ = _t_
  This functionality is provided by specifying the number of trailing rows of _A_ that should be considered as equality constraints (the rest will be treated as inequalities as before)

This all seems to check out in Matlab - although Matlab's `quadprog()` seems to provide higher precision than this implementation (max constraint violation is ~10<sup>-15</sup> with Matlab, ~10<sup>-10</sup> with this PR). 

For testing, feel free to have a go with the `icls` and `icls_text` commands in my [mrtrix_extras repo](https://github.com/jdtournier/mrtrix3_extras/pull/1) (on the `icls_arbitrary_inequality_constraint` branch).


Lestropie:
Would the interface make more sense if inequality and equality constraint matrices were provided separately (even if concatenated internally), rather than giving a single matrix and then an integer specifying how many rows (necessarily from the bottom) are equality rather than inequality?

jdtournier:
Well, that is indeed how most packages do it. Personally, I think the current approach is just as simple, this is only ever going to be used by seasoned developers, so I don't think this is going to cause all that much confusion. But I'm happy enough to change or add to the interface if people feel it's particularly problematic. Probably easiest by providing an alternative constructor.

bjeurissen:
I like the separate arguments :)

jdtournier:
OK, additional constructor in that last commit to accept separate matrices & vectors for equality and inequality constraints. Feel free to break it... 

I might add some tests to the repo just to make sure it all works as expected.

PR #1859, opened 2019-12-20
GitHub: Fix for Windows workflows

Lestropie:
I think this is what's borking Windows tests. Couldn't rename the files in the release to match the fetch calls.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -120,7 +120,7 @@ jobs:
     - uses: actions/checkout@v1
 
     - name: fetch and install MSYS2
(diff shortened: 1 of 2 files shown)


jdtournier:
Whoops! Sorry, that was my mistake - been fiddling with this last night on my repo, forgot this would impact CI on the main one too... 🥺

This is because I've modified the MSYS2 download to use the reduced Qt install I'd prepared earlier. Reduces the download from 8GB to 2GB...

The correct fix is actually `{0,1}`, but if it works, leave it for now, it'll get sorted when I eventually merge what I'm doing.

Lestropie:
For some reason I though you were intentionally padding the indices in case the download beomes larger in the future and you don't want to have to update the workflow to match :rofl:

Issue #1864, opened 2020-01-05
mrview: dark mode issues


maxpietsch:
Text in all `AdjustButton` boxes is not readable in all mrview tools in dark mode:
<img width="1073" alt="Screenshot 2020-01-05 13 26 14" src="https://user-images.githubusercontent.com/10046944/71780119-f5104f80-2fbe-11ea-952d-787e484acb3e.png">

If I take out the hard-coded background colour transition in the `AdjustButton` class:
```
--- a/src/gui/mrview/adjust_button.cpp
+++ b/src/gui/mrview/adjust_button.cpp
@@ -44,7 +44,7 @@ namespace MR
           setStyleSheet ((
               "QLineEdit { "
               "padding: 0.1em 20px 0.2em 0.3ex; "
-              "background: qlineargradient(x1:0, y1:0, x2:0, y2:0.2, stop:0 gray, stop:1 white) url(:/adjustbutton.svg); "
+              "background: url(:/adjustbutton.svg); "
```
it looks like this:

![image](https://user-images.githubusercontent.com/10046944/71780257-a2d02e00-2fc0-11ea-988b-1cf849c70d01.png)

Not exactly eye candy but at least readable.

Personally, I'd take the 3D effect out and emphasize the border:

```
+          const QColor bg (QWidget::palette().color(QWidget::backgroundRole()));
+          const QColor fg (QWidget::palette().color(QWidget::foregroundRole()));
+          const float r (0.4);
+          const QColor hl (fg.red() * (1.0-r) + bg.red() * r, fg.green() * (1.0-r) + bg.green() * r, fg.blue() * (1.0-r) + bg.blue() * r, 255);
           setStyleSheet ((
               "QLineEdit { "
               "padding: 0.1em 20px 0.2em 0.3ex; "
-              "background: qlineargradient(x1:0, y1:0, x2:0, y2:0.2, stop:0 gray, stop:1 white) url(:/adjustbutton.svg); "
+              "background: url(:/adjustbutton.svg); "
               "background-position: right; "
               "background-repeat: no-repeat; "
               "font-size: " + str(font().pointSize()) + "pt; "
-              "border: 1px solid grey; "
-              "border-color: black lightgray white gray; "
+              "border: 1px solid rgb("+str(hl.red())+","+str(hl.green())+","+str(hl.blue())+"); "
               "border-radius: 0.3em }").c_str());
```

[...]

PR #1865, opened 2020-01-05
mrview: adjust button: dark mode modifications

maxpietsch:
To address dark mode issues #1864 for `AdjustButton`s

- removes hard-coded background gradient and uses Widget background colour instead
- border colour emphasized


jdtournier:
Looks good to me. There's a slight concern with what happens if the theme changes at runtime - but not worth investing too much time trying to get it perfect... Feel free to merge when ready!

Issue #1866, opened 2020-01-07
ENH mrview: display full path of loaded files


ansk:
When loading multiple files (via tractography, fixel plot, etc.) the full paths of loaded files are NOT displayed in the toolbar. It becomes particularly frustrating when loading multiple files with differing names only somewhere in the directory tree, since there is further no way to figure out, which file is which. The full file name is not displayed when toolbar is resized. 

It could help to
a) display the full path as the popup tooltip when mouse is hovered over
b) update displayed path to fit to resized toolbar size
c) when clicking at "open", to jump to directory of currently selected file in the toolbar

jdtournier:
Yes, that's been bugging me too... Take a look at #1867, which I think fixes your first two points. 

The last point (about opening in the same place as the currently selected item) should also be possible, but I'm not convinced it's always the best policy. Wouldn't mind hearing other opinions on that one before looking into it.

ansk:
Hi, @jdtournier,

super, it helped!
Could you please also add a popup tooltip of the filename of `.tsf` file used in `colour map and scaling` and `thresholds` in `Tractography` toolbar and also in `Fixel plot`?

Antonin

jdtournier:
OK, I've added tooltips for the .tsf handling. But I'm not sure we need that for the fixel plot tool, since these comboboxes are populated by the contents of the corresponding fixel directory. They're not allowed to be in any other location than the main fixel dataset itself (@Lestropie, correct me if I'm wrong... :grin:). Assuming that's right, there is no point in additional tooltips there.

By the way, I'm closing this issue, but feel free to keep commenting and making suggestions on #1867.

PR #1867, opened 2020-01-07
MRView: better filename elision in List Views

jdtournier:
Attempt at fixing #1866.

### DIFF
--- a/src/gui/dialog/file.cpp
+++ b/src/gui/dialog/file.cpp
@@ -49,8 +49,6 @@ namespace MR
             new_folder = qstring.toUtf8().data();
             if (folder)
               *folder = new_folder;
-            else
-              QDir::setCurrent (new_folder.c_str());
           }
           return new_folder;
         }
@@ -68,8 +66,6 @@ namespace MR
             std::string new_folder = Path::dirname (filename);
             if (folder)
               *folder = new_folder;
-            else
-              QDir::setCurrent (new_folder.c_str());
           }
           return filename;
         }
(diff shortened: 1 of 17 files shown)


maxpietsch:
RE  "when clicking at "open", to jump to directory of currently selected file in the toolbar" suggested in #1866

I am in favour of setting all open dialogues relative to the selected file in the toolbar or the last selected folder for that tool in the absence of selected items (or the main image if none exists so far). For example, a workflow where the current behaviour is counterproductive is opening a tractogram, setting a screenshot directory somewhere else, opening the next tractogram, setting a different screenshot directory, ...

thijsdhollander:
I agree. 👍

jdtournier:
OK, by popular request, the last few commits ensure that each tool keeps track of its own current working directory, and that seems to work as expected. Each tool will default to the current folder, or wherever the file dialog was when last used. Once used for a tool, that tool will keep track of its own folder, with no interaction with the other tools' folders. 

One outstanding question is whether the main window should also keep track of its own folder. Currently, opening an image (File->Open) will pop up the dialog wherever it was last used, tool or otherwise. I can see there would be an argument for ensuring the main window also tracks its own separate folder. 

Thoughts / feedback welcome.

maxpietsch:
I'd also track the folder for the main window for the same reason as the toolbars but also to unify the behaviour of open dialogues across the app.

jdtournier:
That's my feeling too. OK, the last couple of commits should address that. Untested at the moment other than checking it all compiles. I'll check it all works as expected on Monday.

jdtournier:
OK, seems to check out... I'll merge now, feel free to re-open if issues arise.

PR #1876, opened 2020-01-11
mrview: fix view tool

maxpietsch:
Suggested fix to bug introduced in d411e6c5d7928ca (#1850), reported [here](https://community.mrtrix.org/t/mrview-view-options-segmentation-fault/3170). It seems to work but I am not 100% sure if this is a full fix as I've reverted `setValue` back to not trigger the `valueChanged` signals, `setMin` and `setMax` do.

### DIFF
--- a/src/gui/mrview/adjust_button.h
+++ b/src/gui/mrview/adjust_button.h
@@ -66,8 +66,6 @@ namespace MR
               clear();
               is_min = is_max = false;
             }
-            emit valueChanged();
-            emit valueChanged(value());
           }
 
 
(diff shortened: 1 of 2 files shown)


maxpietsch:
(Got the wrong commit hash, should have been d411e6c5d7928ca)

jdtournier:
OK, I think what I've just pushed in a better fix - see what you think...

maxpietsch:
Works. Out of curiosity, it's better because it avoids one `values_changed`?

jdtournier:
Good to hear. No, I think it's better because the purpose of setMin() and setMax() is different from setValue(). Also, depending on what the values might already be (maybe from a prior `-overlay.intensity` option), there's no guarantee that the relevant signal will be emitted. The fact it only gets called once is a bonus...

Issue #1641, opened 2019-07-17
mrview: keyboard shortcuts on macos


maxpietsch:
On opening an image, keyboard shortcuts work but once a tool (view tool for instance) is or was opened they stop working. This behaviour can also be triggered from the command line without opening a tool menu: `mrview 1.mif -overlay.load 2.mif`. 

Opening a popup window (`info-->about Qt`) or expanding the tool to multi-window layout brings back the keyboard shortcuts. Opening a tool docked inside the side bar of the main window removes the shortcuts again irrespective of an open popup window.

Switching between already opened (docked) tools has no effect on the shurtcuts.

Tested with c5c7f9c81037c445892f0, Qt 5.12.0.

maxpietsch:
This is still an issue with Qt 5.14.0

jdtournier:
I assume this was fixed with #1877...?

PR #1877, opened 2020-01-14
mrview keyboard shortcut fix

maxpietsch:
- fixes opening a tool deactivating keyboard shortcuts on macos #1641 
- merge latest updates on master

### DIFF
--- a/src/gui/mrview/adjust_button.h
+++ b/src/gui/mrview/adjust_button.h
@@ -67,8 +67,6 @@ namespace MR
               clear();
               is_min = is_max = false;
             }
-            emit valueChanged();
-            emit valueChanged(value());
           }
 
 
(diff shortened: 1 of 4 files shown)


jdtournier:
For reference: issue was that dock widget was shown as floating in constructor, but immediately docked after construction if the [MRViewDockFloating](https://mrtrix.readthedocs.io/en/latest/reference/config_file_options.html#cmdoption-arg-mrviewdockfloating) config file option is set to false (which is the default). This caused the main window to lose focus, and it never got it back when the tool for docked back in.

If you set that option to true in your config file, there no issue, even after docking the tool back into the main window.

So the fix it to allow the constructor to set the tool's floating status correctly in the constructor before it's shown...

bjeurissen:
Nice!

PR #1880, opened 2020-01-15
Mrview fix cursor jump mac os

maxpietsch:
Probably a leftover from a previous buggy version of Qt...


jdtournier:
For reference: what this fixes is what looks like a skipped screen update after a 'pan gesture' - initiated by a long press and drag on the touchpad. The next click to focus immediately after the pan gesture resulted in the image and focus being in a different location than expected. This seems to fix it.

PR #1884, opened 2020-01-17
header.h: more explicity initialiser braces to avoid compiler (clang …

maxpietsch:
Since it's only in a few places, I'd change the code that triggers the ambiguity warnings ([SO discussion](https://stackoverflow.com/questions/22501368/why-wasnt-a-double-curly-braces-syntax-preferred-for-constructors-taking-a-std)).

### DIFF
--- a/core/header.h
+++ b/core/header.h
@@ -62,8 +62,8 @@ namespace MR
         format_ (nullptr),
         offset_ (0.0),
         scale_ (1.0),
-        realign_perm_ {0, 1, 2},
-        realign_flip_ {false, false, false} {}
+        realign_perm_ {{0, 1, 2}},
+        realign_flip_ {{false, false, false}} {}
 
       explicit Header (Header&& H) noexcept :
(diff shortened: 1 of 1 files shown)


jdtournier:
Given the discussion on SO, I think I'd prefer to use round brackets here, e.g. replace:
```
realign_perm_ {0, 1, 2},
```
with:

```
realign_perm_ ({0, 1, 2}),
```
Then it's clear that the `{0,1,2}` bit is a single argument to a constructor. What do you think?

maxpietsch:
That doesn't suppress the warning: 

```
./core/header.h:65:25: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        realign_perm_ ({0, 1, 2}),
                        ^~~~~~~
                        {      }

```

jdtournier:
Man, I'm not sure I fully understand these initial users... Ok, let's stick to your suggestion if that seems acceptable to the compiler.

PR #1843, opened 2019-12-11
JSON handling on dev

Lestropie:
In trying to resolve the various issues with JSON handling (#1735, #1738, #1771), it looks like there was quite a bit wrong with the relevant code. Given the breadth and magnitude of issues, I'd like to try to get JSON handling working as well as possible on `dev` (essentially expanding #1738), and then once we're happy with the behaviour here, determine what does or does not require immediate rectification on `master`.

-----

The complexity of the problem is as follows. 

1. Input orientation information could be either:

   1. Stored in the image header, in which case such data *may* be reoriented in the header key-value data during the `Header::realign_transform()` function (depending on `-norealign` or `Header::do_not_realign_transform`);

   2. Imported via `-json_import`, which occurs *after* `Header::realign_transform()` has already been completed.

2. Output orientation could be either:

   1. Stored in the header along with the image data;

   2. Exported to JSON to live alongside a non-NIfTI image;

   3. Exported to JSON to live alongside a NIfTI image.

So what I *think* may be two ongoing issues in parallel are:

1. `JSON::read()` was calling `Axes::get_permutation_to_make_axial()`, but doing so on an image header that has already had such permutation applied. So this change stores the permutations & flips applied to a `Header` as class member variables for later reference by `JSON::read()`.

2. On JSON export, realignment of header key-value information based on a predicted transformation of the corresponding output image data I think needs to be predicated on whether or not that output image *is or is not a NIfTI format*. The current function does so based on `Header::do_not_realign_transform` or a manual boolean input parameter. Unfortunately I couldn't find a very clean way of detecting the image format, so it's currently a string comparison against `Header::format()`.

   There's also the issue here of how `mrinfo` should work, as opposed to `mrconvert`, since the former doesn't actually have an output image...

In additon to these:

1. It seems that for JSON import the flipping of phase and slice encoding directions was erroneous, due to the minus signs being in [the](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/core/file/json_utils.cpp#L93) wrong [place](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/core/file/json_utils.cpp#L105)... I think @bjeurissen reported something akin to this at some point? Or was it @dchristiaens?

2. I think that changing from comma-separated to space-separated slice timing on `master` in #1735 may have been a band-aid placed at the wrong location... Other vector data in `.mif` headers are stored as comma-separated. So it might have fixed the export of such data on `master`, but using the wrong approach.

[...]

PR #1888, opened 2020-01-21
Formats::is_nifti(): Fix for images with short names

Lestropie:
Bug in #1843, which results in integer overflow if an image file name is shorter than "`.nii.gz`" (which arises in `./run_tests scripts`).


Lestropie:
> You know about `File::Path::has_suffix()`, right...?

*Ssssshhhhhhhhhhhhhh*

PR #1889, opened 2020-01-21
dwifslpreproc: Fix strides of images input to applytopup

Lestropie:
Bug discovered after report on the [forum](https://community.mrtrix.org/t/dwifslpreproc-eddy-mask-nii-looks-faulty/3164).

Issue was introduced in `3.0_RC1`. Due to the generalisation of `dwipreproc` features, `applytopup` was no longer run on the NIfTI images provided as input to `topup`, instead being run on volumes extracted from the imported DWI series based on phase encoding information. That image possesses the same strides as the input image.


Lestropie:
Input:
screenshot0000

Before fix:
screenshot0001

After fix:
screenshot0002

PR #1891, opened 2020-01-22
Python: Better identification of string data

Lestropie:
I can't remember exactly what it was that caused me to start changing these, but I did eventually recall #1616 where we had an issue with Python2 passing unicode data to `run.command()`: the basic construct removed there for being problematic (testing whether a variable is of string type) is actually present in a number of other locations within the API.

This is the same technique as is used in the Python `six` module; but given that is not a built-in, I didn't want to introduce that dependency.

### DIFF
--- a/lib/mrtrix3/app.py
+++ b/lib/mrtrix3/app.py
@@ -469,7 +469,7 @@ class ProgressBar(object): #pylint: disable=unused-variable
   def __init__(self, msg, target=0):
     from mrtrix3 import run #pylint: disable=import-outside-toplevel
     global EXEC_NAME, VERBOSITY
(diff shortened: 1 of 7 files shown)


jdtournier:
Not enough experience with this to comment... But if that's the approach taken in another module, it can't be all that bad... And if it passes all the testing (including the scripts testing?), no objections. :+1:

Lestropie:
Passes testing; not that `dev` doesn't (if not for `population_template` data divergence). More of an API robustness thing. `six` is *the* Python 2 / 3 compatibility module, so if it's using this approach, it'll be for good reason.

Issue #1868, opened 2020-01-08
ENH tckgen: parameters via config file


Lestropie:
Thought arose while dealing with #1833.

While it's difficult to devise appropriate default tracking parameters that will work across all possible scenarios, it may be the case that some users may consistently use some set of parameters that work for the type of data they are working on at the time; in this case, it would be convenient to not have to manually type out these parameters upon every use, or to modify the macros in the code.

Instead, config file entries could be defined that would override the built-in default tracking parameters/

jdtournier:
I like the general idea, but I'm not sure about requiring users to modify their main config file. Users are not unlikely to forget that their defaults differ - I've been guilty of this in the past. I wouldn't be surprised to find we end up running round in circles trying to debug user issues only to discover they're simply not using our defaults...

What might work is to allow these parameters to be passed through a separate, explicitly specified config file via an environment variable. We could use the same format and backend as for the regular config file(s). Note that this approach would still allow these parameters to reside in the main config, so it doesn't necessarily _prevent_ the issue I'm worried about above. But it might make it (slightly?) less likely. And it would allow users to switch parameters easily between configs (e.g. for different studies). 

We could also extend that to all apps generally - I note that we already have the `MRTRIX_CONFIGFILE` environment variable, which allows users to override the system-wide config file. It'll still look for the user config file in their home folder, but it'll parse this one first _instead_ of the system config (this was added in #1104). We could add another environment variable to allow an additional config files to be parsed _after_ all the standard ones, allowing the user to override any parameters that might have been set in their system-wide or individual config files - but more importantly to preserve all those parameters if they're left untouched in this additional config. All parameters would be as normally used unless explicitly specified in that additional config. To do this, we might have to rename that `MRTRIX_CONFIGFILE` to e.g. `MRTRIX_SYSCONFIG`, and use `MRTRIX_CONFIGFILE` for the additional one to be parsed last. Or shorten it to `MRTRIX_CONFIG`, which might be a good idea to avoid conflict with older versions of _MRtrix3_...

Also: I realise this is pretty tangential to the original point of this issue... But I'm all for it, I just think we would need to be a bit careful with how it might be implemented.

Lestropie:
> Users are not unlikely to forget that their defaults differ

Would this be adequately addressed by a `CONSOLE()` call stating that default tracking parameters have been changed based on the config file (preferably only in cases where at least one such modified default is consequential)?

jdtournier:
[...]

PR #1894, opened 2020-01-22
mrview: add config option to show ortho view in a row

jdtournier:
Adds a config file option `MRViewOrthoAsRow` to show the 3 orthoview panes in a row, e.g.:
Screenshot from 2020-01-23 14-08-27

Any use...?

### DIFF
--- a/src/gui/mrview/mode/ortho.cpp
+++ b/src/gui/mrview/mode/ortho.cpp
@@ -192,6 +192,7 @@ namespace MR
         void Ortho::set_show_as_row_slot (bool state)
         {
           show_as_row = state;
+          frame_VB.clear();
           updateGL();
         }
 
--- a/src/gui/mrview/mode/ortho.cpp
+++ b/src/gui/mrview/mode/ortho.cpp
@@ -188,7 +188,9 @@ namespace MR
           updateGL();
         }
 
-        void Ortho::set_show_as_row (bool state) {
+
+        void Ortho::set_show_as_row_slot (bool state)
+        {
           show_as_row = state;
           updateGL();
         }
--- a/src/gui/mrview/mode/ortho.cpp
+++ b/src/gui/mrview/mode/ortho.cpp
@@ -188,6 +188,11 @@ namespace MR
           updateGL();
         }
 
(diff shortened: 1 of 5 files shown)


maxpietsch:
> Any use...?

Yes, useful for screenshots and widescreen monitors! I'd make it a view menu item or at east view tool option.

jdtournier:
I knew someone would suggest that...

Ok, that would be easy to add, but this is one option that would really benefit from being persistent (i.e. automatically written back to the config file if modified in the UI). That would require a few additions to the config file handling - I'd been meaning to do this for some time, but it's pretty low priority... Also relates to #1868. And I'd be inclined to write that to a different config file than the main one to avoid the risk of inadvertently clobbering it. But that could easily be left for some other time...

jdtournier:
BTW: I reckon this would live as a little check box in the view tool (it's similar to the lightbox mode's settings for number of rows & columns, for instance). Are you interested in taking that on...? :grin:

maxpietsch:
I've spent quite some time on adding a little checkbox... Nearly there I think but I need some help with connecting the QT slot. Does not compile when I uncomment [this line](https://github.com/MRtrix3/mrtrix3/pull/1894/commits/86a483e778d65c9dd08db79a070520a2047e8523#diff-946b56617a4ae8cfde2faecc0c250589R1073).

Lestropie:
@jdtournier: I'd make it a checkbox option in the View tool. :stuck_out_tongue_winking_eye: 

> I need some help with connecting the QT slot.

Why should you not be made to suffer as we have suffered? :smiling_imp:

I'll try to have a peek.

jdtournier:
As far as I can tell, all you were missing was:
```
#include "gui/mrview/mode/ortho.h"
```
in `src/gui/mrview/tool/view.cpp`...

I've made other changes though. Incoming...

maxpietsch:
There is still a bug in it: switching between ortho and other views makes the checkbox and the display inconsistent (until the checkbox is clicked once or twice depending on the default).

jdtournier:
OK, I think that'll fix it. I'd forgotten that modes gets destroyed and constructed between switches - they're not persistent...

PR #1901, opened 2020-01-25
population_template: Keyvals of output data

Lestropie:
- For output warps, utilise the key-values of the warp image within the scratch directory, but erase the "`command_history`" field; the only entry for this field in the output image header will therefore be the `population_template` call.

- For output linear transforms, rather than copying the file from the scratch directory to the output path, import the data and then re-save; this will make the "`command_history`" field in the text file header comment reflect the `population_template` call.

Without this change, output warp images do not contain the requisite transform information within their headers and therefore cannot be used.

### DIFF
--- a/bin/population_template
+++ b/bin/population_template
@@ -17,7 +17,7 @@
 
 # Generates an unbiased group-average template via image registration of images to a midway space.
 
-import collections, math, os, sys, shutil, re
+import collections, json, math, os, re, shutil, sys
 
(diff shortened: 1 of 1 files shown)


maxpietsch:
I could have sworn I fixed this before...

I've removed all but the required key-value entries (`linear1` and `linear2`) as the others are only relevant to the last registration stage (`nl_disp_smooth` `nl_gradient_step` `nl_lmax` `nl_niter` `nl_scale` `nl_update_smooth` ...)

PR #1903, opened 2020-01-25
mrtrix3.matrix: Precision changes

Lestropie:
- When saving numerical data, default to full double-precision floating-point width (erasing trailing zeros); and for matrix / transform data, print with aligned columns; these are to be consistent with C++ API and not accidentally lose precision. Alters what was done in adda703a (#1775), and makes a09ee51c (#1792) redundant.

- New function `phaseencoding.save()`, which will properly write the phase encoding direction values as integers but the readout time as a float.

Twice in a row ended up with numerical data of lower precision than what I wanted. So decided it's better to go full-precision, and in specific cases where a lower precision is wanted to reduce file size, developer can request such. Also discovered the "`g`" specifier for `format()`, which removes trailing zeros from data that don't need it.


Lestropie:
Made sure `dwifslpreproc` tests were passed before merging.

PR #1927, opened 2020-01-30
Python API minor fixes

Lestropie:
Explanations in commit messages.


Lestropie:
Reverted that last one. Forgot the logic behind it: `run.function()` needs to permit passing keyword arguments down to the underlying function; so `show` is the only one it reads, and unlike other API functions, `kwargs` is then *not* verified to be empty.

PR #1928, opened 2020-01-30
Allow parse_ints() to process whitespaces

jdtournier:
Going through old stale branches and came across this. Any use...?

### DIFF
--- a/core/mrtrix.cpp
+++ b/core/mrtrix.cpp
@@ -88,12 +88,13 @@ namespace MR
         }
         else num[i] = to<int> (spec.substr (start, end-start));
 
+        end = spec.find_first_not_of (" \t", end);
         char last_char = end < spec.size() ? spec[end] : '\0';
         if (last_char == ':') {
-          i++;
+          ++i;
+          ++end;
           if (i > 2) throw Exception ("invalid number range in number sequence \"" + spec + "\"");
-        }
-        else {
+        } else {
           if (i) {
             int inc, last;
             if (i == 2) {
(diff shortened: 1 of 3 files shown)


Lestropie:
So it appears to just be jumping over any whitespace it finds and ignoring it? There's probably some interaction with #1794 here, but then again #1823 might well render it redundant?

It feels like something [extensive unit testing](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/testing/cmd/testing_to.cpp) would clarify...

jdtournier:
> So it appears to just be jumping over any whitespace it finds and ignoring it?

Yep, that's pretty much it...

> There's probably some interaction with #1794 here, but then again #1823 might well render it redundant?

... Maybe, but I'm not sure I see how? As far as I can tell, these were about stripping before conversion of single values. This one is specifically about allowing spaces in number sequences, so allowing e.g.:
```
-lmax "0, 0, 8"
-coord 3 "1 : 10"
```
or even (and more importantly):
```
-lmax "0 0 8"
```
(i.e. space is interpreted as a comma). Consecutive spaces are treated as one. At least that's how I remember it - I did this quite a while ago... 

The reason that last one is more important is that this allows more leniency when parsing config file and header entries that we currently parse using that function. 

> It feels like something extensive unit testing would clarify...

Yep, that's definitely a good idea...

jdtournier:
Adding tests was a good idea... There was a slight glitch when handling e.g. "1 :5" - now fixed.

PR #1930, opened 2020-01-30
MRView ODF tool: add option to colour relative to projection

jdtournier:
Another stale branch without a pull request... 

This adds a tick-box to request the directional colouring of the ODF to be relative to the frame of the viewer (i.e. after rotating it) rather than the scanner/world axes. This is useful for applications where the anatomy is acquired in non-standard orientation, particularly neonatal or fetal imaging.


jdtournier:
Not convinced 'colour by projection' on the tick-box does this justice... Maybe we could rename this to 'colour relative to viewer' and add a tooltip to explain what it does?

Lestropie:
Yeah pretty sure most users won't understand what is meant by "projection". "Colour relative to camera"? Or have a combobox with "`Colour: [ <Fixed> <Relative to scanner> <Relative to camera> ]`" (theoretically "relative to image" would also fit in the same framework)?

jdtournier:
Indeed, there's a lot of similarities with the screenshot tool's rotation axis setting. It does suggest a combobox with all 3 settings would be best here. Won't have time to do this for some time though...

Regarding the terminology: 'camera' is the obvious choice here, but there's a difference between that and how it's used in the screenshot tool: in the latter, it really means in eye coordinates. Here, it means relative to the rotation applied to the modelview - as encoded by the orientation quaternion. So the colours change as you rotate the view, but are the same irrespective of whether you're looking at the axial, sagittal or coronal cuts. So it's not strictly relative to camera either (though it would be for the axial projection). Not sure what's best here, but it may be that 'relative to camera' is the cleanest description we're gonna get even if it's not strictly correct...

dchristiaens:
I have been using this branch for 2 years for all fetal data visualisation. It works great; I have never encountered any issues. Very happy to have this included in the release.

jdtournier:
Upon discussion, it would be better long-term to have that option apply to everything that has a colour-by-direction option - not just the ODF. We can merge this now since it's pretty much good to go, maybe just rename the option, but otherwise let's create an issue for the remaining functionality and merge this.

dchristiaens:
Tested once more. This is good to merge as far as I'm concerned.

Just one minor point regarding the new name "colour by camera" though: keep in mind that the colours do not (and should not) change when switching between axial/sagittal/coronal views. The general idea is that the user can align the brain into standard axis in ortho view and have the ODFs coloured relative to that "projection". I think this is the reason why we initially avoided using the name "camera". Personally, I don't care very much how the option is called. I just want to highlight that going for "colour by camera" may also lead to confusion.

jdtournier:
Great, thanks!

PR #1931, opened 2020-01-30
Use thread local on macos

jdtournier:
It looks like XCode finally supports this as of version 8:
https://stackoverflow.com/a/29929949/3244166

This unifies the handling of the RNG instance used in the tractography code between macOS and the other platforms, and avoid having to handle it using a clunky pointer.

### DIFF
--- a/src/dwi/tractography/algorithms/iFOD1.h
+++ b/src/dwi/tractography/algorithms/iFOD1.h
@@ -200,7 +200,7 @@ namespace MR
                 max_truncation = val/max_val;
             }
 
-            if (uniform(*rng) < val/max_val) {
+            if (uniform(rng) < val/max_val) {
               dir = new_dir;
               dir.normalize();
               pos += S.step_size * dir;
--- a/src/dwi/tractography/algorithms/iFOD2.h
+++ b/src/dwi/tractography/algorithms/iFOD2.h
@@ -271,7 +271,7 @@ end_init:
                     max_truncation = val/max_val;
                 }
 
-                if (uniform(*rng) < val/max_val) {
+                if (uniform(rng) < val/max_val) {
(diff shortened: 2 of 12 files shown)


Lestropie:
Fully reverting the code will result in non-compilation on XCode < 8 and iOS <= 8.4 based on that StackOverflow response. I don't follow how Mac updates work, but that seems to be only about five years old. Is there older Mac hardware that no longer receives software updates beyond a certain point, like what happens with Android phones? Is there a risk of us borking previously supported hardware that's still in use?

jdtournier:
> Is there a risk of us borking previously supported hardware that's still in use?

No idea. That'll depend on whether you can install XCode 9 on said older hardware, even if it's no longer receiving OS updates. That said, I have a feeling even the oldest macbooks we can realistically support are at least capable of running relatively recent versions of macOS (and hence XCode), even if not Catalina. 

@maxpietsch: what version of macOS / XCode did you have on your old MacBook Air? And how old was it? I reckon if that's supported, that'd be enough...

jdtournier:
OK, after discussions, I think there's a general agreement this is safe enough to merge. A quick google reveals that XCode 8 can be installed on macOS 10.11. 5 or above (El Capitan), which came out in 2015. And I'm pretty sure if you can't install that on your mac, it's definitely time for an upgrade (I was using an old 2008 macbook that I _think_ ended up running that version)...

More to the point, that's the minimum requirement for Qt 5.1, which is the minimum required to get mrview to run - at least on anaconda. Besides, the Qt docs suggest the [minimum version they support currently is 10.13](https://doc.qt.io/qt-5.14/supported-platforms.html#macos). So I reckon we're fine on all counts...

I'll merge now.

PR #1939, opened 2020-02-05
Ordered thread queue

jdtournier:
As discussed recently. I had indeed looked into extending the Thread::Queue to preserve item order - yet another one of those stable branches gathering dust...

I've not tested this since I cooked it up 10 months ago, so we'll see if it all checks out in testing...

Relates to #1589


jdtournier:
Guess I'd better move that `cmd/test_ordered_queue.cpp` over to the unit tests... :grin:

jdtournier:
@Lestropie: when you have a minute, check out the prerequisites in the commit message for 425e53b, see if they match what's going on in #1589.

Lestropie:
I've got a feeling that parts of #1589 will need to be cherry-picked; not because #1939 isn't feature-complete in its own context, but because there are bits in #1589 that are applicable beyond the scope of maintaining ordering within a multi-threaded queue:

- Adding a streamline index to track scalar data (and indeed providing a named class for storing the combination of those two) may still be useful (& would be consistent with the `Streamline` class);

- Disambiguating between a streamline that should not be written to file, and an empty streamline that should nevertheless be written to file, is still a necessary change due to `tcktransform`. Exploiting the existing `Streamline::weight` member does work, but is not the most intuitive; an explicit `bool Streamline::valid` would be the alternative.

Given the extent of testing of this branch, it's probably good to merge; though I would suggest that modifying the relevant command help pages to remove the warning about streamline reordering belongs here. I will need to produce a subset of #1589 in another branch & PR.

Lestropie:
I've finished integrating the ordered thread queue into the code base and updated the tests accordingly. CI tests pass on my system, including more stringent testing of streamlines-based commands. Code changes are open to independent review.

I believe the outstanding difference to #1589 to resolve is the management of `tcktransform` when an entire streamline resides outside of the FoV. Currently, this will be detected by the file writer as an absent streamline, and will not increment the track counter, resulting in a potential mismatch between input and output streamline count. The more strictly correct behaviour would be to utilise some flag associated with the data for each streamline (in #1589 I used a non-finite weight, but an explicit boolean would be more clear) indicating whether or not it should be written to the output file. In this way, a streamline with no vertices can still be written to the output file. There are probably also some other changes within the code in #1589 that I'll pull out and make use of.

PR #1950, opened 2020-02-12
mrview: better handling of UTF-8 encodings

jdtournier:
Should fix issues identified while trying to use mrview on Brazilian systems...


jdtournier:
Confirmed as working by users in Brazil. I'll merge this now.

PR #1951, opened 2020-02-13
population_template: fix to mapping between mask and images

maxpietsch:
fix to the prefix and suffix removal for masks required to match images and masks under certain circumstances (due to using rstrip which matches a _set_ of characters)


maxpietsch:
Merging as it's a clear bug and I couldn't break the proposed fix on a number of datasets.

PR #1965, opened 2020-03-03
Fix incorrect total readout time when parsing Siemens mosaics

jdtournier:
As reported on forum:
https://community.mrtrix.org/t/dcm2niix-and-mrinfo-different-total-readout-time/3224

Issue was simply that computing the readout time requires knowledge of the matrix size, and in the case of Siemens mosaics, this entry is not the acquired matrix size.

### DIFF
--- a/core/file/dicom/image.cpp
+++ b/core/file/dicom/image.cpp
@@ -254,12 +254,6 @@ namespace MR {
                   }
                 }
                 return;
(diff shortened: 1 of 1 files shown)


jdtournier:
Ran my DICOM tests, everything passes that should pass*.



_*there's one data set with inconsistent intensity scaling that only passes on `dev`._

PR #1968, opened 2020-03-05
matrix.save_numeric(): Fix enforcement of -force option

### LINKED ISSUE #1967: dev: save_numeric broken
`population_template` fails on `dev` on macOS with python 3.7.4 in `save_numeric`'s `with os.open` call. I think there is a context manager wrapper missing in dc64d1a46fd94be96bae8d9c593289712e19a236. 

This reproduces the error on macOS and ubuntu.

```python
import os
with os.open('test', os.O_WRONLY | os.O_CREAT) as f:
    f.write('test')

# Traceback (most recent call last):
#  File "<stdin>", line 1, in <module>
# AttributeError: __enter__
```

Lestropie:
Closes #1967.

Used solution from [here](https://bugs.python.org/issue29214#msg285046).

### DIFF
--- a/bin/population_template
+++ b/bin/population_template
@@ -1411,7 +1411,8 @@ def execute(): #pylint: disable=unused-variable
       matrix.save_transform(os.path.join(linear_transformations_path,
                                          xcontrast_xsubject_pre_postfix[0] + inp.uid
                                          + xcontrast_xsubject_pre_postfix[1] + '.txt'),
-                            trafo)
+                            trafo,
+                            force=app.FORCE_OVERWRITE)
 
   if app.ARGS.transformed_dir:
     for cid, trdir in enumerate(app.ARGS.transformed_dir):
--- a/bin/population_template
+++ b/bin/population_template
@@ -1218,7 +1218,7 @@ def execute(): #pylint: disable=unused-variable
       if average_inv is not None:
         for inp in ins:
           transform = matrix.dot(matrix.load_transform(os.path.join('linear_transforms_%02i' % level, inp.uid + '.txt')), average_inv)
-          matrix.save_transform(os.path.join('linear_transforms_%02i' % level, inp.uid + '.txt'), transform)
+          matrix.save_transform(os.path.join('linear_transforms_%02i' % level, inp.uid + '.txt'), transform, force=True)
 
       for cid in range(n_contrasts):
         for inp in ins:
--- a/bin/responsemean
+++ b/bin/responsemean
@@ -75,7 +75,7 @@ def execute(): #pylint: disable=unused-variable
     weighted_sum_coeffs = [ [ a + multiplier*b for a, b in zip(linea, lineb) ] for linea, lineb in zip(weighted_sum_coeffs, subject) ]
 
   mean_coeffs = [ [ f/len(data) for f in line ] for line in weighted_sum_coeffs ]
-  matrix.save_matrix(app.ARGS.output, mean_coeffs)
+  matrix.save_matrix(app.ARGS.output, mean_coeffs, force=app.FORCE_OVERWRITE)
 
 
 
--- a/lib/mrtrix3/matrix.py
+++ b/lib/mrtrix3/matrix.py
@@ -212,7 +212,8 @@ def save_numeric(filename, data, **kwargs):
   open_mode = os.O_WRONLY | os.O_CREAT
   if not app.FORCE_OVERWRITE:
     open_mode = open_mode | os.O_EXCL
-  with os.open(filename, open_mode) as outfile:
+  file_descriptor = os.open(filename, open_mode)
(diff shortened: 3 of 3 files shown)


maxpietsch:
```
population_template: [ERROR] mrtransform  -reorient_fod no /Users/mp/mrtrix3_dev3/testing/scripts/data/tmp-fa/sub-02.mif -template initial_template_c0.mif -linear linear_transforms_00/2.txt input_transformed_c0/2.mif -nan -datatype float32 (population_template:1231)
population_template: [ERROR] Information from failed command:
population_template:
                     mrtransform: [WARNING] existing output files will be overwritten
                     mrtransform: [ERROR] uneven rows in matrix file "linear_transforms_00/2.txt" (first row: 4 columns; row 5: 2 columns)


cat linear_transforms_00/2.txt
# command_history: /Users/mp/mrtrix3_dev3/bin/population_template tmp-fa/ ../tmp/population_template/fa_masked_template.mif -mask_dir tmp-mask/ -template_mask ../tmp/population_template/fa_masked_mask.mif -force  (version=3.0_RC4-562-g0141d2b0)
 0.999878011470898  0.015492256510744 -0.001988005333866 -0.418593196347680
-0.015520775202990  0.999763451097328 -0.015236383660791  3.891172940951069
 0.001751489109416  0.015265380380649  0.999881943255168  2.054901613767539
 0.000000000000000  0.000000000000000  0.000000000000000  1.000000000000000
31294480907318 -1.9130485547028
0.00248087977945782 0.0131611471833193 0.999910310698084 9.71394892856104
0 0 0 1
```

maxpietsch:
One issue is that save_numeric is tied to the command's overwrite flag. I've added a keyword to allow overwriting files within a script.

The other issue is that files do not get overwritten:

```python
In [14]: open_mode = os.O_WRONLY | os.O_CREAT
In [15]: file_descriptor = os.open('filename', open_mode)
In [16]: with open(file_descriptor, 'wb') as outfile:
    ...:     outfile.write(('ABCDEFG').encode(**encode_args))
    ...:     
    ...:     

In [17]: cat filename
ABCDEFG

In [18]: file_descriptor = os.open('filename', open_mode)
In [19]: with open(file_descriptor, 'wb') as outfile:
    ...:     outfile.write(('abc').encode(**encode_args))
    ...:     
    ...:     
In [20]: cat filename
abcDEFG
```

My current solution is to delete the file before writing but you might want to change that.

Lestropie:
:man_facepalming: Looks like I made a right meal out of this one. Thanks for the fixes. 

I changed to using the `O_TRUNC` flag to truncate any existing file. I also changed the handling of `app.FORCE_OVERWRITE` to be more consistent with the rest of the API.

bjeurissen:
I just tested this branch and get:
```
population_template: [ERROR] Unhandled Python exception:
population_template: [ERROR]   TypeError: coercing to Unicode: need string or buffer, int found
population_template: [ERROR] Traceback:
population_template: [ERROR]   /data/home/bjeurissen/matrix_save_numeric_fix/bin/population_template:1224 (in execute())
population_template: [ERROR]     matrix.save_transform(os.path.join('linear_transforms_%02i' % level, inp.uid + '.txt'), transform, force=True)
population_template: [ERROR]   /data/home/bjeurissen/matrix_save_numeric_fix/lib/mrtrix3/matrix.py:264 (in save_transform())
population_template: [ERROR]     save_matrix(filename, data, **kwargs)
population_template: [ERROR]   /data/home/bjeurissen/matrix_save_numeric_fix/lib/mrtrix3/matrix.py:250 (in save_matrix())
population_template: [ERROR]     save_numeric(filename, data, **kwargs)
population_template: [ERROR]   /data/home/bjeurissen/matrix_save_numeric_fix/lib/mrtrix3/matrix.py:218 (in save_numeric())
population_template: [ERROR]     with open(file_descriptor, 'wb') as outfile:
```

bjeurissen:
should we also use io.open() for the other call in the same script?

maxpietsch:
As far as I can tell there is no other usage of a [file descriptor](https://stackoverflow.com/a/15039662/2389450) in the library that would necessitate calling `io.open`. What `open` call do you mean?

`open` works in both python2 and 3 for path-like arguments, the reason a file descriptor and `os.open` are used here it to avoid race conditions in (not) over-writing existing files. Where this isn't necessary I'd stick to `open`.

bjeurissen:
> As far as I can tell there is no other usage of a [file descriptor](https://stackoverflow.com/a/15039662/2389450) in the library that would necessitate calling `io.open`. What `open` call do you mean?
> 
> `open` works in both python2 and 3 for path-like arguments, the reason a file descriptor and `os.open` are used here it to avoid race conditions in (not) over-writing existing files. Where this isn't necessary I'd stick to `open`.

:+1:

PR #1972, opened 2020-03-10
dwinormalise warp fix

maxpietsch:
fix to `dwinormalise group` necessitated by name changes of temporary files within the `population_tempalte` scratch directory.


Lestropie:
Test runs once merged with #1968.

`dwinormalise group` test fails presumably due to minor changes to `population_template behaviour; can address as a separate issue of lower priority.

PR #1988, opened 2020-03-19
Partial merge of PR#1589

Lestropie:
Manually went through #1589 and determined which items I wanted to keep following review of #1939. This is however a sub-branch of #1939 and should therefore be merged to `dev` after such.

In the end of this process, I have not actually dealt with the main functionality of #1589 that I was originally targeting, being the implementation of an explicit flag to decide which streamlines should be written to file rather than using the absence of vertices as a flag of such. I may instead pursue that as a separate PR as it should be somewhat independent of the other things I've ended up chasing down during this process (#1589 probably has more in there than initially advertised).

So main changes here are:

- Some standardisation of code around track scalar data (even if the common base class with streamlines data isn't actually to be used for its original intended purpose, it's still handy to have unified handling of index across both types of data).

- Modification of `tckedit` behaviour in the instance where both `-inverse` and `-mask` options are used. It's now possible to output only those streamline vertices that are outside of the specified mask.

- Implementation of `tckedit` CI tests.


src/dwi/tractography/streamline.h
```diff
@@ -31,58 +31,99 @@ namespace MR
...
+            weight = that.weight; that.weight = 0.0f;
             return *this;
           }
 
 
           void clear()
           {
             vector<point_type>::clear();
-            index = -1;
+            DataIndex::clear();
```
jdtournier: Shouldn't that be `TrackScalar::clear()`...?

jdtournier: Nope, ignore - my mistake...

PR #1996, opened 2020-03-30
DICOM import & dcminfo: Enhancements

Lestropie:
Back-end changes necessary to support some work I'm doing locally. Details in 3a9ff73 commit message.

@jdtournier: Please make sure this doesn't bork your DICOM test suite.

### DIFF
--- a/cmd/dcminfo.cpp
+++ b/cmd/dcminfo.cpp
@@ -36,7 +36,8 @@ void usage ()
   OPTIONS
   + Option ("all", "print all DICOM fields.")
 
-  + Option ("csa", "print all Siemens CSA fields")
+  + Option ("csa", "print all Siemens CSA fields (excluding Phoenix unless requested)")
+  + Option ("phoenix", "print Siemens Phoenix protocol information")
 
   + Option ("tag", "print field specified by the group & element tags supplied. "
       "Tags should be supplied as Hexadecimal (i.e. as they appear in the -all listing).")
@@ -86,13 +87,17 @@ void run ()
 
   File::Dicom::QuickScan reader;
 
-  if (get_options("all").size())
+  const bool all = get_options("all").size();
+  const bool csa = get_options("csa").size();
+  const bool phoenix = get_options("phoenix").size();
+
+  if (all)
     print (File::Dicom::Element::print_header());
 
-  if (reader.read (argument[0], get_options ("all").size(), get_options ("csa").size(), true))
+  if (reader.read (argument[0], all, csa, phoenix, true))
     throw Exception ("error reading file \"" + reader.filename + "\"");
 
-  if (!get_options ("all").size() && !get_options ("csa").size())
+  if (!all && !csa && !phoenix)
     std::cout << reader;
 }
 
--- a/core/file/dicom/csa_entry.h
+++ b/core/file/dicom/csa_entry.h
@@ -128,6 +128,18 @@ namespace MR {
                   v[m] = NaN;
               }
 
+            vector<std::string> get_string () const {
+              vector<std::string> result;
+              const uint8_t* p = start + 84;
+              for (uint32_t m = 0; m < nitems; m++) {
+                const uint32_t length = Raw::fetch_LE<uint32_t> (p);
+                std::string s (reinterpret_cast<const char*> (p)+16, length);
+                result.push_back (std::move (s));
+                p += 16 + 4*((length+3)/4);
+              }
+              return result;
+            }
+
             friend std::ostream& operator<< (std::ostream& stream, const CSAEntry& item) {
               stream << "[CSA] " << item.name << " (" + str(item.nitems) + " items):";
               const uint8_t* next = item.start + 84;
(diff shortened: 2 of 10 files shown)


jdtournier:
OK, I ran it through my DICOM test suite - everything checks out _apart_ from some of the Siemens localisers. Here's the output from one of those:
```
mrinfo: [ 11%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 22%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 33%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 44%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 56%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 67%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 78%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [ 89%] reading DICOM series "localizer_sf"... 
mrinfo: [WARNING] error reading tag (0029,1020):
mrinfo: [WARNING] incomplete use of string "0x10" in conversion to type "m"
mrinfo: [100%] reading DICOM series "localizer_sf"
mrinfo: [WARNING] slice gap detected (maximum gap: 7mm)
mrinfo: [ERROR] unexpected phase encoding table matrix dimensions
mrinfo: [ERROR] error opening image "../Siemens Trio - mosaic/"
```
I think the warnings will probably end up demoted to `-debug` with #1999. But the main error is at the end. I reckon these shouldn't be treated as errors, but reported with a warning, and the PE information ignored and kept out of the header. It shouldn't be considered fatal if the PE information is inconsistent... What do you think?

Lestropie:
I can't reproduce the warnings... can you send some data?

0029 1020 is a Siemens CSA ("OBCSASeriesHeaderInfo" in the dictionary), and gets parsed by `CSAEntry`. Whatever it is, it looks like it should ideally be going to either a string or a dedicated hexadecimal converter. I'm not sure for what type "`T`" `typeid(T).name()` comes out as "`m`" though. Chris Rorden says [here](https://neurostars.org/t/reading-csa-series-header-info-from-a-pydicom-object/2722/2) that contents of that specific tag are unpredictable, so maybe it should be ignored altogether if our `CSAEntry` parser isn't robust to it.

jdtournier:
OK, it still borks with the same error message... 

I had a bit of a dig around the code to chase this down. The error is emitted in the [PhaseEncoding::get_scheme()](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/phase_encoding.cpp#L118-L119) when it is invoked from [Header::realign_transform()](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/header.cpp#L527) (which is invoked by [Header::open()](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC4/core/header.cpp#L146-L147)), because the PE table at this point is a 3×3 matrix, when it's expected to be *n*<sub>volumes</sub>×4:
```
mrinfo [core/phase_encoding.cpp: 118]: result = 1 0 0                                                                         
1,0,0
0,1,0
```
The code that was enclosed in a try/catch clause in befb66e actually runs successfully, but the PE table just before [this point in DICOM::Frame::get_PE_scheme()](https://github.com/MRtrix3/mrtrix3/blob/befb66e/core/file/dicom/image.cpp#L652-L654) is:
```
mrinfo [core/file/dicom/image.cpp: 651]: pe_scheme = 1 0 0 0
1 0 0 0
0 1 0 0
```
and that last column adds up to zero, so fails the subsequent test, with the (unexpected?!?) result that it returns just the first 3 columns... I'm not sure why that's there, but it's pretty clear that there's an expectation that it should have 4 columns later on...?

Lestropie:
The intent is for all code handling phase encoding information to be compatible with matrices with or without the total readout time column (not too dissimilarly to how the diffusion gradient table handling is intended to work on matrices with at least four columns but optionally more). So the check in `PhaseEncoding::get_scheme()` is contrary to that intent. I'll change that and see if I can get code to bork anywhere else with 3-column data.

PR #2012, opened 2020-04-03
DICOM selector quit aborts cleanly

### LINKED ISSUE #1656: DICOM series selection abort causes error
When providing a DICOM folder to a command as input, as we all know, the folder(s) get scanned for proper series and you get a "select series" menu, where a series can be selected by typing a number (0-...) and hitting enter. Entering anything else than the valid number results in a 

```
[ERROR] no DICOM series selected
mrinfo: [ERROR] error opening image "FOLDER_YOU_PROVIDED"
```

However, the initial prompt also mentions

```
Select series ('q' to abort):
```

But however then again, entering `q` and hitting enter results in the same above error. As a user, I'd expect no error in this scenario, as it's a valid option.

jdtournier:
Closes #1656

### DIFF
--- a/cmd/mrinfo.cpp
+++ b/cmd/mrinfo.cpp
@@ -276,40 +276,46 @@ void run ()
       json_keyval || json_all);
 
   for (size_t i = 0; i < argument.size(); ++i) {
-    auto header = Header::open (argument[i]);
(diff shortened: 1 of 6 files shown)


bjeurissen:
Does the job!

PR #2014, opened 2020-04-08
NIfTI: fix incorrect handling of AnalyzeLeftToRight option

jdtournier:
This has obviously been incorrect on `master` for some time... :man_facepalming: 

I'm creating this pull request here mainly to document the issue. I don't think it's worth merging given that this won't be an issue on `dev`, and we're about to merge anyway.


Lestropie:
Linked in changelog.

PR #2015, opened 2020-04-08
DW scheme handling update

### LINKED ISSUE #1487: fix up handling of bvalue scaling and bvec normalisation
OK, prompted by Chris Rorden, based on [this discussion](https://github.com/rordenlab/dcm2niix/issues/245), I've had a look into our handling of the bvalue scaling, and it's fair to say that it's a bit messy at the moment... 

To start off with, following #60, the DICOM import code and any subsequent conversion will leave the DW scheme unmodified: no normalisation of bvecs, no bvalue scaling. All those modifications happens in the app processing the data. 

Where this is problematic is for 3rd party apps that rely on `mrconvert` for the DICOM to NIfTI conversion: if they expect the information to be usable as-is, they might come unstuck on e.g. multi-shell data acquired on a Siemens.

The other situation where this is annoying is when you have data where the bvalue scaling _shouldn't_ happen: you then need to remember to pass the `-bvalue_scaling false` option every time you use a command that uses the DW encoding. There's no simple way to store the information as you expect it to be used, rather than as the scanner reported it.

So that argues for at least providing options to control that within `mrconvert`. We may also need control over normalisation of the bvecs without bvalue scaling (for instance, I just came across some Bruker data with non-unit vectors, but correct bvalues). At least that would provide users with some control over the interpretation of the DW table that they can apply once at DICOM import and then not have to worry about it, in those cases where it would otherwise be annoying.

But equally pressing is to sort out the glaring inconsistencies in how we handle these things currently. This is what I've observed from a multi-shell dataset that needs bvalue scaling:

---

```
mrinfo DICOM/ -dwgrad
```
shows the correct encoding (scaled by b-value), as it should.

---
```
mrconvert DICOM/ dwi.nii -export_grad_fsl bvecs bvals
mrconvert DICOM/ dwi.nii -export_grad_mrtrix enc.b
```
results in `bvecs`/`bvals` or an `enc.b` with no scaling by b-value, no unit normalisation of gradient vectors.

---

```
mrinfo DICOM/ -export_grad_fsl bvecs bvals
mrinfo DICOM/ -export_grad_mrtrix enc.b
```
results in `bvecs`/`bvals` or an `enc.b` _with_ scaling by b-value, and unit normalised vectors. 

---

```
mrinfo dwi.nii -fslgrad bvecs bvals -dwgrad
mrinfo dwi.nii -grad enc.b -dwgrad
```
both report whatever was stored in the corresponding `bvecs`/`bvals` or `enc.b`: no bvalue scaling, no unit normalisation. Yet this is supposed to report the DW encoding as it would be interpreted in our apps - which it gets right if run on the DICOM directly (see the first call above). I'm not even sure how that's possible, there must be some weird code path in there...

---

I'm not sure how to deal with this properly, but open to suggestions. I think at the very least, we need to ensure consistency in how the `-export_grad_*` options work, and in how the gradient table is interpreted, regardless of where it came from (DICOM headers or provided by `-grad` or `-fslgrad`). I also reckon we should add options to control the bvalue scaling and bvecs normalisation separately. What I'm not sure about is whether these should be provided as part of the generic DW gradient table import (or export?) option group, and whether they should be performed by default (which I'm a little reluctant to do, there was a reason we went along with #60). Thoughts welcome.

jdtournier:
First go at implementing the changes discussed in #1487. There's bound to be niggles to fix, but hopefully this'll be a good enough start. 

One issue is that I don't seem to have suitable test data to verify correct operation. Does anyone have some data that definitely needs b-values scaling that I could run through its paces...? 

Closes #1487


Lestropie:
@jdtournier Changes incoming imminently; trying to avoid working on the same thing in parallel.

Lestropie:
I generated test data by loading an image into Octave, reading the gradient table, scaling the directions by the square root of the b-value, and then writing two output images: one where I reset all b-values to the maximal value in the series (to match the vector norm scaling), and one where I didn't. So in the former case, *MRtrix3* scaling *b*-values by the square of the norm prior to normalising the vectors restores the original table; in the latter case, the explicit command-line option in `mrconvert` is necessary to retain the input *b*-values *despite* the necessity for direction vector rescaling.

These two cases are now both working following a809727. Previous the `-bvalue_scaling` option in `mrconvert` was never read. I also turned it into a command-line flag `-no_bvalue_scaling`, as this should be a rare necessity rather than a "do I choose 0 or 1?". I also put `-raw_dwgrad` back into `mrinfo`, as if someone's raw input data have a dodgey gradient table they need to be able to not just fix it but first diagnose it.

jdtournier:
> Previous the -bvalue_scaling option in mrconvert was never read.

I knew I'd forgotten something...

> I also turned it into a command-line flag -no_bvalue_scaling, as this should be a rare necessity rather than a "do I choose 0 or 1?".

Good to hear - I had the same thought. 

> I also put -raw_dwgrad back into mrinfo, as if someone's raw input data have a dodgey gradient table they need to be able to not just fix it but first diagnose it.

I thought we'd decided not to...? Users can easily inspect their externally-supplied schemes manually anyway. I can see there's an argument to be made for inspecting the table in the mif header though, given that the current implementation invokes `DWI::get_DW_scheme` regardless of whether anything DW-related has been requested.

I think I might make a couple of changes to the order in which the options are processed to ensure that the properties shown are the unprocessed version - i.e. process the `-property` option _before_ `DWI::get_DW_scheme()`, and anything DW-related after that. This would then mean we don't need the `-raw_dwgrad` option, since we can then produce it trivially using `mrinfo -property dw_scheme`. 

If we do decide to reinstate the `-raw_dwgrad` option though, I think we'd need to change what you've got so it matches what used to happen before (or least what should have happened): with `-raw_dwgrad`, _anything_ DW-related is produced using the unmodified DW encoding, including `-shell_bvalues`, `-shell_sizes`, `-shell_indices`, and all the gradient export options. 

On that note, I reckon it would make sense to also provide the `-no_bvalue_scaling` option to `mrinfo`, in case it's used purely as a means of converting the gradient table between formats, or just cleaning it up.

jdtournier:
BTW: I'll make changes along the lines suggested above and commit, see what you think - just to avoid working in parallel...

jdtournier:
OK, latest changes implement what I was talking about in the previous post - broken down into 3 commits. Tested using data generated as suggested by @Lestropie. 

We can easily revert the last commit if it's felt the `-raw_dwgrad` option should stay - although I'd still argue it would need to trigger loading into the same `grad` table, but using `DWI::get_raw_DW_scheme()` rather than `DWI::get_DW_scheme()`, so all subsequent options use that version of the gradient table. 

Tests are currently failing. On inspection, this is because we now only scale b-values if non unit norm vectors are detected (with a generous tolerance of 1e-3), so the b-values are now no longer scaled. To illustrate, here's the first few lines of `mrinfo -dwgrad` using `dev` and this PR for the `dwi.mif` test data:

**`dev`:**
```
0 0 0 0
0 0 0 0
-0.05095409193 0.06175509022 -0.9967898421 2950.000935
0.01190699722 0.9550467768 0.2962159308 3000.001403
-0.5251151245 0.8399851992 0.1366710324 2999.998577
-0.7854452199 -0.6111001711 -0.09814472747 2999.99832
0.06086199581 -0.4567009685 0.8875359389 3000.000413
0.3983250059 0.6676990099 0.6289000094 2999.999911
-0.6806035944 0.689644589 -0.2473238526 3000.003575
0.237398968 0.9699948692 0.05245649293 3000.000809
...
```

**this PR:**
```
0 0 0 0
0 0 0 0
-0.05095409193 0.06175509022 -0.9967898421 2950
0.01190699722 0.9550467768 0.2962159308 3000
-0.5251151245 0.8399851992 0.1366710324 3000
-0.7854452199 -0.6111001711 -0.09814472747 3000
0.06086199581 -0.4567009685 0.8875359389 3000
0.3983250059 0.6676990099 0.6289000094 3000
-0.6806035944 0.689644589 -0.2473238526 3000
```
So we now fail tests because derived data differ in the 3 sig. fig... 

I guess the right thing to do here is to update the test data to match - once we're all satisfied this is behaving as expected...

Lestropie:
> I thought we'd decided not to...? Users can easily inspect their externally-supplied schemes manually anyway.

The principal reason for reinserting the `-raw_dwgrad` option was that if someone's DICOM gradient table is malformed, there needs to be a capability to interrogate that without any *MRtrix3* interpretation. As long as `-property dw_scheme` provides the same functionality (and is documented as such), then the option isn't needed.

> I think we'd need to change what you've got so it matches what used to happen before (or least what should have happened): with -raw_dwgrad, anything DW-related is produced using the unmodified DW encoding, including -shell_bvalues, -shell_sizes, -shell_indices, and all the gradient export options.

The description of the option would need to change, but yes, that would be an appropriate corresponding change.

> Tests are currently failing. On inspection, this is because we now only scale b-values if non unit norm vectors are detected (with a generous tolerance of 1e-3), so the b-values are now no longer scaled.

Yea I saw that. Currently the tolerance threshold for rescaling is based on the premise that we don't want accumulation of precision errors from repeatedly applying it; given that's introduced an issue, maybe it's worth investigating whether or not this is actually an issue, as if it's not we could disable the threshold, and test behaviour would then not change.

Lestropie:
So I wrote some Matlab code to investigate the whole multi-normalisation issue. Generates random directions, then iterates between vector2string and string2vector conversions until a duplicate entry is detected, and collects stats on number of iterations and *b*-value changes; it also repeats for different vector2string precisions.

Regardless of the vector2string precision, if string2vector -> vector2string conversion doesn't immediately give the same result, it oscillates between two different strings. So the concept of having the vectors "creep" over iterations doesn't happen. The largest change in *b*-value is about 3 x 10^(-p), where p is the vector2string precision.

So what I would suggest is:

1. Always normalise vectors.

2. Round *b*-values to the nearest integer.

Point 2 I've thought of previously, as having different volumes with *b*-values of e.g. 3000.000000001 and 2994.9999999999 looks kinda dumb.

jdtournier:
> The principal reason for reinserting the -raw_dwgrad option was that if someone's DICOM gradient table is malformed, there needs to be a capability to interrogate that without any MRtrix3 interpretation. 

Yes, makes sense. It's possible to do this using `dcminfo`, but it is admittedly quite tricky....

> As long as -property dw_scheme provides the same functionality (and is documented as such), then the option isn't needed.

I'm pretty sure the DICOM import code will write the `dw_scheme` entries as found (although it'll flip the x & y coefficients), and defer any modifications for the `DWI::get_DW_scheme()` call. So as far as I can tell, that should work.

> Currently the tolerance threshold for rescaling is based on the premise that we don't want accumulation of precision errors from repeatedly applying it; given that's introduced an issue, maybe it's worth investigating whether or not this is actually an issue, as if it's not we could disable the threshold, and test behaviour would then not change.

So I'm not sure I'm following the argument here, but I assume this refers to the string->vector->string tests you mention in the subsequent post. It's reassuring to see that there is no creep in the b-values upon repeated re-normalisation - one less thing to worry about. But I don't think it avoids the need to re-generate the test data. Looking at your suggestions about how to handle this:

> 1. Always normalise vectors.

Already the case in the current code. The question is purely about whether to also scale the b-values: only when we detect there's significant deviations of vector norms from unity, or always (ignoring the option to prevent b-value scaling). Maybe you're talking about b-values scaling?

> 2. Round b-values to the nearest integer.

Not keen on that one... It'll probably be fine for almost all current purposes, but who's to say someone won't try this on some weird data where much lower b-values are used (maybe even IVIM?), where rounding might introduce non-negligible errors? And users who provide their own gradient table might actually want the full precision in their table to be retained.

More to the point, in the current PR, the original b-values will be left as provided, unless we detect deviations from unit vector norms. This will basically only happen for multi-shell data acquired using the fixed b, modulated grad amplitude approach. In pretty much all other cases, the b-values will remain as originally provided. In most cases, the b-values are already discrete at the input, so there won't be a need to do anything. There may be some rationale for the shell clustering to round the shell's b-value though, that does often report odd values...

The problem we have with the testing is actually because the current PR leaves the b-values unmodified. Your proposal here would also leave them unmodified (for our test data, as far as I can tell), or if it does modify them, they will still end up slightly different from what the current `dev` branch provides since it applies the b-values scaling without rounding. So I don't think this avoids the need to update the test data...

Lestropie:
I wasn't happy overall with the logic here, it felt like trying to do a minor modification whilst preserving existing code rather than solving the problem. I've removed d1af036 and instead done 10c3bff, which I think makes the logic flow much more clear. The thresholds are also set based on Octave number crunching at the relevant string precisions.

The change also makes the question of *b*-value scaling much more clear, and caused me to flip my preference (a8e22d6). Basically, if the gradient vector directions are of unit norm up to the precision expected if they are provided to 6 significant figures, then just do the double-precision normalisation, don't touch the *b*-values, and don't modify the header.

Lestropie:
Worth noting one specific consequence of this.

`mrconvert` now requires that a DW scheme *remain valid* throughout any conversions it performs in order to propagate to the output header. 

This particular CI test now results in removal of `dw_scheme`, whereas currently on `dev` it is preserved:
```
mrconvert dwi.mif -coord 3 1:2:end -axes 0:2,-1,3 - | testing_diff_image - mrconvert/dwi_select_axes.mif
```

The output image has sizes `6 x 8 x 9 x 1 x 34`, whereas `dw_scheme` has 34 rows; since this doesn't match the number of volumes, it is removed.

jdtournier:
OK, just documenting what was discussed recently:

- there's a question mark over when it's appropriate to trigger b-value rescaling. After the changes in 10c3bff, b-value scaling will be triggered if vector norms differ from unity by anything that exceeds precision issues. However, there's no guarantee that incoming regular single-shell data actually store their DW gradient vectors with that level of precision. Case in point below. We definitely want to unit-normalise the vectors, so that happens regardless, but b-value rescaling should only be applied when we're confident that the intention of the user was for their data to be interpreted as such. 

  Most of us felt that if the acquisition was indeed intended to be used as a multi-shell scheme, but acquired at fixed nominal b-value with actual b-value modulation achieved by manipulation of gradient vector amplitudes, there would be variations in the b-values, and hence deviations of the gradient norms, by more than a few percent. Conversely, if the vector norms differ from unity by  less than a percent, this is very unlikely to correspond to a multi-shell scheme. So the agreement was for the b-value rescaling to be applied only when vector norms deviate from unity by something like 1% - clearly well above machine precision, but far below what could reasonable be interpreted as a multi-shell scheme. 

- This also calls for the `-bvalue_scaling` option to be reverted to what it was, taking an on/off argument, so that b-value rescaling can be disabled when it's not desired, but also forced on in those rare cases where the user might in fact have meant for his scheme to be interpreted as multi-shell, despite the deviations in vector norms being very small. 

- We discussed whether the `dw_scheme` entries in the header should always be updated when `DWI::get_DW_scheme()` is called, or only when a substantial change was made. In this context, what constitutes a 'substantial change' is understood to mean that the vector norms deviated sufficiently to trigger a b-value rescale, irrespective of whether the b-value rescaling was actually applied (as it may have been disabled by the user, but the vector norms still need updating nonetheless). I'm personally unsure about what's best here: my feeling is the header should reflect the DW scheme as we'd interpreted it, so that downstream applications get the benefit of our modifications. The flipside of this is that otherwise relatively clean gradient tables may end up with much longer precision string representations for what are really very minor changes in vector coefficients (e.g. adding `000001` or `999999` to the ends of the vector coefficients), prompting users to question what's happened to their DW schemes. 

  I'm still more inclined to write the updated DW scheme in the header regardless, but this is obviously up for discussion. One compromise might be to only update the header if the deviations in the vector norms differ from unity by more than 6 digits of precision (i.e. the threshold currently used to trigger b-value rescaling), rather than the 1% we're planning on using for triggering the b-value rescaling, so that at least we update the vector norms if they're not well normalised...

OK, I think that about covers it... 

---

Just to provide an example where being too stringent doesn't work, here's the (truncated) output of `mrinfo -dwgrad` from b=3000 data acquired on a Siemens Avanto (circa 2005...):

**Using the code as it currently stands:**
```
                   0                    0                    0                    0
                   0                    0                    0                    0
  -0.941302734488416  -0.0407001182340152    0.335100973469742           2999.98257
   0.884643910930935   -0.383375692834769     0.26538317391327           3000.38043
   0.171409259778327   -0.777842020160924   -0.604632663138718           2999.67588
  -0.403958636832073   -0.280140663967972    0.870826405272806           2999.12913
  -0.159295206082908   -0.789076253107489   -0.593282145442495           3000.18057
  -0.195983616454468   -0.307574287864257   -0.931122161440819            3000.5016
  0.0814998329255138    0.298899387256884   -0.950798050865994            3000.0123
  -0.238293722234581    0.760379968053609   -0.604184082979998           3000.15807
  -0.634557923859057    0.766049204961273    0.102493203900967           3000.39786
  -0.992390743396373   -0.122211174889687   0.0150013717131368           2999.45139
                   0                    0                    0                    0
  -0.262841388949602   -0.563588746853504    0.783123316390938           2999.05527
  -0.151083282804211    0.933196742827068   -0.326063921392807           3000.66393
...
```
You'll also note that rounding the b-value to nearest isn't guaranteed to 'fix' the problem either: some of the these will round to 2999, others to 3001...

**If we disable b-value rescaling:**
```
                   0                    0                    0                    0
                   0                    0                    0                    0
  -0.941302734488416  -0.0407001182340152    0.335100973469742                 3000
   0.884643910930935   -0.383375692834769     0.26538317391327                 3000
   0.171409259778327   -0.777842020160924   -0.604632663138718                 3000
  -0.403958636832073   -0.280140663967972    0.870826405272806                 3000
  -0.159295206082908   -0.789076253107489   -0.593282145442495                 3000
  -0.195983616454468   -0.307574287864257   -0.931122161440819                 3000
  0.0814998329255138    0.298899387256884   -0.950798050865994                 3000
  -0.238293722234581    0.760379968053609   -0.604184082979998                 3000
  -0.634557923859057    0.766049204961273    0.102493203900967                 3000
  -0.992390743396373   -0.122211174889687   0.0150013717131368                 3000
                   0                    0                    0                    0
  -0.262841388949602   -0.563588746853504    0.783123316390938                 3000
  -0.151083282804211    0.933196742827068   -0.326063921392807                 3000
...
```
Obviously, these are pretty old data, and newer data may not behave like this. But the point is: it happens...

jdtournier:
OK, I think it's good to go... Any tests we'd need to run to make sure it all behaves as expected...?

PR #2017, opened 2020-04-09
write_mrtrix.m: fix handling of .mih outputs

### LINKED ISSUE #2016: write_mrtrix.m: Not working with .mih target
On both `dev` and `master`, using Octave.

Error on [this line](https://github.com/MRtrix3/mrtrix3/blob/3.0_RC3/matlab/write_mrtrix.m#L143);
```
fseek (fid, dataoffset, -1);
error: fseek: invalid stream number = -1
```
Works fine with `.mif` target.

jdtournier:
Closes #2016 
Given the fix, I've no idea how this ever worked for `.mif` outputs...

### DIFF
--- a/matlab/read_mrtrix.m
+++ b/matlab/read_mrtrix.m
@@ -100,7 +100,7 @@ image.data = reshape (image.data, image.dim(order));
 image.data = ipermute (image.data, order);
 for i=1:size(order,2)
   if layout{i}(1) == '-'
-    image.data = flipdim(image.data, i);
+    image.data = flip(image.data, i);
   end
 end
 
--- a/matlab/write_mrtrix.m
+++ b/matlab/write_mrtrix.m
@@ -139,7 +139,7 @@ end
 
 fclose(fid);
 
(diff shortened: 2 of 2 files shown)


Lestropie:
> `- fid = fopen (datafile, 'r+', byteorder);`
> `+ fid = fopen (datafile, 'w+', byteorder);`

Breaks `.mif` write output. "`w+`" truncates an existing file to zero length.

Lestropie:
I still get slight differences with use of `write_mrtrix.m`, due specifically to truncation of floating-point values in header transformation matrix. Suspect this should have been updated when `default_type = double` was created. But this should maybe technically not be changed on `master`.

jdtournier:
> Breaks .mif write output. "w+" truncates an existing file to zero length.

🤦‍♂️ Thanks for catching that...

> Should consider incresing precision of header floating-point values on dev.

Sounds reasonable. I'll merge this, we can update `dev` as you suggest before release - as long as we don't forget...

PR #2032, opened 2020-04-22
GitHub Actions: don't trigger on pull request edited

jdtournier:
This is to avoid cluster of actions getting triggered every time a pull request is commented on or mentioned in a separate issue, etc... :angry:


jdtournier:
Just making sure commenting on this won't trigger anything...

PR #2021, opened 2020-04-15
MRtrix3 3.0.0 release

Lestropie:
@MRtrix3/mrtrix3-devs: Amend as necessary.

-----

#### To be merged prior to this PR:

- [x] #2020 (last-minute `fixelcfestats` bug report)
- [x] #2015 (DW scheme handling)
- [x] #1985 (documentation updates for release)
- [x] #1540 (NIfTI reverting to sform on transform mismatch, affine header transformation fixes, no Analyze output, NIfTI code refactoring)

#### To be completed prior to this PR:

- [x] At least one approving review (preferably *all* team members)
- [x] Disable CircleCI and AppVeyor tests on `master`
- [x] Update changelog with description of installation instructions update (@jdtournier)
- [x] Write announcement post for community forum

#### To accompany PR merge:

- [ ] Signed tag (@jdtournier or @Lestropie)
- [ ] Create release on GitHub (@jdtournier)
- [x] Merge MRtrix3/MRtrix3.github.io#9 (@jdtournier)

#### Following merge:

- [ ] Verify that binary installers have all been produced correctly by GitHub Actions (@jdtournier)
- [ ] Upload anaconda packages to [MRtrix3 repo](https://anaconda.org/MRtrix3/dashboard) (@jdtournier)
- [ ] Verify updates to website
- [ ] Verify that list of contributors at bottom of website main page is properly updated
- [ ] Delete old branches with no commits beyond new `master` branch
- [ ] Place redirects on readthedocs to handle stale links on forum
- [ ] Delete old GitHub webhooks

#### Once everything checks out:

- [ ] Re-post changelog in unclassified section of community forum (@Lestropie)
- [x] update URL of changelog on announcement
- [ ] Post release announcement on community forum (@jdtournier)
- [ ] Make sure that community forum announcement appears on website as most recent announcement post (@jdtournier)
- [ ] Unlock community forum Wiki section (@Lestropie)
- [ ] Posts on social media

PR #2035, opened 2020-04-22
Update MRTRIX_BASE_VERSION to "3.0.0"

Lestropie:
Necessary for builds triggered for GitHub release to function.

I think that in order to have the CI tests pass at every step:

- Leave `core/version.h` not updated for #2021 for CI tests to pass;

- Create the tag based on the outcome of merging #2021;

- Make a new PR that just makes the update to `core/version.h`;

- Generate the release after that PR is merged.

### DIFF
--- a/core/version.h
+++ b/core/version.h
@@ -9,7 +9,7 @@
 //   git tag -s 3.3
 //   git push --follow-tags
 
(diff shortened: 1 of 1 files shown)


Lestropie:
Rejected; instead pursuing #2036 then #2037.

PR #2037, opened 2020-04-22
Tag update 3.0.0

Lestropie:
This PR is intended to contain only the change to the version number stored in file `core/version.h`, for which the new tag has been simultaneously pushed using `git push --follow-tags`.


Lestropie:
Rejected as merging via GitHub UI will result in an explicit merge commit rather than a fast-forward, precluding the preservation of a tag's commit target through the process.

PR #2044, opened 2020-04-29
Android fixes

bjeurissen:
These were the changes I had to make to enable compilation on android.


jdtournier:
No problem with these, but I wouldn't merge these to `master` - these are going to be included in the next release. I'll rebase these to `dev` instead. 

We'll need to settle on a more formal structure for the git workflow in due course. I'll have a think about this over the next few days...

Issue #2047, opened 2020-05-06
Drop python2 support?


bjeurissen:
Python2 is no longer supported by the Python developers since January 2020: https://www.python.org/doc/sunset-python-2/.

And major Linux distributions like Debian and Ubuntu as well as macOS have now also deprecated python2. This means they ship with python3 instead of 2 and have changed all shebangs in python scripts to #!/usr/bin/python3.

The general idea that seems to be adhered to by both Linux and macos seems to be: Installing 'python' will always install the deprecated python2 and calling 'python' will always run deprecated python2. This also means that any package still using #!/usr/bin/python will have python2 as a (deprecated) dependency. Ultimately, the idea is that you always call 'python3' or install 'python3-foo' and that running 'python' and installing 'python-foo' should become a thing of the past.

E.g.: https://wiki.debian.org/Python/2Removal: 

> port upstream package to python3 (or file upstream bug): make sure package doesn't call python at build- or compile time (as there won't be a python package and no python command in bullseye, only python3.)

and https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3#.2Fusr.2Fbin.2Fpython: 

> /usr/bin/python is a symbolic link to /usr/bin/python2 and belongs to the python2 package. This also means that packages that require /usr/bin/python (usually due to shebangs) de facto require Python 2.
...
Since Fedora 26, there is no python package, only python3 and python2. python is virtually provided from python2.

So the "python-is-python3" package that I mentioned I installed on ubuntu 20.04, is only an ugly workaround that will break software that still relies on the deprecated python2 and the right way forward seems to be to update all our python scripts to use #!/usr/bin/python3 instead... This would also mean that we effectively drop support for python2.

Any thoughts on this?

Lestropie:
[Relevant PEP](https://www.python.org/dev/peps/pep-0394)

It looks like Arch is the outlier here, having [had `python` point to Python3 since 2010](https://www.archlinux.org/news/python-is-now-python-3/).

My summary of the PEP would be to keep "`#!/usr/bin/env python`" for anything intended to run in a virtual environment only, and to shift to "`#!/usr/bin/env python3`" everywhere else.

So I think that I would be in favour of changing; though preferably as part of a minor rather than macro update.

jdtournier:
More relevant musings on the matter from the GNOME devs:
https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/

[...]

PR #2049, opened 2020-05-07
Updates to installation documentation

jdtournier:
As discussed recently - [preview here](https://mrtrix.readthedocs.io/en/doctest/installation/build_from_source.html). I think this addresses everything that we'd talked about, other than the latter half of this specific point:

> - May want to include explicit statement about Python3 being preferable but that depending on the distro there is a chance of Python2 being inadvertently pulled in

I'm not sure we need to go into this amount of detail, especially in light of #2047...

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -160,7 +160,8 @@ jobs:
     - name: install dependencies
       run: |
          sudo apt-get update
-         sudo apt-get install pylint pylint3 python3-sphinx sphinx-rtd-theme-common python3-recommonmark python3-sphinx-rtd-theme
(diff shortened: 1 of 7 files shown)


Lestropie:
Made a few tweaks; I went easy on the Python side of things as I think that's better addressed as a standalone PR along with addressing #2047.

jdtournier:
OK, happy with that, thanks @maxpietsch. Do you want to approve the PR so we can merge it?

maxpietsch:
All good as far as I can tell. I only changed Apple Store to App Store and added a link that on macOS allows opening Xcode in the App Store.

PR #2059, opened 2020-05-20
add 404 page to docs

maxpietsch:
implements a 404 page, replacing the default Maze page

- uses the [sphinx-notfound-page](https://sphinx-notfound-page.readthedocs.io/) plugin
- had to install it explicitly via pip in github workflow, updated the README and added a requirements.txt file in docs/
- added the dependency to a new readthedocs yml config file

With this plugin, loading a nonexistent URL still emits a 404 signal. 

One issue is that on the 404.html page, if one selects a different version of the docs from the menu, one is redirected to a 404.html page relative to that version (which does not exist for older versions) instead of reusing and following the actual link. 
For instance: https://mrtrix.readthedocs.io/en/docs_404/does_not_exist --> https://mrtrix.readthedocs.io/en/latest/404.html instead of 
https://mrtrix.readthedocs.io/en/latest/does_not_exist

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -160,7 +160,8 @@ jobs:
     - name: install dependencies
       run: |
          sudo apt-get update
-         sudo apt-get install pylint pylint3 python3-sphinx sphinx-rtd-theme-common python3-recommonmark python3-sphinx-rtd-theme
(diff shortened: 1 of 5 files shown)


maxpietsch:
cherry-picked d46989 aad948b and 60acb3b into `doctest`

PR #2063, opened 2020-05-25
tck2fixel: detect corruption in index file

jdtournier:
Just came across a case where the user's index file was totally borked. Not sure how that happened, but it clearly contained garbage, while the `nfixels` header entry was preserved and correct (as in, it matched the dimensions of the directions file). This means the `Fixel::get_number_of_fixels()` returns the header value straightaway without actually checking the contents of the index map. Took quite a while to figure out what the problem was... 

I'm not sure my proposed fix is the best way to handle this, it may be better to use a more generic way to detect this across fixel apps more broadly - perhaps in `Fixel::get_number_of_fixels()`? Basically, always check the `nfixel` value against the contents of the file regardless. I have a feeling something like this might already have been discussed, but I can't remember where...

We might also want to have a more informative error message here. So this remains a draft for now.

### DIFF
--- a/cmd/tck2fixel.cpp
+++ b/cmd/tck2fixel.cpp
@@ -147,8 +147,12 @@ void run ()
     Transform image_transform (index_image);
     for (auto i = Loop ("loading template fixel directions and positions", index_image, 0, 3)(index_image); i; ++i) {
       const Eigen::Vector3 vox ((default_type)index_image.index(0), (default_type)index_image.index(1), (default_type)index_image.index(2));
+      index_image.index(3) = 0;
+      index_type num = index_image.value();
       index_image.index(3) = 1;
       index_type offset = index_image.value();
+      if (offset + num > num_fixels)
+        throw Exception ("fixel index \"" + index_image.name() + "\" makes reference to out of bounds fixel!");
       index_type fixel_index = 0;
       for (auto f = Fixel::Loop (index_image) (directions_data); f; ++f, ++fixel_index) {
         directions[offset + fixel_index] = directions_data.row(1);


Lestropie:
> it may be better to use a more generic way to detect this across fixel apps more broadly - perhaps in `Fixel::get_number_of_fixels()`? Basically, always check the `nfixels` value against the contents of the file regardless.

Yep, most definitely not specific to `tck2fixel`.

I'm pretty sure that key was put there purely to avoid the necessity of looping through the whole image to determine the number of fixels; so by checking it against the image data it basically loses its *raison d'etre*. I suppose you'd still keep the key-value there and report a mismatch.

The other option is altering the code for accessing fixel data such that a bounds check is performed during the read operation. But this would require a more sizeable change to the code, and would become more expensive as soon as you access data for each voxel more than once. (I had this all figured out on my huge generalized fixel handling branch back in 0.3.15, but since the format change I've abstained from it because I've no idea how to resolve them)

If it is indeed only a speed thing, we should probably just swallow the up-front whole-index-image scan. Indeed you could probably turn it into more of a "validate fixel index image" function where you also check that every fixel is referenced by exactly one voxel.

jdtournier:
> I'm pretty sure that key was put there purely to avoid the necessity of looping through the whole image to determine the number of fixels; so by checking it against the image data it basically loses its raison d'etre. 

This is not lost on me... 

But in the rare cases where the file is messed up for some reason, not detecting this makes it very hard to track down. Given this experience, it would be good to provide some way to debug this. 

One option that might be worth considering, which I think would offer the best of both worlds, would be to force a full check (even in `nfixels` is set) when operating at debug level (i.e. `App::log_level` >= 3). That would have allowed us to figure out the problem straight away - the first thing the user did was run with the `-debug` flag and send me the full dump...

Lestropie:
> force a full check (even in nfixels is set) when operating at debug level (i.e. App::log_level >= 3).

Although it would have found the issue faster in this case, it would be against the precedent of the rest of the code base where `-debug` is purely about terminal verbosity, not processing. Doing a sweep in the absence of `NDEBUG` would make more sense from a software design perspective: it'd still rely on the user following the advanced debugging instructions, but it'd at least produce a better error message upon doing so.

A third option would be a standalone `fixelvalidate` command, as I did with `tsfvalidate`. There's probably other data types for which a validation command could also be defined:

- `5ttcheck` -> `5ttvalidate`
- `dir`
- `label`
- `mesh`
- `peaks`
- `tck` (testing the `count` field / looking for zero-vertex & one-vertex tracks as `tckstats` already does)

With such in place it might then make sense (& be documented) for `-debug` to trigger data validation functions.

jdtournier:
I see where you're coming from, but I think we can afford a bit of flexibility here. Just on some of your points:

>  it would be against the precedent of the rest of the code base where -debug is purely about terminal verbosity

Well, there's a few places where `App::log_level` is used to trigger other types of behaviour (display of intermediate images, as far as I can tell). But more to the point, I don't think there's anything fundamentally wrong or unexpected with performing additional checks when running with `-debug`, especially if these checks are likely to be expensive. I don't think there's any implicit or explicit contract here that `-debug` only influences terminal verbosity in a very strict sense. It might be more problematic if the option was actually called `-verbose` or similar. So it's probably more unexpected for us developers (since that's not been how we do things to date) than it might be for users... In fact, I would argue that when running under `-debug`, users might actually expect the software to perform a few more checks as appropriate to help with debugging.

> Doing a sweep in the absence of NDEBUG would make more sense from a software design perspective: it'd still rely on the user following the advanced debugging instructions, but it'd at least produce a better error message upon doing so.

Sure, but that requires a fair bit more work from the user, and moving forward, is likely to be beyond  most users once they start relying on the precompiled packages more heavily. More to the point, I don't see why that would be preferable to doing the check when running under `-debug` when we can do so quite easily.

> A third option would be a standalone `fixelvalidate` command, as I did with `tsfvalidate`. 

Yes, this would be a good addition, I think. But I still don't think it means we can't use the `-debug` flag as a signal to trigger a few more checks than we might be performing otherwise. 

> With such in place it might then make sense (& be documented) for -debug to trigger data validation functions.

I don't get why the presence of validate commands would therefore mean it's OK to perform the checks when running with `-debug`, if it wasn't OK to do this on the basis that `-debug` is only supposed to affect terminal verbosity. They strike me as broadly independent things from a software design point of view. What's the rationale here?

Lestropie:
> I don't think there's anything fundamentally wrong or unexpected with performing additional checks when running with `-debug`, especially if these checks are likely to be expensive. I don't think there's any implicit or explicit contract here that -debug only influences terminal verbosity in a very strict sense.

Nothing wrong with the option name; if its use were to trigger additional checks / processing, I would argue that the description message for the command-line option should be updated. 

> Sure, but that requires a fair bit more work from the user, and moving forward, is likely to be beyond most users once they start relying on the precompiled packages more heavily.

Yeah I tend to agree; I maybe didn't qualify well enough that I wasn't fully advocating for that, was just establishing the various options.

> But I still don't think it means we can't use the `-debug` flag as a signal to trigger a few more checks than we might be performing otherwise.

Agreed;

> I don't get why the presence of validate commands would therefore mean it's OK to perform the checks when running with -debug, if it wasn't OK to do this on the basis that -debug is only supposed to affect terminal verbosity. They strike me as broadly independent things from a software design point of view. What's the rationale here?

What it might do is improve the understanding of what's going on. There's ways that various data can be validated, here's an interface to do so explicitly, and if you use the `-debug` option these checks will be included in whatever command you are running.

Alternatively, if we invest the effort in writing functions to check the validity of various forms of data, adding new commands that just invoke these functions would require very little effort, and it would be one more pathway by which users may discover issues without relying on us.

Lestropie:
Superseded by #3300.

PR #2067, opened 2020-05-30
dwifslpreproc: Don't discard modified dw_scheme

Lestropie:
As reported on [forum](https://community.mrtrix.org/t/invalid-dwi-scheme-after-dwifslpreproc/3715).

Bug is the consequence of #1389.

Unfortunately it seems nobody used explicit volume recombination on `dev` for the last two years, I didn't spot the issue, and the script test suite has to call `testing_diff_header` without the `-keyval` flag because `eddy` is stochastic...


jdtournier:
flagellator

Lestropie:
Only 4 hours per charge?

rookienumbers

PR #2068, opened 2020-05-31
mrview Tractography tool: Multiple fixes

Lestropie:
Primarily a fix for [this forum post](https://community.mrtrix.org/t/visualizing-endpoints-colored-according-to-different-rois/3738), but found a few other little things along the way:

- Remove redundant control of scalar-based thresholding UI controls.
- Fix off-by-one error between streamline vertices and scalar / colour by endpoints data.
- Fix erroneous assert() call when loading track scalar file for the purpose of thresholding.
- Fix scaling rate of lower / upper threshold by scalar file AdjustButtons.


jdtournier:
These all look sensible, but I'm a bit perplexed by the off-by-one thing. Can you explain where that one comes from?

Lestropie:
[Confirmed fixed](https://community.mrtrix.org/t/visualizing-endpoints-colored-according-to-different-rois/3738/5) by forum user.

PR #2070, opened 2020-06-04
mtnormalise: warn if only one contrast provided

maxpietsch:
I'd add a warning to prevent accidental use of only one component in `mtnormalise`. This issue came up when debugging a user's analysis.

```
$ mtnormalise wm.mif wm_norm.mif -mask mask.mif
mtnormalise: [WARNING] Only one contrast provided. If multi-tissue CSD was performed, provide all components to mtnormalise.
mtnormalise: [100%] performing log-domain intensity normalisation
```

### DIFF
--- a/cmd/mtnormalise.cpp
+++ b/cmd/mtnormalise.cpp
@@ -487,6 +487,8 @@ void run ()
 {
   if (argument.size() % 2)
     throw Exception ("The number of arguments must be even, provided as pairs of each input and its corresponding output file.");
+  if (argument.size() == 2)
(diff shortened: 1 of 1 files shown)


jdtournier:
Merged, although I think there is a legitimate use case for pure bias field correction. But in most cases, I agree this is overwhelming likely to be due to user error. And for pure bias field correction, the requirement would be that the input image would be expected to have little to no contrast, which is unusual. I'd expect other bias field correction approaches would be more appropriate in such a case. Nevertheless, it's only a warning, so it doesn't prevent use in that case. All good.

PR #2078, opened 2020-06-12
mrview: Fix compilation warnings on Qt 5.15

Lestropie:
Attempted fix for CI testing failures on MacOSX, seemingly due to use of Qt 5.15.

### DIFF
--- a/src/gui/dialog/dicom.cpp
+++ b/src/gui/dialog/dicom.cpp
@@ -103,7 +103,7 @@ namespace MR
             }
 
             Qt::ItemFlags flags (const QModelIndex& index) const {
-              if (!index.isValid()) return Qt::ItemFlags();
+              if (!index.isValid()) return {};
               return (Qt::ItemIsEnabled | Qt::ItemIsSelectable);
             }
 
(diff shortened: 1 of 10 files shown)


Lestropie:
This now passes CI, and will need to be merged prior to any other CIs, but need to make sure that the `mrview` interface still works as expected using both mouse and touchpad on both Qt 5.15 and earlier. I don't have a mouse on me and don't have advanced touchpad features set up.

Lestropie:
Note: Seems to be a partial duplicate of 8322af1c, which wasn't PR'd. Solutions there may be fractionally cleaner, as well as more extensive.

chunhungyeh:
Tested on my macOS laptop with Qt 5.15.0, 5.14.1 and 5.14.2, I did not observe any malfunction when using a mouse or trackpad.

jdtournier:
Turns out I'd made very similar changes recently for the same reason, and pushed to the `qt_fix_obsolete_warnings` branch without creating a pull request... Sorry!

I've taken a slightly different strategy, and there's also a few other fixes in there, so I've merged my changes on top of yours. It should all work out... 🤞

jdtournier:
OK, just noted that you'd already seen my non-PR commit. What about my changes did you think was not sufficiently clean?

Lestropie:
> What about my changes did you think was not sufficiently clean?

Other way around:

> 8322af1 ... Solutions there may be fractionally cleaner

jdtournier:
>Other way around:

Doh! Sorry...

jdtournier:
Closed with #2080

PR #2080, opened 2020-06-16
Fix compiler warnings

Lestropie:
Alternative to #2078.

- Qt 5.15 has marked quite a few bits of their code as obsolete, which
  generated a lot of warnings.

- warning about inaccuracies converting large ints to float.


jdtournier:
From actions log:

> error: no member named 'position' in 'QWheelEvent'

Turns out the member function `pos()` in marked obsolete, and we're strongly encouraged to use `position()` instead. But `position()` was only introduced in Qt 5.14... 

I'll see how I can fix this.

jdtournier:
Ok, seems to work. I'll merge...

PR #2081, opened 2020-06-16
Fix line endings in src/gui/mrview/sync/enums.h

Lestropie:
This seems to have been changed in 439dca78, but is causing all sorts of issues when branch hopping; the file keeps getting flagged as changed, and doesn't resolve through a checkout or even a stash.


jdtournier:
Just came across this issue. It's driving me nuts... The file still shows with changes after a `git reset --hard`, and consequently I can't even merge the fix because my local 'changes' would be overwritten by the merge. Rate of hair loss rapidly increasing...

PR #2086, opened 2020-06-17
mrthreshold: Fix std::cout usage

Lestropie:
If no output image path is specified to `mrthreshold`, and therefore it will be writing the determined threshold values to the standard output stream, then progress bars should not be displayed on the terminal, as these can overwrite the values written to standard output.


Lestropie:
Closed; intended change moved to #2096.

PR #2089, opened 2020-06-17
mrview: add viridis colour maps and fix PET colour map

dchristiaens:
This PR adds two of my favourite colour maps, [viridis and inferno](https://bids.github.io/colormap/) to `mrview` and `mrcolour`. As opposed to _jet_ and others, these are perceptually uniform and also much more friendly to colourblind people.

Viridis:
Screenshot 2020-06-17 at 18 27 34

Inferno:
Screenshot 2020-06-17 at 18 27 42

The implementation uses a 6th order polynomial approximation published [here](https://www.shadertoy.com/view/WlfXRN) (public domain; CC0 license). Using this approximation seems a better fit with the MRtrix3 codebase than defining the [lookup tables](https://github.com/BIDS/colormap/blob/master/colormaps.py) of the original colour maps. I have compared them in python and they seem to be a fairly good match (shown here for inferno).
Figure_1

**Question:** Personally, I strongly prefer _inferno_ and _viridis_ over the _hot_ and _cool_ colour maps currently provided. To avoid overinflating the number of options in `mrview`, I would advocate to remove the _hot_ and _cool_ schemes from the menu. We could still offer them through `mrcolour` if people really want them. How do people feel about deprecating _hot_ and _cool_?

### DIFF
--- a/src/colourmap.cpp
+++ b/src/colourmap.cpp
@@ -98,12 +98,12 @@ namespace MR
               }),
 
           Entry ("PET",
-              "color.r = 2.0*amplitude - 0.5;\n"
+              "color.r = clamp (2.0*amplitude - 0.5, 0.0, 1.0);\n"
               "color.g = clamp (2.0 * (0.25 - abs (amplitude - 0.25)), 0.0, 1.0) + clamp (2.0*amplitude - 1.0, 0.0, 1.0);\n"
               "color.b = 1.0 - (clamp (1.0 - 2.0 * amplitude, 0.0, 1.0) + clamp (1.0 - 4.0 * abs (amplitude - 0.75), 0.0, 1.0));\n",
               [] (float amplitude) { return Eigen::Array3f (clamp (2.0f * amplitude - 0.5f),
-                                                            clamp (0.25f - abs (amplitude - 0.25f)) + clamp (2.0f * (amplitude - 0.5)),
-                                                            clamp (1.0f - 2.0f * amplitude) + clamp (1.0 - 4.0 * abs (amplitude - 0.75))); }),
+                                                            clamp (2.0f * (0.25f - abs (amplitude - 0.25f))) + clamp (2.0f * amplitude - 1.0f),
+                                                            1.0f - (clamp (1.0f - 2.0f * amplitude) + clamp (1.0f - 4.0f * abs (amplitude - 0.75f)))); }),
 
           Entry ("Colour",
               "color.rgb = amplitude * colourmap_colour;\n",
--- a/src/colourmap.cpp
+++ b/src/colourmap.cpp
@@ -50,6 +50,53 @@ namespace MR
                                                             clamp (1.5f - 4.0f * abs (1.0f - amplitude - 0.5f)),
                                                             clamp (1.5f - 4.0f * abs (1.0f - amplitude - 0.75f))); }),
 
+          // The Inferno and Viridis colour maps are implemented using a 6th order polynomial approximation of the originals,
+          // ported from https://www.shadertoy.com/view/WlfXRN (public domain; CC0 license).
+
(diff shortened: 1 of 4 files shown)


jdtournier:
Nice! 

Regarding removal of the hot & cool colourmap: I personally rarely if ever use anything but gray-scale for the main image, but colourmaps are useful for overlays. And these two have a clear use case for things like fMRI blobs, p-values, etc. So I don't think it would be wise to remove them...

dchristiaens:
> these two have a clear use case for things like fMRI blobs

fair point - somehow I always forget about fMRI 😉

jdtournier:
Looks like you'll need to update the tests for `mrcolour`...

dchristiaens:
It appears so. I'll do that later.

On macOS the build fails, though. Looks like that's related to #2078.

dchristiaens:
I don't know why the `mrcolour` tests failed for PET and RGB colour maps yet. Could this somehow be related to the order in which colour maps are declared? I can't see how the order could matter, but it is suspicious that only those maps that are declared _after_ the two new schemes fail testing.

However, I did come across a bug in the declaration of the PET scheme. Prior to a320f9a, GLSL and C++ implementations did not match, meaning that `mrcolour X.mif pet - | mrview -` looked completely different than setting the colour map to PET in `mrview X.mif`. As far as I can tell, there were different bugs in both implementations; I have fixed them to be at least _the same_ but I am not 100% sure that it is _right_. Can @jdtournier or others double-check?

Tests will obviously fail now. Once we agree that the PET map is correct, we will need to update the `mrcolour` testing data for PET and also add tests for Viridis and Inferno colour schemes.

src/colourmap.cpp
```diff
@@ -98,12 +98,12 @@ namespace MR
...
           Entry ("PET",
-              "color.r = 2.0*amplitude - 0.5;\n"
+              "color.r = clamp (2.0*amplitude - 0.5, 0.0, 1.0);\n"
               "color.g = clamp (2.0 * (0.25 - abs (amplitude - 0.25)), 0.0, 1.0) + clamp (2.0*amplitude - 1.0, 0.0, 1.0);\n"
               "color.b = 1.0 - (clamp (1.0 - 2.0 * amplitude, 0.0, 1.0) + clamp (1.0 - 4.0 * abs (amplitude - 0.75), 0.0, 1.0));\n",
               [] (float amplitude) { return Eigen::Array3f (clamp (2.0f * amplitude - 0.5f),
-                                                            clamp (0.25f - abs (amplitude - 0.25f)) + clamp (2.0f * (amplitude - 0.5)),
-                                                            clamp (1.0f - 2.0f * amplitude) + clamp (1.0 - 4.0 * abs (amplitude - 0.75))); }),
+                                                            clamp (2.0f * (0.25f - abs (amplitude - 0.25f))) + clamp (2.0f * amplitude - 1.0f),
+                                                            1.0f - (clamp (1.0f - 2.0f * amplitude) + clamp (1.0f - 4.0f * abs (amplitude - 0.75f)))); }),
```
Lestropie, line 106: If I did indeed make a meal of this, I'd like to cherry-pick and include in `3.0.1`.

dchristiaens: There is definitely an issue in 3.0.0. I was in two minds about creating a separate issue for this one; maybe I should have.. Happy for it to be pushed to 3.0.1.

jdtournier:
Happy for it to go into 3.0.1. As long as it's been tested thoroughly...

jdtournier:
> By all means, try to break it! 😄

As soon as I've caught up with the backlog from the BrainHack...

See you in a couple of months. 👍

dchristiaens:
The tests succeed! Can I get a review?

PR #2102, opened 2020-06-26
fix transformcalc header

maxpietsch:
As reported on the [forum](https://community.mrtrix.org/t/2-questions-about-mrview-and-mrregister/3824), `transformcalc header` fails for large rotations in `mrview`'s transformation tool due to the axis interpretation. 

The proposed solution is to set `Header::do_realign_transform = false;`. I think this is sensible given the purpose of the tool but I am not sure if this brings any negative side effects for other use cases. Any opinions?

### DIFF
--- a/cmd/transformcalc.cpp
+++ b/cmd/transformcalc.cpp
@@ -67,7 +67,8 @@ void usage ()
              "")
 
   + Example ("Calculate the transformation matrix from an original image and an image with modified header",
-             "transformcalc mov mapmovhdr header output",
+             "transformcalc original.mif modified.mif header output",
+             "This transformation can be used for instance: mrtransform original_too.mif -linear output modified_too.mif"
              "")
 
   + Example ("Calculate the average affine matrix of a set of input matrices",
@@ -202,6 +203,7 @@ void run ()
     case 3: { // header
       if (num_inputs != 2)
         throw Exception ("header requires 2 inputs");
+      Header::do_realign_transform = false;
       auto orig_header = Header::open (argument[0]);
       auto modified_header = Header::open (argument[1]);
 
--- a/docs/reference/commands/transformcalc.rst
+++ b/docs/reference/commands/transformcalc.rst
@@ -36,7 +36,9 @@ Example usages
 
 -   *Calculate the transformation matrix from an original image and an image with modified header*::
 
-        $ transformcalc mov mapmovhdr header output
+        $ transformcalc original.mif modified.mif header output
+
+    This transformation can be used for instance: mrtransform original_too.mif -linear output modified_too.mif
 
 -   *Calculate the average affine matrix of a set of input matrices*::
 


jdtournier:
Yes, I can see this is going to be an issue. I'm not sure what the best way to fix this is going to - I imagine setting `Header::do_realign_transform = false` is likely to cause trouble in other use cases - for example with NIfTI images that often have very large rotations in their headers. It's likely to alter behaviour that would have previously been considered 'normal'. We'll have to think about this one...

Lestropie:
- `Transform(orig_header).voxel2scanner.inverse()` seems an unusual call given the presence of `Transform::scanner2voxel`.

- The fact that the variable named "`forward_transform`" has to be inverted when writing to file comes off as suspicious. I'm not accustomed to dealing with transforms myself - I tend to just tinker until I get the right answer and then stop - but this nevertheless sets off alarms in my head.

- The transformations that were performed during header load are now stored [here](https://github.com/MRtrix3/mrtrix3/blob/3.0.0/core/header.h#L380-L383). You could theoretically use this info to construct the effective corresponding affine transform matrix if necessary. Don't have a sense of whether or not this actually solves anything that `RealignTransform false` doesn't; just making sure you're aware.

Lestropie:
Suggested alteration: In #2526 I propose that the `Header` class should have a method that reconstructs, from the transformation and the set of axis flips & permutations that were applied at image load, what the image transformation for the image header as stored on the filesystem was. I think this is what is wanted here, but by doing it this way it would not be necessary to manipulate that static class member, which would otherwise be assumed to be fixed for the duration of command execution.

Lestropie:
Pinging @ppruc.

This needs to be revisited following #3011. What I expect is necessary is that any calculations that involve header transforms of input images need to be made with respect to the transform of the image as it is stored on disk, not as it is interpreted by MRtrix, which has *possibly* experienced a realignment through shuffling of transform axes & strides. With the changes in #3011 there is *explicit* tracking of the transform of the image on disk, as well as what such shuffling was applied at the point of image load; the former should I think be what *any* header-transform-dependent calculations are based on. It's possible that the scope of erroneous transform calculations extends beyond just `transformcalc header`. Having some sample tests with which the prblem can be demonstrated & fix verified would be useful.

Lestropie:
- [ ] Ensure that current code using `Transform(orig_header).voxel2scanner.inverse()` is replaced with `Transform(orig_header).scanner2voxel` if the relevant code remains
    (as the code currently in this PR may need to be replaced entirely to fully resolve the issue)

cmd/transformcalc.cpp
```diff
@@ -202,11 +215,23 @@ void run ()
     case 3: { // header
       if (num_inputs != 2)
         throw Exception ("header requires 2 inputs");
-      auto orig_header = Header::open (argument[0]);
-      auto modified_header = Header::open (argument[1]);
-
-      transform_type forward_transform = Transform(modified_header).voxel2scanner * Transform(orig_header).voxel2scanner.inverse();
-      save_transform (forward_transform.inverse(), output_path);
```
Lestropie, line 209: Not sure why this composed the transform from modified header to original header and then inverted it, as opposed to just computing the transform from original heder to modified header...

PR #2109, opened 2020-06-30
Docs: Move fixel format to dedicated page in FBA

### LINKED ISSUE #2107: Docs: Move fixel format?
Currently the [information on the fixel directory format](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#fixel-image-directory-format) is buried within the "[Images and other data](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html)" page. 

I'm thinking it might be better for this to instead have its own dedicated page, probably within the "Fixel-Based Analysis" group, with the "Images and other data" page maybe just having a hyperlink to it.

Lestropie:
Closes #2107.

@maxpietsch: Two things: I had to change the installation instructions for locally building the docs; also there will be a lot of forum links to the old location of this documentation, can we easily set up something so that attempts to deference the link to the old location redirects to the new location?


maxpietsch:
Yes, `sphinx-notfound-page` is correct (also tested on readthedocs). 

Full URL redirects seem to not work but I think I found the solution for relative page redirects [here](https://github.com/readthedocs/readthedocs.org/issues/1826#issuecomment-247995569). However, I think this won't work for the Fixel format as redirects require 404 status and the page `https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#fixel-image-directory-format` still exists and a wrong anchor does not cause a 404. The best we can do besides manually cleansing the forum is to keep the anchor in `image_data` and add a cross-reference to the new location so that users can find the relevant information.

Issue #2099, opened 2020-06-24
Allow image dimensions greater than 2^31


jdtournier:
As [reported on the forum](https://community.mrtrix.org/t/fixelfilter-error-opening-image-and-loading-path-as-fixel-fixel-connectivity-data/3821?u=jdtournier)

PR #2113, opened 2020-06-30
Support images with axis size greater than 2^31

Lestropie:
Work toward addressing #2099. 

Compiles, but I'd like to make an attempt at producing a fixel matrix that exceeds 4 billion entries and see that `fixelfilter` behaves as intended.

### DIFF
--- a/cmd/amp2response.cpp
+++ b/cmd/amp2response.cpp
@@ -244,20 +244,18 @@ void run ()
     }
   }
 
-  vector<int> lmax;
-  int max_lmax = 0;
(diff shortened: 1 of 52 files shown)


Lestropie:
Confirmed `fixelfilter` producing correct result on a 0.9mm template with 1.5M fixels and 5.5B fixel-fixel connections.

Lestropie:
@jdtournier: Make sure you're happy with 8216779. I don't see it being a problem unless the adapter were to wrap something that only implemented `move_index()` and not the non-const `index()`.

PR #2118, opened 2020-07-01
run_tests: allow different location for MRtrix executables

jdtournier:
This is to allow binary packages to be tested. To use, make sure you
have a clone of the MRtrix3 repo checked out, and the binary package you
want to test is installed. Then set the MRTRIX_BINDIR environment
variable to point to the location of the executables to be tested, and
invoke run_tests. For example:
```
MRTRIX_BINDIR=~/miniconda3/bin/ ./run_tests binaries
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,8 @@
 [submodule "testing/data"]
 	path = testing/binaries/data
 	url = https://github.com/MRtrix3/test_data
+	ignore = dirty
 [submodule "testing/script_data"]
 	path = testing/scripts/data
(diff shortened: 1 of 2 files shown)


jdtournier:
Found out how to avoid the version being marked dirty just because there's untracked files in the testing data submodules :tada: 

See [this blog post](http://www.nils-haldenwang.de/frameworks-and-tools/git/how-to-ignore-changes-in-git-submodules), changes in 4756a858.

Lestropie (approved):
Looks pretty simple and safe to me.

Issue #1392, opened 2018-07-11
`run.command()`: Support for command-line piping


Lestropie:
When the Python script library was first provided, piping in/out of image data actually worked just as it does for binary executables; this was due to the use of `shell=True` in the `run.command()` function. This capability was revoked as this function became more advanced; this however now prevents the use of piped images as the input / output of Python scripts.

To restore support for this will require some gymnastics:

1. Command-line arguments will need to have a flag indicating if they are an input or an output image, much like the `.type_*()` C++ functions.

2. The `app` module will need to replace the value of any '`-`' input image argument with the contents of `stdin`. The `run.command()` function could either just display this, or replace any appearances of temporary image paths with '`-`'.
This I think would be easier than trying to detect the '`-`' within `run.command()` (where information about whether it's an input or output would have been lost) and connect the processes's `stdin` to that of the main process.

3. The `run.command()` function will need to be informed if the command call includes an output to a pipe, and store the contents of `stdout` from that call somewhere for the `app` module to write to `stdout` of the main process.
This can't strictly be done based on the "last" `run.command()` call, since a script may contain multiple outputs and an output piped image may not necessarily appear as part of the very last one.

Lestropie:
Rethought this a little since it was re-raised:

Could have keyword arguments to the `run.command()` function indicating that the command string contains at least one user image input / output (would need to be one or the other). Then the `run` module could itself manage the gymnastics between "`-`" characters temporary image paths, terminal outputs, and `stdin` / `stdout` contents in an entirely enclosed way.

Alternatively, this could potentially be managed within the `fsys.fromUser()` function (which would also need to be informed as to whether the path is an input or output). This would technically be more "general" since there's no guarantee that a temporary image piped to/from a script will necessarily be an input to / output from `run.command()`. Such handling though probably exceeds what one would intuitively expect from the "filesystem" module.

Too many possible solutions without a clear winner :-/

Lestropie:
In retrospect, after looking at the documentation more closely, it might be possible to support this without going to the extent of #1608.

`argparse` permits the input to the `type` kwarg to be any callable that accepts a string. Therefore custom callables could be provided for input and output images.

-   For input images, this would mean that a "`-`" input could be detected by the relevant function and be filled with content provided at `stdin`.

[...]

PR #2125, opened 2020-07-07
Pipe handler: ensure deletion of piped images even if not loaded

jdtournier:
This is in response to an [issue raised on the forum](https://community.mrtrix.org/t/unix-tmp-becomes-full/3879?u=jdtournier), where the user reports `/tmp` filling up within a day of use. As I was responding, I noticed that `mrinfo` wasn't deleting images piped into it. I'm pretty sure this wasn't ever the intention... Yet this has been the behaviour for a very long time, as far as I can tell. :facepalm: 

This applies to any command that opens the header only, without actually loading the image itself (e.g. I expect the `-template` option used in various commands would also be affected). This should fix the issue everywhere it occurs... Note that the same images are (presumably correctly) deleted in the signal handler, but this is only invoked when a signal is raised - not in normal operation.

I think this is a fairly critical bug to fix, but I'd also want to make sure that the fix isn't going to break expectations elsewhere. Can anyone think of any reason why `mrinfo` and similar commands _shouldn't_ delete images piped into them....? 

Also, am I correct in assuming that some of our scripts may be using pipes into `mrinfo` as part of normal operation? If so, that would definitely explain the `/tmp` filesystem filling up fast...

### DIFF
--- a/bin/dwifslpreproc
+++ b/bin/dwifslpreproc
@@ -982,7 +982,7 @@ def execute(): #pylint: disable=unused-variable
       if os.path.isfile('dwi_post_eddy.eddy_residuals'):
         eddyqc_options += ' -g ' + bvecs_path
       if do_topup:
-        eddyqc_options += ' -f ' + fsl.find_image('field_fieldcoef')
+        eddyqc_options += ' -f ' + fsl.find_image('field_map')
       if eddy_mporder:
         eddyqc_options += ' -s slspec.txt'
       if app.VERBOSITY > 2:
--- a/cmd/mrinfo.cpp
+++ b/cmd/mrinfo.cpp
@@ -24,6 +24,7 @@
 #include "file/json.h"
 #include "file/json_utils.h"
 #include "dwi/gradient.h"
+#include "image_io/pipe.h"
 
 
 using namespace MR;
@@ -106,7 +107,10 @@ void usage ()
     +   Option ("shell_indices", "list the image volumes attributed to each b-value shell")
 
     + PhaseEncoding::ExportOptions
-    +   Option ("petable", "print the phase encoding table");
+    +   Option ("petable", "print the phase encoding table")
+
+    + OptionGroup ("Handling of piped images")
+    +   Option ("nodelete", "don't delete images passed to mrinfo via Unix pipes");
 
 }
 
@@ -242,6 +246,9 @@ void run ()
         return true;
     return false;
   };
+
+  if (get_options("nodelete").size())
+    ImageIO::Pipe::delete_piped_images = false;
 
   const bool export_grad = check_option_group (GradExportOptions);
   const bool export_pe   = check_option_group (PhaseEncoding::ExportOptions);
--- a/cmd/mrinfo.cpp
+++ b/cmd/mrinfo.cpp
@@ -110,7 +110,7 @@ void usage ()
     +   Option ("petable", "print the phase encoding table")
 
     + OptionGroup ("Handling of piped images")
-    +   Option ("nodelete", "don't delete images passed to mrinfo via Unix pipes");
+    +   Option ("nodelete", "don't delete temporary images or images passed to mrinfo via Unix pipes");
 
 }
 
--- a/core/formats/pipe.cpp
+++ b/core/formats/pipe.cpp
@@ -43,7 +43,8 @@ namespace MR
       if (H.name().empty())
         throw Exception ("no filename supplied to standard input (broken pipe?)");
 
-      SignalHandler::mark_file_for_deletion (H.name());
+      if (ImageIO::Pipe::delete_piped_images)
+        SignalHandler::mark_file_for_deletion (H.name());
 
       if (!Path::has_suffix (H.name(), ".mif"))
         throw Exception ("MRtrix only supports the .mif format for command-line piping");
--- a/core/formats/pipe.cpp
+++ b/core/formats/pipe.cpp
@@ -14,6 +14,8 @@
  * For more details, see http://www.mrtrix.org/.
  */
 
+#include <unistd.h>
+
 #include "signal_handler.h"
 #include "file/utils.h"
 #include "file/path.h"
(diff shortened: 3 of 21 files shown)


maxpietsch:
> Can anyone think of any reason why mrinfo and similar commands shouldn't delete images piped into them

When a temporary image is fed into a script, the script might want to query the input header information before doing anything with it. Adding an option in `mrinfo` to not delete temporary files could be useful.

On a related note: Would it be possible to delay the deletion of temporary images to once the command finishes? Currently, the below fails:
```
a=$(mrconvert 0.mif -)
b=$(mrcat $a $a -)
```
while 
```
a=$(mrconvert 0.mif -)
b=$(mrcalc $a $a -add -)
```
works.

Lestropie:
> When a temporary image is fed into a script ...

Can't be done; no piping support.

jdtournier:
> When a temporary image is fed into a script, the script might want to query the input header information before doing anything with it. 

Yes, that's precisely the kind of usage I'm worried the current 'fix' might conflict with. So the question is: does anything we do rely on this currently?

> Adding an option in mrinfo to not delete temporary files could be useful.

That's an option, although it'll be messy - would need hooks deep into the backend. But realistically, I think usage like that above should be handled with a proper named output - not a dangling pipe. So I'm not enormously keen on offering the option.

> Would it be possible to delay the deletion of temporary images to when the command finishes? 

Yes, but that's up to the individual commands to manage the lifetime of the relevant objects. The example you show is a case in point: the fix here is to open all headers (or images) up front, and not allow them to go out of scope until processing is finished.

An alternative would be to register the files with an additional `atexit()` handler. This is actually what I was going to try first, by hooking into the signal handling backend (I've actually just pushed a branch that was originally supposed to do that, but ended up different). The issue there is that this would need to be a different handler than the signal handler, since we also register output piped images to be deleted on error - we can't assume all these files should be deleted on normal exit. But it can certainly be done - and might admittedly be a cleaner solution for the reasons you raised here.

jdtournier:
> Can't be done; no piping support.

Ok, good to hear. One less thing to worry about...

jdtournier:
>> Can't be done; no piping support.
>
> Ok, good to hear. One less thing to worry about...

Though this doesn't rule out internal commands _within_ a script using pipes fed into `mrinfo`...?

Lestropie:
> Though this doesn't rule out internal commands within a script using pipes fed into mrinfo...?

Technically no; but given there's both an `image.Header` class for capturing header information and an `image.mrinfo()` function, I doubt such is done anywhere.

maxpietsch:
> Can't be done; no piping support.

I did't mean piping images into scripts but using temporary images stored in bash variables as a workaround that automatically deletes temporary files upon completion.

The current implementation of `image.Header` class calls `mrinfo` which would delete these: https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/image.py#L34. If `mrinfo` had a flag to not delete files, this could be used there. 

If we decided that we do not support temporary files as input to scripts that'd be fine but I think that's an unnecessary limitation.

Not related to this PR but one more issue with scripts is the usage of `mrconvert_keyval` on output (for instance in `dwishellmath`). Supporting temporary input files would require parsing this information at the beginning of the script.
```
dwishellmath $a mean mean.mif
dwishellmath: Generated scratch directory: /Users/mp/data/dwishellmath-tmp-A7VUCH/
Command:  mrconvert /var/folders/0_/x_fhlh1n76bd5hnk6rcmyzg40000gn/T/mrtrix-tmp-Qt9Lyw.mif /Users/mp/data/dwishellmath-tmp-A7VUCH/in.mif -strides 0,0,0,1
dwishellmath: Changing to scratch directory (/Users/mp/data/dwishellmath-tmp-A7VUCH/)
Command:  dwiextract -shells 0 in.mif - | mrmath -axis 3 - mean shell-00.mif
Command:  dwiextract -shells 400 in.mif - | mrmath -axis 3 - mean shell-01.mif
Command:  dwiextract -shells 1000 in.mif - | mrmath -axis 3 - mean shell-02.mif
Command:  dwiextract -shells 2600 in.mif - | mrmath -axis 3 - mean shell-03.mif
Command:  mrcat -axis 3 shell-00.mif shell-01.mif shell-02.mif shell-03.mif out.mif
Command:  mrconvert out.mif /Users/mp/data/mean.mif

dwishellmath: [ERROR] mrconvert out.mif /Users/mp/data/mean.mif (dwishellmath:60)
dwishellmath: [ERROR] Information from failed command:
dwishellmath:
              mrconvert: [ERROR] Unable to obtain header key-value entries from spec "/var/folders/0_/x_fhlh1n76bd5hnk6rcmyzg40000gn/T/mrtrix-tmp-Qt9Lyw.mif"
dwishellmath:
dwishellmath: [ERROR] For debugging, inspect contents of scratch directory: /Users/mp/data/dwishellmath-tmp-A7VUCH/
dwishellmath: Scratch directory retained; location: /Users/mp/data/dwishellmath-tmp-A7VUCH/
```

jdtournier:
Based on this morning group meeting, this is the list of changes we discussed:

- [X] Need to mark piped images for deletion and cleanup at command exit
- [X] Also prevent output to TTY on platforms that support it (Linux / macOS)
- [X] Add option to not delete piped images? Might be useful in scripts and various other situations. 
- [ ] Add functionality to discard outputs - e.g. add special NULL filename?
- [ ] Update to docs to clarify what happens with dangling pipes

jdtournier:
OK, I reckon this one is ready to go. Slight concern with the NULL image feature, hard to argue that's a bug fix...

jdtournier:
OK, I've moved the `NULL` image feature over to its own pull request (#2131) to `dev` for inclusion in the next minor release - it's really not a bug fix...

Everything else here is good to go, I reckon.

Lestropie:
The topic of not deleting piped images when specifically `mrinfo` is invoked ([ed2f35e](https://github.com/MRtrix3/mrtrix3/pull/2125/commits/ed2f35e2a056055195f98695dd83bd4fe1a1df71) in particular) will be relevant for #1392.

PR #2128, opened 2020-07-08
Fix matlab write mrtrix

### LINKED ISSUE #2126: write_mtrix.m writing out incorrect values
Hello,
Thanks for fixing the problem with write_mtrix.m (see #2092) . This fixes the error in Matlab, but the output file has incorrect values (at least when writing out integers). 
Below is an example  of a file written with either the current version of write_mrtrix.m (incorrect) or with an older version (95b144d) that writes the correct values to the file. 95b144d is the last commit that writes out correct values. All versions I have tried after that either fail with a Matlab error or write incorrect values to the file.

This is the correct output. Written values are integers between 8100 and 8233 in this example.
image

This is the incorrect output: note the small value were the crosshair is located. Written values range between 10^-42 and 10^-39.
image

_Originally posted by @julienbesle in https://github.com/MRtrix3/mrtrix3/issues/2092#issuecomment-654769366_

jdtournier:
This should fix #2126. See commit message in 71696f3e for details.

### DIFF
--- a/matlab/read_mrtrix.m
+++ b/matlab/read_mrtrix.m
@@ -40,24 +40,17 @@ while 1
   if isempty(d)
     disp (['invalid line in header: ''' L ''' - ignored']);
   else
-    key = lower(strtrim(L(1:d(1)-1)));
+    key = strtrim(L(1:d(1)-1));
     value = strtrim(L(d(1)+1:end));
-    if strcmp(key, 'dim')
-      image.dim = str2num(char(split_strings (value, ',')))';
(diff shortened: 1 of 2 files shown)


julienbesle:
This fixed it for me. Thanks!

About the data type, this is a volume I converted from a freesurfer-generated parcellation of the thalamus. I checked this original volume and it's encoded as 32-bit floats, even though it only contains integers. I suppose I could change the type to uint16 in my Matlab script to save some disk space, since I'm always expecting integers anyway...

PR #2136, opened 2020-07-22
mrtransform: Tweak -modulate option help

Lestropie:
For options that receive multiple command-line inputs, the names of those arguments are space-separated; it is therefore preferable for the description of an option that receives only one command-line input to not have space separators in its name.


jdtournier:
Agree this needs to be fixed, but I don't think `-method` is sufficiently clear as to its purpose in the context of `mrtransform`. How about `-modulate` instead?

Lestropie:
> I don't think `-method` is sufficiently clear as to its purpose in the context of `mrtransform`. How about `-modulate` instead?

:man_facepalming: :rofl: 

> **-modulate method** Valid choices are: fod and jac.

jdtournier:
Ok, moral of the story: don't work while on holiday...

Issue #2134, opened 2020-07-15
CI: Conditional testing of scripts


Lestropie:
Scripts can currently be tested via `./run_tests scripts`, but due to long run times and external dependencies, they are not integrated into CI.

What should be possible is that for a PR that forms a pending tag update, a specific label would be added to the PR, which could then trigger a specific GitHub Action, which would install the external dependencies and run the full gamut of Python script tests.

Lestropie:
Idea from @dchristiaens: Instead of installing and utilising external packages, provide dummy commands with names corresponding to those of external softwares that compare input data to that expected and place pre-computed output data in expected locations.

PR #2143, opened 2020-08-21
CI: Test scripts w. no external dependencies

Lestropie:
Adds "`scripts_nodep`" input to `run_tests` script, which executes only tests for those Python scripts absent from the blacklist identifying those scripts with external software dependencies.

Related to #2134, but seeking to integrate more regular testing of as many scripts as possible; testing of those scripts *with* external dependencies could potentially be done only for merges to `master`. Will find out from the checks on this PR how much additional runtime is required.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -37,6 +37,9 @@ jobs:
 
     - name: binary tests
       run: ./run_tests binaries || { cat testing_binaries.log; false; }
+      
+    - name: script tests
+      run: ./run_tests scripts_nodep || { cat testing_scripts_nodep.log; false; }
 
 
 
@@ -98,6 +101,9 @@ jobs:
 
     - name: binary tests
       run: ./run_tests binaries || { cat testing_binaries.log; false; }
+      
+    - name: script tests
+      run: ./run_tests scripts_nodep || { cat testing_scripts_nodep; false; }
 
     - name: check command documentation
       run: ./docs/generate_user_docs.sh && git diff --exit-code docs/
@@ -143,6 +149,10 @@ jobs:
     - name: binary tests
       shell: cmd
       run: msys64\msys2_shell.cmd -c "./run_tests binaries || { cat testing_binaries.log; false; }"
+      
+    - name: script tests
+      shell: cmd
+      run: msys64\msys2_shell.cmd -c "./run_tests scripts_nodep || { cat testing_scripts_nodep.log; false; }"
 
 
 
--- a/run_tests
+++ b/run_tests
@@ -15,14 +15,19 @@
 #
 # For more details, see http://www.mrtrix.org/.
 
+
+# Scripts to be blacklisted from testing due to external software dependencies
+SCRIPTS_EXTDEP="5ttgen dwibiascorrect dwifslpreproc labelsgmfix"
+
+
 if test "$#" -ne 1; then
-  echo "Script must be provided with single input argument: \"binaries\", \"scripts\", \"units\", or the name of a specific test to be run."
+  echo "Script must be provided with single input argument: \"binaries\", \"scripts\", \"scripts_nodep\", \"units\", or the name of a specific test to be run."
   exit 1
 fi
 
 
 if [ -d testing/data/ ]; then
-  echo -n "Deleting old testing data directory testing/data/"
+  echo -n "Deleting old testing data directory testing/data/ ... "
   rm -rf testing/data/
   echo OK
 fi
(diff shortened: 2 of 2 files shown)


Lestropie:
A potential alternative solution would be to define a separate text file that contains only those script tests intended to be run as part of CI. So for instance only one variant of each `dwi2response` algorithm would be invoked, only one `population_template` test would run, etc.. It may not catch all potential interactions between C++ code modifications and Python scripts, and it would require additional maintenance for the contents of that test list file, but it would enable a reasonably quick test of the scripts to be integrated into CI, and it would allow for those tests to be run on all three OS's within CI.

bjeurissen:
It definitely makes sense to add scripts to CI. I feel only adding it for Mac could offer a good balance between test coverage and runtime.

What is the reason for the large discrepancies between the platforms? Is that because they have different amounts of resources allocated?

bjeurissen:
> What is the reason for the large discrepancies between the platforms? Is that because they have different amounts of resources allocated?

I had a look and the resources for each platform are the same. I believe those massive Windows Python runtimes are indicative of using "msys/python3", which uses POSIX emulation, rather than "mingw64/mingw-w64-x86_64-python3", which is native Windows python. I think we should be using the latter?

maxpietsch:
`population_template` test runtime can be reduced by a factor 2 by deactivating the multi-contrast tests.

Instead of additional tests files, how about deactivating non-essential standalone script tests stochastically for CI testing?  This speeds up testing but we'd still eventually catch test failures.

Something along the lines of `([ ! -z "$CI" ] || (($RANDOM % 2))) && <test>`? Could also include some git-fu to check if a file of interest was changed which would trigger to run all tests: 
```run_all=$(git --no-pager diff --name-only $(git branch --show-current) $(git merge-base $(git branch --show-current) master) | grep 'mrtransform\|mrregister\|registration\|population_template')```

**Comment by jdtournier** [comment]
>> What is the reason for the large discrepancies between the platforms? Is that because they have different amounts of resources allocated?
>
> I had a look and the resources for each platform are the same. I believe those massive Windows Python runtimes are indicative of using "msys/python3", which uses POSIX emulation, rather than "mingw64/mingw-w64-x86_64-python3", which is native Windows python. I think we should be using the latter?

Just to follow up on that: I'm not sure the python runtime should make all that much difference, the bulk of the work should be performed within the binaries. But more to the point, I had a go at checking the difference in performance on my Windows 10 laptop for some of these scripts between MSYS2 and WSL2 (which I've recently installed). This was done using this branch, and just timing the tests for `dwi2response` (timing the second run to ensure all data are synchronised and testing commands built first):

### using MSYS2:
```
$ time ./run_tests dwi2response
logging to "testing_dwi2response.log"
checking for requisite test data... Synchronized OK
building testing commands... OK
Erasing old script test output directory... OK
running "testing/scripts/tests/dwi2response"... 43 of 43 passed

real    11m44.009s
user    0m2.616s
sys     0m6.493s
```

### using WSL2:
```
$ time ./run_tests dwi2response
logging to "testing_dwi2response.log"
checking for requisite test data... Synchronized OK
building testing commands... OK
Erasing old script test output directory... OK
running "testing/scripts/tests/dwi2response"... 43 of 43 passed

real    51m44.493s
user    13m57.978s
sys     10m55.893s
```

😮 not what I was expecting (are at least hoping for)! I was hoping WSL2 would bring performance improvements, not a 4x drop on the exact same hardware... I might run it again just to make sure, this is really weird.

**Comment by bjeurissen** [comment]
Wow... What happens if you benchmark dwi2fod between the two?

**Comment by bjeurissen** [comment]
https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configure-global-options-with-wslconfig maybe you have to allocate all cores to wsl?

**Comment by jdtournier** [comment]
OK, it looks like my custom WSL2 setup was probably to blame... I'd set it up to use the same home directory as the regular Windows one, using [these instructions](https://jeremyskinner.co.uk/2018/07/27/sharing-home-directory-between-windows-and-wsl/). If I revert to a vanilla WSL2 setup, the numbers are actually far more impressive:
```
$ time ./run_tests dwi2response
logging to "testing_dwi2response.log"
checking for requisite test data... Synchronized OK
building testing commands... OK
Erasing old script test output directory... OK
running "testing/scripts/tests/dwi2response"... 43 of 43 passed

real    5m56.313s
user    10m31.748s
sys     0m29.153s
```
So it looks like WSL2 is a substantial improvement after all - as long as you don't mess with the setup...

Lestropie:
Closing in facvor of #2154.

PR #2150, opened 2020-08-26
dwifslpreproc: -topupfiles option

### LINKED ISSUE #2144: ENH dwifslpreproc: Manually provide topup estimated field
Re-raised on [forum](https://community.mrtrix.org/t/synb0-disco/3954), but has been discussed on multiple occasions.

Principally, providing the output of topup (field coefficients and movement parameters) explicitly at the command line would permit utilisation of [Synb0-DISCO](https://github.com/MASILab/Synb0-DISCO). I think this is probably the most sensible interface, rather than providing a T1w as input to `dwifslpreproc` and trying to execute that tool or its components internally.

Lestropie:
Request has been raised a couple of times, and may well prove useful locally for processing some historical data, so figured I'd just do it.

`-topup_files` option (and indeed if they specify `-eddy_options " ... --topup= ..."`) allows the user to specify a location where `topup` has already been run, and `dwifslpreproc` will import the two necessary files, and otherwise proceed as normal as if `topup` had been run internally, e.g. performing an initial EPI distortion correction for the sake of producing a brain mask for `eddy`. The path specified by the user can be the `topup` output prefix, or one of the two files, and `dwifslpreproc` will find both files automatically.

Primarily to enable interfacing with Synb0-DISCO, as its principal output is the results of `topup` but its interface just about precludes its execution from within `dwifslpreproc`, so running that first and `dwifslpreproc` second makes more sense.

Also added some option groups to the `dwifslpreproc` help page while I was at it, since it's likely to continue getting more complex over time.

Closes #2144.


jdtournier (approved):
Seems fine, though I've not had a chance to test. I guess since this is going to `dev`, it'll get further testing before being released, so happy enough to merge. :+1:

PR #2151, opened 2020-08-29
dwi2response tournier: Use os.path.join()

Lestropie:
Created whilst doing a quick investigation regarding [this forum post](https://community.mrtrix.org/t/dwi2response-error-iter0-fixel-peaks-mif/4010); don't think it is actually a solution, but would nevertheless prefer that the appropriate function be used rather than the "`/`'" character.

### DIFF
--- a/lib/mrtrix3/dwi2response/tournier.py
+++ b/lib/mrtrix3/dwi2response/tournier.py
@@ -84,15 +84,15 @@ def execute(): #pylint: disable=unused-variable
     # Run CSD
     run.command('dwi2fod csd dwi.mif ' + rf_in_path + ' ' + prefix + 'FOD.mif -mask ' + mask_in_path)
     # Get amplitudes of two largest peaks, and direction of largest
-    run.command('fod2fixel ' + prefix + 'FOD.mif ' + prefix + 'fixel -peak peaks.mif -mask ' + mask_in_path + ' -fmls_no_thresholds')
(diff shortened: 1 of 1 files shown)


jdtournier:
Changes seem fine to me. :+1:

Looking at the orignal post though, I'm trying to figure out what might be happening that could explain the error, especially given that it shows up with one data set and not the other. I'm thinking that the issue might be that the file in question (`iter0_fixel/peaks.mif`) genuinely doesn't exist. Looking at possible situations where this might happen, [this line in `fod2fixel`](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/fod2fixel.cpp#L178-L179) definitely introduces the potential for there to be no output at all from the command, with no error or warning. Personally, I reckon this should lead to an error and a non-zero exit status. 

But as to what might cause a zero fixel count, the simplest explanation here is an empty mask in the previous `dwi2fod` call, but that's not it, given that the `mrstats` call directly before that reports a count of 159691 voxels in the mask. So the alternative is that the `dwi2fod` call returns an image of zeros or NaNs for some reason. It's not the initial RF since [that's hard-wired](https://github.com/MRtrix3/mrtrix3/blob/master/lib/mrtrix3/dwi2response/tournier.py#L75), so there's something not right with the data themselves. I'm not sure we're going to figure this out without taking a look at the data directly... Though one option is Philips' tendency to include a synthetically generated isotropic weighted image at the end of their DICOM series with the same (non-zero) b-value, but a zero direction vector - this can tend to mess things up. I'll enquire about that on the forum.

Issue #2142, opened 2020-08-19
Drop Qt4 support


Lestropie:
Superseding #2120.

Believe the current plan is:

- Update `master` documentation only to indicate that Qt5 is required, to be included in `3.0.2` update.

- Enforce Qt5 requirement, potentially dropping relevant precompiler directives and code, on `dev` to be included in `3.1.0` update.

Issue #2120, opened 2020-07-03
CI: Test different Qt versions


Lestropie:
Have raised previously, but it seems to have [bitten us](https://community.mrtrix.org/t/mrtrix-3-0-1-update-errors/3864) after all.

CI should ideally test the most recent Qt4, the most recent Qt5, and any prior Qt5 versions where API changes have been introduced.

Lestropie:
Linking to #2121.

bjeurissen:
One could also consider dropping Qt4 support. I think any distribution that supports gcc5 and up also supports qt5, as qt5 predates gcc5 with 2-3 years.

jdtournier:
I'd also be minded to drop Qt4 support - I doubt it's even supported by the maintainers themselves anyway.
But as we currently still support Qt4, that was definitely a goof on my part...

bjeurissen:
I've had a look at packages.ubuntu.com, and the oldest version that we can reasonably expect people to run is 16.04, which comes with GCC 5.3.1 and Qt 5.5.1. So requiring Qt>=5.5, might be sensible.

bjeurissen:
For Debian, the oldest supported version that comes with GCC>=5 appears to be stretch, which comes with GCC 6.3 and Qt 5.7.1. For fedora, it is 22, which comes with GCC 5.3.1 and Qt 5.5.1. So I have the impression that a distribution with GCC >= 5, but Qt < 5.5 is pretty hard to find.

bjeurissen:
> I'd also be minded to drop Qt4 support - I doubt it's even supported by the maintainers themselves anyway.

Support for Qt4 ended on December 19th 2015: https://www.qt.io/blog/2014/11/27/qt-4-8-x-support-to-be-extended-for-another-year

bjeurissen:
Also had a quick look at centos, as it seems to be popular for scientific computing and at universities. There, gcc>=5 is only included since the very latest release (8), which comes with Qt 5.12.5. The first version that would be able to compile MRtrix using packages from the package manager is version 7 as it comes with clang. There Qt is at 5.9.7.

So I have the impression that we can really rely on the availability of Qt>=5.5 as soon as a C++ compliant compiler comes with the distribution.

PR #2153, opened 2020-09-02
Docs: State Qt5 as GUI requirement

Lestropie:
As described in #2142. 

I have some recollection of a discussion where Qt 5.05 would be a more specific minimum requirement; anyone recall?

### DIFF
--- a/docs/installation/build_from_source.rst
+++ b/docs/installation/build_from_source.rst
@@ -35,7 +35,7 @@ Required dependencies:
    due to `deprecation of Python2 <https://www.python.org/doc/sunset-python-2/>`__);
 -  The `zlib <http://www.zlib.net/>`__ compression library;
 -  `Eigen <http://eigen.tuxfamily.org>`__ version >= 3.2 (>= 3.3 recommended);
--  `Qt <http://www.qt.io/>`__ version >= 4.7 (>= 5.0 recommended) *[GUI components only]*;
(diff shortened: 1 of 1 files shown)


jdtournier:
Yes, that rings a bell. I think it might have been to do with this being the minimum version that works for macOS. @bjeurissen might be able to give the specifics, I must admit I can't remember...

bjeurissen:
I can only remember https://github.com/MRtrix3/mrtrix3/issues/2120 where I argued that  Qt >=5.5 should be available in  all versions of popular linux distributions that are realistically still being used.

That said, there might be reasons to opt for more recent qt versions, e.g. better hdpi support or better dark mode support, ... But as long as it compiles and works properly, I think we can deem it supported..

So I would just look at the oldest linux distribution versions that are still supported by their maintainers and see what Qt version they package to decide a minimal requirement. At least then we are sure it will work out of the box on popular distributions.

jdtournier:
I'm happy enough to mandate Qt>5.5, it will simplify the OpenGL handling substantially (that's what a lot of those directives will relate to). 

We'd need to update the Qt checks in the `configure` script, and I'dd probably add a compile time check in there somewhere, probably in `src/gui/opengl/gl.h` (which is included pretty much everywhere GUI related), of the form:
```
#if QT_VERSION < 0x050500
# error "MRtrix requires Qt version 5.5 or later"
#endif 
```

jdtournier (approved):

PR #2157, opened 2020-09-04
Enforce Qt5

Lestropie:
Second half of #2142: where the first half (#2153) only alters documentation on `master`, this goes to `dev`, and involves removing support for anything prior to 5.5 and performing corresponding checks.


jdtournier (approved):
Seems to work fine on my Arch system. But then it's running Qt 5.15, and it's hard to force Qt4 back onto it, so I can't realistically check how it handles a Qt4 system. 

But then, the main thing is that it still works with Qt >= 5.5, and tries to prevent the use of older versions, so it's all good as far as I'm concerned.

Issue #2165, opened 2020-09-09
dwifslpreproc: Python version for `eddy_quad`


Lestropie:
`dwifslpreproc` currently explicitly invokes the Python3 interpreter to run FSL EddyQC `eddy_quad` command. This was due to my original testing, where EddyQC was installed from source, where I found that essential.

I've now received a report where `eddy_quad` fails within `dwifslpreproc`, and removing the `python3` invocation rectifies the issue.

My guess is that it is related to the integration of EddyQC into FSL6, and is possibly related to "FSLPython".

PR #2166, opened 2020-09-10
dwifslpreproc: For EddyQC, don't invoke Python3 interpreter

Lestropie:
This I think makes sense as a solution.

I don't know for certain, but what I think has happened is that prior to integration in FSL6, `eddy_quad` required Python3, whereas now in FSL6 it gets an FSLPython shebang and so needs to be executed as-is.

This approach should succeed on first attempt for anyone with FSL6 installed, whereas for any systems that happen to have FSL5 combined with an EddyQC installation from source it'll still make a second attempt by explicitly invoking the Python3 compiler.

### DIFF
--- a/bin/dwifslpreproc
+++ b/bin/dwifslpreproc
@@ -970,8 +970,7 @@ def execute(): #pylint: disable=unused-variable
 
   # Run eddy qc tool QUAD if installed and one of -eddyqc_text or -eddyqc_all is specified
   if eddyqc_path:
-    eddyqc_exe = find_executable('eddy_quad')
-    if eddyqc_exe:
+    if find_executable('eddy_quad'):
       eddyqc_options = ' -idx eddy_indices.txt -par eddy_config.txt -m eddy_mask.nii -b bvals'
       if os.path.isfile('dwi_post_eddy.eddy_residuals'):
         eddyqc_options += ' -g ' + bvecs_path
@@ -983,9 +982,10 @@ def execute(): #pylint: disable=unused-variable
         eddyqc_options += ' -v'
       try:
         run.command('eddy_quad dwi_post_eddy' + eddyqc_options)
-      except run.MRtrixCmdError as exception_eddyquad:
+      except run.MRtrixCmdError as exception:
         with open('eddy_quad_failure_output.txt', 'w') as eddy_quad_output_file:
-          eddy_quad_output_file.write(str(exception_eddyquad))
+          eddy_quad_output_file.write(str(exception))
+        app.debug(str(exception))
         app.warn('Error running automated EddyQC tool \'eddy_quad\'; QC data written to "' + eddyqc_path + '" will be files from "eddy" only')
         # Delete the directory if the script only made it partway through
         try:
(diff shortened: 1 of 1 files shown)


Lestropie:
Closes #2165.

jdtournier (dismissed):
Seems sensible. I've not tested, but I assume there's already been quite a bit of testing on your side?

bjeurissen:
This looks good, I am just wondering if it should not simply work without providing an interpreter at all, end of story?

As far as I know, eddyqc is set up to use the python interpreter that comes with FSL, so If that does not work, I would assume this indicates the FSL installation has been messed with.

Is this fix not hiding an issue with the FSL installation, that then later will pop up as soon as the user tries to run eddyqc directly?

Lestropie:
When I first tried EddyQC myself (installed from source), running without an interpreter did *not* work; it would only work if explicitly invoked via Python3. So I expect that if I were to simply remove the Python3 interpreter invocation, it would then work for users with FSL6, but for anyone using FSL5 in conjunction with EddyQC installed from source (as I did in my original testing), `dwifslpreproc -eddyqc_*` would then be broken.

If running `eddy_quad` without an explicit interpret does not work due to some fault with FSL installation, then running with Python3 invoked explicitly will *also* not work, with the following error message:
```
     File "/*****/bin/fsl_6.0.3/bin/eddy_quad", line 53, in <module>
       from eddy_qc.QUAD.quad import main
   ImportError: No module named 'eddy_qc'
```
So I don't see this "hiding" an FSL installation issue. `dwifslpreproc` will report failure of EddyQC, the default terminal output will show both of the command strings that were attempted, and the error messages that were generated in the two attempts are written to separate text files in the scratch directory.

bjeurissen:
Hiding was the wrong word, sorry. I just meant that this will enable running something that is still intrinsically broken as far as I can tell. If a straight "./eddy_quad" does not work, why should it work in the python script?

Lestropie:
> ... this will enable running something that is still intrinsically broken ...

I don't see the logic behind that. Even if it were somehow possible that `eddy_quad` could fail under default usage due to a broken FSL installation yet simply executing explicitly under Python3 were to resolve the issue, then at that specific moment the fact that the FSL installation is somehow "broken" would not be concern given the command in question just completed successfully. If such a broken installation were to expose itself elsewhere then so be it; but I'm not in the business of validating people's non-*MRtrix3* softwares, if I call the command and it produces what it's supposed to I don't care beyond that.

> If a straight "./eddy_quad" does not work, why should it work in the python script (not technically, but just conceptually)?

All I recall right now is that when I first installed EddyQC from source, that didn't work, and neither did explicitly invoking Python2, but explicitly invoking Python3 did. I'll hopefully be able to reconstruct those scenarios tomorrow when my FSL5-installed system comes back online.

Lestropie:
On the system where I first implemented EddyQC support:
```
$ which eddy_quad
/usr/bin/eddy_quad
$ cat $(which eddy_quad)
#!/usr/bin/python
# EASY-INSTALL-SCRIPT: 'eddy-qc==1.0.0','eddy_quad'
__requires__ = 'eddy-qc==1.0.0'
__import__('pkg_resources').run_script('eddy-qc==1.0.0', 'eddy_quad')
```

I'm guessing that `1.0.0` had installation issues, and I've somehow muscled it into a working state. With a clean config, I now can't get `eddy_quad` to work on that particular system with any version of Python: both have issues with dependencies, perhaps because I did a system update. So perhaps my strong arming is indeed better omitted.

jdtournier (approved):

PR #2167, opened 2020-09-10
Ensure -force works for fixel directory outputs

### LINKED ISSUE #2164: *2fixel: -force operation
Reported for `fod2fixel`, but may well apply to other commands as well.

Function:
```
Fixel::check_fixel_directory (const std::string &path, bool create_if_missing = false, bool check_if_empty = false)
```
does not take into account usage of the `-force` command-line option. If the directory exists and is non-empty, and `check_if_empty` is true, the Exception "`Expected fixel directory ... to be empty.`" will be thrown, despite usage of the `-force` option.

Lestropie:
Only putting as draft request for now as I've committed the code untested; plus there's some chance that a more extensive refactor may be preferable over the proposed change.

I found the structure of this function quite strange. I think it may be in part due to the way that fixel commands provide their command-line interface, e.g. specifying both input & output fixel directories explicitly & separately but allowing them to be the same path. In any other context I'd expect separate functions to be defined for checking input vs. output path specifications. Probably if I were to attack #1043, that one would involve altering command-line interfaces and pipeline documentation, and the code altered here would probably be refactored in the process; but maybe for the sake of a hotfix we should go for easiest solution.

Closes #2164.

### DIFF
--- a/cmd/fixel2sh.cpp
+++ b/cmd/fixel2sh.cpp
@@ -60,10 +60,11 @@ void usage ()
 void run ()
 {
   auto in_data_image = Fixel::open_fixel_data_file<float> (argument[0]);
+  const std::string in_directory = Fixel::filename2directory (argument[0]);
 
-  Header in_index_header = Fixel::find_index_header (Fixel::get_fixel_directory (argument[0]));
-  auto in_index_image =in_index_header.get_image<index_type>();
-  auto in_directions_image = Fixel::find_directions_header (Fixel::get_fixel_directory (argument[0])).get_image<float>().with_direct_io();
+  Header in_index_header = Fixel::find_index_header (in_directory);
+  auto in_index_image = in_index_header.get_image<index_type>();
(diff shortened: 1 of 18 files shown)


Lestropie:
Removing this PR from `3.0.2` milestone for now. The philosophy that has been used for fixel directory handling is proving to be difficult to resolve with a desire for the `-force` option to behave as expected... and with the magnitude of changes necessary there are various use cases that could potentially be broken if they are not all tested comprehensively. So it might be preferable to tag `3.0.2` without these changes, and resolve in the next tag.

I'll try to figure out a better description elsewhere; but for now just know that it's doing my head in trying to find the right logic that will work in all use cases.

Lestropie:
This PR will *not* be merged, as it has not been possible to modify the existing fixel command interface to support the `-force` option. What will happen instead is:

- For `3.0.2`: there will be modifications to the error messages produced if attempting to overwrite an existing non-empty directory, including instruction to manually erase any directory rather than attempting to use `-force`.

- For `3.1.0`: will pursue more extensive alterations as discussed in #1043.

PR #2173, opened 2020-09-16
Tweak error message on failed fixel overwrite

Lestropie:
Argument "`bool check_if_empty`" of function "`Fixel::check_fixel_directory()`" was only set to `true` for commands `fod2fixel` and `peaks2fixel`. Everything else is currently set up to be permissive of existing & non-empty output fixel directories. I have decided that `fixelconvert` and `fixelcrop` should also not be permissive of such. Other commands cannot enforce this requirement due to allowing the same directory to be defined for both input & output.

### DIFF
--- a/cmd/fixelconvert.cpp
+++ b/cmd/fixelconvert.cpp
@@ -114,7 +114,7 @@ void convert_old2new ()
   const bool output_size = get_options ("out_size").size();
 
   const std::string output_fixel_directory = argument[1];
-  Fixel::check_fixel_directory (output_fixel_directory, true);
+  Fixel::check_fixel_directory (output_fixel_directory, true, true);
 
   index_type fixel_count = 0;
   for (auto i = Loop (input) (input); i; ++i)
--- a/cmd/fixelcrop.cpp
+++ b/cmd/fixelcrop.cpp
@@ -61,7 +61,7 @@ void run ()
   Fixel::check_fixel_size (in_index_image, mask_image);
 
   const auto out_fixel_directory = argument[2];
-  Fixel::check_fixel_directory (out_fixel_directory, true);
+  Fixel::check_fixel_directory (out_fixel_directory, true, true);
 
   Header out_header = Header (in_index_image);
   index_type total_nfixels = Fixel::get_number_of_fixels (in_index_header);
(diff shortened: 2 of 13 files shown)


jdtournier:
Looks good. I made a minor edit to the error message to give a hint as to why we don't like `-force` on directories, but nothing earth-shattering. 

However, should we do anything about the suggestion to use `-force` when the index and/or directions files don't match, as found on [this line](https://github.com/MRtrix3/mrtrix3/blob/c4164867ec86c7035e10974036213743dfe24fd4/core/fixel/helpers.h#L352) and [this line](https://github.com/MRtrix3/mrtrix3/blob/c4164867ec86c7035e10974036213743dfe24fd4/core/fixel/helpers.h#L373)...?

Lestropie:
I think it's more than just the error message that needs to be tweaked in the `Fixel::copy_index_file()` and `Fixel::copy_directions_file()` functions. Currently, consistency between expected and existing index / directions images is only enforced if `-force` is not used. So if `-force` *is* used, the index / directions files could be overwritten with new content, completely agnostic to the fact that there may be pre-existing fixel data files within that directory that are based on the original index / directions, and would hence become invalid. e86ad11a precludes such.

jdtournier:
I've tried to reproduce the assert failure, but it works fine on my system with both `clang++ 10.0.1` and `g++ 10.2.0`... I'll see if I can install `clang 9.0.0` as used in the Linux build.

jdtournier:
OK, I eventually managed to reproduce by running _all_ tests, not just those for `fixelfilter` - means it had something to do with leftover files from a previous test. 

Assert failure itself was caused by ignoring the result of `headers_match()` in the `images_match_abs()` call (fixed in f34963ee2). The wider problem was that `run_tests` didn't remove the `tmp/` folder, created in an earlier set of tests - though it deletes anything else that _starts_ with `tmp`... (fixed in d50ca31d1). Hopefully it'll run to completion now.

jdtournier:
I've also cleaned up the tests to ensure all temporary output filenames start with `tmp` - this ensures they get deleted by `run_tests` before the next set of tests. I noticed there were a few tests that manually deleted their temporaries, which shouldn't be necessary as long as these temporaries are named appropriately.

I also note that it seems the `scripts` tests also sidestep that feature, for some reason... Why are all the temporaries created in the parent folder? Is that so they can be inspected afterwards?

jdtournier (approved):

jdtournier (approved):

PR #2180, opened 2020-09-28
[BUG] tckgen: Fix absent CST with ACT

Lestropie:
As originally reported directly by @cbajada, then posted on [forum](https://community.mrtrix.org/t/unwanted-results-using-act-tractography-with-hcp-data/2419), but I neglected to list on GitHub and hence a fix was erroneously omitted from `3.0.0`.

I second-guessed myself a bit as to how exactly I wanted to go about this, as there were a few different options, but I won't bother boring with the full details unless there are questions or concerns.

One outstanding issue is that I'm going to have to modify the test data. Can't have the transition from >=50% WM, <50% GM to <50% WM, >=50% GM align perfectly with the voxel transition from present FOD to absent FOD, as this change now means that there's no valid DWI data just inside of the GM.

### DIFF
--- a/core/interp/base.h
+++ b/core/interp/base.h
@@ -203,6 +203,11 @@ namespace MR
                                     pos[2] <= -0.5 || pos[2] >= bounds[2])));
         }
 
+        template <>
+        bool set_out_of_bounds (const bool& value) {
+          return ((out_of_bounds = value));
+        }
+
         template <class VectorType>
         Eigen::Vector3 intravoxel_offset (const VectorType& pos) {
           if (set_out_of_bounds (pos))
--- a/core/interp/base.h
+++ b/core/interp/base.h
@@ -197,22 +197,17 @@ namespace MR
 
         // Some helper functions
         template <class VectorType>
-        bool is_out_of_bounds (const VectorType& pos) const {
-          if (pos[0] <= -0.5 || pos[0] >= bounds[0] ||
-              pos[1] <= -0.5 || pos[1] >= bounds[1] ||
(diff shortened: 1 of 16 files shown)


Lestropie:
> One outstanding issue is that I'm going to have to modify the test data.

Cancel that: looks like earlier failures were due to what I thought were unrelated kinks I hadn't yet ironed out. All passed locally.

-----

Proof of fix (from top to bottom: `fact`, `ifod1`, `ifod2`, `sd_stream`, `tensor_det`, `tensor_prob`):

all_small

Lestropie:
@jdtournier: There are in the end some "non-compulsory" code changes through here, so see if you're happy as-is or whether you'd prefer a stripped version for `master`. 

- I changed a few things around `Interp::Base::out_of_bounds`, as I found it unclear which functions were responsible for altering it or reading from it and which were not. It's still not great, but it's maybe a little better than it was.

- My first solution involved explicitly generating a binary voxel mask image within the `Interp::Masked` class constructor, and then reading from that in the `.voxel()` function. This became super slow because there was a whole lot of unnecessary re-initialisation happening, so there's some changes around the place to ensure that the series of copy-constructions meant that that mask only needed to be derived once. No longer completely necessary given the alternative implementation in 
7a8ba57 doesn't have an expensive constructor, but it's hopefully OK to leave those changes in place.

jdtournier (approved):
:sweat_smile: OK, quite a bit more in there than strictly what's required to fix the bug, but all good nonetheless!

PR #2184, opened 2020-09-30
mrhistmatch: Use normal equation

Lestropie:
Proposing to squeeze into `3.0.2` given that the `dwicat` script fails on the BATMAN tutorial data.


Lestropie:
CI fails as the precision of the derived solution is not as good.

jdtournier (approved):

PR #2185, opened 2020-09-30
NIfTI: Fix verification of rigid transform

Lestropie:
@jdtournier: This removes the warning for me on a dataset where the error was appearing.

I was just printing out variable contents and noted that the 3x3 matrix produced from Q was identical to R until the `Q.w` check resulted in negating the vector, which changed the matrix (some elements unchanged, others changed by maybe 1e-3).

I always lose my understanding of quaternions as soon as I'm finished dealing with whatever code uses them, plus there's the confound of NIfTI peculiarities, so I'm handing this one off to you if it needs to be resolved in a different way.

### DIFF
--- a/core/file/nifti_utils.cpp
+++ b/core/file/nifti_utils.cpp
@@ -513,7 +513,7 @@ namespace MR
           Eigen::Quaterniond Q (R);
 
           if (Q.w() < 0.0)
-            Q.vec() = -Q.vec();
+            Q.coeffs() *= -1.0;
 
           if (R.isApprox (Q.matrix(), 1e-6)) {
             Raw::store<code_type> (NIFTI_XFORM_SCANNER_ANAT, &NH.qform_code, is_BE);


jdtournier:
OK, that's interesting. Negating the quaternion should be equivalent to negating both the angle of rotation, and the axis about which the rotation is taking place, which should lead to the same operation. And looking at [the code that converts the quaternion to its 3x3 rotation matrix](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Geometry/Quaternion.h#L590-624), I have to say I can't see any reason why there would be discrepancy - mathematically the result should be identical. 

Looking at the issue more closely though, I think the issue is that [this line](https://github.com/MRtrix3/mrtrix3/blob/b84d16d40a0109023cce299ee52f1e1069f2b9cc/core/file/nifti_utils.cpp#L516) only negates the _x_,_y_,_z_ components of the quaternion, leaving _w_ unchanged. Presumably me trying to avoid the needless negation of _w_, since the qform only stores the _x_,_y_,_z_ components anyway - but leading to an inconsistent quaternion that fails the rigid-body check that was introduced subsequently (also by me... 😬).

So the proposed fix is fine. but I'm guessing replacing that `Q.vec() = -Q.vec();` line with `Q = -Q;` will also fix the issue, and might also avoid further confusion down the track...?

Lestropie:
> I'm guessing replacing that Q.vec() = -Q.vec(); line with Q = -Q; will also fix the issue

No such operator exists; one could instead replace it with `Q *= Eigen::Quaterniond (-1, 0, 0, 0)`, but would that be less clear rather than more?

jdtournier:
Yes, just noticed that... This works:
```
 Q.coeffs() = -Q.coeffs();
```

jdtournier:
Just [asked users on the forum](https://community.mrtrix.org/t/nifti-contains-non-rigid-information-qform-will-not-be-stored/4107/7?u=jdtournier) to confirm whether that change would fix it. Let's see...

Lestropie:
Changed solution to 4606bfc, which contains less characters than yours :upside_down_face:

jdtournier (approved):
Provisional pending [confirmation from the affected users on the forum](https://community.mrtrix.org/t/nifti-contains-non-rigid-information-qform-will-not-be-stored/4107/7?u=jdtournier).

PR #2187, opened 2020-09-30
labelconvert: Error if -spine used but absent from LUT

Lestropie:
As reported on [forum](https://community.mrtrix.org/t/tracks-passing-through-brainstem-counted-for-connectome/427/4).

Even if this isn't the actual source of confusion for the user, it nevertheless makes sense that if the command-line option is not going to achieve its purpose properly due to not having a target parcel index, the command should not be executing.

I've also promoted the case where "`Spinal_column`" is specified *multiple times* in the output parcellation from a warning to an error.

### DIFF
--- a/cmd/labelconvert.cpp
+++ b/cmd/labelconvert.cpp
@@ -129,32 +129,31 @@ void run ()
   auto opt = get_options ("spine");
   if (opt.size()) {
 
-    if (duplicates) {
-      WARN ("Could not add spine node: \"" + SPINE_NODE_NAME + "\" appears multiple times in output LUT");
+    if (duplicates)
+      throw Exception ("Cannot add spine node: \"" + SPINE_NODE_NAME + "\" appears multiple times in output LUT");
+    if (!spine_index)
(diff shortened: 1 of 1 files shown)


jdtournier (approved):
Seems OK. Is this destined for 3.0.3...?

Lestropie:
> Is this destined for `3.0.3`...?

Yes, going to `master`. Calling it a "bug" is maybe excessive, but it's catching and reporting a clearly erroneous use case.

The GitHub diff is slightly misleading because it doesn't properly catch the change to indentation. Really the change is from:

```
    if (duplicates) {
      WARN ("Could not add spine node: \"" + SPINE_NODE_NAME + "\" appears multiple times in output LUT");
```
to:
```
    if (duplicates)
      throw Exception ("Cannot add spine node: \"" + SPINE_NODE_NAME + "\" appears multiple times in output LUT");
    if (!spine_index)
      throw Exception ("Cannot add spine node: \"" + SPINE_NODE_NAME + "\" not present in output LUT");
```

PR #2189, opened 2020-10-01
Tractography::tangent() function

Lestropie:
This was giving me the irrits while working on other things, so I defined a library function and introduced a typedef to just standardise things a little.


Lestropie:
Considering renaming this function to reflect that this is an *approximate* tangent estimation, based on the adjacent streamline vertices but ignoring inter-vertex distances. A more precise quantification of the tangent would involve taking the derivative of the Hermite interpolator.

PR #2193, opened 2020-10-07
Transparently handle UTF-8 dashes as regular ones

jdtournier:
Prompted by [recent forum discussion](https://community.mrtrix.org/t/trouble-with-mrmath/4158/2?u=jdtournier)... 

Seems to work for just about [any Unicode dash](https://www.compart.com/en/unicode/category/Pd) I could reasonably throw at it. Any combination of dashes will also be handled as if they were each regular dashes. For example:
- `-option`
- `–option`
- `﹣option`
- `—option`
- `﹣-‒option`

are all interpreted equivalently.

One slight question mark is whether to treat a negative number where the negative sign is not a regular hyphen as a number. Initial commit prevents a token like `‒1` from being interpreted as an option (which I think is correct), but this will still fail down the track when the command tries to convert this to a number, e.g.:
```
$ mrcalc anat.nii ‒1 ‒mult out.nii
mrcalc: [ERROR] Could not interpret string "‒1" as either an image path or a numerical value
mrcalc: [ERROR] As image: 
mrcalc: [ERROR] unknown format for image "‒1"
mrcalc: [ERROR] error opening image "‒1"
mrcalc: [ERROR] As numerical value: 
mrcalc: [ERROR] error converting string "‒1" to complex float (no valid conversion)
```
I think it's going to be too hard to support this type of conversion though, and the output is clear enough in this case anyway. 

One other use case that isn't currently handled is the use of these characters for piping. This currently fails:
```
$ mrconvert anat.nii ‒ | mrview -
mrconvert: [ERROR] unknown format for image "‒" (no file extension specified)
mrconvert: [ERROR] error creating image "‒"
```
Should be simple enough to add support for it, and it's probably a good idea to do so before merging...


Lestropie:
1a634d9 is intended to help a little with error determination as per your [response](https://community.mrtrix.org/t/trouble-with-mrmath/4158/2). May not have been sufficient in this instance, but I generally try to add missing information to error messages when I see it.

Lestropie:
Bumping due to [appearing again on forum](https://community.mrtrix.org/t/dwi2response-error-voxels-command-not-recognized/4339).

Lestropie (dismissed):
Suggest changing function name `is_dash()` to something that better reflects the fact that it's not just a binary test that does not modify the data: it also increments the pointer. So maybe something like "`consume_dash`"?

Lestropie:
- Consider changing target branch to `master` and shifting to `3.0.3` milestone

- Add similar handling for piped images

core/app.h
```diff
@@ -289,13 +289,19 @@ namespace MR
             opt (option), args (arguments)
         {
           for (size_t i = 0; i != option->size(); ++i) {
```
jdtournier, line 291: @Lestropie: this bit warrants closer inspection to make sure it still works as originally intended...

Lestropie: Presumably you mean the whole code block below? You can click & drag on the "new comment" + button to select a set of lines & associate a comment thread with it.

Will do some tests. The message could potentially be expanded a bit also.

Lestropie: Basic test performs as expected; this is including 505adef8:

```
mrcat sub-109639_T2w.nii.gz sub-109639_desc-preproc_T2w.nii.gz -datatype -axis 3 test.mif
mrcat: [WARNING] Value "-axis" is being used as the expected argument for option "-datatype", yet this itself looks like a separate command-line option; the requisite input to command-line option "-datatype" may have been erroneously omitted, which may cause other command-line parsing errors
mrcat: [ERROR] unknown format for image "3"
mrcat: [ERROR] error opening image "3"
```

bjeurissen (approved):
Tested the cases listed above on my system. Works!

PR #2201, opened 2020-10-14
Docs: afdconnectivity: Link to preprint

Lestropie:
It was always my intention that the explanations provided in the "quantitative structural connectivity" manuscript would supersede the contents of this documentation page. Once the manuscript is properly published we can discuss how the software documentation may or may not change; but for now I think it's worth throwing a hyperlink to the preprint in there, since for anyone interested in such concepts it's going to be a better reference than that text-only documentation page.


jdtournier (approved):
Fair enough.

PR #2202, opened 2020-10-14
GLM: Change check for rank deficiency

Lestropie:
Change inspired by [forum thread](https://community.mrtrix.org/t/fba-rank-deficient-matrix/4162).

My suspicion is that I added a threshold to the QR decomposition in order to catch near-deficient cases, but then later on added the condition number check. With the latter in place, I think it makes more sense to have an up-front check for genuine deficiency only, and near-deficiency should be exposed in the condition number check (as per my response in that thread).

Open to alternatives from those with better linear algebra than mine though.

### DIFF
--- a/core/math/stats/glm.cpp
+++ b/core/math/stats/glm.cpp
@@ -79,8 +79,7 @@ namespace MR
 
         void check_design (const matrix_type& design, const bool extra_factors)
         {
-          Eigen::ColPivHouseholderQR<matrix_type> decomp;
-          decomp.setThreshold (1e-5);
+          Eigen::FullPivHouseholderQR<matrix_type> decomp;
           decomp = decomp.compute (design);
           if (decomp.rank() < design.cols()) {
             if (extra_factors) {


Lestropie:
Slightly related to this following contemplation on [forum thread](https://community.mrtrix.org/t/connectomestats-issue/4501/2):

Reporting the design matrix condition number to the user at the commencement of execution is not the only location at which the condition number is quantified. In the case of *variable* GLM tests, where the design matrix varies between each element tested, the condition number is checked once the element-wise design matrix is constructed (could have rows removed and/or additional columns added).

1. If the condition number is a poor marker for "validity" of the model (whether issuing a warning to the user, or excluding elements from generating test statistics), that needs to be revised in both locations.
2. If calculation of the condition number requires execution time that is comparable to the actual inversion of the model, then the calculation is essentially being doubled-up. It might be preferable to use a matrix inversion method that is rank-revealing (with customisable threshold) or that additionally provides eigenvalues for condition number calculation.

Issue #2199, opened 2020-10-13
[./run_tests]: [bug report title]


bjeurissen:
After making some changes that could have repercussions elsewhere, I ran the regressions tests and noticed the following unexpected behavior:

```
./build bin/tckgen
./run_tests tckgen
 head -n 13 testing_tckgen.log
```
outputs
```
------------------------------------------
  Testing MRtrix3 binary "tckgen"
------------------------------------------


-------------------------------------------

## building test commands...


PATH is set to /Users/ben/.cargo/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands

Testing executables in: /Users/ben/mrtrix3/bin
```

which looks fine. But when I remove bin/tckgen and re-run the same test:

```
rm bin/tckgen
rm testing_tckgen.log
./run_tests tckgen
head -n 13 testing_tckgen.log
```

The test still runs fine, even though the tckgen binary has been removed. Clearly another binary (that is on the PATH?) is being used instead, even though the output log suggests that the same ''local'' tckgen binary is being used:

```
------------------------------------------
  Testing MRtrix3 binary "tckgen"
------------------------------------------


-------------------------------------------

## building test commands...


PATH is set to /Users/ben/.cargo/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands

Testing executables in: /Users/ben/mrtrix3/bin
```

Is this behavior expected?
If so, is the test log misleading?
If not,  should the test simply fail?

Note that this behavior is observed on macos.

jdtournier:
Nuts, yes, I can see why that happens... All the `run_tests` script does is make sure the MRtrix folder being tested appears first in the `PATH`, but if your `PATH` already includes the command being tested, then the fact that it's missing from the `bin/` folder being tested isn't picked up. Not sure how to detect that without screwing up the `PATH` for other commands that might be used in the testing, like `diff`, etc...

Lestropie:
One option might be to do a text substitution of the command name (based on the name of the text file that contains the test string) with an absolute expected path, ensuring that anything that's already a filesystem path but also contains that string *doesn't* get substituted (a regex would probably suffice).

PR #2203, opened 2020-10-14
run_tests: check location of executable matches expectation

jdtournier:
Proposed fix for #2199. Simply check that full path to executable being tested matches the `bin/` folder of the MRtrix installation being tested. Should catch most of these issues...

Note that since the test applies to the command as named in the corresponding test, it won't catch mismatches due to other commands used in the test. I expect this is less of an issue, but it might cause problems in isolated cases. It would be much harder to detect this case though...

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -169,6 +169,7 @@ cat >> $LOGFILE <<EOD
 Testing executables in: $MRTRIX_BINDIR
 EOD
 
+export PATH="$(pwd)/testing/bin:${MRTRIX_BINDIR}:${PATH}"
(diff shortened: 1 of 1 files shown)


Lestropie:
This should probably be in `3.1.0`, but not to `dev` given that this script no longer exists there. It will need to be cherry-picked onto the pending branch that will form `3.1.0` to the exclusion of `cmake`.

Lestropie:
Closing as following #2689 script `run_tests` no longer exists on `dev`.

PR #2207, opened 2020-10-21
mrtransform: Throw exception on no operation

Lestropie:
As [suggested on forum](https://community.mrtrix.org/t/feature-request-mrtransform/4198).

Need to make sure that fix is compatible with the logic flow and doesn't bork any possible use cases I'm not seeing.


maxpietsch:
I've changed the error message to reflect that regridding without transformation can be performed via the `-template` option. Otherwise good to go I reckon.

PR #2208, opened 2020-10-22
mrcalc: -modulo operand

Lestropie:


jdtournier:
Sounds good on the real numbers side, but I'm not sure whether we should support this for complex numbers - is the operation even well defined? Also, does your implementation of the operator for complex numbers handle negative values correctly? If I put pure real numbers a=-5, b=2 into the complex number version, I'd get 1 as the result. As far as I can tell, the result of that operation should match [fmod()](https://en.cppreference.com/w/cpp/numeric/math/fmod), which would give a -1. Personally, I'd mark that as `NOT_IMPLEMENTED`...

On a different note: homebrew seems to be broken on the CI:
```
Run brew update
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Error: Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-openssl failed!
Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-python2 failed!
fatal: invalid upstream 'origin/master'
fatal: invalid upstream 'origin/master'
```
Any thoughts, @maxpietsch?

maxpietsch:
Looks like a known github actions failure, workaround: https://github.com/actions/virtual-environments/issues/1811#issuecomment-708480190

Lestropie:
Modulo of complex numbers reverted; I'd done quite a bit of searching, found some implementations, but no hard evidence / proof. Wolfram Alpha supports it, whereas Octave doesn't. Happy leaving out.

PR #2194, opened 2020-10-08
Docs: DW handling: Describe shells

Lestropie:
As prompted by [forum question](https://community.mrtrix.org/t/question-about-b-values-and-its-fitness-to-the-software/4156).

Given the centrality of the shell interpretation of DW volumes, I think it warrants a reasonable amount of explanation.

PR #2209, opened 2020-10-22
mrtrix3.utils.RunList: Pass kwargs to run.command()

Lestropie:
This class isn't currently used anywhere in *MRtrix3*, it's only in place intending to do so at some point.
But turns out I failed to update it in accordance with `run.command()` moving to keyword arguments.


jdtournier (dismissed):

Lestropie:
Due to the combination of a3a7389 & d54dd1f on this PR being not quite a duplicate of fix 4e0f5b1 in #2194, I've force-pushed back to the solitary novel content of this PR (dc5d004) rather than doing a weird nearly-duplicate-fix merge conflict in `.github/workflows/checks.yml` in a PR that's not to do with CI. Another good example of why novel CI issues are best addressed in dedicated PRs <:-P

PR #2212, opened 2020-10-24
Write binary vtkPolyDataFiles

blezek:
This change enhances the functionality of `tckconvert` by enabling writing `vtkPolyData` files (`.vtk`) in a binary format.  In addition to being faster because floats do not need to be converted to ASCII, the files are smaller.

This PR does not have a corresponding test at this moment.

To use the new format:

`tckconvert -binary streamlines.tck streamlines.vtk`

Writing the new VTK XML format (`.vtp`) is preferred and may be the subject of a future PR.

@jdtournier @Lestropie

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -94,9 +94,9 @@ void usage ()
     + Option ("radius", "radius of the streamlines")
     +   Argument("radius").type_float(0.0f)
 
-    + OptionGroup ("Options VTK writer")
+    + OptionGroup ("Options specific to VTK writer")
 
-    + Option ("binary", "write a binary VTK file");
+    + Option ("ascii", "write an ASCII VTK file (binary by default)");
 
 }
 
@@ -108,15 +108,15 @@ void usage ()
 
 class VTKWriter: public WriterInterface<float> { MEMALIGN(VTKWriter)
   public:
-    VTKWriter(const std::string& file, bool write_binary = false) : VTKout (file, std::ios::binary ), write_binary(write_binary){
+    VTKWriter(const std::string& file, bool write_ascii = false) : VTKout (file, std::ios::binary ), write_ascii(write_ascii){
       // create and write header of VTK output file:
       VTKout <<
-        "# vtk DataFile Version 4.2\n"
+        "# vtk DataFile Version 3.0\n"
         "Data values for Tracks\n";
-      if ( write_binary ) {
-          VTKout << "BINARY\n";
-      } else {
+      if ( write_ascii ) {
           VTKout << "ASCII\n";
+      } else {
+          VTKout << "BINARY\n";
       }
       VTKout << "DATASET POLYDATA\n"
         "POINTS ";
@@ -130,15 +130,15 @@ class VTKWriter: public WriterInterface<float> { MEMALIGN(VTKWriter)
       size_t start_index = current_index;
       current_index += tck.size();
       track_list.push_back (std::pair<size_t,size_t> (start_index, current_index));
-      if ( write_binary ) {
+      if ( write_ascii ) {
+          for (const auto &pos : tck) {
+              VTKout << pos[0] << " " << pos[1] << " " << pos[2] << "\n";
+          }
+      } else {
           float p[3];
           for (const auto& pos : tck) {
               for (auto i = 0; i < 3; ++i) Raw::store_BE(pos[i], p, i);
               VTKout.write((char*)p, 3 * sizeof(float));
-          }
-      } else {
-          for (const auto& pos : tck) {
-              VTKout << pos[0] << " " << pos[1] << " " << pos[2] << "\n";
           }
       }
       return true;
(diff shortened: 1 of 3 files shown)


cmd/tckconvert.cpp
```diff
@@ -89,8 +91,12 @@ void usage ()
 
   + OptionGroup ("Options for both PLY and RIB writer")
 
-  + Option ("radius", "radius of the streamlines")
-  +   Argument("radius").type_float(0.0f);
+    + Option ("radius", "radius of the streamlines")
+    +   Argument("radius").type_float(0.0f)
+
+    + OptionGroup ("Options VTK writer")
```
jdtournier, line 97: Could we make that "Options **specific to** VTK writer" for consistency with the other sections...?

cmd/tckconvert.cpp
```diff
@@ -153,6 +192,7 @@ class VTKWriter: public WriterInterface<float> { MEMALIGN(VTKWriter)
 
   private:
     File::OFStream VTKout;
+    bool write_binary;
```
jdtournier, line 195: I'd make that a `const bool` if that's OK.

cmd/tckconvert.cpp
```diff
@@ -102,13 +108,17 @@ void usage ()
...
   public:
-    VTKWriter(const std::string& file) : VTKout (file) {
+    VTKWriter(const std::string& file, bool write_binary = false) : VTKout (file, std::ios::binary ), write_binary(write_binary){
       // create and write header of VTK output file:
       VTKout <<
-        "# vtk DataFile Version 1.0\n"
-        "Data values for Tracks\n"
-        "ASCII\n"
-        "DATASET POLYDATA\n"
+        "# vtk DataFile Version 4.2\n"
```
jdtournier, line 114: I note the version number has changed. Is that necessary for the binary output, and does that impact on the ASCII output in any way...?

jdtournier:
Looks good, thanks for the contribution! I've just suggested a couple of tiny changes, but otherwise I'd be happy to merge to `dev` with a view to release with 3.1.

Test data would be good if you could provide some, though adding these to the test repo is a bit fiddly... Happy to help with that if required. 

We may want to discuss making binary the default output format, and modify the `-binary` option to `-ascii`? Is there any particular reason to prefer VTK ASCII to binary...? As far as I can tell, `tckconvert` currently [doesn't even support importing VTK ASCII](https://github.com/MRtrix3/mrtrix3/blob/0a154e88cbdd001097268a6475646d511abcf849/cmd/tckconvert.cpp#L220), which seems a bit backwards...

cmd/tckconvert.cpp
blezek: I looked into this a bit more, the VTK documentation is conflicting.  The [Users Guide](https://www.kitware.com/products/books/VTKUsersGuide.pdf) on pg 470 says the current version is `3.0`.  I updated the code to write `3.0` (the 4.2 came from [this document, i.e. the first google hit](https://vtk.org/wp-content/uploads/2015/04/file-formats.pdf)).

cmd/tckconvert.cpp
blezek: Sure, also swapping the sense to `write_ascii` and making `binary` the default, related to above comment.

cmd/tckconvert.cpp
blezek: Done.

blezek:
> Test data would be good if you could provide some, though adding these to the test repo is a bit fiddly... Happy to help with that if required.

Happy to contribute, I would need some assistance writing the test, but can provide an example of one streamline saved as a binary and ascii VTK file.

> We may want to discuss making binary the default output format, and modify the `-binary` option to `-ascii`? Is there any particular reason to prefer VTK ASCII to binary...? As far as I can tell, `tckconvert` currently [doesn't even support importing VTK ASCII](https://github.com/MRtrix3/mrtrix3/blob/0a154e88cbdd001097268a6475646d511abcf849/cmd/tckconvert.cpp#L220), which seems a bit backwards...

Agreed, the `-ascii` is more portable especially because the documentation is ambiguous on the endian-ness (it's big endian, for the record).  I swapped the default to be binary.

Writing ascii and reading binary where to accommodate some other VTK based software I needed to use.  Writing 1M streamlines in ASCII was becoming prohibitive, so I wanted to add the binary.

Additional commits coming soon!  Much appreciated!
-dan

blezek:
@jdtournier attached is a single streamline in `tck`, `vtk` ASCII and `vtk` BINARY formats.  I verified VTK `v8.x` can read the binary and ascii files.
[Archive.zip](https://github.com/MRtrix3/mrtrix3/files/5440063/Archive.zip)

blezek:
@jdtournier I'm pretty happy with the branch, and got the tests passing.  Please consider it ready to go.

jdtournier:
OK, I'm looking at it now - all checks out as far as I can tell. 

I see the test modifications work out, but only consider the ASCII output for VTK. I can add the data you sent the testing repo and make sure all of that is consistent, assuming you have a test or two to add for the binary output...?

blezek:
> I see the test modifications work out, but only consider the ASCII output for VTK. I can add the data you sent the testing repo and make sure all of that is consistent, assuming you have a test or two to add for the binary output...?

I added two new tests for conversions from `tck` -> `vtk`, and `tck` -> `vtk` -> `tck`.

No need to add more data.

Thanks for all the help!

jdtournier:
Great! Thanks for your contribution, I'll merge now.

PR #2214, opened 2020-10-27
tckmap: Revise help page

Lestropie:
Was addressing [forum question](https://community.mrtrix.org/t/tckmap-options/4212/3) and felt like `tckmap` was a command begging for exploitation of the `DESCRIPTION` help page mechanism. Could probably be expanded even further, but I might put that off given I'm hoping to go back and finish the fixel TWI work at some point and that will involve some changes to the interface here.


jdtournier:
Looks good. I've added a commit to slightly tweak the formatting of the operations, since it was looking a bit crowded on the help page. This is what it looks like with my edits on the terminal:
```
DESCRIPTION

     The -contrast option controls how a value is derived for each streamline
     that is subsequently contributed to the image elements intersected by that
     streamline, and therefore strongly influences the contrast of that image.
     The permissible values are briefly summarised as follows: 

     - tdi: Each streamline effectively contributes a value of unity to the
     final map (equivalent to the original Track Density Imaging (TDI) method)

     - length: The length of the streamline in mm

     - invlength: The reciprocal of streamline length

     - scalar_map: Values are sampled from a scalar image (which must be
     provided via -image)

     - scalar_map_count: If a non-zero value is sampled from a scalar image (as
     provided via -image), the streamline contributes a value of 1, otherwise
     it contributes 0, such that an image can be produced reflecting the
     density of streamlines that intersect such an image

     - fod_amp: The amplitudes of a Fibre Orientation Distribution (FOD) image

     - curvature: The curvature of the streamline

     - vector_file: A value for each streamline has been pre-calculated, and
     these are provided in a text file via the -vector_file option

     A "super-resolution" output image can be generated using the -vox option,
     whether or not a template image is provided using the -template option. If
     -template is used in conjunction with -vox, the image axes and FoV will
     still match that of the template image, but the spatial resolution will
     differ.

     Note: if you run into limitations with RAM usage, make sure you output the
     results to a .mif file or .mih / .dat file pair - this will avoid the
     allocation of an additional buffer to store the output for write-out.
```
and online:
Screenshot from 2020-11-11 12-29-09

See what you think...

jdtournier:
One more point: 

>      Note: if you run into limitations with RAM usage, make sure you output the results to a .mif file or .mih / .dat file pair - this will avoid the allocation of an additional buffer to store the output for write-out.

Technically, that only works if the file is being written to a local drive. For network drives (or even USB drives), we'd be using delayed write-back.

Issue #2191, opened 2020-10-04
run.command(): Refine text decoding


Lestropie:
Following from #2190. Even if Python2 support is dropped, this could still do with a little tweaking to account for the fact that it's possible for a command to be placing non-text data (not just e.g. accented characters, but even binary data) on `stdout` in particular, and such data should ideally not be decoded until the point at which an attempt to display it as text is made.

- When using `-info` or `-debug`, `stderr` is currently split between terminal output that is decoded, and return value that is not; this should remain as-is.

- For `stdout` in this case, as well as both `stdout` / `stderr` when neither `-info` nor `-debug` is used, text decoding should *not* occur.

- `stdout` and `stderr` data are accessible through two means: the `CommandReturn` `namedtuple` and the `MRtrixCmdError` exception. Each of these would need to provide access to both the raw undecoded binary data, and decoded data in string form:

    - For `CommandReturn`, any alteration to the `.stdout` and `.stderr` members would result in backward-incompatibility. One option to avoid this would be to keep those members as providing the decoded data, and have separate members of the `namedtuple` providing the undecoded data. But this may result in divergence with the behaviour of `MRtrixCmdError` depending on what is done there...

    - For `MRtrixCmdError`, I'd need to check how may locations read from the individual `.stdout` / `.stderr` contents, but if such usages are rare, the `.__str__()` member could perform decoding there instead of expecting those two members to be already decoded. That would also be consistent with how binary data vs. string data is intended to be handled in Python3 (as far as I can tell).

PR #2215, opened 2020-10-28
Drop python2 support

### LINKED ISSUE #2047: Drop python2 support?

### LINKED ISSUE #2190: dwifslpreproc: a python inconsistancy.
Please use this form for *bug reports only*; for installation problems, and general questions and support, please consult the [community forum](http://community.mrtrix.org/search?expanded=true).

The bug is described at https://community.mrtrix.org/t/dwifslpreproc-error-on-a-cluster/4144/2

A clear and concise description of what the bug is; including if appropriate
how the observed behaviour differs from that expected.

dwifslpreproc does not run with python2.7

Please provide the following information:

OS =  Linux on the HPC.

### LINKED ISSUE #2210: distutils.spawn.find_executable to shutil.which
The *MRtrix3* Python API depends on the `distutils` module for the `find_executable()` function. Apparently usage of this module is discouraged in favour of `setuptools`, and it is indeed [intended for deprecation](https://www.python.org/dev/peps/pep-0632/#id13). But even then, with `setuptools` being to "download, build, install, upgrade, and uninstall Python packages", it's not necessarily a sensible dependency to have.

As of Python 3.3, the widely-used `shutil` module has [the `which()` function](https://docs.python.org/3.8/library/shutil.html#shutil.which), which serves the same purpose.

So if following through with #2047, I would suggest making the minimum dependency 3.3 ([released Sep 2012](https://www.python.org/downloads/release/python-330/)), and making use of this function throughout.

Lestropie:
Closes #2210.

Closes #2047.

Will eventually resolve #2190.

Should consider resolving #2191 here.

Made a bit of a list in 3a856ce as to what changed, but would appreciate some reflection on whether or not there's anywhere else in the code base that required gymnastics to support both Python 2 and 3 that I may have missed.


jdtournier:
All looks sensible, other than the first thing: changing the interpreter on the shebang to `python3`. I don't think that's the recommended way of doing things...? It might work on Ubuntu, but it's not guaranteed to work on other standard distros. Are there recommendations around this from the python community?

jdtournier:
Ok, forget that last comment, it is indeed the recommendation on this PEP:
https://www.python.org/dev/peps/pep-0394/

jdtournier:
CI is failing due to lots of these errors during `pylint`:
```
Bad option value 'import-outside-toplevel' (bad-option-value)
```
A quick look online reveals that [this check is available as of pylint version 2.4](http://pylint.pycqa.org/en/latest/whatsnew/2.4.html). But the version dragged in by `apt-get` on the GitHUb runners is 1.8.3:
```
Unpacking pylint3 (1.8.3-1) ...
```
which explains the errors. 

Previously, `bad-option-value` was also disabled in the config, but that was removed in 3a856ce1 - which is why the issue only arises now. 

So, two options:
- reinstate `bad-option-value` in `pylint.rc`
- modify all occurrences of `#pylint: disable=import-outside-toplevel` to additionally have `bad-option-value`, e.g. `#pylint: disable=bad-option-value,import-outside-toplevel` (as [suggested here](https://github.com/PyCQA/pylint/issues/3312#issuecomment-571082824)).

Lestropie:
Installing `pylint` via `pip` rather than `apt-get` brings in 2.6.0 rather than 1.8.3.

Lestropie:
Minor clarification here: While the Python scripts in `bin/` will as of `3.1.0` explicitly invoke `python3`, do we want to retain compatibility with Python2 in auxiliary scripts (e.g. `build`, `configure`), or force Python3 there also? This has ramifications for what to do with `configure`: it either needs to ensure that it is itself running on Python3, or if running on Python2 (and such is permitted at our end), it needs to do an explicit check for the presence of `/usr/bin/python3` as a compulsory dependency.

Issue #1975, opened 2020-03-11
ENH mrfilter: Abort on unused options


Lestropie:
Raising in the context of `mrfilter`, but there's probably other commands where the same argument is applicable.

In commands where there are discrete set of operations available, each of which has its own unique set of command-line options, but the user provides a command-line option that isn't applicable to the operation selected, the command should at the very least issue a warning, if not abort outright (probably my preference). In this specific case, `mrfilter smooth -extent` was used: `-extent` gets ignored and the default smoothing FWHM is used, both silently.

I'm sure I've used code somewhere that generates a list of all available command-line options, removes only those options applicable to the operation selected, and then verifies that the user hasn't specified any of the other options. Maybe there should be something along these lines embedded deeper within `MR::App`?

Lestropie:
Renaming and bumping in response to #2220. 

All this needs is a binary flag for each option that gets set whenever `get_options()` finds it (irrespective of the content of any Arguments associated with the option), and upon command completion, scan through `MR::App::OPTIONS` and issue a warning listing those options that were specified but never accessed through `get_options()`.

Lestropie:
I would also want this functionality incorporated in Python CLI: #1608

Lestropie:
My intention was indeed to perform the check at command exit, for the reasons you describe. To do it before primary processing takes place would require a lot of code modification, a new explicit function invocation for each command, and developer diligence to ensure that the required execution order is not violated. But I think a warning would be a considerable enhancement over the current silent absence of use even if it is only at command exit.

PR #2221, opened 2020-11-11
get_option_value(): Restrict usage

Lestropie:
Came across this while thinking about #1975. While the current `get_option_value()` code is permissive, this theoretically enables erroneous usage of the function by developers. This change constrains usage of the function to only those command-line options for which the usage of this function makes sense.

### DIFF
--- a/core/app.h
+++ b/core/app.h
@@ -442,8 +442,8 @@ namespace MR
 
 
     //! Returns the option value if set, and the default otherwise.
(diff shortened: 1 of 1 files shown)


jdtournier (approved):
Looks good. I've added a minor tweak to the switch statement to avoid duplicating the error message, but it's pretty much cosmetic. Looks fine either way.

PR #2227, opened 2020-11-17
CI: Try installing "python3-pylint" rather than "pylint3" (Take 2)

Lestropie:
See if this addresses outdated version of pylint pulled by GitHub runner as commented in #2215.

Opening a second PR in the hope that it triggers the "`secondary-checks`" GitHub Action, for which I've created a dedicated branch protection rule specifically for this PR.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -160,7 +160,7 @@ jobs:
     - name: install dependencies
       run: |
          sudo apt-get update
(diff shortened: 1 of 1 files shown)


Lestropie:
Never mind: The PR target branches for which checks are submitted is defined in the YAML config file rather than GitHub settings, and therefore I can't trivially activate checks for a specific CI test like this. So will just push to #2215 and test proposed CI changes there.

PR #2230, opened 2020-11-20
dwidenoise: limit patch size to image dimensions

dchristiaens:
As reported on the [forum](https://community.mrtrix.org/t/dwidenoise-system-fatal-code-sigsegv-11-segmentation-fault-invalid-memory-access/2302/31).

If the user sets a patch size larger than the image dimension, this will throw an error.

If the patch size is selected automatically, it will be truncated to the image dimensions.


jdtournier (approved):
Seems good to me. :+1:

PR #2233, opened 2020-11-24
mrtransform: write fewer comments 

maxpietsch:
Small tweak to prevent 'regridded to template image...' and 'resliced using warp image...'  from being written to the image comments. For `.mif`, this information is in the command history anyway and `mrmath` or `mrcat` tend to accumulate these comments when operating on multiple files.


Lestropie (approved):
Agreed; these give me the irrits too since `command_history` was implemented.

PR #2236, opened 2020-11-27
dwifslpreproc: fix help for -align_seepi option

jdtournier:
@Lestropie: hope I've got that right - and that you haven't already submitted a fix elsewhere...


Lestropie (approved):
:facepalm:

PR #2237, opened 2020-12-02
DICOM: Tweaks to import

Lestropie:
Tweaks based on assessing data provided as a consequence of [this forum thread](https://community.mrtrix.org/t/dwifslpreproc-error-b-value-must-be-non-negative/4308).

Details in cf83f19.

While the negative *b*-value one is reasonably trivial, I'm not 100% on the sequence numbering change. In the case of these data, the frame (?) named "`ep_b-16383`" was being allocated as sequence number +16383, while "`ep_b1000#1`" - "`ep_b1000#12`" were allocated the expected sequence numbers. My guess is that the relevant code was *intended* to parse the sequence numbers in this way but was not made sufficiently robust to the absence of the "`#`", but I don't have the breadth of experience with DICOM to know if there are other cases for which the original code was intended to handle and that may be broken by this change.

### DIFF
--- a/core/file/dicom/image.cpp
+++ b/core/file/dicom/image.cpp
@@ -70,8 +70,11 @@ namespace MR {
                       return;
                     while (c >= 0 && isdigit (sequence_name[c]))
                       --c;
-                    ++c;
-                    sequence = to<size_t> (sequence_name.substr (c));
+                    if (c >= 0 && sequence_name[c] != '#') {
+                      sequence = 0;
+                      return;
+                    }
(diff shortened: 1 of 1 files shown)


jdtournier (approved):
Looks sensible.

jdtournier:
You're spot on with the way the sequence name is used to order the slices. This only applies to Siemens, by the way. I'm also not sure what consequence your change may have for how these problematic data are handled, but hopefully that issue will be fixed reasonably quickly and we don't have to worry about it a great deal... 

I'll run the changes through my DICOM tests now just to make sure, though...

PR #2239, opened 2020-12-04
Fix enable_if_image_type() method

jdtournier:
Required for the [`save()` method](https://github.com/MRtrix3/mrtrix3/blob/dev/core/image.h#L529-L536) to work correctly (only useful when debugging). 

There's still an issue when using `save()` to dump a temporary file to `.mif` format: for some reason, the data in the file produced is corrupted. I'm not sure what's causing it, but this only happens when going through the [`dump_to_image_file()` method](https://github.com/MRtrix3/mrtrix3/blob/dev/core/image.h#L454-L499). Might be worth looking into at some point...

### DIFF
--- a/core/image_helpers.h
+++ b/core/image_helpers.h
@@ -145,7 +145,7 @@ namespace MR
             std::declval<ImageType>().ndim() +
             std::declval<ImageType>().size(0) +
             std::declval<ImageType>().name().size() +
-            std::declval<ImageType>().value() +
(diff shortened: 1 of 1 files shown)


Lestropie:
This looks to me like a `master` hotfix?
Also should be easy to produce an example command that misbehaves, and that can sit in `testing/bin/`?

Lestropie:
Superseded by #3334.

PR #2246, opened 2020-12-12
Fixes issue when double-clicking quarantined files on macos.

### LINKED ISSUE #2245: mrview: filename association woes on macos
On macos, I am getting these weird crashes when opening mif files through double clicking when MRview is not open yet. 

So for most files it just works, but for some it crashes... When it does crash on a file, creating an entire new file with the same contents (e.g. by means of mrconverting it) no longer results in a crash, but simply copying the file to another filename does not help... But there really seems to be nothing wrong with the original file, so I really don't understand what is going on...

Has anyone else experienced this before? I have had this issue many times before, but ignored it so far, but now it is finally getting to me ...

You can see the issue in this screen recording: https://youtu.be/7l775DmD4Ks

bjeurissen:
This fixes #2245

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -977,16 +977,18 @@ namespace MR
     void sort_arguments (int argc, const char* const* argv)
     {
       for (int n = 1; n < argc; ++n) {
-        const Option* opt = match_option (argv[n]);
-        if (opt) {
-          if (n + int (opt->size()) >= argc)
-            throw Exception (std::string ("not enough parameters to option \"-") + opt->id + "\"");
+        if (argv[n]) {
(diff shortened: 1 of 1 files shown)


jdtournier:
The OS sends a null argument...??? How does that make any sense?!? 

In any case, if that fixes it, I can't see any problem with your changes. :+1:

jdtournier (approved):

bjeurissen:
> The OS sends a null argument...??? How does that make any sense?!?
> 
No idea... The weird thing is that the original argument is still present, so double-clicking opens quarantined files just fine as long as you ignore the null pointers...

Issue #2248, opened 2021-01-03
dwi2response: Unhandled Python exception ValueError: could not convert string to float: N/A


GerardYu:
Hi there, i've encounter the error below when running dwi2response. It appears that this error only occurs on 1 out of 91 of my subjects. If its helpful, i've uploaded beddytensors.mif here https://drive.google.com/file/d/1omAdS-2td2PwVZjIPEVlmEB4FWdQvYjB/view?usp=sharing

junhong@junhong-Precision-5820-Tower:/bcs/junhong/MAP/data/017/time1/DTImrt2$ dwi2response dhollander beddytensors.mif wm_response.txt gm_response.txt csf_response.txt -nthreads 20 -force
dwi2response:
dwi2response: Note that this script makes use of commands / algorithms that have relevant articles for citation. Please consult the help page (-help option) for more information.
dwi2response:
dwi2response: Generated scratch directory: /bcs/junhong/MAP/data/017/time1/DTImrt2/dwi2response-tmp-KZJOBX/
dwi2response: Importing DWI data (/bcs/junhong/MAP/data/017/time1/DTImrt2/beddytensors.mif)...
dwi2response: Changing to scratch directory (/bcs/junhong/MAP/data/017/time1/DTImrt2/dwi2response-tmp-KZJOBX/)
dwi2response: Computing brain mask (dwi2mask)...
dwi2response: -------
dwi2response: 3 unique b-value(s) detected: 0,1000,3000 with 14,60,60 volumes
dwi2response: -------
dwi2response: Preparation:
dwi2response: * Eroding brain mask by 3 pass(es)...
dwi2response:   [ mask: 49228 -> 32396 ]
dwi2response: * Computing signal decay metric (SDM):
dwi2response:  * b=0...
dwi2response:  * b=1000...
dwi2response:  * b=3000...
dwi2response: * Removing erroneous voxels from mask and correcting SDM...
dwi2response:   [ mask: 32396 -> 32393 ]
dwi2response: -------
dwi2response: Crude segmentation:
dwi2response: * Crude WM versus GM-CSF separation (at FA=0.2)...
dwi2response:   [ 32393 -> 17770 (WM) & 14623 (GM-CSF) ]
dwi2response: * Crude GM versus CSF separation...
dwi2response:   [ 14623 -> 10032 (GM) & 4591 (CSF) ]
dwi2response: -------
dwi2response: Refined segmentation:
dwi2response: * Refining WM...
[...]

PR #2249, opened 2021-01-06
python: fix handling of N/A output in image.statistics()

jdtournier:
Discussed in #2248

I consider this a bug fix, but I suggest merging to `dev` given there is scope for this to change behaviour of existing apps. Up for discussion.

### DIFF
--- a/lib/mrtrix3/image.py
+++ b/lib/mrtrix3/image.py
@@ -247,7 +247,7 @@ def statistics(image_path, **kwargs): #pylint: disable=unused-variable
   stdout_lines = [ line.strip() for line in stdout.decode('cp437').splitlines() ]
   result = [ ]
   for line in stdout_lines:
-    line = line.split()
+    line = line.replace('N/A', 'nan').split()
     assert len(line) == len(IMAGE_STATISTICS)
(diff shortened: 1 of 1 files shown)


Lestropie (changes requested):
Would prefer to do the copyright message update in its own PR so that it doesn't corrupt the diff of unrelated changes. It looks like other stuff unrelated to the topic at hand has also made its way into here, likely due to a difference in base between the local branch and the PR, which would also make more sense in its own PR.

> I consider this a bug fix, but I suggest merging to dev given there is scope for this to change behaviour of existing apps.

If the only "change of behaviour" is provision of a more interpretable error message where continuation of processing is not possible in any case, I don't think that merging to `master` is a problem.

jdtournier:
Fair enough to split it from the changes in `master`. I'll clean this up once #2255 is merged. 

>> I consider this a bug fix, but I suggest merging to dev given there is scope for this to change behaviour of existing apps.
>
> If the only "change of behaviour" is provision of a more interpretable error message where continuation of processing is not possible in any case, I don't think that merging to master is a problem.

Well, in the case of `dwi2response`, I assume processing would stop in any case. But there may be other cases where a single-voxel mask would have caused a failure prior to these changes, but would complete successfully now. But if that's the case, I would agree that these would still nonetheless class as bug fixes. So I'll rebase it onto `master` as you suggest.

Lestropie (approved):

PR #2250, opened 2021-01-06
mrview: fix handling of -focus and -target options

jdtournier:
- The `-focus` option was not labelled as type_sequence_float(), leading to
  a warning if the first value is negative, e.g.:

      mrview: [WARNING] Value "-1,2,3" is being used as the expected argument
      for option "-focus"; is this what you intended?

- It should be allowed to use the `-target` option multiple times.


Lestropie (changes requested):
Given that `-focus` can be *either* a floating-point sequence *or* a boolean, it would be better flagged as `.type_various()`. That will still remove the erroneous warning if the first character of the provided argument is a dash.

jdtournier:
> it would be better flagged as `.type_various()`.

I guess you learn something new every day... :+1:

Lestropie (approved):

PR #2252, opened 2021-01-06
dwi2response dhollander: fix handling of data with negative values

jdtournier:
This simply clamps values of the input data at zero before taking their mean per-shell, which seems to fix issues raised in #2248.

### DIFF
--- a/lib/mrtrix3/dwi2response/dhollander.py
+++ b/lib/mrtrix3/dwi2response/dhollander.py
@@ -116,7 +116,7 @@ def execute(): #pylint: disable=unused-variable
   for ibv, bval in enumerate(bvalues):
     app.console(' * b=' + str(bval) + '...')
     meanpath = 'mean_b' + str(bval) + '.mif'
-    run.command('dwiextract dwi.mif -shells ' + str(bval) + ' - | mrmath - mean ' + meanpath + ' -axis 3', show=False)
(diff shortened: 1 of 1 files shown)


Lestropie (approved):
Should be fine as a local band-aid, but unclear whether or not it will resolve the case where no values in a voxel are positive (e.g. maybe caused by a negative Jacobian in pre-processing), and hence the subsequent sivision / log transform won't be finite. Presumably the issue with the sample data was that *some* volumes were negative, and therefore the mean was misleadingly close to zero?

jdtournier:
> Presumably the issue with the sample data was that some volumes were negative, and therefore the mean was misleadingly close to zero?

Yes, that's exactly what the problem was. I think it would have been fine if the mean had actually been _below_ zero, as that would have given rise to non-finite values, which I'm pretty sure would have been ignored in the subsequent `mrthreshold` operation. 

So you're right that this 'fix' is certainly not guaranteed to fix all cases. But I'm pretty sure that if no values were positive, that wouldn't actually be a problem, by the same rationale as above (I'd need to run tests to make sure). 

But realistically, this is only supposed to catch the rare cases where regridding operations in e.g. `dwifslpreproc` produce a lot of negative values in problematic low signal areas (particularly CSF, as was the case here), leading to unrealistically high ADC values, which is what was throwing off the automatic thresholding. I'm not sure how else we could guard against that, other than implementing some form of outlier rejection in `mrthreshold`...

In any case, I'll merge now, but I agree this doesn't preclude adding more robust safeguards down the track if necessary.

PR #2257, opened 2021-01-12
tckglobal: Fix bug that caused particles in a voxel to be duplicated when tried to be removed

sertopexamgio:
I observed particles inside a voxel getting duplicated after the  ParticleGrid::remove and ParticleGrid::shift. After some digging, the reason seem to be the [erase-delete-idiom](https://en.wikipedia.org/wiki/Erase–remove_idiom). After these changes, no duplicates of particles were observed and removal of shifting seems to work as expected.


dchristiaens (approved):
Very good catch, and thanks for contributing!

This is clearly a bug, which I seem to have introduced 4 years ago. I tested your fix and it works exactly as it should.

PR #2258, opened 2021-01-12
tckglobal: Fix erroneous -niter parsing for values bigger that 1.4e9

sertopexamgio:
The tckglobal command when run with `-niter 10B` could only run for `1410065408` iterations instead of `10000000000` iterasions. These changes fixed the parsing issue.

### DIFF
--- a/cmd/tckglobal.cpp
+++ b/cmd/tckglobal.cpp
@@ -281,7 +281,7 @@ void run ()
     }
   }
 
-  uint64_t niter = get_option_value("niter", (uint64_t)DEFAULT_NITER);
+  uint64_t niter = get_option_value<uint64_t>("niter", DEFAULT_NITER);
   double t0 = get_option_value("t0", DEFAULT_T0);
   double t1 = get_option_value("t1", DEFAULT_T1);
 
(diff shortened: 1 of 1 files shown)


Lestropie:
Hey @sertopexamgio,

Awesome to see some external pull requests coming in!

For this one, rather than expanding the code out to multiple lines, you can instead exploit the fact that [`get_option_value()` is a templated function](https://github.com/MRtrix3/mrtrix3/blob/master/core/app.h#L444-L460). This is typically inferred from the type of the default value, which in this case is a [macro](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/cmd/tckglobal.cpp#L44) that the compiler will presumably interpret as an `int`. If you were to instead explicitly invoke `get_option_value<uint64_t>()`, that would be fractionally neater.

Cheers!
Rob

sertopexamgio:
Hey @Lestropie,

thanks for your review, that was a neat suggestion.
I hope it looks good now.
Have a good one!

Cheers!
Serxhio

dchristiaens (changes requested):
I concur with @Lestropie that
```
uint64_t niter = get_option_value<uint64_t>("niter", DEFAULT_NITER);
``` 
would be the cleaner syntax. Nice addition and thanks for contributing!

dchristiaens (approved):

PR #2259, opened 2021-01-16
Multithreading: Proposed alteration to functor duplication

Lestropie:
This one needs a little bit of background to justify.

In mucking around with some of the new fancy GLM features on full FBA datasets, I've encountered a couple of things I'd like to resolve:

1. Currently, calculation of properties of the default permutation is done through a [different set of functions](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/core/math/stats/glm.h#L124-L202) to the GLM test class. While it has the advantage of multi-threading within a single permutation, it's a lot of duplicated code.

2. In order to be thread-safe, the GLM test classes allocate all requisite memory for processing each permutation within [the functor itself (which is defined `const`)](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/core/math/stats/glm.h#L243). This results in e.g. `top` reporting memory use being larger than it actually is.
This is necessary with the current code design because the statistical inference commands construct a `std::shared_ptr<GLM::TestBase>`, and this gets passed to the relevant heavy-lifting functions, none of which are templated. 

What I'd prefer is to have a number of instances of the GLM test class equal to the number of working threads, where each instance retains all of the relevant intermediate matrix data as member variables. In addition to preventing re-allocation of memory as per point 2, it should also be possible to, following running the default permutation through the GLM test class, run a virtual function that would return access to whatever statistical properties the command may wish to write to the file system, thus removing the necessity of the code in point 1.

The trap here is that, if you have a pointer to the base class, you can't then just pass that to the multi-threading back-end, as it'll run the copy-constructor of the base class rather than the derived class that's responsible for the actual heavy-lifting.

The generally accepted solution to generating new instances of a derived class from a base class pointer is to define a virtual function that returns a pointer to the new instance, with any derived classes providing custom implementations of such duplication. So what this PR is proposing is that whenever a functor is to be cloned for the sake of running multiple threads, the threading back-end checks for the presence of such a function (which must have an accepted pre-determined name). 

In the absence of this change or something comparable, to achieve what I want to do I'd have to resort to templating the bulk of the permutation testing functions & classes, which seems like a lot of unnecessarily duplicated machine code to me.

### DIFF
--- a/core/thread.h
+++ b/core/thread.h
@@ -22,8 +22,8 @@
 #include <mutex>
 
 #include "debug.h"
-#include "mrtrix.h"
 #include "exception.h"
+#include "mrtrix.h"
 
 /** \defgroup thread_classes Multi-threading
  * \brief functions to provide support for multi-threading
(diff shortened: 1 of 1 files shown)


Lestropie:
P.S.:

- For my particular application currently, think I can get away with using this *concept* but without necessitating the specific changes in the threading back-end. Nevertheless still worth considering.

- If pursuing these changes, I would suggest that `MR::copy_ptr<>` would probably need to receive similar changes.

Lestropie:
Closing this given an alternative solution was used for #2269. If however this design pattern re-appears elsewhere could reconsider a deeper integration.

(Rather than the threading back-end detecting the presence of the `_clone()` member function, `PermTest::(Pre)Processor` have a `std::unique_ptr<>` to a base class, for which the virtual `_clone()` function is invoked during (copy-)construction of `PermTest::(Pre)Processor`)

Lestropie:
It occurs to me now that there's a few places in the code where class B, which inherits from class A, is overriding A's `clone()` virtual function in order for copy-construction of class C, which contains a pointer to an instance of A, can properly duplicate class B accounting for inheritance.

This is the sort of scenario where `MR::copy_ptr<>` would ideally be used, so that the default copy-constructor will do a deep copy. However I'm guessing that with the code as it currently stands, that deep copy would yield an instance of A rather than B. With some SFINAE, `MR::copy_ptr<>` could be modified to detect the presence of a `clone()` function and invoke that for copy-construction.

PR #2263, opened 2021-01-20
fod2fixel: Order fixels by FD

Lestropie:
As recently re-raised on [forum](https://community.mrtrix.org/t/fixel2voxel-how-to-extract-primary-secondary-and-tertiary-voxel-based-fd-maps/4394).

Here's the output of `fod2fixel` then `fixel2voxel none` on FD from the test data:

Old code:
screenshot0003

New code:
screenshot0002

Kinda hard to see, but there are some voxels in the top image that get brighter going from left to right, whereas in the lower image FD always decreases from left to right.

### DIFF
--- a/src/dwi/fmls.h
+++ b/src/dwi/fmls.h
@@ -160,9 +160,6 @@ namespace MR
       };
 
 
-      bool order_by_fd (const FOD_lobe&, const FOD_lobe&);
-
-
 
       class FOD_lobes : public vector<FOD_lobe> { MEMALIGN(FOD_lobes)
         public:
--- a/src/dwi/fmls.h
+++ b/src/dwi/fmls.h
@@ -160,6 +160,9 @@ namespace MR
       };
 
 
+      bool order_by_fd (const FOD_lobe&, const FOD_lobe&);
+
+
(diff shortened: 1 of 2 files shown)


src/dwi/fmls.h
```diff
@@ -160,6 +160,9 @@ namespace MR
       };
 
 
+      bool order_by_fd (const FOD_lobe&, const FOD_lobe&);
```
jdtournier, line 163: Wouldn't it be worth declaring this function `inline` and define it fully at this point? It's unlikely to make that much difference to your use case given that the function is defined in the same cpp file as it's going to be used, so the compiler will likely inline the code anyway (even though you've basically told it not to). But it might impact on the code if used in other cpp files. 

On the other hand, if you don't intend for that function to be used outside of `src/dwi/fmls.cpp`, then I'd declare and define it `inline` within an anonumous namespace in `src/dw/fmls.cpp` - no need to expose internals of your implementation if you have no intention to allow other parts of the code to make use of it...

Lestropie: 39ab87b?

PR #2264, opened 2021-01-20
mrview show rgb values

maxpietsch:
In RGB-mode, this adds info for the volumes used for RGB rendering to mrview's main image and overlay:

image


jdtournier:
Seems good to me - I reckon we can merge this to `master`? It would be a good candidate for inclusion in the next minor release...

jdtournier (dismissed):

Lestropie:
Will try to rebase PR onto new branch that will be `dev` without `cmake` that will form `3.1.0`.

PR #2278, opened 2021-02-07
mrfilter enhancements

Lestropie:
Candidate changes for getting into `3.1.0`, but still requires a bit of work and cleanup, hence listed as draft.

Main goal was to support arbitrary 3D convolution kernels. In the process of doing that, ended up reworking quite a bit of the adapter back-end that's shared with e.g. gradient calculations. Previously, certain demands were implemented only for specialised cases, for which here I attempt to provide a more generalised solution, e.g.:

- Modifying 3D convolution to be defined in scanner rather than image space;
- Using three orthogonal kernels and from that producing an image containing the norm of the resulting 3-vector;
- Dealing with convolution at the outskirts of the image FoV.

Apart from re-checking the code that was written a while back, including going through the various remaining TODO notes, there's a couple of things that I'd intended on pursing before merging:
- Kernels that are based on a combination of prefilter & derivative might be possible to apply as three 1D convolutions rather than building a single 3D convolution
- Binary dilation & erosion that is superior to multiple iterations of 6-nearest-neighbour dilation / erosion by instead generating a spherical kernel of a specified radius.


Lestropie:
- [Prewitt kernel](https://en.wikipedia.org/wiki/Prewitt_operator) would be an easy one to add.

- What's currently referred to here as "unsharp" is actually just a sharpening filter; should rename, and add the actual "unsharp mask" technique, which is a combination of smoothing and edge detection, sometimes with some form of thresholding (though I've not yet found a good source for how best to perform such).

Lestropie:
Rename "`EdgeHandlers`" to "`BoundaryHandlers`"; more consistent with [`scipy`](https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#interpolation-boundary-handling).

Lestropie:
It would be more elegant if the convolutional filters could be generated and applied using `Eigen::Tensor<>`.

PR #2197, opened 2020-10-12
dwi2mask overhaul (replacement PR)

Lestropie:
#2088 is broken due to me wiping my fork (completely forgetting that that's where the `dwi2mask` updates were...) and GitHub refusing to re-establish the PR link; so relisting.

Given that the other OHBM Hackathon *MRtrix3* project has matured to a preview release and should become a part of `3.1.0`, I wanted to clear this off of my desk; there's also been a couple of private discussion threads where having this finalised would be beneficial. Listing as a full PR rather than a draft in the hope of getting some fresh eyes & opinions on it, but I will probably try to make some test data before merging.

There's some brain masking approaches that I tried to get working as a `dwi2mask` algorithm but couldn't get anything reasonable out of them, I still have the code if anyone wants to try themselves:
https://github.com/GUR9000/Deep_MRI_brain_extraction
https://pypi.org/project/deepbrain/
I'm open to implementing further algorithms to wrap existing approaches, but such approaches need to be compatible with T2-weighted images, and many newer brain extraction approaches are T1w tailored.

Because it's coming from my fork I don't have things set up to preview the documentation, but you can see the relevant raw pages:
https://github.com/Lestropie/mrtrix3/blob/dwi2mask/docs/dwi_preprocessing/masking.rst
https://github.com/Lestropie/mrtrix3/blob/dwi2mask/docs/reference/commands/dwi2mask.rst

@jdtournier: I thought about having hyperlinks in the `dwi2mask` help pages to the new dedicated docs page on the topic. But this would require parallel gymnastics to those used for getting version-matched hyperlinks into the SH-related binaries prior to tagging. Want to do it anyway?
**Edit**: If doing so, should also add links to the [`dwi2response` docs page](https://mrtrix.readthedocs.io/en/latest/constrained_spherical_deconvolution/response_function_estimation.html) in the `dwi2response` help pages.

Tagging @wtsyeda and @RicardoRios46 so that they're aware of the progress given their contributions.

PR #2279, opened 2021-02-07
Container recipes

Lestropie:
As discussed in MRtrix3/containers#14.

Plan is to have the primary container building recipes inside the main repository, and for the "`containers`" repository to only be responsible for building the minified dependencies.

1. In order for the recipe file contained within any particular tagged version of *MRtrix3* to be guaranteed to clone that very same version of *MRtrix3* during the build process, the recipe files must be manually updated to reflect the upcoming tag name, similarly to the gymnastics already done with `core/version.h`. The [Wiki instructions](https://github.com/MRtrix3/mrtrix3/wiki/Publishing-a-new-release#merge-to-master) have already been updated accordingly.

2. I have moved some of the contents of `README.md` from the `containers` repo into a new page in the online documentation page. 

3. Reflecting on the contents of `README.md` in *this* repository, there's quite a lot of stripped-down instructions for compilation from source. Not only is it less comprehensive that the online documentation for building from source, but we expect over time that less people will be building from source. It might be worth contemplating removing these instructions altogether and instead relying on selection of which hyperlink to follow based on the desired installation route.

4. It would be preferable to add instructions for running `mrview` via containers on OSX and Windows prior to merge.

5. This is currently targeted for `master`. I don't think there should actually be any problem with either merging this to `master` prior to tagging, or with bundling it up with whatever hotfixes go into the next macro release, though I don't actually recall the result of discussion on the latter. Documentation may not cover all bases if it's included in a macro release, but that can nevertheless be revised on `master` if necessary. Note that #2197 will necessitate the construction of new container dependency downloads, but that targets `3.1.0`, so can be addressed on a separate branch on the `MRtrix3/containers` repo.

6. This PR does not yet address the prospect of automatic building of containers. Given the relative occurrence of updates, such automation is not necessarily a prerequisite for this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,100 @@
+ARG MAKE_JOBS="1"
+ARG DEBIAN_FRONTEND="noninteractive"
+
+FROM python:3.8-slim AS base
+FROM buildpack-deps:buster AS base-builder
+
+FROM base-builder AS mrtrix3-builder
+
+# Git commitish from which to build MRtrix3.
+ARG MRTRIX3_GIT_COMMITISH="master"
+# Command-line arguments for `./configure`
+ARG MRTRIX3_CONFIGURE_FLAGS=""
+# Command-line arguments for `./build`
+ARG MRTRIX3_BUILD_FLAGS="-persistent -nopaginate"
+
+RUN apt-get -qq update \
+    && apt-get install -yq --no-install-recommends \
+        libeigen3-dev \
+        libfftw3-dev \
+        libgl1-mesa-dev \
+        libpng-dev \
+        libqt5opengl5-dev \
+        libqt5svg5-dev \
+        libtiff5-dev \
+        qt5-default \
+        zlib1g-dev \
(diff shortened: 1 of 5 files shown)


jdtournier (dismissed):

Lestropie:
@MRtrix3/mrtrix3-devs: Could consider merging this into `3.0.3` but not immediately broadcasting publicly, and the code is then there in `master` that we can go through the process of setting up on DockerHub etc., get feedback from any sysadmins who try to use it, and then automate those processes via Actions to happen from the next update onwards.

jdtournier (approved):
Quite happy to merge this - it adds functionality without altering anything else, and it would be good to get feedback from container admins as early as possible. :+1:

dkp:
docker pull mrtrix/mrtrix:latest
Error response from daemon: pull access denied for mrtrix/mrtrix, repository does not exist or may require 'docker login': denied: **requested access to the resource is denied.**

PR #2280, opened 2021-02-07
README: Update to reflect revised dependencies

Lestropie:
Fixes divergence between online documentation / actual internal requirements and the contents of `README.md`. Though as noted in #2279, there is some question as to whether such instructions should actually still remain in the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ Support and general discussion is hosted on the [MRtrix3 Community Forum](http:/
 ## Quick install
 
 1. Install dependencies by whichever means your system uses. 
-   These include: Python (>=2.6), a C++ compiler with full C++11 support (`g++` 4.9 or later, `clang++`), 
-   Eigen (>=3.2.8), zlib, OpenGL (>=3.3), and Qt (>=4.8, or at least 5.1 on MacOSX).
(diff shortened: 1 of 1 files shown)


jdtournier (approved):
Agree with your changes in #2279, but happy for this to be patched up in the interim.

Lestropie:
> Agree with your changes in #2279, but happy for this to be patched up in the interim.

This might be slightly back-to-front given #2280 is to `dev` whereas #2279 is to `master`. The change here can still be applied to `dev` only given that that is where the software dependencies change, but will nevertheless put off merging in case this detail wasn't appreciated at the time.

jdtournier:
Yes, I had noticed, and I wasn't sure exactly what the rationale was, but I think I can see what you're suggesting. Version of dependencies has changed on `dev`, but not `master` - hence this PR. But #2279 suggests further changes to `README.md` to strip out install from source instructions which would make these changes redundant - but that was only a suggestion rather than already part of the PR, as far as I can tell? 

So I _think_ I get it. And I also agree we can remove the install from source instructions from the `README.md` in future version - we  can make it part of #2279 if you want, that would be fine by me.

PR #2285, opened 2021-02-10
fod2fixel: Wipe transform / spacing info from fixel data files

Lestropie:
As raised on [forum](https://community.mrtrix.org/t/a-warning-in-the-fixel-pipeline/4531).

I had in 61bfc64d prevented non-identity transform / voxel spacing info from being written to fixel data files, but only in the helper function `data_header_from_index()`, whereas `fod2fixel` does this same process manually.


jdtournier (approved):
Looks like you might need to update the testing data though...

Lestropie:
New test data generated by `mrtransform | mrconvert` to remove unwanted transform / voxel spacing info, rather than a complete re-generation.

PR #2286, opened 2021-02-10
dwifslpreproc: Compatibility between image padding and EddyQC

### LINKED ISSUE #1781: dwipreproc: Odd number of slices
Downstream of #1084 and #1169. Mostly the result of discussion in #1773 with @ansk and communication with `eddy_qc` developer. 

For an input image with an odd number of voxels specifically in the slice encoding direction, currently `dwipreproc` pads the image by duplicating and concatenating the last slice (as it does for the other two spatial axes), and also adds that slice to the vector of slice acquisition times, with the same time as the duplicated slice. This is to prevent a `topup` crash when it tries to naively downsample the image by a factor of 2. When this is done, in subsequent generation of the slspec file for eddy, the multiband group corresponding to the duplicated slice will have one more entry than all other groups.

It however seems as though having a different number of slices between different slice groups is considered strictly invalid. Though `eddy` currently successfully processes such data, in `eddy_quad` it leads to a NumPy error upon attempting to import these data. Therefore an alternative solution is required.

Options:

1. If `eddy` is happy to process such data, then `dwipreproc` could instead not only crop the padded slice for the sake of the output DWI having the same dimensionality as the input DWI, but also remove the excess slice from any and all other outputs from `eddy` prior to running `eddy_quad` (not only images but also text files). In reality this should be done anyway if such image padding is performed in the slice encoding axis and `-eddyqc_text` or `-eddyqc_all` are specified, independently of getting the `eddy_quad` command to execute.

2. Do not pad the images in this case; instead, use a modified `topup` config file that does not downsample the images during its optimisation iterations ([FSL thread](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1702&L=FSL&D=0&P=120897)). 

3. Delete `dwipreproc` from the repository, tell people to process their data the hard way, and restore a healthy sleeping pattern for myself.

Lestropie:
Closes #1781. Details in 004f884.

Opening as draft PR until it can be tested more thoroughly, particularly with S2V correction and an odd number of slices.


Lestropie:
Can confirm pre-existing `dwifslpreproc` tests succeed at 004f884.

jdtournier:
OK, minor issue when copying back `eddy_mask.nii` when folder contains spaces using problematic data with this PR:
```
mrconvert eddy_mask.nii /home/jdt13/data/Rob EddyQC test/mb3odd_PR_eddyQC/eddy_mask.nii -coord 2 0:80

mrconvert: [ERROR] Expected exactly 2 arguments (4 supplied)
mrconvert: [ERROR] Usage: mrconvert input output
mrconvert: [ERROR] Yours: mrconvert eddy_mask.nii /home/jdt13/data/Rob EddyQC test/mb3odd_PR_eddyQC/eddy_mask.nii
```
Issue [springs from this line](https://github.com/MRtrix3/mrtrix3/blob/004f884a4eca9c62c93d56a009787d86d59b9fcb/bin/dwifslpreproc#L1266) as far as I can tell. 

---

Otherwise, I'm running with FSL 6.0.4 and it looks like everything works out of the box - though the `topup` call takes 1h40 evey time... No issue running on the problematic data with `master` or the PR code (other than the white space issue above). But I also don't get either PDF or JSON files in the eddyQC output folder. This is the full listing I get running with `master` and FLS 6.0.4 on the non-problematic data:
```
total 5.7M
-rw-r--r-- 1 jdt13 perinatal 5.3M May  6 12:42 eddy_mask.nii
-rw-r--r-- 1 jdt13 perinatal 193K May  6 12:42 eddy_movement_over_time
-rw-r--r-- 1 jdt13 perinatal 2.7K May  6 12:42 eddy_movement_rms
-rw-r--r-- 1 jdt13 perinatal  17K May  6 12:42 eddy_outlier_map
-rw-r--r-- 1 jdt13 perinatal  67K May  6 12:42 eddy_outlier_n_sqr_stdev_map
-rw-r--r-- 1 jdt13 perinatal  67K May  6 12:42 eddy_outlier_n_stdev_map
-rw-r--r-- 1 jdt13 perinatal 3.7K May  6 12:42 eddy_outlier_report
-rw-r--r-- 1 jdt13 perinatal  24K May  6 12:42 eddy_parameters
-rw-r--r-- 1 jdt13 perinatal 1.5K May  6 12:42 eddy_post_eddy_shell_alignment_parameters
-rw-r--r-- 1 jdt13 perinatal  587 May  6 12:42 eddy_post_eddy_shell_PE_translation_parameters
-rw-r--r-- 1 jdt13 perinatal 2.8K May  6 12:42 eddy_restricted_movement_rms
```

I'll try again with FSL 6.0.3 and see whether that makes any difference...

Lestropie:
Concerning if EddyQC is not working even on the non-problematic data. Though it sounds like from your text that you've not gotten those data for *any* test? If that's the case it might be a system configuration issue. 

Also the example command included `-nocleanup`, so you should be able to go into scratch and see the contents of `eddy_quad_failure_output.txt`.

jdtournier:
OK, you're right, there was a configuration issue - makes a bit more sense now... I didn't realise I needed to run the post_install scripts - the FSL installer script won't work on Arch, so I had to reverse-engineer what's needed for install (didn't realise it'll install a whole new miniconda environment!). Seems to be sorted, at least I can see the `eddy_quad` command now, which was previously missing. I'm running the whole thing again with `6.0.3` - expect results in ~4 hours...

jdtournier:
Just figured out I could finally get the FSL installer to work by requesting the right version manually:
```
python2 fslinstaller.py --fslbuild=centos7_64
```
I'll have to make a note of that...

jdtournier:
In other news, it looks like the `-continue` option is borked, because the `dirstat` command is skipped, yet its results are used to validate the DW encoding... 
```
$ dwifslpreproc mb3odd_in.mif mb3odd_out_PR.mif -rpe_header -eddyqc_all mb3odd_PR_eddyQC -nocleanup -eddy_options " --mporder=6" -continue dwifslpreproc-tmp-GGTUFC/ dwi_post_eddy
dwifslpreproc: 
dwifslpreproc: Note that this script makes use of commands / algorithms that have relevant articles for citation; INCLUDING FROM EXTERNAL SOFTWARE PACKAGES. Please consult the help page (-help option) for more information.
dwifslpreproc: 
Skipping command: mrconvert '/home/jdt13/data/Rob EddyQC test/mb3odd_in.mif' '/home/jdt13/data/Rob EddyQC test/dwifslpreproc-tmp-GGTUFC/dwi.mif' -json_export '/home/jdt13/data/Rob EddyQC test/dwifslpreproc-tmp-GGTUFC/dwi.json'
dwifslpreproc: Changing to scratch directory (/home/jdt13/data/Rob EddyQC test/dwifslpreproc-tmp-GGTUFC)
dwifslpreproc: No slice encoding direction information present; assuming third axis corresponds to slices
Skipping command: dirstat dwi.mif -output asym

dwifslpreproc: [ERROR] Number of b-values reported by mrinfo (3) does not match number of outputs provided by dirstat (0)
dwifslpreproc: Scratch directory retained; location: /home/jdt13/data/Rob EddyQC test/dwifslpreproc-tmp-GGTUFC
```

Lestropie:
> This seems to stem from this line, which I've changed to:

Yep, that looks correct.

In retrospect I should have made the `image.Header()` function return a `namedtuple`, which would allow direct access to immutable member variables without going through accessor functions; but I'm slightly hesitant to change the API as of `3.0.0`.

> In other news, it looks like the `-continue` option is borked, because the dirstat command is skipped, yet its results are used to validate the DW encoding...

Yep, that's one of many situations where the simplicity of the current `-continue` operation limits things.

I've thought about dumping internal variables to a file in the scratch directory upon an exception and then re-loading them on use of `-continue`. But that might not resolve all scenarios; e.g. there's no guarantee that the contents of that variable at the point of the exception being raised is equivalent to its contents when that variable was read at any prior point in the script. Better would be to load the variables and then do some kind of "goto" to the specific line where the error occurred; but I don't see any way of doing that.

The only way to make this kind of functionality truly robust would be to have a fully-blown `nipype`-like system where the pipeline is broken into individual steps and their dependencies, and completion of each individual step is logged.

jdtournier:
OK, take 2, still no joy:
```
...
Command:  eddy_cuda --imain=eddy_in.nii --mask=eddy_mask.nii --acqp=eddy_config.txt --index=eddy_indices.txt --bvecs=bvecs --bvals=bvals --topup=field --mporder=6 --slspec=slspec.txt --out=dwi_post_eddy --verbose
dwifslpreproc: [  0%] Removing image padding prior to running EddyQC... 
dwifslpreproc: [ERROR] Unhandled Python exception:
dwifslpreproc: [ERROR]   AssertionError: 
dwifslpreproc: [ERROR] Traceback:
dwifslpreproc: [ERROR]   /home/jdt13/exp/mrtrix3/bin/dwifslpreproc:1016 (in execute())
dwifslpreproc: [ERROR]     assert all(len(group) == len(slice_groups[0]) for group in slice_groups[1:])
```
Not sure how to address that one...

---

> The only way to make this kind of functionality truly robust would be to have a fully-blown nipype-like system where the pipeline is broken into individual steps and their dependencies, and completion of each individual step is logged.

Yes, it might be worth re-thinking the `-continue` thing. Something to think about for `3.1`...

jdtournier:
OK, tried a few times, still get this failure:
```
dwifslpreproc: [ERROR] Unhandled Python exception:
dwifslpreproc: [ERROR]   AssertionError:
dwifslpreproc: [ERROR] Traceback:
dwifslpreproc: [ERROR]   /home/jdt13/exp/mrtrix3/bin/dwifslpreproc:1017 (in execute())
dwifslpreproc: [ERROR]     assert all(len(group) == len(slice_groups[0]) for group in slice_groups[1:])
```
Originating from[ line 1016](https://github.com/MRtrix3/mrtrix3/blob/dwifslpreproc_padding_eddyqc/bin/dwifslpreproc#L1016). 

Original command:
```
dwifslpreproc mb3odd_in.mif mb3odd_out_PR.mif -rpe_header -eddyqc_all mb3odd_PR_eddyQC -nocleanup -eddy_options " --mporder=6"
```

Had a go at printing out the `group` variable by adding `app.console (str(group))` at [line 1014](https://github.com/MRtrix3/mrtrix3/blob/dwifslpreproc_padding_eddyqc/bin/dwifslpreproc#L1014):
```
dwifslpreproc: [0, 27, 54]
dwifslpreproc: [1, 28, 55]
dwifslpreproc: [2, 29, 56]
dwifslpreproc: [3, 30, 57]
dwifslpreproc: [4, 31, 58]
dwifslpreproc: [5, 32, 59]
dwifslpreproc: [6, 33, 60]
dwifslpreproc: [7, 34, 61]
dwifslpreproc: [8, 35, 62]
dwifslpreproc: [9, 36, 63]
dwifslpreproc: [10, 37, 64]
dwifslpreproc: [11, 38, 65]
dwifslpreproc: [12, 39, 66]
dwifslpreproc: [13, 40, 67]
dwifslpreproc: [14, 41, 68]
dwifslpreproc: [15, 42, 69]
dwifslpreproc: [16, 43, 70]
dwifslpreproc: [17, 44, 71]
dwifslpreproc: [18, 45, 72]
dwifslpreproc: [19, 46, 73]
dwifslpreproc: [20, 47, 74]
dwifslpreproc: [21, 48, 75]
dwifslpreproc: [22, 49, 76]
dwifslpreproc: [23, 50, 77]
dwifslpreproc: [24, 51, 78]
dwifslpreproc: [25, 52, 79]
dwifslpreproc: [26, 53, 80, 81]
```
Not too sure why that last line didn't get truncated. Running again with a few more debugging statements, hopefully I'll have an answer tomorrow...

Lestropie:
> Not too sure why that last line didn't get truncated.

Python :facepalm: Try with bf3515e?

jdtournier:
OK, it looks like that last commit fixes it for me... :+1:

jdtournier (approved):

Lestropie (commented):
Was doubting 64158a1, but it is indeed necessary specifically because of [this line](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/bin/dwifslpreproc#L768): variable `dwi_num_slices` is incremented to reflect the padded slice (specifically to better support [slice encoding direction reversal](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/bin/dwifslpreproc#L882-L883)), and therefore the index of the padded slice is *not* equivalent to the content of that variable.

PR #2287, opened 2021-02-13
Fix script tests on master

Lestropie:
Script tests had a couple of residual issues when `3.0.0` was merged (nothing problematic from a user perspective, just test failures). These are fixed on `dev` (both in this code repo and in the corresponding branch of the `script_test_data` branch). Ahead of `3.0.3` I would prefer for these fixes to be propagated, e.g. for testing containers in #2279.


Lestropie:
Promoted to `3.0.3` milestone: testing discrepancies turned out to be much easier to resolve than expected. Confirmation of successful tests pending.

Lestropie:
Confirm all current script tests passing (b19fceb4; script test data at b89990f5), *as long as* the fix in #2318 is included.

Lestropie:
Confirm `dwifslpreproc` tests failing due to #2308 as reported in #2348.

The problem at least in part stems from the differences in code [here](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/bin/dwifslpreproc#L786) and [here](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/bin/dwifslpreproc#L800-L806): generation of the phase encoding table for running `topup` is based on having manipulated the strides to LAS in the exported NIfTI, whereas the extraction of volumes with specific phase encoding directions for running `applytopup` (in order to produce a mask for `eddy` in "undistorted space") is done based on the `.mif` where no permutation has taken place. 

What I'm not sure about is whether or not this is the *only* problem that #2308 will introduce into `dwifslpreproc`...

Lestropie:
Tests pass following 77f43472; but I want to check the scratch contents before & after and make sure the mask derivation is doing as it should.

Lestropie:
Just needs a review.

Also need to remember to update the `master` branch pointer in the `script_test_data` repo upon merge.

jdtournier (approved):

PR #2154, opened 2020-09-03
CI: Perform limited script testing

Lestropie:
Proposed alternative to #2143.

In #2143, blacklisting the entire test list of specific scripts with external dependencies, but executing the complete test list of all other scripts, results in a long execution time. This alternative cherry-picks the set of specific tests to be executed as part of CI for merges to both `dev` and `master`. Comprehensive testing of all scripts, including those with external dependencies, is deferred to #2134.

PR #2288, opened 2021-02-13
dwishellmath: Compatibility with one unique b-value

Lestropie:
Spotted while following up on #2154. `dwi2response` tests for algorithms that pre-select only the highest shell and no *b*=0 data would internally run `dwi2mask legacy`, which in turn would invoke `dwishellmath`, within which `mrcat` would fail due to being provided with only one input image. While the fix is required there specifically because of the new Python version of `dwi2mask` (#2197), it is nevertheless a limitation that can lead to an unhandled Exception on `master`, hence proposing to merge there.


dchristiaens (approved):
Makes sense, happy with this.

Issue #2289, opened 2021-02-23
fixelcfestats: abnormal termination when doing t-tests + f-test


diagiraldo:
**Describe the bug**
I am trying to run the `fixelcfestats` to perform hypothesis testing for multiple contrasts + an f-test, but the process terminates abruptly while calculating the basic properties of the default permutation, even when I use the `-notest` flag:

```
fixelcfestats -force smoothed_fixel_data list_of_files.txt design_matrix.txt contrasts.txt wm_fixel_connectivity stats_results -mask fixel_mask.mif -exchange_within blocks.txt -variance blocks.txt -ftests ftests.txt -notest
fixelcfestats: [WARNING] existing output files will be overwritten
fixelcfestats: Number of fixels in template: 574477
fixelcfestats: Number of fixels in mask: 498510
fixelcfestats: [done] Configuring data import from files listed in "list_of_files.txt" as found relative to directory "smoothed_fixel_data"
fixelcfestats: Number of inputs: 8
fixelcfestats: Number of factors: 6
fixelcfestats: Design matrix condition number: 14.3611
fixelcfestats: Number of variance groups: 2
fixelcfestats: Number of hypotheses: 5
fixelcfestats: [WARNING] A total of 15032 fixels in the provided mask do not possess any streamlines-based connectivity; these will not be enhanced by CFE, and hence cannot be tested for statistical significance
fixelcfestats: [100%] Loading fixel data (no smoothing)
fixelcfestats: [  6%] Calculating basic properties of default permutation...double free or corruption (out)
Aborted (core dumped)
```
It should be mentioned I am including two different measures in the testing so I have 2 exchangeability blocks and variance blocks while the design matrix also has 2 blocks, however It works without problem when:
- I add the `-fonly` flag -> Only F-test
- I don't add the `-ftests`  option -> Only t-tests 

**To Reproduce**

Data I used to reproduce the error: https://drive.google.com/file/d/1U3rIOQ9T8DvVXH3wCNycKuohx4pM_VLD/view?usp=sharing 

**Platform/Environment/Version**

-  OS: Ubuntu 18.04.4 LTS

-  *MRtrix3* version: 3.0.2-246-g861c004f

**Advanced debugging information**

This is what happens when debugging:
```
[...]

PR #2294, opened 2021-03-03
GLM: Fix all_stats() segfault

Lestropie:
As reported in #2289.

Fixes segmentation fault in case where:

1. Either:
    1. Input data contain non-finite values (regardless of whether or not those data are or are not within the statistical inference processing mask);
    2. Element-wise design matrix columns are added via `-column` option;
2. At least one F-test is performed.

Fairly obvious blunder even without explaining the surrounding code.

### DIFF
--- a/core/math/stats/glm.cpp
+++ b/core/math/stats/glm.cpp
@@ -471,7 +471,7 @@ namespace MR
                 local_stdev = matrix_type::Zero (num_vgs, 1);
                 for (size_t ih = 0; ih != hypotheses.size(); ++ih) {
                   if (hypotheses[ih].is_F())
-                    local_abs_effect_size (1, ih) = local_std_effect_size (1, ih) = NaN;
+                    local_abs_effect_size (0, ih) = local_std_effect_size (0, ih) = NaN;
                 }
               }
           };


jdtournier (approved):
I'm not sure I fully appreciate what each row of your `local_abs_effect_size` and `local_std_effect_size` is supposed to contain, but I'll take your word for it that it's appropriate to set the first row to NaN rather than the second... :grin:

Lestropie:
> I'm not sure I fully appreciate what each row of your local_abs_effect_size and local_std_effect_size is supposed to contain ...

Normally rows correspond to elements (e.g. fixels). In cases where the GLM can't be fit to the exhaustive set of data in a single matrix operation, the relevant code [loops over elements](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/core/math/stats/glm.cpp#L476-L479), each thread processes one element at a time. So any GLM derivative that would normally have one row per element becomes a row vector in that context. But they're treated as matrices with one row rather than explicit vectors so that the [same function](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/core/math/stats/glm.cpp#L437-L438) can be used as that used when all data are solved at once.

PR #2296, opened 2021-03-05
Support for Qt6

bjeurissen:
Not sure if this should go to master.

Today homebrew upgraded my qt installation to qt6, so I tried rebuilding mrtrix3 against it, which did not work out-of-the-box.

I believe that all of the changes here are backward compatible with Qt5. Most of the things I had to change were things that were already deprecated in Qt5.

I tested this on:
* MacOS Big Sur (arm64) with Qt 6.0.2
* Ubuntu 20.04.2 LTS (x86_64) with Qt 5.12.8

One caveat is the ``configure`` script:
- for Qt6 it needs to add module ``openglwidgets`` to the project file (as in this branch)
- for Qt5 it needs to add module ``opengl`` to the project file (as in master)
but we detect the Qt-version by building a binary using this project file...

What would be the preferred solution here?

### DIFF
--- a/configure
+++ b/configure
@@ -1251,7 +1251,7 @@ int main() { Foo f; }
       filetext = 'CONFIG += c++11'
       if debug:
         filetext += ' debug'
-      filetext += '\nQT += core gui opengl svg network\n'
+      filetext += '\nQT += core gui openglwidgets svg network\n'
       filetext += 'HEADERS += qt.h\nSOURCES += qt.cpp\n'
       if system == "darwin":
         filetext += 'QMAKE_MACOSX_DEPLOYMENT_TARGET = ' + macosx_version + '\n'
--- a/configure
+++ b/configure
@@ -1251,7 +1251,7 @@ int main() { Foo f; }
       filetext = 'CONFIG += c++11'
       if debug:
         filetext += ' debug'
-      filetext += '\nQT += core gui openglwidgets svg network\n'
+      filetext += '\nQT += core gui opengl svg network\n'
       filetext += 'HEADERS += qt.h\nSOURCES += qt.cpp\n'
       if system == "darwin":
         filetext += 'QMAKE_MACOSX_DEPLOYMENT_TARGET = ' + macosx_version + '\n'
--- a/core/algo/loop.h
+++ b/core/algo/loop.h
@@ -194,9 +194,9 @@ namespace MR
         const std::tuple<ImageType&...> vox;
         const ssize_t size0;
         FORCE_INLINE Run (const size_t axis, const std::tuple<ImageType&...>& vox) :
-          axis (axis), vox (vox), size0 (std::get<0>(vox).size(axis)) { apply (set_pos (axis, 0), vox); }
+          axis (axis), vox (vox), size0 (std::get<0>(vox).size(axis)) { MR::apply (set_pos (axis, 0), vox); }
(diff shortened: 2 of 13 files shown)


bjeurissen:
Not convinced that this will work with older Qt versions.

The issue appears to be that:
-QGLWidget no longer exists in Qt6, but is fully replaced with QOpenGLWidget
-Both QGLWidget and QOpenGLWidget are available in Qt>=5.4, and we seem to be currently mixing both.

whereas in this branch I have replaced all instances of QGLWidget by QOpenGLWidget, so this will probably only work in Qt>=5.4 and will break both Qt4 and older Qt5 versions.

jdtournier:
Thanks Ben! It'll take me a while to get through that... 

Just one comment though:
> -Both QGLWidget and QOpenGLWidget are available in Qt>=5.4, and we seem to be currently mixing both.

I'm pretty sure we're using one or the other, depending on the Qt version - but we shouldn't be _mixing_ both at the same time, right...?!? There's a version check based on `QT_VERSION` (e.g. [this line](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/opengl/gl.h#L68)), and after that it _should_ all be based on one version or the other. Given it switches to `QOpenGLWidget` if Qt >= 5.4, I would have thought it would cover Qt6 already...

bjeurissen:
> > -Both QGLWidget and QOpenGLWidget are available in Qt>=5.4, and we seem to be currently mixing both.
> 
> I'm pretty sure we're using one or the other, depending on the Qt version - but we shouldn't be _mixing_ both at the same time, right...?!? There's a version check based on `QT_VERSION` (e.g. [this line](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/opengl/gl.h#L68)), and after that it _should_ all be based on one version or the other. Given it switches to `QOpenGLWidget` if Qt >= 5.4, I would have thought it would cover Qt6 already...

I think ``src/gui/opengl/gl.h`` is indeed fine to that respect, but there were some other files that were trying to include include and use QGLWidget regardless of the Qt version (see the list of Files changed). This worked fine in Qt5 because QGLWidget was still there (but deprecated), but fails on Qt6 where QGLWidget is removed.

bjeurissen:
Noticed that the Tool widgets don't work properly when docked when using Qt6. You can see them being rendered for a split second, but then the entire widget becomes empty and no interaction whatsoever is possible (e.g. undocking or changing between different Tools, by clicking the vertical tabs). Everything works fine when using mrview with ``MRViewDockFloating: 1``, but as soon as you dock the widget, it becomes empty and one cannot interact with it.

Lestropie:
Have not followed along here or even looked at Qt6, but want to explicitly link to #2157, and make sure it's recognised that those changes were made on `dev` whereas this PR is currently targeting `master`. The existing changes on `master` are only to the *documentation* (#2153).

While there's a strong argument for adding support for Qt6 as a hotfix on `master`, if embedded within that is an increase in minimum version for Qt5, that probably shouldn't go directly to `master`; though if it were to change immediately prior to a macro version increment it'd be fine.

bjeurissen:
Thanks Rob, I did not realize that the 5.5 requirement was only on master in the documentation. I think it makes a lot of sense to move Qt6 support to dev given the overlap with the Qt5 requirement work.

bjeurissen:
Moved here: #2297

PR #2298, opened 2021-03-09
mtnormalise: add resilience to non-finite values in mask

jdtournier:
Addresses [issue raised on the forum](https://community.mrtrix.org/t/mtnormalise-nan-values-as-balance-factors/4552/3?u=jdtournier), caused by presence of non-finite values within data voxels included in the analysis mask. Was previously relying on outlier rejection to deal with this, but this can't work since Inf*0 = NaN...


Lestropie:
So to be precise, if any individual ODF image contains a non-finite value in a voxel within the user-specified mask, the magnitude of that particular ODF in that particular voxel will instead be interpreted as 0. While this will allow execution, it's maybe not the ideal solution. Perhaps preferable would be for:
- The initial mask in the first iteration to be the intersection between the user-specified mask and the set of voxels for which all ODFs are finite;
- A warning to be issued to the user if the mask they provided contains any non-finite values in any ODF.

This could be pursued on `dev` independently of a quick fix to `master`?

thijsdhollander:
Hi @jdtournier, it might be useful as part of this to update the reference with our recent abstract ( https://www.researchgate.net/publication/350248989_Multi-tissue_log-domain_intensity_and_inhomogeneity_normalisation_for_quantitative_apparent_fibre_density ), so users have up to date access to the description and performance evaluation of the current algorithm.  This wouldn't affect behaviour of course, so is a safe change. Thanks, Thijs.

jdtournier:
Sure, will do.

jdtournier:
> ... if any individual ODF image contains a non-finite value in a voxel within the user-specified mask, the magnitude of that particular ODF in that particular voxel will instead be interpreted as 0

OK, I had a good look into addressing that comment, and I'm pretty sure 1fabe21 does it. Basically, non-finite will still be loaded as-is (not set to zero), but will inherently be sent to the bottom (for `-Inf` & `NaN`) or bottom (for `Inf`) of the range, and be excluded by the outlier rejection (have its weight set to zero). The code then has to handle the outliers properly, since multiplying by zero is not sufficient to get rid of non-finite values - the changes explicitly sets these values to zero rather than multiplying by zero, which then ensures the balance factors are computed correctly (this is what was previously failing). 

Finally, I've added a warning if any non-finite values are detected within the mask. I've also verified that the command produces identical results to what it did before. This remains the case even if a non-finite value is included in the mask, but was previously excluded as an outlier (so the final mask is the same as before anyway). If I introduce a NaN in a voxel that would otherwise have been included in the final mask, the command runs as expected, with changes commensurate with the omission of that voxel (changes of the order of around 1e-8 or so in that case). 

So I think this one is good to go...

Lestropie (approved):
Given the way in which it is now used, i.e. not directly multiplicative, "`weights`" isn't really a great variable name; it's a mask that imposes the outlier rejection & non-finite value masking on top of the pre-defined user mask. It's fine, but would be worth revising if the command is modified further in the future, e.g. using a two-pass approach to pre-exclude voxels with non-finite values.

PR #2311, opened 2021-04-21
DICOM: simplify handling of parsing in context of multi-frame

jdtournier:
Changes to multiframe handling prompted by discussion on forum:
https://community.mrtrix.org/t/mrconvert-directions-in-0018-9089-not-detected/

The idea is to parse all tags regardless of which level in the hierarchy they appear in, rather than only parsing when they're top-level or within a Functional Group Sequence. Instead, we just need to ignore the IconImageSequence, which contains tags that can conflict with the main image; indeed, this seems to be the only reason the parsing worked as it did so far.

All DICOM tests pass fine with these changes.

Setting this up as a draft request for now as any changes to DICOM handling require quite a bit of double-checking...


Lestropie:
Opened a bunch of random DICOMs across my file system and didn't find an issue; outstanding question is whether or not you're concerned about changes in the parsing of key-values, and whether or not you want to check for that in your tests.

jdtournier:
> outstanding question is whether or not you're concerned about changes in the parsing of key-values, and whether or not you want to check for that in your tests.

Working on that as we speak. Lots of differences obviously, especially in the DW scheme as we're handling things slightly differently than previously, all of which requires careful double-checking... I'll hopefully have all this worked out by COB today. I'll be committing this to #2346 though. I'll merge this into #2346 so we can test everything in one go before merging.

PR #2313, opened 2021-05-06
dwifslpreproc: Handle non-ascii data in eddy error outputs

Lestropie:
The FSL `eddy` commands sometimes contain a small block of binary data at the end of their outputs. If, when attempting to write those data to a text file in the scratch directory, an encoding operation occurs, such data may yield a `UnicodeEncodeError`. This commit changes the output text files to be opened in binary mode, so such an encoding should no longer be applied.

(Note that `__str__()` is [defined](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/lib/mrtrix3/run.py#L192-L193) for `MRtrixCmdError` as a straight concatenation of `stdout` and `stderr` contents, so there shouldn't be any `ascii` conversion happening there)

### DIFF
--- a/bin/dwifslpreproc
+++ b/bin/dwifslpreproc
@@ -912,13 +912,13 @@ def execute(): #pylint: disable=unused-variable
     except run.MRtrixCmdError as exception_cuda:
       if not eddy_openmp_cmd:
         raise
-      with open('eddy_cuda_failure_output.txt', 'w') as eddy_output_file:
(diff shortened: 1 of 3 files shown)


jdtournier (dismissed):
Seems fairly innocuous to me. My experience is that using non-binary mode on text files eventually creates problems anyway, so quite happy with the suggested change.

jdtournier (approved):

PR #2317, opened 2021-05-12
Compatibility with Pylint 2.8

Lestropie:
Address new refactoring message "`consider-using-with`".

See: http://pylint.pycqa.org/en/latest/whatsnew/2.8.html


build
```diff
@@ -942,7 +933,7 @@ def execute (message, cmd, working_dir=None):
 
   try:
     start = timer()
-    process = subprocess.Popen (cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=environ, cwd=working_dir)
+    process = subprocess.Popen (cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=environ, cwd=working_dir) #pylint: disable=consider-using-with
```
Lestropie, line 936: Performing this operation with context management in Python2 resulted in compilation error messages not being written to the `error_text` variable, and hence no compilation debugging information printed. However in Python3 it worked for me. So should consider modifying the code here for `3.1.0` where Python2 support is dropped.

build
```diff
@@ -195,62 +195,54 @@ def colorize(s):
...
-    fid = os.fdopen (fid, 'wb')
-    fid.write (colorize(error_stream).encode (errors='ignore'))
-    fid.close()
-    os.system ("less -RX " + name)
-    os.unlink (name)
-
+    with tempfile.NamedTemporaryFile() as tf:
+      tf.write (colorize(error_stream).encode (errors='ignore'))
+      tf.flush()
+      os.system ("less -RX " + tf.name)
```
Lestropie, line 201: I've a sneaking suspicion that this *won't work* on Windows; requires testing.

Lestropie:
It seems that context manager support for `subprocess.Popen` was only added in Python 3.2. So for `master` I will need to avoid such, and I'll do further revision when merging to `dev`.

build
Lestropie: Compiler errors appear fine on MSYS2.

jdtournier (approved):

PR #2321, opened 2021-05-17
dwigradcheck: explain how to obtain corrected gradient table

bjeurissen:
Some users pointed out to me that it was not obvious how to obtain the corrected gradient table with dwigradcheck.

### DIFF
--- a/bin/dwigradcheck
+++ b/bin/dwigradcheck
@@ -23,6 +23,7 @@ def usage(cmdline): #pylint: disable=unused-variable
   from mrtrix3 import app #pylint: disable=no-name-in-module, import-outside-toplevel
   cmdline.set_author('Robert E. Smith (robert.smith@florey.edu.au)')
   cmdline.set_synopsis('Check the orientation of the diffusion gradient table')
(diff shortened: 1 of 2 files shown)


bin/dwigradcheck
```diff
@@ -23,7 +23,8 @@ def usage(cmdline): #pylint: disable=unused-variable
   from mrtrix3 import app, _version #pylint: disable=no-name-in-module, import-outside-toplevel
   cmdline.set_author('Robert E. Smith (robert.smith@florey.edu.au)')
   cmdline.set_synopsis('Check the orientation of the diffusion gradient table')
-  cmdline.add_description('Note that if the -mask command-line option is not specified, the MRtrix3 command dwi2mask will automatically be called to '
+  cmdline.add_description('Note that the corrected gradient table can be output using the -export_grad_{mrtrix,fsl} option. \n\n'
```
jdtournier, line 26: I think if you need to have several paragraphs in the description, the right way to do it is to provide each paragraph within its own `cmdline.add_description()` call - not explicitly hard-code the break as two newlines.

bjeurissen: Sorry... Thanks for fixing!

jdtournier (approved):

PR #2325, opened 2021-05-21
Tckgen i fod1 fix max angle at seed

jdtournier:
Just got a report that when using iFOD1, the angle between track segments can be up to twice the stated max angle at the seed point. This was due to an interaction between iFOD1 and the way the track reversal is handled, which this PR hopefully fixes. More detail in commit 5237d466c.

Note that I don't think this affects iFOD2 or other algorithms, though it would be worth double-checking, maybe by adding some tests to verify correct behaviour.


Lestropie:
Superseded by #3335.

Issue #2231, opened 2020-11-20
b-value scaling (again)


dchristiaens:
My apologies in advance to bring the b-value scaling up once again... 😉 

I came across a GE dataset where interspersed b-zero volumes are encoded like this:
```
$ mrinfo dwi.mif -dwgrad
...
   0.0475479964418854    -0.752325943629832    -0.657072950726055             2000.0003
    0.124123021599946     0.977383170399578     0.171218029799926           1999.999303
                    0                     0                     0                  2000
    0.603679064182912      0.28296303009199    -0.745321079178902           1999.999575
    0.422664756213416    -0.886664488528144     0.187564891805954           2000.002307
...
```
It appears that these volumes are not detected as b-zero scans. Since I can't be bothered reading through the lengthy discussions in the past threads on the subject, I just want to ask if there is any reason why this line should _not_ be interpreted as a b-zero volume?

The problem, I think, stems from these lines in the recent modifications:
https://github.com/MRtrix3/mrtrix3/blob/c4f5e0c995656cde8ea005c1a2a5105011c808e5/core/dwi/gradient.cpp#L259-L260
`squared_norms[row]` need to stay 0 to scale the b-value to 0 too. My suggestion would be to simply delete these lines. Any objections?

jdtournier:
My one concern is the handling of Philips' isotropic-weighted volume that they sometimes include in the main series - [came up recently on the forum](https://community.mrtrix.org/t/dwi2response-error-iter0-fixel-peaks-mif/4010/4?u=jdtournier). I just had a look into our DICOM handling, and they're definitely included in the main series - there's nothing to really tell them apart from the other volumes in the series. But the only real option to deal with these is to remove them altogether, they simply shouldn't be included in the first place. With your changes, they'd be interpreted as b=0 volumes. Currently, they're stuff things up in other ways. Either way, they cause trouble, so that's actually not an argument against the change you're proposing...

So to cut a long story short: I think your changes are the only way to properly handle a single-shell scheme with interspersed b=0 volumes produced on systems that won't otherwise allow it (same as multi-shell handling, really). I'm guessing this change was added so that zero vectors wouldn't affect the automatic detection of whether b-value scaling should be applied, with the assumption that zero vectors won't come with a non-zero b-value? Either way, I reckon the correct fix here is to change that line 260 to:
```
squared_norms[row] = grad(row,3) > BZeroThreshold ? 0.0 : 1.0; 
```
[...]

PR #2329, opened 2021-06-01
preserve b-value scaling for b-zero #2231

dchristiaens:
Please see #2231 for discussion. Do I have permission to close this off?


Lestropie:
It seems to me as though the line [`0 0 0 >BZeroThreshold`] is fundamentally ambiguous. It could be a post-acquisition reconstructed mean intensity within a shell, or it could be a *b*=0. Perhaps the former could only possibly apply if that volume is the last in a series? So perhaps:

1. Incorporate this proposed change, so that volumes with [`0 0 0`] direction are rescaled to *b*=0;
2. If such a rescaling occurs for the last volume in a series, issue a warning level message that the data need to be checked in case this is in fact a trace-weighted image that needs to be removed via `mrconvert -coord 3`.

?

dchristiaens:
> 2\. If such a rescaling occurs for the last volume in a series, issue a warning level message that the data need to be checked in case this is in fact a trace-weighted image that needs to be removed

I am hesitant about any sort of behaviour that depends on a particular ordering in the series. If you think a warning message is needed, I would prefer to issue one for every dataset that contains at least one ambiguous gradient table entry.

Lestropie:
> If you think a warning message is needed, I would prefer to issue one for every dataset that contains at least one ambiguous gradient table entry.

That was actually my first thought, but I pulled it back for some reason; happy to go with that. Perhaps just the content of the warning message itself could be changed if it's the final volume that's identified; i.e. there's a stronger hypothesis that interpretation of that volume as a b=0 is likely to be erroneous.

jdtournier:
Happy enough with this, but we need to figure out why tests currently fail for `dwi2tensor` and `tckgen`. Might be that our test data are actually affected by the very issue this is dealing with...?

jdtournier (dismissed):

jdtournier (approved):

PR #2333, opened 2021-06-16
mrview: accept drop events

maxpietsch:
File drop into mrview works currently but the gesture appears to be rejected. This change accepts the drop gesture for opening new main images and for fixels, overlays, rois, tractograms.


jdtournier:
Why not merge to `master`? Seems like a bug fix to me...

Issue #937, opened 2017-03-16
mrview gui / tools: responsive (elided) text


maxpietsch:
Currently, we shorten labels (filenames) in tools to sometimes unusable length and resizing the toolbar does not change the text size. I am thinking for instance about tractograms with long descriptive filenames.

Qt supports automatic text [shortening](https://doc.qt.io/qt-5/qfontmetricsf.html#elidedText) which combined with the tool widget's [resize event](https://doc.qt.io/qt-5/qwidget.html#resizeEvent) should make the text responsive. I am not familiar enough with Qt and mrview's code but here are two examples for widgets and docks that make it look simple:

https://stackoverflow.com/questions/7381100/text-overflow-for-a-qlabel-s-text-rendering-in-qt
https://stackoverflow.com/questions/15331854/qt-resize-event-for-docks

ansk:
I would **very** appreciate this feature. When loading multiple tracktograms/fixels in FBA I am sometimes losing awarenes which file is which and I did not find a way to learn that once the file is loaded.
Other option (maybe easier to implement since it is also used in `mrview`?) would be to add popup (or whatever this GUI object is called) of full path when focusing the mouse pointer on the relevant filename/radio button in the toolbar (similarly as it is now implemented when user focus mouse pointer on the icon in main upper toolbar - File, Image, etc.).
Antonin

thijsdhollander:
@ansk : you're referring to "tool tips", I suppose.  I've had this request from many other people as well indeed.  It's not even just long filenames that cause this for them, but even just long paths (which are quite common in a complex analysis, where all files sit in deeply nested complex folder structures).  The only unique thing per file loaded would in fact be the file name, but the text in the tool is cut off at a point where you can just see the start of the path; so everything looks the same.

I'm not sure any of us has the time to look into this quickly/soon, but I agree it's indeed a heavily requested feature.

maxpietsch:
Thematically related: Currently, there is no way of showing the full filename of the main image if the scree size is too small to expand mrview's main window title and the image names in the image menu are relatively short and fixed-length. 

Overlay image names in the image pane are not eladed, possibly hiding the overlay value:
![image](https://user-images.githubusercontent.com/10046944/122211595-3345ba00-cea7-11eb-8402-5865266112da.png)

PR #2334, opened 2021-06-16
mrview: elide ODF file names 

maxpietsch:
Currently, ODF path and file names are truncated and fixed length:
image

Eliding text and responsive widget width (this PR):
Screenshot 2021-06-16 at 12 15 19

I tried scrolling full file names (only modify `src/gui/mrview/tool/odf/model.h`):
but the scroll bar is not shown and it's located left on load:
image horizontal scroll:
Screenshot 2021-06-16 at 12 17 15

So I'd go with the sliding option. This is at least a partial fix for #937.


jdtournier (approved):

jdtournier:
Adding this to 3.0.3 since it's a trivial change, won't affect processing, and works fine as far as I can tell...

Issue #2322, opened 2021-05-19
ENH dwidenoise: Warning on few directions


Lestropie:
It's currently possible to run `dwidenoise` on DWI data containing a tiny number of directions. I expect that determination of the MP threshold will be highly discretised in this case, which may make the influence of denoising quite inconsistent between adjacent voxels. Perhaps if the number of volumes is smaller than what would result in use of a 5x5x5 kernel, i.e. fewer than 27, such that a 3x3x3 kernel is utilized, this should be stated at a warning level, along with the potential consequences of such. This would also catch people trying to denoise individual volumes, which probably happens.

dchristiaens:
Good suggestion, although a cutoff at 27 would catch a lot of old DTI data. The commit triggers a warning on a matrix rank of less than 15, but I'm fully prepared to admit that this is more of a gut feeling and not at all the result of a thorough evaluation.

> This would also catch people trying to denoise individual volumes, which probably happens.

This would already trigger an error in `dwidenoise`.

This change has been committed to the branch of PR #2340.

PR #2340, opened 2021-06-25
Denoising updates for 3.1.0

dchristiaens:
The PR adds the option to output the selected denoising rank in each patch as an image.


dchristiaens:
Also add a warning on a small matrix rank in the denoising, as suggested in #2322.

jdtournier (approved):

PR #2346, opened 2021-06-29
DICOM improvements

jdtournier:
Initially prompted by very long compile times for the DICOM dictionary, but testing showed a few other issues that needed to be dealt with. 

With 3b6ba30, compile time for `core/file/dicom/dict.cpp` went from about 1½ minutes down to ~3s. 

But testing on a copy of my DICOM testing repo stored on a different system came up with a number of failures, mostly down to different ordering of the different patients & studies when scanning a large DICOM folder. Basically, the previous handling ordered patients and studies as they appeared on the file system when traversing the directory tree, which won't be reproducible. Commit 130f83c makes sure patients and studies are properly and reproducibly sorted, so that the indices entered when selecting DICOM series aren't affected by the order of traversal of the filesystem. 

Finally, it turns out there was one failing series with two echoes, with identical attributes in all fields that we rely on for sorting frames. So that meant that slices ended up randomly assigned to the first or second echo - again depending on the order of traversal of the filesystem. Commit 6500c00 now takes the echo number and echo time into account when sorting. 

With these changes, we get faster compile times (and presumably faster runtimes too), and the testing matches across the different systems I ran it on - as it should.


Lestropie:
- Consider regenerating 3b6ba30f using @MRtrixBot 

- I'd hoped that this would resolve the issue I've been having, being that when opening a multi-echo fMRI series the three echoes are not guaranteed to be presented in sorted order. But that appears to still be present. Have never had the slices-interspersed-between-frames issue you describe, but I still get:
    ```
    [rob@BRI-Pub-122 AEP]$ mrinfo DICOM_in/007001/func_task-pseudoword_ses-01/
    mrinfo: [done] scanning DICOM folder "DICOM_in/0...unc_task-pseudoword_ses-01/"
    Select series ('q' to abort):
       0 -  221 MR images 14:02:17 func_align-AxObl_dir-AP_task-pseudoword_ses-01 (epfid2d3_72) [15] ORIGINAL PRIMARY M MB TE1 ND MOSAIC
       1 -  221 MR images 14:02:17 func_align-AxObl_dir-AP_task-pseudoword_ses-01 (epfid2d3_72) [15] ORIGINAL PRIMARY M MB TE3 ND MOSAIC
       2 -  221 MR images 14:02:17 func_align-AxObl_dir-AP_task-pseudoword_ses-01 (epfid2d3_72) [15] ORIGINAL PRIMARY M MB TE2 ND MOSAIC
    ```

jdtournier:
OK, I've now added full header checks for my DICOM testing, which required a few changes to the code that does the header key/value checks (ec2c2da). Also merged #2311 in here which allows the checks to run on all proposed DICOM changes. I've had to update my DICOM test data to match, after verifying all the differences between my previously converted data and the data produced with this PR. Basically, the new versions either contains fields that weren't recorded before; have slight differences in the rounding for the DW scheme (due to the unit normalisation that now happens by default); or have differences in the comments field where we now record the ImageType array. Everything checks out, though it did raise a few issues with some (very) badly formatted Philips data (see 8c86d6b for details). 

As far as I can tell, everything now checks out... :sweat_smile:

jdtournier:
Spoke too soon... I think that last commit gets rid of the last inconsistency.

Would appreciate someone giving this the thumbs-up!

bjeurissen (approved):

Issue #2349, opened 2021-07-14
mrview: incorrect current volume index displayed for indices > 99


julienbesle:
When displaying a 4D overlay file and setting the volume index to a number larger than 99, the current volume index is not retained correctly when clicking away from the overlay (or selecting another overlay), and then selecting again the original overlay.

To reproduce:
- load a 4D file with more than 100 volumes as an overlay, and increase the volume index to a number > 99 (e.g. 104) using the volume index field at the bottom right. The volumes will display correctly, including for volumes > 99. 
- Click anywhere below the overlay name (in the empty overlay list) or select another, previously loaded, overlay name, so that the volume index field is either greyed out or shows the volume index of the other overlay.
- select the original overlay: the index will be set to 99, instead of 104 as expected, but the displayed slice will correspond to the 105th volume  (index = 104). 
- Incrementing or decrementing the index will go to volume 100 or 98 and the corresponding volume will be displayed (instead of the 104th or 106th as expected)

-  OS: Ubuntu 16.04
-  *MRtrix3* version 3.02-193-gd63cc20
- Qt version: 4.8.7

Cheers
Julien

jdtournier:
Good catch. I think this is because when we recreate the volume spin boxes after an image switch, the [constructor for the spin box](https://doc.qt.io/qt-5/qspinbox.html#QSpinBox) defaults to min / max = 0 / 99. We do [set the min / max appropriately](https://github.com/MRtrix3/mrtrix3/blob/master/src/gui/mrview/tool/overlay.cpp#L666), but the max was being set _after_ we set the volume index back to where it was. I think the fix here is simply to swap the order around so the value is set after the max is set - which is what I've just pushed to the `mrview_fix_overlay_volume_selection` branch. Could you give that a shot and see if it fixes it for you?

julienbesle:
This fixed it. Thanks!

jdtournier:
Great, thanks for confirming. Will merge shortly. :+1:

PR #2352, opened 2021-07-15
mrview: fix volume index reset after image switch

jdtournier:
Addresses #2349


Lestropie (approved):
Looks innocent enough (#famouslastwords), and fix confirmed by reporter of issue.

PR #2355, opened 2021-07-16
Release 3.0.3

jdtournier:
Final PR before release, just to make sure all CI tests run clean. 

DO NOT MERGE. This will need to be pushed as a fast-forward from the command-line...

### DIFF
--- a/core/version.h
+++ b/core/version.h
@@ -9,7 +9,7 @@
 //   git tag -s 3.3.0
 //   git push --follow-tags
 
-#define MRTRIX_BASE_VERSION "3.0.2"
+#define MRTRIX_BASE_VERSION "3.0.3"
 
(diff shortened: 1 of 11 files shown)


Lestropie:
Didn't follow the updated [instructions](https://github.com/MRtrix3/mrtrix3/wiki/Publishing-a-new-release): the container recipes should be hard-wired to the new tag, as otherwise checking out an older tag and building a container will not build the source code corresponding to that tag.

jdtournier:
Oops. I'd actually seen that and tried to figure out what needed to be updated in the `Dockerfile` and `Singularity` files, but didn't see the previous release tags in there. I assumed the release instructions were out of date as you'd changed how  these were handled recently... Sorry!

PR #2356, opened 2021-07-17
Release 3.0.3 (take 2)

Lestropie:
Replacement of #2355.

f89e89c4 replaces e14597c6 by including requisite change to container recipes.

NOT TO BE MERGED; is only to be used for a final CI run.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM buildpack-deps:buster AS base-builder
 FROM base-builder AS mrtrix3-builder
 
 # Git commitish from which to build MRtrix3.
-ARG MRTRIX3_GIT_COMMITISH="master"
+ARG MRTRIX3_GIT_COMMITISH="3.0.3"
 # Command-line arguments for `./configure`
 ARG MRTRIX3_CONFIGURE_FLAGS=""
 # Command-line arguments for `./build`
--- a/Singularity
+++ b/Singularity
@@ -58,7 +58,7 @@ Include: apt
     ln -s /usr/bin/python3 /usr/bin/python
 
 # MRtrix3 setup
-    git clone -b master --depth 1 https://github.com/MRtrix3/mrtrix3.git /opt/mrtrix3
+    git clone -b 3.0.3 --depth 1 https://github.com/MRtrix3/mrtrix3.git /opt/mrtrix3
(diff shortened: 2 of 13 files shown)


Lestropie:
PR shows as out of date as branch `master_pre_3.0.3` commences at 6a2e5658, which was the same merge base as #2355, but due to the prior fast-forward merge to `master` that was done associated with #2355 `master` now contains additional commits that this branch does not. But that should be fine given that this PR is only used for CI tests.

jdtournier:
OK, given that last release was a dud, I need to urgently delete that release before people start installing it...

jdtournier:
OK, I don't get how the CI passes when I can't build this on my system due to tag mismatch. The [3.0.3 tag on here](https://github.com/MRtrix3/mrtrix3/releases/tag/3.0.3) still points to 19cbb1e, which was pushed in #2355. But this PR is based off `3.0.2`, so this shouldn't work. What am I missing...?!?

jdtournier:
Right, I get why the CI tests pass: because what's being tested is a merge of this branch with the current `master`, which has itself already been tagged... Makes sense, but that will mean the final force-push will leave the tag pointing elsewhere. I've gone ahead and fixed this all up: reverted `master` to what it was, and moved the tag over to the tip of this branch, as it should be. I'll re-run the CI just to make sure...

Lestropie:
There's a couple of other components to this particular update:

- Manually build & push Docker container to DockerHub (I've commenced building this locally just now). Following `3.0.3` I'd like to have this automated, but I wanted to do it manually first time out.

- On contemplation, I think that following the commits that perform version updating & documentation rebuild for the purpose of the tag, which now include hard-wiring of the commitish to be pulled by the container recipes, there should be a PR to `master` immediately *after* the tagged commit, which reverts the container recipes back to pulling `master`. This way, building a tagged container pulls the code corresponding to that tag, whereas building a container off of `master` will pull the latest `master` code. In combination with the first point above, this will mean that the container tagged `latest` can be continuously updated on DockerHub based on hotfixes to `master`.

- The [Wiki instructions](https://github.com/MRtrix3/mrtrix3/wiki/Publishing-a-new-release) for managing updates should itself be updated accordingly.

Lestropie:
Update: Docker containers now [on DockerHub](https://hub.docker.com/repository/docker/mrtrix3/mrtrix3).

PR #2358, opened 2021-07-25
CONTRIBUTING.md: Revise

Lestropie:
In the process of helping @remitoburrito generate a novel contribution, realized that there was elements missing from this document (notably running `docs/generate_user_docs.sh`). Then in adding that, started to realize that the structure was a little bit off; in particular the selection of a base branch and generation of a novel branch off of that. So in thinking about it again, I decided to alter the relevant heading structure to:

1. Classify the type of contribution that is to be made (since this has consequences in various places);
2. Produce the actual content;
3. Create the Pull Request.

I hope this works better. But I'm also leaving it as a draft PR for now, as there's a reasonable chance of further revisions as we finish off the process.


Lestropie:
P.S. Would be ideal to have this merged before Friday, as I hope to make reference to it during a presentation.

jdtournier:
Made a few changes, happy to merge if you're OK with that.

jdtournier (approved):

Lestropie:
Don't know why in my mind semantic versioning was "major; minor; macro" rather than "major; minor; patch" 🤔

PR #2359, opened 2021-07-27
New command "mask2glass"

remitoburrito:
New command "mask2glass" written, which automatically generates a "glass brain" image for visualisation purposes, using a mask image as input. This visualisation technique has been used in publications using fixel-based analysis, and has been requested from members to apply to their own independent research.

### DIFF
--- a/bin/mask2glass
+++ b/bin/mask2glass
@@ -18,20 +18,22 @@
 import warnings
 
 def usage(cmdline): #pylint: disable=unused-variable
-  from mrtrix3 import app #pylint: disable=no-name-in-module, import-outside-toplevel
   cmdline.set_author('Remika Mito (remika.mito@florey.edu.au) and Robert E. Smith (robert.smith@florey.edu.au)')
   cmdline.set_synopsis('Create a glass brain from mask input')
   cmdline.add_description('The output of this command is a glass brain image, which can be viewed '
-                         'using the volume render option in mrview, and used for visualisation purposes to view results in 3D. '
-                         'This script takes either a binary mask image or floating point image (e.g. a mean mask image computed using mrmath) as input.')
+                          'using the volume render option in mrview, and used for visualisation purposes to view results in 3D.')
+  cmdline.add_description('While the name of this script indicates that a binary mask image is required as input, it can '
+                          'also operate on a floating-point image. One way in which this can be exploited is to compute the mean '
+                          'of all subject masks within template space, in which case this script will produce a smoother result '
+                          'than if a binary template mask were to be used as input.')
   cmdline.add_argument('input', help='The input mask image')
   cmdline.add_argument('output', help='The output glass brain image')
-  cmdline.add_argument('-scale', type=float, default=2.0, help='Provide resolution usscaling value; default = 2.0')
+  cmdline.add_argument('-dilate', type=int, default=2, help='Provide number of passes for dilation step; default = 2')
+  cmdline.add_argument('-scale', type=float, default=2.0, help='Provide resolution upscaling value; default = 2.0')
   cmdline.add_argument('-smooth', type=float, default=1.0, help='Provide standard deviation of smoothing (in mm); default = 1.0')
 
 
 def execute(): #pylint: disable=unused-variable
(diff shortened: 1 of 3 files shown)


Lestropie:
@MRtrix3/mrtrix3-devs Would appreciate an independent code review on this given that some of the commits are mine.

bin/mask2glass
```diff
@@ -0,0 +1,83 @@
...
+  smooth_option = ' -stdev ' + str(app.ARGS.smooth)
+  threshold_option = ' -abs 0.5'
+
+  # check whether threshold should be fixed at 0.5 or computed automatically from the data
+  if image.Header('in.mif').datatype() == 'Bit':
+    app.debug('Input image is bitwise; no need to check image intensities')
+  else:
+    app.debug('Input image is not bitwise; checking distribution of image intensities')
+    result_stat = image.statistics('in.mif')
+    if result_stat.min < 0.0 or result_stat.max > 1.0:
```
jdtournier, line 55: Would suggest amending to:
```
if result_stat.min != 0.0 or result_stat.max != 1.0:
```
in case image contains small intensities -- e.g. [ 0 0.001 ] would not trigger the automatic determination of the threshold, but the default of 0.5 would clearly not be appropriate.

bin/mask2glass
```diff
@@ -0,0 +1,83 @@
...
+      app.warn('Input image contains values outside of range [0.0, 1.0]; threshold will not be 0.5, but will instead be determined from the image data')
+      threshold_option = ''
+    else:
+      app.debug('Input image values reside within [0.0, 1.0] range; fixed threshold of 0.5 will be used')
+
+  # run scaling step
+  run.command('mrgrid in.mif regrid upsampled.mif' + scale_option)
+
+  # run upsampling step
```
jdtournier, line 64: Comment should be `# run smoothing step`

bin/mask2glass
```diff
@@ -0,0 +1,83 @@
...
+  run.command('mrfilter upsampled.mif smooth upsampled_smooth.mif' + smooth_option)
+
+  # threshold image
+  run.command('mrthreshold upsampled_smooth.mif upsampled_smooth_thresh.mif' + threshold_option)
+
+  # dilate image for subtraction
+  run.command('maskfilter upsampled_smooth_thresh.mif dilate upsampled_smooth_thresh_dilate.mif' + dilate_option)
+
+  # create border
+  run.command('mrcalc upsampled_smooth_thresh_dilate.mif upsampled_smooth_thresh.mif -subtract out.mif')
```
jdtournier, line 74: Might suggest using `-xor` rather than `-subtract` here, since it's probably more appropriate for binary inputs, but it's a very minor point...

bin/mask2glass
```diff
@@ -0,0 +1,83 @@
...
+
+  # create output image
+  run.command('mrconvert out.mif ' + path.from_user(app.ARGS.output),
```
jdtournier, line 77: Worth requesting `-datatype bit` if the output is binary...? 

I guess this depends on whether `out.mif` is already binary by that stage, I'm not sure whether `mrcalc` will produce a binary output for the previous subtraction...

jdtournier (commented):
Looks pretty good, just a few very minor points...

bin/mask2glass
Lestropie: I'm sure I've made an argument somewhere for a `maskcalc` command...

bin/mask2glass
Lestropie: Resolved by 7b97f13d.

bin/mask2glass
Lestropie: Resolved by 7b97f13d.

bin/mask2glass
Lestropie: Resolved by 7b97f13d.

bin/mask2glass
Lestropie: Resolved by 7b97f13d.

Lestropie:
@remikamito Given this will appear in the `3.1.0` release changelog, and all listed PRs by are automatically filled with contributor avatars (as is the *MRtrix3* website), would you like to add a portrait to your GitHub profile?

PR #2363, opened 2021-08-18
dwidenoise: Add option -demean

Lestropie:
This adds command-line option `-demean`, which subtracts the mean signal across the shortest matrix axis, essentially pre-allocating it as a signal component.
This improves stability of rank determination when the coefficient of variance of the input data is small (e.g. fMRI data).
Re-introduces code from 99d71856.

#### Magnitude data denoising

screenshot0000

Rank images:
1. Double precision calculations, no demeaning
2. Double precision calculations, with demeaning
3. Single precision calculations, no demeaning
4. Single precision calculations, with demeaning

screenshot0001
screenshot0002
screenshot0003
screenshot0004

#### Complex data denoising

screenshot0010

Rank images:
1. Double precision calculations, no demeaning
2. Double precision calculations, with demeaning
3. Single precision calculations, no demeaning
4. Single precision calculations, with demeaning

screenshot0011
screenshot0012
screenshot0013
screenshot0014

(Big bright spot at focus point is issue introduced by use of sub-of-squares reconstruction; should have used the SENSE1 data for demonstration...)


Lestropie:
Base branch for merge is base of #2340, since much of the experimentation done around this was based around observations only made by utilising the enhancements included there. I would however suggest on resolving the pair of these that the command-line options be arranged into groups.

Lestropie:
[Suspiciously relevant Twitter post](https://twitter.com/gabrielpeyre/status/1428095472850325504) appeared on my feed...

jdtournier (approved):
Seems good to me. @dchristiaens, are you happy with these changes?

dchristiaens:
Sorry for not responding earlier. I meant to test this at some point, but I forgot. The code looks OK to me, except for the minor issue about the `rankmap` that I mentioned in my review. I'm also not 100% sure, though.

jdtournier:
No worries, it's not like I've been particularly diligent of late... What issue are your referring to related to rankmap? I don't see any comment here that would match the review you mention...

Lestropie:
The other option here (from another suspiciously relevant twitter thread, though I didn't grab the link at the time) is to explicitly Z-transform all input data prior to PCA, which would supersede this proposal.

dchristiaens:
> No worries, it's not like I've been particularly diligent of late... What issue are your referring to related to rankmap? I don't see any comment here that would match the review you mention...

Perhaps I messed up Github reviews... I was referring to [this line](https://github.com/MRtrix3/mrtrix3/blob/11cc9e04794e5f402c41b930ae2bf7caca94b3c4/cmd/dwidenoise.cpp#L239), on which I had commented:

> I'm not sure if this is right. When subtracting the mean, the rank is reduced by 1 during the EVD, but on the output and with the mean added again, it shouldn't make a difference. So I think the output rank map should still be `r - cutoff_p`. We could probably test this on an image that contains pure Gaussian noise with 0 mean; in that instance the behaviour with and without demeaning should be identical.

---

> The other option here (from another suspiciously relevant twitter thread, though I didn't grab the link at the time) is to explicitly Z-transform all input data prior to PCA, which would supersede this proposal.

Do you mean this "double centring" [you mentioned earlier](https://twitter.com/gabrielpeyre/status/1428095472850325504)? I'm not convinced this adds any value over the current approach of demeaning on the shortest matrix dimension, but I haven't done any testing on this...

Lestropie:
> Do you mean this "double centring" you mentioned earlier?

The double-centering would address the issue of selecting which matrix dimension to demean, and therefore potentially having sudden changes in behaviour if the number of volumes is just below vs. just above the kernel size. But it comes with a bit of extra computational cost because you have to iterate to do so. I was thinking instead just doing a single Z-transform along the shortest dimension; so not just demeaning as currently implemented here but also multiplying to unit variance as a kind of pre-conditioning. The link I came across was essentially just stating that it's a very common step prior to PCA.

I'm not the expert here, I'm just verifying that we make an informed decision.

Also, currently demeaning is done via an explicit command-line option, whereas if we were to double-centre or just Z-transform, we'd probably do it all the time and remove the command-line option.

jdtournier:
My 2¢...

First off, it's important to note we're not doing a PCA analysis, at least not in the regular sense of the term. We won't be looking at the principal components themselves, we're only interested in those that look like noise. I think that means some of the considerations for PCA won't be applicable here. At least, that's my understanding, might be wrong here...

So with that in mind: if demeaning is only there to help with numerical stability, simply demeaning one of the dimensions should be plenty enough to get 99% of the desired benefit. Double centering strikes me as overkill in this context, especially if it comes with additional computational overhead.

As to going for a full Z transform, I think that would explicitly break the assumption of constant noise variance in the data (at least, over the patch being denoised). Again, might be appropriate for a PCA, but in our case, I expect doing this would lead to the distribution of singular values for pure noise no longer being characterised by the MP distribution...

Regarding @dchristiaens's concerns with the rank when demeaning applied, I'm not sure either, but I agree that a quick simulation with random matrices should confirm one way or the other.

Lestropie:
It occurred to me that there's an additional possibility here for removing non-zero mean effects prior to decomposition.

If the input data possess a DWI shell structure, then what one could choose to do is: for each voxel, rather than regressing out the mean across all volumes, instead regress out the mean per shell.

Lestropie:
Closing in favour of #3023.

Lestropie:
Some notes:

-   For multi-echo fMRI data, might be advantageous to demean each echo individually, rather than computing a single mean per voxel across all non-spatial axes.

-   Regarding the rank issue:
    -   My assessment while thinking about & implementing the demeaning that is now a part of the "preconditioning" block was that, for particularly the MP distribution based estimation, the distribution of signal eigenvalues is effectively irrelevant: what matters instead is the appearance of the MP part of the distribution. As such, anything that makes the noise part of the data look most like the MP distribution should be the preference.
        (Similar to this is the concept of making the signal portion of the data as low-rank as possible. In addition to getting a more dense sampling of the MP portion of the distribution this likely also increases the separation between the MP and non-MP parts of the distribution.)
    -   Since the noise level estimation is based on manifestation of the MP distribution, that aspect should I think not be affected by having performed "preconditioning" of the data prior to PCA. What would be expected to be affected is an inference of the rank based on that noise level (the `Exp` estimators kind of do this backwards, choosing a rank and computing a noise level from that, but this is not ordered out of necessity). My current solution is to say that if *either* demeaning or phase demodulation is applied, treat that as a signal component; but this doesn't affect the estimator, only *reporting* of signal rank.
        This may require some cross-checking to make sure it does as it should in all use cases. Eg. The fixed-rank estimator requested in #3046 seems to not currently be accounting for that additional component.

Lestropie:
Some further notes / questions:

-   Demeaning across all volumes results in one ideally zero singular value at the start of the ascending list. With the BDCSVD decomposition #2906 this is commonly (but not always) exactly zero. The SelfAdjointEigenSolver is however less precise: this first eigenvalue can be "only" several orders of magnitude smaller than the next eigenvalue (as opposed to being barely distinguishable from zero), and is sometimes negative. This in turn is detrimental to the estimator. #3029 will have more explicit handling of singular values that should be presumed zero as a consequence of data preconditioning.

-   Within #3029 the code is set up to consider demeaning as a component of "preconditioning", which is applied as a single step to the whole input series before any PCA takes place. It is therefore done on a per voxel basis. This means that in the scenario where the patch size is smaller than the number of input volumes, the demeaning will be occurring along the longer dimension of the matrix. This would therefore conflict with @dchristiaens [comment](https://github.com/MRtrix3/mrtrix3/pull/2363#issuecomment-962467934) that demeaning should occur along the shorter dimension of the matrix. Anyone know from first principles to what extent this might be a problem?

PR #2366, opened 2021-08-23
dwicat: Support 3D images

Lestropie:
If an input image to `dwicat` is a 3D image containing a single *b*=0 volume, it should be possible to continue processing; but it is necessary to validate such, and to bypass the `dwiextract` step.


jdtournier (approved):
Looks all good to me :+1:

jdtournier (approved):

PR #2368, opened 2021-08-24
 Eigen 3.4 compatibility

jdtournier:
Remove `Eigen::Vectors/4` typedefs since these are defined in Eigen 3.4 and cause conflicts. Instead, just use the standard `Eigen::Vector3d` typedef, which is identical currently. This may cause issues if we ever decide to revert the `default_type` typedef from `double` (its current value) to float or something, but I don't see that happening any time soon. Besides, even if we do change `default_type`, there's a good argument that this shouldn't influence the vast majority of these instance of `Vector3`, since this is likely to break all kinds of operations. 

These changes are required to pass CI on macOS, since homebrew now ships with Eigen 3.4.


jdtournier:
Looks like Eigen 3.4 introduces subtle differences in precision. Actually not so subtle in the case of `dwidenoise`, but that only happens with the `Exp1` estimator. Not sure what to make of that... 

For now, I reckon we pin the Eigen version to 3.3 in homebrew for the macOS CI - that'll at least allow the CI to complete, we can investigate the issues with precision at a later time.

jdtournier:
@maxpietsch: any idea how to request version 3.3.X of `eigen` in `homebrew`...?

jdtournier:
I had a look into the test failures with Eigen 3.4, and they're really due to quite minor differences in precision. For `dwidenoise` with the `Exp1` estimator, the tests pass OK if I reduce the absolute precision to 1e-3 - which equates to a _relative_ precision of around 1e-5 given that the mean DWI values in there are around the 200-300 mark, so I reckon that's actually a fair level of tolerance. I don't think we need to duplicate the test data at this point. Is that OK with everyone?

bjeurissen:
Looks good!

bjeurissen (approved):

jdtournier:
OK, I'll merge this to `master` as it is, and then merge master back to `dev`. When I do, I'll update the test data with results using Eigen 3.4. Rationale being to avoid making unnecessary changes to `master` that might look more significant than they really are (which modifying the test data might suggest...).

Issue #2360, opened 2021-08-05
mrtrix3 docker - fslsplit missing


chiara111p111:
Hi everyone,

I've recently started using MRtrix3 container to do a dMRI analysis (eddy preprocessing and quad analysis). 
In particular I've used dwifslpreproc command:

`docker run --rm -it \
                        -v $input_folder/$pos/$pzt/$weight/dMRI_SS/1:/input \
                        -v $output_folder/$pos/$pzt/$weight/:/work \
                        mrtrix3 dwifslpreproc \
                        /input/$nifti \
                        /work/dwi.mif \
                        -rpe_none \
                        -eddy_options " --slm=linear --cnr_maps" \
                        -pe_dir AP \
                        -fslgrad /input/$bvec /input/$bval \
                        -eddyqc_all directory -force`

The problem presented after I insert _--cnr_maps_ option. Doing so, I obtained this warning:

`[WARNING] Error running automated EddyQC tool 'eddy_quad'; QC data written to "/work/directory" will be files from "eddy" only`

and quad analysis wasn't done.

Rerunning in debug mode I've found that the error is given by fslsplit missing:

`Running $FSLDIR/bin/fslsplit dwi_post_eddy.eddy_cnr_maps.nii.gz dwi_post_eddy.qc/cnr
/bin/sh: 1: /opt/fsl/bin/fslsplit: not found
Traceback (most recent call last):
  File "/opt/fsl/bin/eddy_quad", line 54, in <module>
    main(args.eddyBase, args.eddyIdx, args.eddyParams, args.mask, args.bvals, args.bvecs, args.output_dir, args.field, args.slspec, args.verbose)
  File "/opt/fsl/fslpython/envs/fslpython/lib/python3.7/site-packages/eddy_qc/QUAD/quad.py", line 377, in main
    quad_cnr_maps.main(pp, data, eddyOutput)
  File "/opt/fsl/fslpython/envs/fslpython/lib/python3.7/site-packages/eddy_qc/QUAD/quad_cnr_maps.py", line 38, in main
    fslpy.fslsplit(eddy['cnrFile'], data['qc_path'] + '/cnr')
  File "/opt/fsl/fslpython/envs/fslpython/lib/python3.7/site-packages/eddy_qc/utils/fslpy.py", line 28, in fslsplit
    jobout = check_output(cmd, shell=True)
  File "/opt/fsl/fslpython/envs/fslpython/lib/python3.7/subprocess.py", line 411, in check_output
[...]

PR #2370, opened 2021-09-02
Container recipe updates following 3.0.3 release

Lestropie:
Pairs with MRtrix3/containers#21.

Plan is to push update to `latest` tag on DockerHub in order to make the fix to #2360 available.

Note that this update does *not* address the additional dependencies that will need to be satisfied in order to resolve #2279 (addition of container recipes) with #2197 (new `dwi2mask`).

Instructions on DockerHub will also need to be updated to reflect expectation of obtaining file `fslinstaller.py`.


Lestropie:
I have checked that both Docker and Singularity containers build, and execute `dwifslpreproc` including the flag that was causing failure in #2360.

Unusually, I found now that when I tried to use the Singularity container [via direct invocation](https://mrtrix.readthedocs.io/en/3.0.3/installation/using_containers.html#id1), i.e. "`MRtrix3.sif mrinfo file.mif`", it didn't work, I would instead get the help page for the *MRtrix3* command. The change in 489bffe seems to deal with this.

I have one more small modification to test before proposing, then will seek review on this so that I can move on cleanly to #2428 and #2388.

Lestropie:
Okay, final thing I wanted to experiment with (#2458) did not work as initially intended, so I'm not going to add anything to this PR. I'll look into a solution for that separately, it's not requisite for `3.0.4`. So seeking review.

Lestropie:
As of a635c0d I get warning-less `mrview` execution on both Docker and Singularity (built both from `Singularity` recipe file and from conversion from DockerHub). If anybody wants to test on a different system, in addition to the code here you can try `docker://mrtrix3/mrtrix3:test_3.0.4`.

Lestropie:
Turns out I'm having a bunch of problems getting `mrview` working on Xubuntu 20.04 that I didn't have previously. Could not get it working on Docker or Singularity, on 3.0.3 or what will become 3.0.4 (including suggested changes in the documentation), and regardless of how the Singularity container was built. So I need to keep experimenting here. Not sure what's changed.

Quick observation:

```
libGL error: No matching fbConfigs or visuals found
libGL error: failed to load driver: swrast
```
can seemingly be fixed with:
```
export LIBGL_ALWAYS_INDIRECT=1
```

Lestropie:
On my Arch Linux laptop, can confirm `mrview` works without error for both Docker and Singularity (regardless of whether the latter was built from the Singularity recipe file or from the Dockerhub container). However my Xubuntu desktop at work failed across the board. Would very much appreciate an independent test to find out which of these is the odd one out.

Lestropie:
The system on which I was having trouble with container-executed `mrview` is now dead. So in the absence of anyone independently reproducing the same issue, I can only advocate for merging this and re-assessing if any user reports an issue with `3.0.4`.

jdtournier (approved):

PR #2373, opened 2021-09-10
dwi2response manual: avoid checking and calculating a brain mask

bjeurissen:
For the manual response function estimation, a brain mask is not needed or supported. However, currently, a brain mask is always calculated if one is not provided. This is unnecessary and can also lead to crashes whenever dwi2mask fails (e.g. exotic data, simulations, ... exactly the use-cases where you would use manual response estimation).


bjeurissen:
The eigen 3.4 issue has now also made it into the CI :(

Lestropie:
Testing string equivalence of the selected algorithm within `dwi2response` is not the ideal solution from a code hierarchy perspective. It would be preferable for algorithms to define a function that returns a value indicating whether or not that particular algorithm wishes for a brain mask to be defined / calculated for it.

bjeurissen:
> Testing string equivalence of the selected algorithm within `dwi2response` is not the ideal solution from a code hierarchy perspective. It would be preferable for algorithms to define a function that returns a value indicating whether or not that particular algorithm wishes for a brain mask to be defined / calculated for it.

Agreed!

jdtournier (approved):
Seems OK to me. Ready to merge, @Lestropie?

Issue #2383, opened 2021-10-20
CUDA not working in Singularity


araikes:
**Describe the bug**

I downloaded the official Singularity image to use with some data. Given that CUDA paths were defined in the Singularity recipe, I anticipated `dwifslpreproc` to make use of eddy_cuda. However, this is not the case because there are no CUDA folders in the actual Singularity container and `eddy_cuda9.1` requires CUDA 9.1 drivers, which are not available in Ubuntu 20.04 (the base OS for the container).

If you are running this on a system with an older OS that has CUDA 9.1 toolkit, it may work by default when you load the container with the `--nv` flag, but if you're running anything newer (e.g., our HPC is running CUDA 11.4 with no backports) then this does not work out of the box.

I tested locally and including the following (created by others, not me) in the recipe enables CUDA support on our HPC. https://gist.github.com/DaneGardner/accd6fd330348543167719002a661bd5#install-cuda-toolkit

**To Reproduce**

Try running eddy_cuda on a GPU running anything later than CUDA 9.1.

**Platform/Environment/Version**

Please provide the following information:

-  OS: CentOS 7; CUDA 11.4 toolkit

-  *MRtrix3* version: `master` branch, based on what the recipe indicates.

---

PR #2388, opened 2021-10-27
Singularity: Update Singularity recipe for eddy_cuda9.1 support

araikes:
Cuda support in the official Singularity container is currently not functional https://github.com/MRtrix3/mrtrix3/issues/2383#issue-1031635708). FSL's Eddy requires CUDA 9.1 or 8. However, Ubuntu 20.02 does not inherently support CUDA < 10. 

Modifications to the Singularity recipe installs CUDA 9.1 toolkit only as well as relevant patches. I build this locally (CentOS 7, CUDA 11.4 installed) and eddy_cuda9.1 runs for `dwifslpreproc`.

Ubiquitous testing requires multiple OS options, which I do not have access to.

### DIFF
--- a/Singularity
+++ b/Singularity
@@ -26,8 +26,8 @@ Include: apt
     export FSLDIR FSLOUTPUTTYPE FSLMULTIFILEQUIT FSLTCLSH FSLWISH
     
 # All
-    LD_LIBRARY_PATH="/usr/local/cuda/lib64:/usr/local/cuda/bin:/.singularity.d/libs:/usr/lib:/opt/fsl/lib:$LD_LIBRARY_PATH"
-    PATH="/opt/mrtrix3/bin:/opt/ants/bin:/opt/art/bin:/opt/fsl/bin:/usr/local/cuda/bin:$PATH"
+    LD_LIBRARY_PATH="/usr/local/cuda-9.1/lib64:/usr/local/cuda-9.1/bin:/.singularity.d/libs:/usr/lib:/opt/fsl/lib:$LD_LIBRARY_PATH"
+    PATH="/opt/mrtrix3/bin:/opt/ants/bin:/opt/art/bin:/opt/fsl/bin:/usr/local/cuda-9.1/bin:$PATH"
     export LD_LIBRARY_PATH PATH
 
 %post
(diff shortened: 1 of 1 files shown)


Singularity
```diff
@@ -62,7 +81,7 @@ Include: apt
     cd /opt/mrtrix3 && ./configure && ./build -persistent -nopaginate && rm -rf testing/ tmp/ && cd ../../
 
 # apt cleanup to recover as much space as possible
-    apt-get remove -y build-essential ca-certificates curl git libeigen3-dev libfftw3-dev libgl1-mesa-dev libpng-dev libqt5opengl5-dev libqt5svg5-dev libtiff5-dev qt5-qmake qtbase5-dev-tools wget zlib1g-dev && apt-get autoremove -y && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    apt-get remove -y build-essential ca-certificates curl git libeigen3-dev libfftw3-dev libgl1-mesa-dev libpng-dev libqt5opengl5-dev libqt5svg5-dev libtiff5-dev qt5-qmake qtbase5-dev-tools wget zlib1g-dev && apt-get autoremove -y && apt-get clean && rm -rf /var/lib/apt/lists/* #/tmp/* #/var/tmp/*
```
Lestropie, line 84: @araikes Do you have any recollection of whether it was necessary to no longer wipe the contents of these directories, or whether this is just residual from testing?

Lestropie:
Sorry for taking forever to take a look at this @araikes.

I'd wanted to merge this to `master` and slip it into the pending `3.0.4`, given I've been doing a bunch of other fixes around the new containers as well (#2461, #2370, #2360). Couple of concerns though:

- This approximately triples the size of the container. I'm guessing that only a fraction of the libraries brought in are actually necessary to make use of `eddy_cuda9.1`. It would be preferable to make use of `neurodocker-minify` to produce a reduced download file. But not only would this require getting CUDA working in the container used for detection of dependencies (https://github.com/MRtrix3/containers), there would also potentially be a risk of that operation omitting dependencies that change depending on the host system and/or target hardware. I can hopefully look into this at some point, but it won't happen for `3.0.4`.

- The Linux system I'd been intending to use for testing the proposed change in retrospect is not a good choice, as it's an Arch laptop with non-functional dual-GPU hardware. I think I only have easy access to one other system with adequate GPU hardware to run `eddy_cuda`. I can test on a couple of HPCs prior to merging, but those are slow to test on and each will likely have their own requisite wrapper configurations.

I'll try to test what I can soon and post back here.

Singularity
Lestropie: Note: This was originally addressed in [comment](https://github.com/MRtrix3/mrtrix3/issues/2383#issuecomment-952446849) in #2383.

Lestropie:
I haven't had any success getting this or alternative solutions to work on either of my testing systems. So I'll transition from `3.0.4` to `3.1.0` milestone. But I do want to get this working.

araikes:
Hi @Lestropie,

I'll try and circle back around to this next week. There may be some opportunity for container size optimization, watching what's happening with FastSurfer development.

PR #2392, opened 2021-11-06
pylint: updates for pylint 2.11.1 to pass

maxpietsch:
- `open` requires an encoding. I use `io.open` with `utf8` enoding wherever the encoding wasn't specified. This works for python 3 but causes issues with python 2. Looks like the only portable solution is to use binary mode and encode/decode everything as 'utf8' but that clutters the code quite a bit. I've done that in `configure` but in other places I left the Python 3 compatible `io.open` calls. Would you rather deactivate the warning, use explicit encode and decode calls, or are you happy to abandon Python 2 support in this PR?
- global statements for global variables that are not assigned to were present mostly in `build` and `configure`. Removed these but left them in `app.py` and deactivated the warning there as they seem useful for documentation purposes.
- disabled a few `consider-using-f-string` (many in `popluation_template`) and `consider-using-with` warnings
- some python dictionary `.items` and `.keys` cleanup
- generators instead of lists for `any` and `all` calls
- couldn't figure out why `except Exception as e:` throws so many `unused-variable` warnings about `e`. Looks like a pylint bug but deactivated these

note to self: update master before working on the code


Lestropie:
> Would you rather deactivate the warning, use explicit encode and decode calls, or are you happy to abandon Python 2 support in this PR?

I would prefer for `3.0.x` to retain Python2 support. But we shouldn't invest a lot of effort in such. So I would prefer to:

- Disable newer pylint checks that fail in Python2;
- Focus on modifying `dev` to conform to recommendations for Python3 for `3.1.0`.

> couldn't figure out why `except Exception as e`: throws so many `unused-variable` warnings about `e`. Looks like a pylint bug but deactivated these

If "`e`" is being used, it may be a red herring warning string, with the actual issue being that the variable name does not conform to the specified requirements as it is too short.

jdtournier:
Sorry, looks like we might have been doubling up... I'd already pushed changes to fix these errors to `dev` in 9752aa3a58 - looks pretty much identical to yours. And yes, I think there is a bug in `pylint`'s handling of those exceptions...

maxpietsch:
`-.- ` 

feel free to close unless we want to make master pylint-compatible before merging dev or cherry-picking https://github.com/MRtrix3/mrtrix3/commit/9752aa3a580d1accdf2410686085c683eb9b4f9e.

jdtournier:
OK, did a bit of tidying up. I kept most of your changes, especially the explicit use of the `io` module, which I think allows us to handle locale issues more consistently between python 2 & 3? Otherwise, I've kept a couple of my changes, but nothing major. I'll need to merge this back to `dev` once we merge this to `master`...

@maxpietsch, @Lestropie: can I ask you to double-check you're happy with the changes in aa5a354? These are to fix 'unused variable' warnings from my compiler (clang 13.0).

cmd/warp2metric.cpp
```diff
@@ -87,11 +87,6 @@ void run ()
       Fixel::copy_directions_file (template_fixel_directory, output_fixel_directory);
     }
 
-    uint32_t num_fixels = 0;
```
maxpietsch, line 90: is this supposed to be merged to master?

cmd/tckedit.cpp
```diff
@@ -142,7 +142,7 @@ void run ()
 
   // Get the consensus streamline properties from among the multiple input files
   Tractography::Properties properties;
-  size_t count = 0;
+  size_t count = 0, total_count = 0;
```
maxpietsch, line 145: is this supposed to be merged to master?

jdtournier: This depends on whether the same warnings pop up during CI... If they do, the CI will fail. But you're right, it's probably best to leave this alone on `master` unless the CI does fail. We can fix it on `dev` in the meantime...

maxpietsch:
Correct me if I am wrong but I don't think we can follow pylint rules by using `io.open(path, 'r', encoding='utf8')` and make the code compatible with python 2 and 3.

Python2
```python
with io.open('file', 'r', encoding='utf8') as f: type(f.read())
<type 'unicode'>
```

Python3
```python3
with io.open('file', 'r', encoding='utf8') as f: type(f.read())
<class 'str'>
```

In `configure` I used binary mode and encoded/decoded every string manually to make it compatible with Python2 and 3 but that's ugly. So currently, we're breaking Python2 runtime compatibility (not build).

jdtournier:
> Correct me if I am wrong but I don't think we can follow pylint rules by using io.open(path, 'r', encoding='utf8') and make the code compatible with python 2 and 3.

I've no idea... I don't know whether the type being a regular `str` in python 2 is necessarily going to cause any issues... I'll give some of these commands a shot under python2 now.

jdtournier:
Yep, it's a problem. Just tried to run `dwi2response`:
```
dwi2response: [ERROR] Unhandled Python exception:
dwi2response: [ERROR]   TypeError: write() argument 1 must be unicode, not str
dwi2response: [ERROR] Traceback:
dwi2response: [ERROR]   /home/donald/mrtrix3/bin/dwi2response:86 (in execute())
dwi2response: [ERROR]     app.make_scratch_dir()
dwi2response: [ERROR]   /home/donald/mrtrix3/lib/mrtrix3/app.py:310 (in make_scratch_dir())
dwi2response: [ERROR]     outfile.write(WORKING_DIR + '\n')
```

Lestropie:
> can I ask you to double-check you're happy with the changes in aa5a354?

Changes in `tckedit` are not related to fixing unused variable warnings. Others look fine.

maxpietsch:
>I would prefer for 3.0.x to retain Python2 support. But we shouldn't invest a lot of effort in such. So I would prefer to:
> - Disable newer pylint checks that fail in Python2;
> - Focus on modifying dev to conform to recommendations for Python3 for 3.1.0.

Of the two options
1. reverting the changed `open` calls and ignoring these warnings just to merge them in again with `dev`, or
2. abandon python 2 in this PR and tag it appropriately

I initially preferred the latter but as the removal of python 2 is more thoroughly addressed in `dev` and as the milestone list of the current 3.1.0 is still far from complete and releasing and announcing versions takes time, I'd go with Rob's suggestion. Is everyone happy with that?

jdtournier:
OK, I have to admit I was wondering how we were going to maintain python 2/3 compatibility _and_ pass all the `pylint` tests... I agree it's far simpler just to disable the failing `pylint` tests on `master`, and worry only about Python2 compatibility on `dev` for release in `3.1.0`. I'll make the necessary changes in a bit. 

Other than that:

> Changes in tckedit are not related to fixing unused variable warnings. 

They are, in that `clang++` version 11 throws this warning:
```
cmd/tckedit.cpp:173:28: warning: variable 'this_total_count' set but not used [-Wunused-but-set-variable]
    size_t this_count = 0, this_total_count = 0;
```
Not a problem on CI yet, but I guess it won't be long before that causes failures if we don't address it. And it is true that while this variable [gets incremented in the loop](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/tckedit.cpp#L179), it ends up completely unused. So either we ditch it or we use it - whichever is most appropriate. But this can (and should) all be done in a separate PR since I have a feeling that it doesn't _yet_ cause any failures...

jdtournier:
OK, as far as I can tell, these are all the warnings that we currently get on `master` with `pylint` 2.11.1:
```
consider-using-f-string
unspecified-encoding
use-dict-literal
global-variable-not-assigned
consider-iterating-dictionary
consider-using-with
redundant-u-string-prefix
redefined-outer-name
unused-variable
consider-using-in
arguments-renamed
```
Of these, I reckon we won't be able to fix `consider-using-f-string`, `redundant-u-string-prefix`, and `unspecified-encoding` without breaking Python 2 compatibility. I'll try to make all the changes we can on this branch and disable these 3 warnings on `master`.

Lestropie:
> > Changes in tckedit are not related to fixing unused variable warnings.

> They are, in that clang++ version 11 throws this warning:

Compilation may well throw that warning with some version of the code, but you specifically requested a review of commit aa5a354, and that commit *introduces* that variable. So that commit *in isolation* couldn't possibly be responsible for solving that unused variable warning. I'm guessing that you stripped out eh variable altogether in a previous commit, then re-introduced it including not only the definition and increment but also the later utilisation.

Lestropie:
Making changes to text file / string data handling / encoding on `master` does make me a bit nervous. It took a lot of care to have everything working cross-platform and on Python 2 & 3, and it's been accidentally broken on a couple of occasions. That's the primary aspect where I'd prefer to just disable warnings on `master` and focus on overhauling to Python3 best practise for `3.1.0`.

jdtournier:
> > > Changes in tckedit are not related to fixing unused variable warnings.
> 
> > They are, in that clang++ version 11 throws this warning:
> 
> Compilation may well throw that warning with some version of the code, but you specifically requested a review of commit [aa5a354](https://github.com/MRtrix3/mrtrix3/commit/aa5a35486ada74bddf5e4103f12cf6c339cb69fd), and that commit _introduces_ that variable. So that commit _in isolation_ couldn't possibly be responsible for solving that unused variable warning. I'm guessing that you stripped out eh variable altogether in a previous commit, then re-introduced it including not only the definition and increment but also the later utilisation.

No, what I mean is there is an unused variable `this_total_count` in the code on `master` - this is what the compiler complains about. My fix did introduce another variable `total_count` to capture these values and include them in the output header. The alternative would have been to simply remove the `this_total_count` variable altogether (which arguably would have been the most correct since that wouldn't change the current behaviour at all). But I don't think it's worth worrying about at this stage anyway since it doesn't (yet) cause issues in the CI. We can fix this on `dev` and leave `master` alone for now.



> Making changes to text file / string data handling / encoding on `master` does make me a bit nervous. It took a lot of care to have everything working cross-platform and on Python 2 & 3, and it's been accidentally broken on a couple of occasions. That's the primary aspect where I'd prefer to just disable warnings on `master` and focus on overhauling to Python3 best practise for `3.1.0`.

Yes, I think you're right. Having looked at the range of changes we'd have to make to fix all the warnings, even if do ignore the encoding and string handling warnings, it's probably best to just disable all (or at least most) of the failing tests on `master` for now, as you say. There are one or two minor issues that would be worth fixing though, I'm going through these as we speak. I'll push a more streamlined set of changes when ready.

Lestropie:
Closing due to being superseded by #2401.

Still wish to go through the code and remove any hacks that were put in place for Python2 compatibility; #2047 is already part of [`3.1.0` milestone](https://github.com/MRtrix3/mrtrix3/milestone/5).

PR #2398, opened 2021-11-19
Documentation (Fixels and Dixels): double negative typo

fionaEyoung:
Tiny fix affecting [Fixel docs page ](https://mrtrix.readthedocs.io/en/latest/concepts/fixels_dixels.html)

Based on previous versions of this file I'm assuming this double negative snuck in by accident. Could also be rewritten as _"are ~not~ applied inconsistently"_

Sidenote, I figured it would be overkill to open an issue about this, but do let me know if that would have been preferred. The contributing [guidelines](https://github.com/MRtrix3/mrtrix3/blob/master/CONTRIBUTING.md) are somewhat vague when it comes to documentation, and since the `doc` branch hasn't been updated in two years, I went with `master` as base.


jdtournier (approved):

jdtournier:
Thanks for the fix, @fionaEyoung. For a minor change to the documentation, it's appropriate to merge to `master` - all good!

PR #2400, opened 2021-11-23
New command vendorscheme

bjeurissen:
Creates a vendor-specific gradient scheme from the output of gen_scheme


Lestropie:
- Vendor selection should be a compulsory positional argument rather than an option, since you can't do anything without it; it should also be configured as `choices=` in the `argparse` configuration so that it's enforced at the parsing stage

- Not sure whether it makes sense to have this as a standalone tool to be executed after `gen_scheme`, or whether it would make more sense to have `gen_scheme` able to write its results to a specific format rather than the format in which it currently writes those data. It would depend on the extent to which the outputs of `gen_scheme` are currently being used in that format. I suppose the exception would be the case where you want to generate one scheme and then put that same scheme on multiple machines of different vendors.

- I also have my [`dvsgen` script](https://gist.github.com/Lestropie/aee5d9f13ba60dc34593df5e4ccf1125), which does something akin to `gen_scheme` but with a few tweaks and with immediate export to Siemens `.dvs` format. Having both of these available to the community in parallel might cause confusion. 

`gen_scheme` was put there "as-is" and probably should have been a Gist or external project, and this might be kind of accentuating that. So we should consider how best to provide such tools to the community. Personally I'd rather this all be a single command rather than multiple.

bjeurissen:
> * I also have my [`dvsgen` script](https://gist.github.com/Lestropie/aee5d9f13ba60dc34593df5e4ccf1125), which does something akin to `gen_scheme` but with a few tweaks and with immediate export to Siemens `.dvs` format. Having both of these available to the community in parallel might cause confusion.

I was not aware of this gist. I think it would make sense to replace the code of `gen_scheme` with the corresponding code of `dvsgen` so that it uses the python API and better integrates with the rest of mrtrix.

> * Not sure whether it makes sense to have this as a standalone tool to be executed after `gen_scheme`, or whether it would make more sense to have `gen_scheme` able to write its results to a specific format rather than the format in which it currently writes those data. It would depend on the extent to which the outputs of `gen_scheme` are currently being used in that format. I suppose the exception would be the case where you want to generate one scheme and then put that same scheme on multiple machines of different vendors.

I think a separate script to convert the output of gen_scheme to a particular vendor's format makes more sense. Depending on the scheme running `gen_scheme` can take quite long, whereas converting to the format of a specific vendor takes less than a second. And like you mentioned, for multi-site projects, you would want to take the same scheme and output it to different formats.

thijsdhollander:
I can also see people wanting to get and convert a scheme from elsewhere; e.g., using a scheme from an existing dataset and converting it to a vendor format for implementation of similar acquisitions. Design-wise this seems in line with other conversion commands, e.g., `mrconvert` and `tckconvert`.  So this could be something like `schemeconvert`, or whatever name sounds right.

Lestropie:
Being able to import and export from multiple different storage conventions for acquisition schemes would indeed be ideal, and probably not too hard. The difficulty is in the representation of phase encoding information. My concern in having this script designed to operate on the output of `gen_scheme` is that it may be promoting the convention used in the output of `gen_scheme` to do so to a privileged status, where in reality I don't know whether the target destination for which that script was actually written actually required further modification down the road.

`dvsgen` writes multiple schemes to a single DVS file only if each has a unique number of directions (which can only occur if splitting the scheme into two and with an odd total volume count); otherwise it splits into multiple DVS files. `gen_scheme` writes a single file, with split number being an additional column.

Ideal for me would be to have a new `schemegen`, written in the Python API and merging features of `gen_scheme` with a trick or two from `dvsgen`, and a `schemeconvert`, which could take multiple input files in order to facilitate conversion between those formats that don't store everything in one file. The existing output format of `gen_scheme` could be a format option within both of these.

jdtournier:
I wouldn't worry about preserving the current output format of `gen_scheme` - it's served its purpose, and was always a bit of a hack. The only reason I'd added it to the repo was because of demand from some users, but in hindsight, this isn't the right way to do this (same with `convert_bruker`). I think it's time to figure out a more formal system for others to make their scripts and additions available alongside the core MRtrix install - something that's been discussed a few times over the years, but it may now be time to do something about it... Separate issue to what's discussed here though.

Lestropie:
My `dvsgen` Gist is now a repo from which one can build a Docker image to utilise that tool.
https://github.com/Lestropie/VendorScheme

If the code in this PR for exporting to GE can be integrated there:
https://github.com/Lestropie/VendorScheme/issues/1
, I think that'd be a more suitable place overall for the functionality.

bjeurissen:
What bothers me with this approach is that it combines generation of the gradient table (which is independent of the vendor), with vendor-specific conversion. Decoupling both would provide much more flexibility.

Lestropie:
If the code were split out to separate generation of the table from export of that table in vendor-specific formats, then that would expose, in code at least, the intermediate representation. That could be written to disk in some other format, or alternative commands could be written that would import that intermediate representation from disk and just to the conversion to vendor-specific format.

Part of my issue here is that the form of data produced by `gen_scheme` does not support my most common use case. For multiple phase encoding directions it produces a table with varying phase encoding direction per volume, which the Siemens sequences are not capable of doing. (see https://github.com/Lestropie/VendorScheme/issues/6)

PR #2327, opened 2021-05-24
Update connectome2tck.cpp

nicdc:
changed tck2connectome to connectome2tck

Please enter a concise description of the changes proposed in the pull request,
along with any relevant links to external discussions.

-  If your pull request involves fixing a bug, then the relevant code
   should be in a branch that is a derivative of `master`, and you should
   select `master` as the target branch in the pull request; for all other
   pull requests, the relevant code should be in a branch that is a
   derivative of `dev`, and you should select `dev` as the target branch
   of the pull request (assuming no other more appropriate branch exists).

-  If your proposed changes address an [existing Issue](https://github.com/MRtrix3/mrtrix3/issues)
   listed on GitHub, please add a reference to that Issue (a "`#`"
   character followed by the issue number).
#2326
-  If your proposed changes are not yet ready to be merged, but you are
   creating a pull request in order to initiate communications on such,
   please flag it as a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

-  You may optionally add `@mentions` of a person or team appropriate
   for reviewing proposed changes.
@jdtournier

PR #2402, opened 2021-11-24
CONTRIBUTING.md: Revise text RE: target branch for docs changes

Lestropie:
Based on feedback in #2398.

@fionaEyoung: Would this text have been more appropriate in your opinion? Eager to remove barriers to contribution as much as possible.

@jdtournier: I figure the main barrier is not `dev` vs. `master` target so much as whether we need to check formatting; in which case we probably need to merge `master` into `docs` at our end, merge the external PR into `docs`, then it's our obligation to fix then merge `docs` into `master` or `dev`. Unless it's easy to configure docs building from external forks?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,9 +145,13 @@ to be provided in subsequent sections of this document.
 
    -  If only making changes to the *MRtrix3* *documentation*, then the
       appropriate base branch will depend upon the nature of the changes
-      being proposed and the urgency of the update. In this scenario,
-      please generate an [Issue](https://github.com/MRtrix3/mrtrix3/issues)
-      to facilitate communication with the core development team.
+      being proposed and the urgency of the update. If providing a minor
+      correction to spelling / grammar, then a direct merge to `master`
+      is appropriate. If however your changes are sufficiently complex that
+      they would require verification that the formatting of the online
+      documentation is correct, then this will require coordination with
+      the *MRtrix3* core development team; in this case, please generate an
+      [Issue](https://github.com/MRtrix3/mrtrix3/issues) to facilitate such.
 
 1. Generate one or more Git commits that apply your proposed changes to
    the repository.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,13 +145,17 @@ to be provided in subsequent sections of this document.
 
    -  If only making changes to the *MRtrix3* *documentation*, then the
       appropriate base branch will depend upon the nature of the changes
-      being proposed and the urgency of the update. If providing a minor
-      correction to spelling / grammar, then a direct merge to `master`
-      is appropriate. If however your changes are sufficiently complex that
-      they would require verification that the formatting of the online
-      documentation is correct, then this will require coordination with
-      the *MRtrix3* core development team; in this case, please generate an
-      [Issue](https://github.com/MRtrix3/mrtrix3/issues) to facilitate such.
+      being proposed and the urgency of the update. For instance, if providing
+      a minor correction to spelling / grammar, or fixing an erroneous
+      hyperlink, then a direct merge to `master` is likely appropriate.
+      Conversely, if the changes are non-trivial, and it makes more sense for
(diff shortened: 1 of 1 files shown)


fionaEyoung:
I think that's a good clarification, and sends the message that no contribution is too big or small as long as the appropriate route is taken. It might also be nice, for the various scenarios, to have links to some example issues/PRs (e.g. #2327) that demonstrate the process.

For more complicated changes that might require core team input, maybe making use of the "[Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" feature would help? Its a fairly obscure option that I've never made use of before, but might offer an easier way of changing external PRs without a whole bunch of merges. That may not answer the docs building question though.

While we're on the subject of contributing guidelines, I noticed an outdated link in the guidelines for opening a new [bug report](https://github.com/MRtrix3/mrtrix3/issues/new?assignees=&labels=bug&template=bug_report.md&title=%5Bcommand%5D%3A+%5Bbug+report+title%5D). https://mrtrix.readthedocs.io/en/latest/troubleshooting/advanced_debugging.html is dead, I guess the equivalent wiki page is [this one](https://community.mrtrix.org/t/advanced-debugging-of-mrtrix3-binaries/2555)?

jdtournier:
Looks good, no issues from my point of view. The only thing I would say is that starting with an issue before making changes might be advisable to avoid the faff that switching merge targets can otherwise entail. If the user makes changes based on `master` and submits a PR, but we subsequently suggest this should be merged to `dev`, this can sometimes be a messy process (typically best done with a cherry-pick and a forced push, or a whole new PR...). It would be good to somehow convey this issue so users understand why starting with an issue might save them a lot of time down the track... 

And I also agree that "allow edits by maintainers" feature should be encouraged! It would make all of our lives much easier if we could push suggested changes directly to the relevant branch... 

In terms of generating the docs on other branches: it's trivial for branches on our repo, not so trivial if they're on other repos, as far as I can tell. It can be done, but it essentially involves setting up a whole new project on readthedocs - not difficult, mind you, maybe we can provide that as a suggestion? Otherwise, the simplest would indeed be to duplicate the user's branch on our repo, and yes, the simplest might be to re-use the `doctest` branch since it is automatically built (though I might disable it on the main readthedocs site, and enable it on the [mrtrix-dev project](https://mrtrix3-dev.readthedocs.io/en/latest/) instead. We can have any branch we want built, we just need to enable it on the readthedocs site. 

And thanks for spotting the dead link, @fionaEyoung! I think the link you've identified is indeed the relevant one. We'll get that fixed ASAP - unless you beat us to it... ;)

fionaEyoung:
Aha! I hadn't realised editing an issue template was even an option...

Lestropie:
> The only thing I would say is that starting with an issue before making changes might be advisable to avoid the faff that switching merge targets can otherwise entail.

Agreed; revised.

> And I also agree that "allow edits by maintainers" feature should be encouraged! 

Given this can be engaged after the fact, I don't think it needs to appear in the contribution document; we can simply make the request if need be.

> In terms of generating the docs on other branches:

I've removed that component entirely. Somehow forgot the fact that one can do a [local rendering](https://github.com/MRtrix3/mrtrix3/blob/master/docs/README.md) for such verification purposes.

jdtournier:
Looks good to me, thanks! 

One minor point though: while local rendering is possible, I've certainly had issues with it from time to time (probably mostly due to using a bleeding-edge distro like Arch Linux, though). Pushing to a branch and having readthedocs render it is much easier, especially for users less familiar with `python` and `pip`. But then, the right place to mention this would be in that `docs/README.md` document, not necessarily in the issue template...

jdtournier (approved):

Lestropie:
If there's a reasonable chance of a rendering issue, we can always at our end generate a new branch here based on the fork and set that up to render on readthedocs. But it's maybe too esoteric an issue to be waffling on about in the contributing guidelines.

PR #2406, opened 2021-11-29
dev update

jdtournier:
Regular merge of `master` into `dev`, this time to deal with the changes made to deal with the many `pylint` 2.11.1 warnings. 

One point worth mentioning is that I've left `consider-using-f-string` disabled, since these were introduced in Python 3.6. As far as I can tell, we can't use them without breaking compatibility with Python 3.5 or earlier. We are dropping support for Python 2, but I'm not sure we should immediately require Python 3.6+. Open for discussion...


Lestropie:
> I've left consider-using-f-string disabled, since these were introduced in Python 3.6.

:+1:

Lestropie (approved):
Looks right to me.

PR #2423, opened 2022-02-07
tensor2metric: avoid crash when supplying input with wrong size

bjeurissen:
Tensor2metric was segfaulting with multiple users on the forum when provided DWIs instead of tensors. On my own system the command even became unresponsive... This change checks if the input consists of 6 volumes and thus can be considered a tensor.

### DIFF
--- a/cmd/tensor2metric.cpp
+++ b/cmd/tensor2metric.cpp
@@ -145,7 +145,7 @@ class Processor { MEMALIGN(Processor)
 
       /* input dt */
       Eigen::Matrix<double, 6, 1> dt;
-      for (auto l = Loop (3) (dt_img); l; ++l)
+      for (dt_img.index(3) = 0; dt_img.index(3) < 6; ++dt_img.index(3))
         dt[dt_img.index(3)] = dt_img.value();
 
       /* output adc */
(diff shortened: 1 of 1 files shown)


jdtournier:
I agree this needs a fix, and this may be the simplest way to dealing with this. Would this cause any issues when producing DKT outputs from `dwi2tensor`? I think they're currently kept as separate files, but I have a feeling we'd discussed storing them within the same file at some point in the past... (?)

I also reckon it might be worth trying to identify what it is that's causing the program to hang or segfault. We don't typically have issue with too many volumes, normally the problem is too few causing overruns. Do we have any insights into what's actually causing the failures? And is capping the number of input volumes sufficient to avoid these problems?

jdtournier:
Quick look at the code: I reckon [this would be the problematic line](https://github.com/MRtrix3/mrtrix3/blob/aba024b0f7600f9e60e51331fdbc85b76af6a092/cmd/tensor2metric.cpp#L148): it'll immediately cause a buffer overrun if `dt_img` has more than 6 volumes. So maybe that fix is enough to guarantee correct operation. I would be tempted to modify that loop to explicitly only pick out the first 6 volumes, just in case...?

jdtournier (approved):

bjeurissen (commented):
Looks great!

PR #2424, opened 2022-02-07
Avoid invalid memory access when ROI is malformed in tck header

jdtournier:
As [discussed on the forum](https://community.mrtrix.org/t/tck2fixel-segmentation-fault/2497/11), if a `tck` file header contains an entry for the ROI that doesn't have the expected 2 fields, this leads to a segfault when using e.g. `tckinfo` or `tck2fixel`. For some reason, the user had a header entry of the form:
```
ROI: none
```
Whereas we would normally write something like:
```
ROI: seed mask.mif
```
i.e. two separate entries. 

The change here consists simply of making sure that if only one token is parsed, a second empty string is added as the second, which then avoids the invalid memory access, and permits these headers to be used (the alternative is simply to ignore them, but I'm not sure that's the best thing to do here).

### DIFF
--- a/src/dwi/tractography/file_base.cpp
+++ b/src/dwi/tractography/file_base.cpp
@@ -36,6 +36,8 @@ namespace MR {
           if (key == "roi" || key == "prior_roi") {
             try {
               vector<std::string> V (split (kv.value(), " \t", true, 2));
+              if (V.size() < 2)
+                V.push_back (std::string());
               properties.prior_rois.insert (std::pair<std::string,std::string> (V[0], V[1]));
             }
             catch (...) {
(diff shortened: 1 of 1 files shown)


Lestropie:
Especially when the content is filling `prior_roi` with the string "`none`", ignoring makes more sense to me than conditioning. There are multiple other instances in the software where a malformed key-value specification results in a `WARN()`-level notification that the entry is being ignored.

jdtournier:
@Lestropie, just to make sure we're on the same page: is your preference here to ignore the entry altogether, rather than preserve the `none` token even if malformed? Happy with that too - in fact, that was the first thing I did. Just thought it might be better to preserve the info, but I agree if it's malformed, there's a pretty good argument for stripping it out!

Lestropie:
I would indeed ignore malformed tokens, regardless of the content. But the specific malformed token "`none`" is a particularly ironic token to go to the effort of preserving, given that an empty list conveys exactly that information (and arguably more accurately, since the content of `prior_rois` could stack across sequential commands, and then "`none`" wouldn't make much sense).

Lestropie (approved):
:+1:

PR #2427, opened 2022-02-10
DICOM: Change slice timing priority

Lestropie:
Reported by @bjeurissen.

Between two DWI series, one stored in mosiac format and one not, the current code produces a clearly erroneous slice timing vector in the latter case. The proposed change results in the use of DICOM field `AcquisitionTime` in the latter case, which produces an identical result between the two datasets (ignoring floating-point precision, since slice timing for mosaics is reported and propagated as a string).

Outstanding questions:

-   `dcm2niix` reports fractionally different slice timing for the non-mosaic dataset. I don't see any obvious source for this discrepancy. It's possible that it is deriving timings to a greater precision than what is reported in the DICOM headers, but I'm not sure.

-  I don't recall whether the CSA `TimeAfterStart` source was prioritised over `AcquisitionTime` due to behaviour in some other dataset, and therefore whether this change may result in erroneous behaviour elsewhere. @jdtournier Would your private DICOM test suite pick up such a change?

### DIFF
--- a/core/file/dicom/mapper.cpp
+++ b/core/file/dicom/mapper.cpp
@@ -308,13 +308,6 @@ namespace MR {
               slices_timing_str.push_back (temp);
             }
           }
-        } else if (std::isfinite (frame.time_after_start)) {
-          DEBUG ("Taking slice timing information from CSA TimeAfterStart field");
-          default_type min_time_after_start = std::numeric_limits<default_type>::infinity();
-          for (size_t n = 0; n != dim[1]; ++n)
-            min_time_after_start = std::min (min_time_after_start, frames[n]->time_after_start);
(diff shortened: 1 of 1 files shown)


Lestropie:
I wonder if it's worth generating slice timing vectors from all plausible sources, and then checking for supra-threshold differences between them. A warning could be issued if there's a discrepancy greater than some threshold, and the source to prioritise in that instance could be set via a config file entry. Might then get some community feedback on whether certain sources are better than others, or receive exemplars of when one fails.

Lestropie:
@jdtournier This needs a run against your private DICOM tests; particularly interested if anything spits out a different slice timing vector to what it had previously.

Also @bjeurissen if you have any sample data where you recall that MRtrix was getting this wrong, would be useful to add it to that test suite.

PR #2430, opened 2022-02-16
fixelconnectivity: Normal image handling

Lestropie:
Hopefully results in better behaviour in the scenario reported on the [forum](https://community.mrtrix.org/t/fixelfilter-error/5469); also removes a lot of unnecessary custom code.


jdtournier (approved):
Sounds sensible - though I can't really assess correctness properly, I'm relying on the CI tests here...

PR #2431, opened 2022-02-16
labelconvert: Throw exception on empty mapping

Lestropie:
Triggered by [forum post](https://community.mrtrix.org/t/about-how-to-use-the-template-to-create-the-connectome/5028), though it's hardly the first time it has happened.

### DIFF
--- a/cmd/labelconvert.cpp
+++ b/cmd/labelconvert.cpp
@@ -97,7 +97,7 @@ void run ()
 
   // Build the mapping from input to output indices
   const auto mapping = get_lut_mapping (lut_in, lut_out);
-  if (!*std::max_element (mapping.begin(), mapping.end()))
+  if (*std::max_element (mapping.begin(), mapping.end()) == 0)
     throw Exception ("Mapping between input and output LUTs is empty, i.e. no common node names between these two LUTs");
 
   // Modify the header for the output file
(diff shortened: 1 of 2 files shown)


cmd/labelconvert.cpp
```diff
@@ -97,6 +97,8 @@ void run ()
 
   // Build the mapping from input to output indices
   const auto mapping = get_lut_mapping (lut_in, lut_out);
+  if (!*std::max_element (mapping.begin(), mapping.end()))
```
jdtournier, line 100: Shouldn't that be:
```
if (std::max_element (mapping.begin(), mapping.end()) == mapping.end()) 
```
From [cppreference](https://en.cppreference.com/w/cpp/algorithm/max_element):

> **Return value**
> Iterator to the greatest element in the range `[first, last)`. If several elements in the range are equivalent to the greatest element, returns the iterator to the first such element. **Returns `last` if the range is empty**.

jdtournier: OK, so you already know it's non-empty container, and that if the mappings don't match at all, the largest value stored in there will be the zeros you've initialised it with. Makes sense. 

Minor (optional) suggestion to change to:
```
  if (*std::max_element (mapping.begin(), mapping.end()) == 0)
```
To make the intention a touch clearer (i.e. we know we're not checking a pointer here). 

Also: if you find the time to add a test for this to the CI, that would be great too. :+1:

jdtournier (approved):

Lestropie:
PS. Extra `labelconvert` test added in b6c4f496 fails on `master`.

PR #2433, opened 2022-02-18
tensor2metric: Check input data

Lestropie:
Implemented based on [forum post](https://community.mrtrix.org/t/tensor2metric-system-fatal-code-sigsegv-11-segmentation-fault-invalid-memory-access/5521).

If there were a wider set of relevant commands, it would be preferable to have this as an e.g. `check()` function in a `Tensor` namespace. The only other context I can think of is loading tensor images in the `mrview` ODF tool.

### DIFF
--- a/cmd/tensor2metric.cpp
+++ b/cmd/tensor2metric.cpp
@@ -273,8 +273,10 @@ class Processor { MEMALIGN(Processor)
 
 void run ()
 {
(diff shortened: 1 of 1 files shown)


bjeurissen:
I believe this was already addressed in #2423

Lestropie:
OK; need to do a merge of `master` into `dev` then.

Issue #2261, opened 2021-01-19
Python3 shebang for 3.1.0


Lestropie:
_Originally posted by @bjeurissen in https://github.com/MRtrix3/mrtrix3/issues/2047#issuecomment-763199900_

See #2047 and #2215.

D'Oh. Better to discover such while it's only on `dev` though.

1. We can just revert to `#!/usr/bin/env python3`, despite the recommendations against use of `env`;
2. [This comment](https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/#comment-504) on the page Donald linked earlier suggests using some shell logic within the shebang itself to determine the appropriate interpreter;
3. The recommendation to incorporate into the build system isn't ideal for us given we're now distributing pre-compiled code, i.e. we can't rely on `configure` or `build`. If however we had a script that set script shebangs, then `configure` could use logic of any arbitrary complexity to try to decide this, and those installing from pre-compiled packages could still use it if the default doesn't work on their system.
4. Still not a fan of requiring users to explicitly specify the interpreter each time...
5. Any other options still on the table?

daljit46:
Given that we merged #2958, can we close this?

Lestropie:
#2958 only deals with how a Python script already being executed will behave as it executes other Python scripts. It doesn't modify what will happen when an MRtrix3 Python script is first executed by the user from a shell.

With the CMake-generated Python executables produced in #2920, there is technically the option of hard-coding the Python interpreter of choice into the shebang of those short files; this is how FSL gets their code to be interpreted by their custom Conda environment. However I don't think we want to be doing the same thing ourselves. For one, it wouldn't work for shipped compiled tarballs.

Given [your own prior comment](https://github.com/MRtrix3/mrtrix3/issues/2261#issuecomment-2100332887), I had presumed that you would prefer those CMake-generated executables to change to `#!/usr/bin/env python3`. `3.0.x` runs `#!/usr/bin/env python`, which has become problematic; on `dev` we shifted to `#!/usr/bin/python3` rather than `#!/usr/bin/env python3` based on [this comment](https://github.com/MRtrix3/mrtrix3/pull/2215#issuecomment-718131291), but you had made an argument to the contrary at some point. So I'm quite content to change them all to `#!/usr/bin/env python3`, which should now only require a change in the CMake Python executable generation function.

daljit46:
[...]

PR #2437, opened 2022-02-25
Npy file format support

Lestropie:
Superset of #2436. I'll leave that open for now, but showing this PR might help to more clearly provide a justification for those changes.

Listed as draft PR for now as I would like to add some unit tests, which will require additions to the test data repo.

Discussion points:

1.  The functions for loading & saving matrix data from / to the filesystem are here proposed to be moved from the `::MR` namespace in file `core/math/math.h` to the `::MR::File::Matrix` namespace in file `core/file/matrix.h`. I always found it unusual to have those functions in our base namespace, as well as having those functions defined in a header file where the content of the first half is so drastically different and in no way related to filesystem interfacing. While this change would break backward compatibility with any external C++ projects, it to me feels like a much more appropriate location, and external projects could get away with testing the content of `MRTRIX_MINOR_VERSION` at compile time if preserving `3.0.x` compatibility is desired.

1.  Support for 16-bit floating-point data is via a single header include file, which is integrated into the repository, just as was done previously with JSON for Modern C++. It however occurs to me that the utilisation of such libraries is not broadcast to the same extent as are external dependencies that are listed as prerequisites for compilation. I wonder if we should have somewhere listed all packages on which the software depends, regardless of whether they are compile-time dependencies or imported directly into the repository.

1.  16-bit floating-point support is currently implemented in a manner specific to the NPY file format for matrix or vector data. It does however occur to me that `Float16` could theoretically be just another type in `MR::DataType`, with get/set functions that use the new 16-bit floating-point library, and the MRtrix image format would inherit support for such. That would potentially simplify the NPY code slightly, but would open up a whole bunch of requisite testing to make sure it doesn't break something else (e.g. no idea how easy or difficult this data type might be to support in `mrview` without trying it).

1.  There is scope to speed up the NPY load / save operations in cases where data type /  endianness / row vs. column ordering is identical between disk and RAM, by doing a direct `memcpy()` with the memory-mapped region.


jdtournier:
> The functions for loading & saving matrix data from / to the filesystem are here proposed to be moved from the `::MR` namespace in file `core/math/math.h` to the `::MR::File::Matrix` namespace in file `core/file/matrix.h`. 

:+1: Happy with that.

>  I wonder if we should have somewhere listed all packages on which the software depends, regardless of whether they are compile-time dependencies or imported directly into the repository.

Fully agree. I've had similar thoughts in the past, but not done anything about them... External code that we import wholesale includes the [original NIfTI headers](https://nifti.nimh.nih.gov/), glLoadGen to generate the `src/gui/opengl/gl_core_3_3.*` files (though the project seems defunct now - [this is the only reference](https://github.com/SFML/SFML-glLoadGen) I can find to it on the web...), along with [JSON for Modern C++](https://github.com/nlohmann/json). There may be others I've forgotten about... 

>  It does however occur to me that `Float16` could theoretically be just another type in `MR::DataType`, with get/set functions that use the new 16-bit floating-point library, and the MRtrix image format would inherit support for such.

Yes, that sounds like the most sensible way to do it!

> but would open up a whole bunch of requisite testing to make sure it doesn't break something else (e.g. no idea how easy or difficult this data type might be to support in mrview without trying it).

I wouldn't worry about MRView - as long as the internal in-RAM representation is one it can deal with, it doesn't matter how it was stored on file. I'd be surprised if that broke anything...

core/datatype.h
```diff
@@ -144,8 +145,9 @@ namespace MR
       static constexpr uint8_t     UInt16        = 0x03U;
       static constexpr uint8_t     UInt32        = 0x04U;
       static constexpr uint8_t     UInt64        = 0x05U;
-      static constexpr uint8_t     Float32       = 0x06U;
-      static constexpr uint8_t     Float64       = 0x07U;
+      static constexpr uint8_t     Float16       = 0x06U;
+      static constexpr uint8_t     Float32       = 0x07U;
+      static constexpr uint8_t     Float64       = 0x08U;
```
Lestropie, line 150: I don't think there's any context in which these numerical identifiers are important, the datatype is always translated to string and back again; so putting them in this order shouldn't be a problem?

core/fetch_store.cpp
```diff
@@ -29,14 +32,14 @@ namespace MR
     template <typename TypeOUT, typename TypeIN>
       inline typename std::enable_if<std::is_floating_point<TypeOUT>::value, TypeOUT>::type
       round_func (TypeIN in, typename std::enable_if<std::is_arithmetic<TypeIN>::value>::type* = nullptr) {
-        return in;
+        return TypeOUT(in);
```
Lestropie, line 35: `half_float::half` for whatever reason didn't like these being implicit.

Lestropie:
16-bit floating-point images working, and appear properly in `mrview`:

```
$ mrconvert dti_FA.nii.gz -datatype float16le - | mrinfo -
mrconvert: [100%] uncompressing image "dti_FA.nii.gz"
mrconvert: [100%] copying from "dti_FA.nii.gz" to "./mrtrix-tmp-Oi6GFd.mif"
Image name:          "./mrtrix-tmp-Oi6GFd.mif"
  Dimensions:        104 x 104 x 72
  Voxel size:        2.01923 x 2.01923 x 2
  Data strides:      [ -1 2 3 ]
  Format:            Internal pipe
  Data type:         16 bit float (little endian)
  Transform:               0.9953    -0.08673     0.04235      -90.83
                          0.09543      0.9501     -0.2969      -106.5
                         -0.01449      0.2996       0.954        -127
  command_history:   C:\msys64\home\rob\src\mrtrix3\bin\mrconvert.exe dti_FA.nii.gz -datatype float16le -  (version=3.0.3-346-g0dc2a396-dirty)
  comments:          FSL5.0
  mrtrix_version:    3.0.3-346-g0dc2a396-dirty
```

Lestropie:
Worth noting on this PR, given its import of the [`half`](https://half.sourceforge.net/) header-only library, that [C++23 integrates a 16-bit floating-point format into the standard](https://en.cppreference.com/w/cpp/types/floating-point).

Lestropie:
Still having some outstanding issues with CI on this PR.

I had a couple of issues compiling on Mac that have been resolved. One of them I'm still a little confused by: it didn't like compiling `File::NPY::load_matrix<size_t>()` because it claimed `__set_store_functions<unsigned long>()` didn't exist, even though the functions for `uint64_t` should all have been explicitly instantiated for the sake of image IO. But I got around it by just not calling `load_matrix<size_t>()`, and instead typedef-ing an `index_type` as `uint32_t` for use all throughout the relevant code.

Where it's still stuck now is my new unit tests for NPY support, which require importing `numpy` in CI (I both generate NPYs in C++ and read them in Python, and generate NPYs in Python and read them in C++).
-   On Mac, it's added to the list of packages to install, and looks to be installed just fine, but then when the corresponding script is executed Python claims there's no such module. Maybe a Mac person can help me with that one.
-   On MSYS2, `numpy` refuses to install. By default, it's using the `msys` version of Python. There's no `pacman` package for an `msys` version of `numpy`. So I'm trying to install `numpy` through `pip`. But it fails, regardless of whether or not I upgrade `pip` beforehand. I have alternatively tried instead installing both Python and `numpy` through the `mingw64` channel; but in that case Python scripts just don't execute at all.

fionaEyoung:
> On Mac, it's added to the list of packages to install, and looks to be installed just fine, but then when the corresponding script is executed Python claims there's no such module. Maybe a Mac person can help me with that one.

the shebang in `testing/bin/testing_check_npy`: 
```
#!/usr/bin/python3
```

⬆️ will be the macos-latest xcode python version (3.8 or 9 I think?) and will look for packages in somewhere like 
/Library/Frameworks/Python*.framework/Versions/3.9/lib/python3.9/site-packages 

but it looks like macos-latest also has homebrew python (3.11) already installed to /usr/local/bin/python3 and the associated `pip3` will install to /usr/local/lib/python3.11/site-packages ( or possibly /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages I'm pretty lost about where / why homebrew puts things sometimes.)

this will be the python3 invoked by `python3 -m ensurepip && pip3 install numpy`

So either

`/usr/bin/pip3 install numpy ` and  `#!/usr/bin/python3`

or

`pip3 install numpy` and `#!/usr/bin/env python3`

should sort it out

Lestropie:
Thanks @fionaEyoung and @daljit46 for tips on the Python multiple version issue; Mac checks are now working. There, I've chosen to stick to the same shebang for these unit test files as the Python executable scripts (`#!/usr/bin/python3`).

Unfortunately MSYS2 has a similar issue, and that one may be harder to fix. `/usr/bin/python3` corresponds to the MSYS2 version of Python, which unlike the MinGW64 version, not only does not have a numpy package that can be installed through pacman, but pip fails to install the module even with `python-dev` installed. Possible strategies:

1.  On the CI machine, uninstall the MSYS2 version of Python and softlink `/usr/bin/python3` to `/mingw64/bin/python3`.
2.  Permit the NPY unit tests to fail if `numpy` is not installed.
3.  Find some other solution to get `numpy` working on the MSYS2 version of Python.

Regardless, we will need to reconsider the hardcoded shebang on `dev` (#2261, #2047).

daljit46:
Note, that we are no longer relying on `MINGW64`, instead the CI checks now use the newer `UCRT64` environment (which uses the `ucrt` C library instead of the old `msvcrt`).

I think symlinking could be the least disrupting option here. Also even if we manage to successfully install the `numpy` package for MSYS2, it would noticeably increase the CI checks time as the package needs to be built from source (there is no pre-packaged `numpy` for `MSYS`).

I personally think we should change the shebang for the scripts as mentioned in #2261. Relying on hardcoded paths is not appropriate in my opinion, and I would rather use `#!/usr/bin/env python3` to give the user the possibility to choose what python executable they wish to use. I see very little benefit in using `#!/usr/bin/python`.

Lestropie:
> even if we manage to successfully install the numpy package for MSYS2, it would noticeably increase the CI checks time

If that's the case, that would IMO be an argument for bypassing those checks in the case where `numpy` is not installed. As long as the relevant unit tests get executed in at least one environment it should be fine; it's not something for which I expect the behaviour to differ between OS's.

> I personally think we should change the shebang for the scripts as mentioned in https://github.com/MRtrix3/mrtrix3/issues/2261.

I think I'm leaning in the same direction, though I haven't re-familiarised myself with the debate that led to the current decision. Importantly "`#!/usr/bin/python3`" is only on `dev` and is not included in any tag. But we can flesh that out in #2261, and figure out there what the consequences might be in the context of these changes.

Issue #2172, opened 2020-09-15
Robust handling of Philips data w. trace image


Lestropie:
As reappeared on [forum](https://community.mrtrix.org/t/dwi2response-error-iter0-fixel-peaks-mif/4010), but has apparently been observed many times.

If a volume is present with a null vector direction but a *b*-value equivalent to a valid shell, particularly if it is the last volume in the series, it's likely to be Philips data with a reconstructed trace-weighted image. Such a volume needs to be excluded from any A2SH transformation at least; possibly other components would need to detect & handle such data as well, someone would need to run a suite of commands to see what breaks and what doesn't.

jdtournier:
OK, after looking into this in more detail, prompted by [requests from the forum](https://community.mrtrix.org/t/dwipreproc-problem-dwipreproc-853/2887/6?u=jdtournier), I think we may have a solution. 

First off, I don't think we'll be able to keep these trace-weighted volumes in the data and handle them downstream, relying on the specific applications to detect and ignore them. The problem is that following the changes in #2015, the DW encoding table gets cleaned up and stored in a format that can be assumed ready for analysis; these trace-weighted volumes with encoding `[ 0 0 0 b ]` are indistinguishable from a _b_=0 that might have been acquired via modulation of the gradient amplitudes on systems that only allow for a single _b_-value - and will hence be immediately converted to actual _b_=0 volumes if _b_-value scaling is on (the default). The user will be a warning about it, but I don't expect many users to really look at this too closely. In short, I think these have to be kept separate from the actual measured DWI data or removed altogether. 

Having rummaged around these types of data a bit, it turns out there's a special Philips private DICOM entry `2001 1004 DiffusionDirection` which takes value `'O'` for regular orientational DWIs, and `'I'` for isotropically-weighted images. Only slight snag is that genuine _b_=0 images are also labelled `'I'`, so we need to be a bit more careful when stripping out the synthetic trace-weighted images. Basically, look for those labelled `'I'` with _b_>0.

Ideally, the best way to do this would be to split the series into the DWI data as measured, and the synthetic trace-weighted images, as separate outputs that the user can select. However, while technically possible, this would require a fairly drastic overhaul of the current approaches we use to do this (which is currently based solely on the `0008 0008 ImageType` entry). Notably, the backend assumes that there is one ImageType per file, and so isn't designed to allow a multi-frame image to be split - and yet, I have examples of Philips multiframe data with these trace-weighted images in it... 

[...]

PR #2442, opened 2022-03-04
Various DICOM updates

jdtournier:
Simple fix for [issue reported on the forum](https://community.mrtrix.org/t/problem-with-sorting-of-dicom-images-using-mrconvert/5493?u=jdtournier), which lead to incorrect sorting of dMRI data for some Philips datasets.


Lestropie (commented):
Looks limited to Philips and DWI, and the field name (already present in our dictionary) certainly checks out.

Having multiple DICOM fields write to the same member variable however makes me nervous. In similar circumstances for other types of data (e.g. slice timing comes to mind), what I've done is to have different DICOM fields write to different member variables, and then only at the point of constructing the series, interrogate which fields have been filled comprehensively by which sources, and set the precedence of which shall be utilised if more than one source is available. Would this be a safer approach here?

jdtournier:
I had similar reservations, but convinced myself that this particular variable (`sequence`) is currently only used in one other contexts, which is specific to older Siemens data - where the only way to sort the series correctly was by interpreting the sequence name string (0x0018, 0x0024 - `SequenceName`), which had a format like `b1000#2` where the last number indicates which volume the slice belongs to. 

But thinking about it further, you might be right that in some contexts where the `SequenceName` entry is set to a parsable value, this might cause issues. I'll add another variable for this to avoid this kind of issue.

jdtournier:
We've also got much bigger problems in that the DICOM test suite fails currently, on `master` and this pull request. I have a feeling this might date back to #2015 - which was quite a while ago... Maybe there's some other more recent change that would explain this, but nothing stands out right now...

jdtournier:
You'll note that I've added a solution to #2172 in that last commit (5df4c98) - see [this comment](https://github.com/MRtrix3/mrtrix3/issues/2172#issuecomment-1064307374) for a full description of the thought process...

jdtournier:
Turns out that last change breaks a lot of the testing, since the DWI dataset we use for testing was affected... Just to clarify:

#### current master:
```
$ mrinfo testing/binaries/data/dwi.mif -dwgrad
                   0                    0                    0                    0
                   0                    0                    0                    0
 -0.0509540919293102    0.061755090218525   -0.996789842117064     2950.00093450932
  0.0119069972166564    0.955046776751157    0.296215930757461       3000.001402542
   -0.52511512453343    0.839985199206294    0.136671032412154       2999.998577073
  -0.785445219860643    -0.61110017105824  -0.0981447274725243     2999.99832048927
  0.0608619958084041   -0.456700968546777    0.887535938874958       3000.000413223
   0.398325005929864    0.667699009940035    0.628900009362435       2999.999910678
  -0.680603594422654     0.68964458903505   -0.247323852617658       3000.003575451
   0.237398967996455    0.969994869235849   0.0524564929283866     3000.00080885475
...
```


#### this branch:
```
$ mrinfo testing/binaries/data/dwi.mif -dwgrad
                   0                    0                    0                    0
                   0                    0                    0                    0
 -0.0509540919293102    0.061755090218525   -0.996789842117064                 2950
  0.0119069972166564    0.955046776751157    0.296215930757461                 3000
   -0.52511512453343    0.839985199206294    0.136671032412154                 3000
  -0.785445219860643    -0.61110017105824  -0.0981447274725243                 3000
  0.0608619958084041   -0.456700968546777    0.887535938874958                 3000
   0.398325005929864    0.667699009940035    0.628900009362435                 3000
  -0.680603594422654     0.68964458903505   -0.247323852617658                 3000
   0.237398967996455    0.969994869235849   0.0524564929283866                 3000
...
```

Guess we're going to have to update some of the test data as well...

jdtournier:
Right, looks like this is pretty much good to go. Decision required as to whether to go with this PR (see description in 4223ae6) or #2450. I personally favour #2450, I expect everyone else will too... Good news is both approaches check out, which gives us confidence that the updated test data on #2450 is correct. Can I get someone to approve #2450 (or this one if that's the preferred option)?

Quick recap of the changes here: 
- fix sorting issue on some Philips data (92e4885, f48553c3)
- strip out trace-weighted scan from Philips data (with option to preserve) (5df4c98d)
- fix bug in detecting of whether b-value scaling should be applied (e105aec0)

I realise these should probably all be merged as independent PRs, but this should be OK as a single PR if I edit the title to e.g. 'various DICOM updates', right...? :pray:

Lestropie:
@jdtournier Close?

jdtournier:
:+1:

PR #2444, opened 2022-03-08
build: fix pylint warnings

jdtournier:
To fix pylint warnings which currently cause CI to fail


jdtournier:
Right, once #2445 is merged (assuming we're all happy), I'll merge all changes on `master` back to `dev`, and merge this into it - should sort out the remaining copyright issues. Once that's done, we can try merging #2443...

PR #2445, opened 2022-03-08
configure: ignore CFLAGS & LDFLAGS from qmake

jdtournier:
The LTO options introduced with Qt 5.15 cause non-trivial conflicts with clang++ (at least it does on Arch Linux). I expect this will start to become problematic for other users soon. 

Looking at the flags added by Qt, it looks like they don't add much over what we normally use. It's probably just as easy to ignore the additional flags that `qmake` suggest (`CFLAGS` & `LDFLAGS`), and stick to grabbing only the main ones (`DEFINES`, `INCPATH` & `LIBS`). 

Works fine on Linux, will need checking on other OS's.

### DIFF
--- a/configure
+++ b/configure
@@ -1291,21 +1291,15 @@ int main() { Foo f; }
               line = line.strip()
               if line.startswith ('DEFINES'):
                 qt_defines = shlex.split (line[line.find('=')+1:].strip())
-              elif line.startswith ('CXXFLAGS'):
-                qt_cflags = shlex.split (line[line.find('=')+1:].strip())
               elif line.startswith ('INCPATH'):
                 qt_includes = shlex.split (line[line.find('=')+1:].strip())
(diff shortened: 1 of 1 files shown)


jdtournier:
Just to clarify: this will only affect the compiler and linker flags used while compiling Qt / GUI apps. No impact on command-line only apps. I'll check this on Windows 10 as well, that's the OS most likely to break with these changes.

maxpietsch:
works with:
```
OS: darwin
Apple clang version 13.0.0 (clang-1300.0.29.30)
Qt 5.15.2
```

jdtournier:
Works fine on Windows 10 under MSYS2 as well. I think it's safe to merge. Can I get a review please...? :grin:

maxpietsch (approved):

PR #2447, opened 2022-03-10
tcksift2: Fix fixel masking

Lestropie:
Collaborator encountered the situation that when `tcksift2` is run with `-nthreads 0`, it does nothing. It turns out that I buggered up the calculation of fixel track counts & track densities: the multi-threaded class destructor would mutex and increment the track density of the shared class, but this was erroneously interpreted as being the track density (i.e. intersection length) of a single streamline. As a result, the model thought that the number of streamlines in each fixel was equal to the number of threads. This is only consequential at the commencement of iterations, where any fixel with only 1 solitary streamline is excluded from the optimisation. Because of the bug, this meant that if only 1 thread was used to map streamlines to fixels, the model would exclude all fixels from the optimisation.

To demonstrate, the `tcksift2` test added here fails on `master`.


jdtournier (approved):
OK, not sure I get the subtleties here. But everything seems sensible, and as long as the tests check out, it's all good as far as I'm concerned. 

Just one quick question: does this not impact on the validity of previous results, even when running with more than one thread? It seems like quite a fundamental change if the quantity being updated in the destructor is not the right one...

Lestropie:
It means that previously, fixels to which only one streamline was assigned would nevertheless still be included in the optimisation, despite my intention for them not to be. So technically yes, it may change outcomes slightly, but I nevertheless think it's a patch fix.

Issue #2435, opened 2022-02-21
ENH Universal interface for scalar vector files


Lestropie:
This is something I've had in my mind for quite some time, but seemingly never got around to listing.

There are a range of applications that involve a 1D array of scalar data. Fixel data files, SIFT2 track weights, `tcksample` outputs, etc.. Just like for images, and has been proposed for some time for tractography data (#411), it would be good to have format handlers for such data, such that (where coded for such) different file formats could be used by *MRtrix3* commands depending on user input. This could additionally be extended to support NumPy arrays, which are an appealing alternative to reduce the storage space utilised for data of this nature (and floating-point precision of such on save could be a config file entry).

Issue #1664, opened 2019-07-31
Better handling of < 3D images


Lestropie:
There are a number of places in the core libraries that make assumptions about any image with less than three dimensions being somehow invalid. This leads to e.g. fixel data files that have sizes `Nx1x1` as opposed to just a 1D image with `size(0) = N`, and can lead to erroneous warnings about invalid transforms / voxel sizes. I'm encountering similar when going over #1541, where (greyscale) PNGs are by design 2D.

It would be nice if at some point we could find the areas in the library that either break or explicitly intervene when provided with < 3D data, and update them to either handle such appropriately or abort due to their incompatibility. E.g. One could envisage `mrview` issuing an immediate error message if attempting to open a fixel data file as the main image, rather than displaying a 500,000x1x1 texture; or the image transform being ignored entirely at both read & write in any < 3D image header rather than being set to identity.

Lestropie:
I think missing from this issue that I can't find posted anywhere else:

If <3D, I'd prefer for the resulting images to not contain voxel spacings or affine transform. These make no sense, and being absent entirely rather than set to "default" / "invalid" would to me make more sense. But it'd require a fair bit of combing through the code and example usages to find situations where new branches need to be introduced.

jdtournier:
Just my 2 cents: voxel and transform both make sense in general for 2D slices - that's how they're provided in DICOM. I'd be be uncomfortable with removing that information.

Lestropie:
> voxel and transform both make sense in general for 2D slices - that's how they're provided in DICOM. I'd be very uncomfortable with removing that information.

That would still be preserved in the case of e.g. XxYx1; I'm thinking specifically data where the dimensionality has been explicitly reduced below 3.

-----

Bumping due to [forum post](https://community.mrtrix.org/t/fba-how-to-display-one-subjects-the-value-of-fd-fc-and-fdc/4094): attempting to load a fixel data file in the Overlay tool should, IMO, produce an error.

Lestropie:
P.S. There's potentially an alternative solution here.

Define a different file format, with a different file extension, but that shares almost all of the code with the MRtrix format. The only points at which the handling would diverge would be:

- One demands >= 3 axes, whereas the other is entirely OK with being a 1D file;
- One can be opened directly for visualisation in `mrview`, the other cannot;
- One must write & read transform & voxel spacing from the header, the other must not;
- One can be opened for the purpose of e.g. per-fixel information, the other cannot.

[...]

PR #2448, opened 2022-03-10
tcksift2: -output_debug to fixel directory format

Lestropie:
-   When `-output_debug` is specified, outputs will:
    - Be written to a user-specified sub-directory, rather than the current working directory.
    - Be in the fixel directory format if fixel data export is enabled (which it is by default at compile time).
    - No longer produce duplicates of data that are identical before and after filtering.
-   Fixes erroneous formatting of `tcksift2` terminal progress.
-   Adds an `INFO`-level message regarding the decrease in entropy resulting from the modulation of streamline weights. This is the appropriate way to quantify the way in which a filtered tractogram with many streamlines, but very few streamlines with large weights and most streamlines with very small weights, is comparable to a tractogram that simply has very few streamlines. Nothing is done with the data, it's just printed, but it may be an alternative way in which to look at the differential effects of regularisation.


Lestropie:
f82756f was necessary to pass binary test suite. For the test data, 1D fixel data files have the same voxel spacing as the index image, but contain the identity transform. This change simply reforms the modifications to the content of `core/fixel/helpers.h` to duplicate this behaviour.

If there is a desire for fixel data files to contain the same spatial position information as the fixel image to which they refer, then this would require setting both the voxel spacing and the transform, which currently is not achieved even in the test data. My own position is that these data are irrelevant for fixel data, and should either be ignored entirely or indeed expected to not be present in image headers for fixel data files (#1664), or it should be possible to use other formats for fixel data (#2435).

jdtournier:
Had a quick look, but 8137340 is a bit of a monster... Everything looks OK to me, but this all code that I'm not very familiar with, so I'm going to defer to your judgement on that one - I won't be able to review this in any meaningful sense of the word! But this is on `dev`, feel free to merge if you're happy with your changes.

Lestropie:
I might generate some test data for `-output_debug` for both `tcksift` and `tcksift2`.

jdtournier:
The main thing is to make sure normal operation is left unaffected. I don't think you necessarily need to add tests for debug outputs!

Lestropie:
Won't bother verifying contents of debugging outputs, but having the test executions use that option at least proves that it executes successfully, and also makes it easy for anyone to look at the contents generated for the test data.

Issue #2386, opened 2021-10-24
Matlab: Handle multi-line key-values


Lestropie:
[Re-appeared](https://community.mrtrix.org/t/write-mrtrix-m-error-using-fprintf/2580/3) recently on forum.

All Matlab functions should transparently handle key-value data that are multi-line.

PR #2455, opened 2022-03-21
Matlab: multiline key-value metadata #2386

fionaEyoung:
Added handling of struct field with cell array values, which otherwise cause an error as `fprintf` doesn't take cell array arguments. MRtrix3/mrtrix3#2386
Additionally switched to [dynamic fieldnames syntax](https://uk.mathworks.com/help/matlab/matlab_prog/generate-field-names-from-variables.html). 

Note I haven't made any changes to `write_mrtrix.m` which already [does the same thing](https://github.com/MRtrix3/mrtrix3/blob/ed9558402834e9fa97b21b941783bf02765ef30d/matlab/write_mrtrix.m#L122-L128) in a slightly different way.


jdtournier (approved):
Seems good to me, thank you for the contribution!

And I didn't know about the struct.(fieldname) Matlab syntax - guess there's still plenty to learn...

PR #2460, opened 2022-03-29
mrgrid and mrtransform: check spatial stride order

maxpietsch:
As raised on the [forum](https://community.mrtrix.org/t/mrtransform-does-not-output-same-data-strides-as-template-volume/1577/7), this checks the spatial strides of output and template image and, if no stride option was provided, warns if their order does not match.


jdtournier:
I like the idea of better documenting the strides issue, but I'm not sure a full-blown warning is a good idea. I have a feeling that will produce a lot of unnecessary warnings in users' existing pipelines, and raise a lot more questions on the forum. 

How about, instead of issuing warnings, adding a more explicit paragraph in the command description, making mention of NIfTI specifically and the issue of interoperability with other packages? Even better, add a few example usages, like we do with mrconvert ([code](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mrconvert.cpp#L85), [docs](https://mrtrix.readthedocs.io/en/latest/reference/commands/mrconvert.html#example-usages)), including how to deal with the NIfTI situation in external packages (i.e. the use case that triggered these changes on the forum). That way, there is a documented way to deal with this that users faced with this issue should hopefully come across (maybe with a quick search), and we don't issue potentially unnecessary warnings. 

Alternatively, I would be OK with issuing the warning if both reference and output images are in NIfTI format...

maxpietsch:
No strong opinion either way but I think a warning message is justified here, as the explicit operation to regrid to a reference image might be understood in different ways, the most conservative and potentially most common one including matching strides. As the definition of this operation is somewhat ill- or undefined and debugging these issues can take a long time (recent experience....) I think it's fair to use a warning level message in the same spirit as `-force`. 

Only triggering warnings for NIfTI gets rid of some unnecessary warnings but `.mif` get eventually converted to NIfTI for use in other packages and `mrconvert` and other commands changes the strides silently if necessary so not sure it solves the issue.

thijsdhollander:
For what it's worth, I think Romain has a good point, fundamentally.

On the one hand, yes, the common logic among commands might be that properties of the input are preserved as much as possible, which from that point of view "in principle" applies to the strides.

However, on the other hand, one can argue that the template here is an important part of the input: the fact that it happens to be provided via an option is just interface semantics; or a consequence of the fact that the "input" that tells us what/how the final grid should look like can come in a range of ways, of which some are inherently present in the other input, and hence the other possible ways are "optional", which then dictates the interface.  But if you forget the interface practicalities, then the scenario where a template is provided means that the template provides important input; i.e. the final grid.  The question then becomes whether you regard the strides as a property of "the grid" or not.  If a property of the grid, then one can argue the strides of the template should by default end up in the output, just like the grid's resolution, dimensions and transform do.  For the latter properties it's of course unavoidable, whereas for the strides it works both ways... in MRtrix.

Anyways, all that to say that it can easily be justified either way.  That said, I keep running into users that expect the strides to come along from the template; just last week in a group discussion at an event (and these were methods developers).  Some like Romain expect this explicitly indeed, knowing the ins and outs of strides very well.  Others, who don't know what strides even are, expect this implicitly and don't understand why their outputs don't work in other software; especially when those outputs come directly from `mrtransform` or `mrgrid`... i.e. when there's no notion of "another command might have messed it up in the meantime".

Yes, you can solve this with a warning in theory, but for those users who don't know what strides are, it won't give them peace of mind.  Even worse, you'll end up having to explain strides to those users that didn't even have a problem to begin with (i.e. when the warning wasn't practically relevant to them).  Then only making that warning appear for some file types is a bit like a patch on a patch; i.e. it only gets more complicated, as do the explanations when someone asks for it ("I only get this warning for file type X, and not Y; why?" ==> question isn't even about strides, yet the answer is a long story involving strides).

If you do go for a warning, I would argue you better simply _always_ make it appear (regardless of output file type), but maybe make it verbose enough so it says something like "this (warning) is typically only relevant in context of interoperability with other software"; maybe with an optional link to the history of file types, software, and their ways of handling strides... but I think the "accessible" warning message will be the most valuable part in this for most users.

jdtournier:
I have to admit, I'm also leaning towards matching the strides by default when the `-template` option is provided. It works as expected out of the box for NIfTI, and besides I don't think `-template` implies a great deal as to whether that we'll be using the strides of the input or of the template - if anything, I'd probably tend to lean towards using the strides of the template, to be honest. 

But I am wary of unwittingly turning volume-contiguous data into spatially-contiguous data, simply because the template provided was 3D or a spatially-contiguous 4D image (e.g. a NIfTI). I think it might be enough to grab the ordering of the spatial strides of the template, and apply them to the output, preserving the strides of the non-spatial axes. It'll make no difference for NIfTI outputs anyway (they don't support volume-contiguous storage). And if users do want to preserve the strides of the input, it would be trivial to supply that through the existing `-stride` option anyway. 

Since this would change behaviour a bit, I reckon such a modification would probably best be made on the `dev` branch. Thoughts?

thijsdhollander:
> But I am wary of unwittingly turning volume-contiguous data into spatially-contiguous data, simply because the template provided was 3D or a spatially-contiguous 4D image (e.g. a NIfTI). I think it might be enough to grab the ordering of the spatial strides of the template, and apply them to the output, preserving the strides of the non-spatial axes. It'll make no difference for NIfTI outputs anyway (they don't support volume-contiguous storage). And if users do want to preserve the strides of the input, it would be trivial to supply that through the existing -stride option anyway.

I think that sounds very sensible. There is a justifiable logic here too, if need be: there is no interpolation/resampling along the non-spatial dimensions in these commands, so the non-spatial dimensions can be seen as separate from the (spatial 3D) grid.  Put differently, only the spatial grid (which would then include the spatial strides) is pulled from `-template`.

Lestropie:
Replaced with #3186.

PR #2461, opened 2022-03-30
Container: Additional fixes

Lestropie:
Changes over and above #2370. Using that as the base branch in order to show differences, but will change base branch to `master` after that is merged.

3af9f6f6 contains two fixes:

- Singularity was refusing to run `mrview` due to an error I'd never seen previously. Don't know whether it is related to a change to the base container OS, or a change to the host system (which is the same one I first generated the container definition files); planning to also test on a different host system before merging.

- Building the Singularity container via the DockerHub image resulted in `mrview` not showing icons; yet executing `mrview` via Docker *did* show icons. Can't claim to know how that works, but it nevertheless looks like a missing SVG dependency.

I am still chasing issues with Docker running `mrview`. This morning it was giving me a bunch of error messages but nevertheless executing, whereas now this evening it's not working at all:

```
No protocol specified
qt.qpa.xcb: could not connect to display :1
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
```

Will continue chasing over the next few days.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,7 @@ RUN apt-get -qq update \
         libqt5core5a \
         libqt5gui5 \
         libqt5network5 \
+        libqt5svg5 \
         libqt5widgets5 \
         libquadmath0 \
         libtiff5 \
--- a/Singularity
+++ b/Singularity
@@ -39,7 +39,7 @@ Include: apt
     apt-get update && apt-get upgrade -y
 
 # Runtime requirements
-    apt-get update && apt-get install -y --no-install-recommends dc less libfftw3-bin liblapack3 libpng16-16 libqt5network5 libqt5widgets5 libtiff5 python3 python3-distutils zlib1g
+    apt-get update && apt-get install -y --no-install-recommends dbus dc less libfftw3-bin liblapack3 libpng16-16 libqt5network5 libqt5widgets5 libtiff5 python3 python3-distutils zlib1g
 
 # Build requirements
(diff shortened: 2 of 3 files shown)


Lestropie:
Error I'm currently getting when executin `mrview` through Docker (though seems to function fine):
```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
libGL error: MESA-LOADER: failed to retrieve device information
libGL error: Version 4 or later of flush extension not found
libGL error: failed to load driver: i915
libGL error: failed to open /dev/dri/card0: No such file or directory
libGL error: failed to load driver: i965
```

Lestropie:
Okay, cb64041 seems to resolve residual issues I was having running `mrview` through Docker on my Arch Linux laptop. Switching from draft to PR, though currently targets the #2370 branch, so can either merge this then #2370, or change target branch here to `master` and close #2370 without merging, or merge #2370 and then change the destination branch here to `master`.

Would appreciate testing of `mrview` on a wider range of host systems from:
- Docker build from `Dockerfile` recipe file
- Singularity build from `Singularity` recipe file
- Singularity build from `docker://mrtrix3/mrtrix3:debug_3.0.x

Lestropie:
Given that neither this nor #2370 have been reviewed, and there is no logical distinction between the two, I'm going to merge this into #2370, and have that as a single review of container modifications (in addition to a single set of images to test, particularly for `mrview` support).

PR #2465, opened 2022-04-12
PNG: handle single sagittal or coronal slices correctly

jdtournier:
Came across this while answering a question on the forum. On current `master`, this happens:
```
$ mrconvert ~/data/anat.nii -coord 0 70 out-[].png 

mrconvert: [SYSTEM FATAL CODE: SIGSEGV (11)] Segmentation fault: Invalid memory access
```
Or alternatively:
```
$ mrconvert ~/data/anat.nii -coord 0 70 - | mrconvert - out-[].png 
mrconvert: [100%] copying from "/home/donald/data/anat.nii" to "/tmp/mrtrix-tmp-ck4o7Y.mif"
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```
Basically any attempt to convert a single sagittal or coronal slice using the square bracket syntax. Works OK if output file is simply `out.png` (not using square bracket syntax). Also works OK if selecting a single axial slice. 

Changes in the PR seem to fix the problem.

### DIFF
--- a/core/formats/png.cpp
+++ b/core/formats/png.cpp
@@ -146,15 +146,9 @@ namespace MR
         case 1:
           throw Exception ("Cannot generate PNG image with only 1 axis");
         case 2:
-          if (H.ndim() == 3) {
-            // Which axis do we remove?
-            // If axis 0 or 1 is size 1, then let's leave the header as 3D
-            // If however all 3 axes are greater than 1, we're going to remove
-            //   the final axis, which will then be looped over via the NameParser
-            if (H.size(0) > 1 && H.size(1) > 1)
-              H.ndim() = 2;
-          }
-          break;
+          if (H.ndim() == 2)
+            break;
+          // otherwise proceed to case 3:
         case 3:
           if (H.size(1) == 1) {
             axis_to_zero = 1;


Lestropie:
PR fails existing binary test. Being an intrinsically 2D format, there is a more complex relationship between the nature of the input data and the utilisation of square-bracket notation in the output path for PNG. I'll have to remind myself of how it works and then figure out why this edge use case fails.

Lestropie:
Need to clarify the intended behaviour in this circumstance.

If `anat.nii` is a 3D image, let's say 128x128x128, and `-coord 3 70` is specified, then you have an image that is 1x128x128. If you were to request to write that image to a PNG, it would detect that the first axis is of unity size, and so write a 2D image based on axes 0 and 1, which would be sagittal assuming `RealignTransform` is true.

But you're then specifying an output image path with the square-bracket syntax. This would logically take the last non-unity axis---the third axis in this case---and generate one PNG file per coordinate, resulting in 128 images of dimension 1x128. Suspect that this is not what the user intended, but it should nevertheless work rather than segfaulting. I don't think it makes sense that the square-bracket syntax would detect a unity dimension and "attach itself" to that dimension given that by its very nature it's intended to write multiple image files.

Lestropie:
Proposed that this be deferred from `3.0.4` to `3.1.0`, given that the usage resulting in the segfault is arguably erroneous.

Lestropie:
Closing in favour of #2535.

Issue #2467, opened 2022-04-21
population_template: bogus average space


maxpietsch:
For some data, `population_template`'s seems to crop the FOV. Deactivating the scale and drift correction seems to fix it:
- https://community.mrtrix.org/t/population-template-fov-cropped-issue/5681
- https://community.mrtrix.org/t/population-template-fov/4199

I think I'll need test data where this is an issue to debug this.

PR #2471, opened 2022-04-28
population_template initialisation and drift correction fixes

maxpietsch:
1. proposed fix to #2467 
2. also fixes that the intialisation was not used in the first registration level
3. while at it, also added the option to delete intermediate templates which could be handy if many levels are requested and storage space in scratch is limited

For context regarding 1.:

The initial template space is computed after the initial trafos `R_i` (default: translations to centre of mass) are applied to the input images. 
The drift correction is supposed to prevent changes in appearance to the template over template levels and used the average transformation `A = E [T_i ]` across the image trafos `T_i` which include the initialisation. Hence for large "outliers" in the initialisation (i.e. one subject is far form the others in scanner space), the initial drift correction moves the template content out of the template FOV (as reported in #2467). The proposed fix computes `E [R_i ]` and corrects drift via `T_i <- T_i E [R_i ] (E [T_i ])^{-1}`. This _approximately_ preserves the size and shape of the image relative to the (average) original input images. Suggestions on how to improve drift correction are very welcome. 

BTW, on my test data (neonatal FA maps), deactivating drift correction via `-linear_no_drift_correction` leads to visible distortion in the affine levels that increase with the number of levels iterations. It's also worth thinking about implementing some form of (approximate) average shape preservation for the nonlinear registration.

This PR alters the behaviour and output of `population_template` for all types or registration except for `nonlinear` so we'd need to tag it. Test data needs to be updated as well. Should this go to master?

### DIFF
--- a/bin/population_template
+++ b/bin/population_template
@@ -585,7 +585,12 @@ def parse_input_files(in_files, mask_files, contrasts, f_agg_weight=None, whites
 
 
 def execute(): #pylint: disable=unused-variable
-  from mrtrix3 import MRtrixError, app, image, matrix, path, run #pylint: disable=no-name-in-module, import-outside-toplevel
+  from mrtrix3 import MRtrixError, app, image, matrix, path, run, EXE_LIST #pylint: disable=no-name-in-module, import-outside-toplevel
+
+  expected_commands = ['mrgrid', 'mrregister', 'mrtransform', 'mraverageheader', 'mrconvert', 'mrmath', 'transformcalc', 'mrfilter']
+  for cmd in expected_commands:
+    if cmd not in EXE_LIST :
+      raise MRtrixError("Could not find " + cmd + " in bin/. Binary commands not compiled?")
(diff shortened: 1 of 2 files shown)


Lestropie:
> Should this go to master?

Given #2467 states that use of an existing command-line option resolves the issue for problematic data, I would tend to think no? But I don't have a sense of the prevalence or magnitude of the problem to make a confident call.

maxpietsch:
> Given https://github.com/MRtrix3/mrtrix3/issues/2467 states that use of an existing command-line option resolves the issue for problematic data

Don't think there is a way to fix it via command-line options, setting `average_inv = None` is done by patching the code. Currently, for certain data, `population_template` produces cropped templates. So it's a bugfix but it also affects outputs for other data, hence a new tag is required.

Lestropie:
>  I've amended the PR to merge to dev ...

... and then changed it back to `master`?

PR #2478, opened 2022-05-26
update mime & icon handling on linux

jdtournier:
On Linux, the `./install_mime_types.sh` command should now produce a working launcher for `mrview`, with recognised mime types that launch correctly and show up with the MRtrix icon


Lestropie (dismissed):
Worked for me on Xubuntu 20.04.

PR #2484, opened 2022-06-22
Python: don't cleanup on SIGINT if -nocleanup option has been provided

jdtournier:
When debugging long-running scripts (such as `dwifslpreproc`), it's sometimes useful to interrupt the process when it's already moved on from the problematic stage, but hitting Ctrl-C currently deletes the temporary folder even if the process was invoked with `-nocleanup`, which IMO is unexpected. This seems to fix it.


Lestropie (approved):
This should now do as initially desired but also print the location of the scratch directory if the handler is invoked but -nocleanup` was specified.

Lestropie:
Does anybody else want to review this given I changed all of the code?

jdtournier:
> Does anybody else want to review this given I changed all of the code?

I would, but GitHub won't let me as I'm the originator for this pull request... But I have looked at the code, and I approve!

Issue #1433, opened 2018-08-17
mrview: allow custom colourmaps


jdtournier:
As [discussed on the forum](http://community.mrtrix.org/t/custom-user-specified-colourmap/1820), and raised previously in #165. 

This is going to require a bit of an overhaul, but might actually make our lives easier. Most packages (including Matlab) expect a table of RGB triplets, and will interpolate that assuming the samples are regularly spaced between 0 & 1. It turns out that all of our current (scalar) colourmaps fit in with that trivially (bearing in mind that each component gets clamped between 0 & 1):

- greyscale:
   ```
   0 0 0
   1 1 1
   ```

- hot:
   ```
        0       -1  -2.7727
   2.7213   1.7213        1
   ```

- cool:
   ```
   -1.7213  -0.7213   0.0
    1.0      2.0      3.7727
   ```

- jet:
   ```
      0  -0.5   0.5
   -0.5   0.5   1.5
    0.5   1.5   0.5
    1.5   0.5  -0.5
    0.5     0     0
   ```

- pet:
   ```
   -0.5   0.0  0.0
      0   0.5  0.5
    0.5   0.0  1.0
    1.0   0.5  0.0
    1.0   1.0  1.0
   ```

So really, all we'd need is a 1D texture, and a texture lookup followed by a `clamp()` operation. Actually way more simple than the current implementation, and immediately compatible with just about every colourmap users are likely to add. It also simplifies the colourmap_button implementation: we would no longer need both shader coder and an anonymous function to use on the C++ side. It's embarrassing, it's just gotta be done...

thijsdhollander:
This would be great indeed, if anyone has the time to implement of course.  User interface wise, what I'd think would be quite nice, is that users can predefine their own colour maps in the above format in e.g. the configuration file, along with a name, and that they'd be accessible in the colour map list(s) in `mrview`.  I can imagine people using `mrview` in very particular environments an use cases where they often need a same (selection of) colour map(s).  Allowing them to configure this as such up front, would be very convenient.

julienbesle:
I'd like to voice support for this feature to be implemented.
I currently need a color map that contains successive random colors in order to visualize parcellations and there is no colormap in mrview that can be used for this. I can think of a lot of other situations in which I would need to use other custom colormaps that are unlikely to ever be added to mrview.
Thanks

Lestropie:
[...]

PR #2487, opened 2022-06-23
MRview: add support for custom colourmap

jdtournier:
This adds support for loading custom colourmaps from file at runtime. 

This addition will scan through the folder `~/.mrtrix/colourmaps/` (if it exists) and load each file as a distinct colourmap, making it available on the GUI using the name of the file as the colourmap name. The format of each colourmap file is as a matrix of 4 columns, with each column corresponding to:
```
[ control_point red green blue ]
```
The first and last row control point values should be zero & one respectively, and all intermediate control points should increase monotonically between these values. For example:
```
0 0 0 0
0.5 0 0 1
1 1 0 0
```
increases from black to pure blue at 50%, then to pure red at 100%. For intermediate intensities, the colour is derived by linear interpolation from the nearest control points.

---

This required some refactoring, all of which is self contained within 8a389425. Worth double-checking all functionality still works, including overlays, fixel rendering, colouring by TSF, etc.


bjeurissen:
Very cool!

I did get a crash when using negative control points, e.g. [-1; 0; 1] instead of [0; 0.5; 1]

jdtournier:
Yes, I could add some more thorough error-checking here... Might do that if I find the time.

Lestropie:
Relates to #1433.

Forwarding prior discussion points, I'm slightly uneasy about this one for multiple reasons:

1. It's a feature addition / enhancement rather than a bug fix, and so would be more comfortable with a merge to `dev` rather than `master`.
2. It's a bit of a hack in that it introduces a code-branch-based linear interpolation, which is slow on GPUs. I would very much prefer a GPU-texture-based / CPU-lookup-based solution from the same originating data, which would be used for the majority of built-in colour maps also; it's only the "special" maps like RGB / complex that necessitate code injection. I'm not sure if merging this proposal would kill the motivation to implement such.
3. The format of `<control_point R G B>` is non-standard. It's more conventional to just provide 256 rows. We could have the capability of less than 256 rows resulting in an equal distribution of control points and that could be reasonably interpreted.

lconcha:
Sorry to revive an old thread, but I just tried this branch for custom colourmaps and I find it great. In response to @Lestropie's comment about the hard-coded 256-row format, I have to say I prefer the control-point approach taken by @jdtournier in his current implementation. This way it is much easier to create custom colourmaps. 

Looking forward to seeing this implemented in the main branch and release.

Lestropie:
Worth noting that it would be relatively easy to create a script that would take as input this control point style format, and produce a colour map that can then be read as a texture and interpolated. Indeed in the example above:
```
0   0 0 0
0.5 0 0 1
1   1 0 0
```
, there wouldn't be any need for producing a 256-line colour map. The control points are equally spaced. So that could just be read as a 3-line texture with linear interpolation between the three specified colours. It's only if the control points have some unusual spacing that one would need to figure out how many lines are necessary for generating a texture so that linear interpolation can be done without knowledge of control point positions. But in the back end, having everything as a texture lookup would be so much simpler.

PR #2488, opened 2022-06-24
tckconvert write binary VTK

blezek:
This pull request enables mrtrix to read VTK binary files when converting to .tck format.

PR #2489, opened 2022-06-24
Tckconvert write binary vtk

jdtournier:
replaces #2488, this time merging to `master` with VTK output still defaulting to ASCII. 

@blezek, could do with a second pair of eyes to double-check none of my changes break anything...

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -234,7 +234,7 @@ class VTKReader: public ReaderInterface<float> { MEMALIGN(VTKReader)
 
         if (sscanf (line.c_str(), "POINTS %d float", &number_of_points) == 1) {
           points.resize (3*number_of_points);
-          input.read ((char*) &points[0], 3*number_of_points * sizeof(float));
+          input.read ((char*) points.data(), 3*number_of_points * sizeof(float));
 
           // swap
           for (int i = 0; i < 3*number_of_points; i++)
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -32,64 +32,64 @@ using namespace MR::ByteOrder;
 void usage ()
 {
   AUTHOR = "Daan Christiaens (daan.christiaens@kcl.ac.uk), "
-           "J-Donald Tournier (jdtournier@gmail.com), "
-           "Philip Broser (philip.broser@me.com), "
-           "Daniel Blezek (daniel.blezek@gmail.com).";
+    "J-Donald Tournier (jdtournier@gmail.com), "
+    "Philip Broser (philip.broser@me.com), "
+    "Daniel Blezek (daniel.blezek@gmail.com).";
 
   SYNOPSIS = "Convert between different track file formats";
 
   DESCRIPTION
-  + "The program currently supports MRtrix .tck files (input/output), "
+    + "The program currently supports MRtrix .tck files (input/output), "
     "ascii text files (input/output), VTK polydata files (input/output), "
     "and RenderMan RIB (export only)."
 
-  + "Note that ascii files will be stored with one streamline per numbered file. "
+    + "Note that ascii files will be stored with one streamline per numbered file. "
     "To support this, the command will use the multi-file numbering syntax, "
     "where square brackets denote the position of the numbering for the files, "
     "for example:"
 
-   + "$ tckconvert input.tck output-'[]'.txt"
+    + "$ tckconvert input.tck output-'[]'.txt"
 
-   + "will produce files named output-0000.txt, output-0001.txt, output-0002.txt, ...";
+    + "will produce files named output-0000.txt, output-0001.txt, output-0002.txt, ...";
 
   ARGUMENTS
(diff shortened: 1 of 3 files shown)


blezek:
@jdtournier LGTM.  Preserving backwards compatibility by using `ascii` as the default makes sense.

cmd/tckconvert.cpp
```diff
@@ -21,76 +21,83 @@
...
 
-  + OptionGroup ("Options for both PLY and RIB writer")
+    + Option ("radius", "radius of the streamlines")
+    +   Argument("radius").type_float(0.0f)
 
-  + Option ("radius", "radius of the streamlines")
-  +   Argument("radius").type_float(0.0f);
+    + OptionGroup ("Options specific to VTK writer")
+
+    + Option ("ascii", "write an ASCII VTK file (this is the default)")
```
Lestropie, line 99: If one is the default, then a command-line option to activate such is redundant.

cmd/tckconvert.cpp
```diff
@@ -102,41 +109,76 @@ void usage ()
 
 class VTKWriter: public WriterInterface<float> { MEMALIGN(VTKWriter)
```
Lestropie, line 110: Having the boolean switch between ASCII and binary as a template parameter rather than a constructor input would be theoretically more efficient. Speculative execution will make up some of the ground, but has famously led to its own issues... nevertheless, performance is maybe not a concern here, but for #411 would want to template it out.

testing/binaries/tests/tckconvert
```diff
@@ -1,8 +1,15 @@
...
 tckconvert tckconvert/out2-[2:9].txt tmp.tck -force && testing_diff_tck tmp.tck tckconvert/out3.tck
 echo 1 2 3 > tmp.txt && tckconvert -force -quiet tmp.txt tmp.tck && tckconvert -quiet -force tmp.tck tmp.rib && [ $(wc -l < tmp.rib ) == 4 ]
 tckconvert -force -quiet tckconvert/empty.vtk tmp.tck
+
+# convert tck -> vtk -> txt and compare to tck -> txt
+tckconvert tracks.tck tmp.vtk -binary -force -quiet && tckconvert tmp.vtk tmp-vtk-streamlines-[].txt -force && cat tmp-vtk-streamlines-*.txt > tmp-vtk-all.txt && tckconvert tracks.tck tmp-mrtrix-streamlines-[].txt -force -quiet && cat tmp-mrtrix-streamlines-*.txt > tmp-mrtrix-all.txt && testing_diff_matrix tmp-vtk-all.txt tmp-mrtrix-all.txt
+
+# convert tck -> vtk -> txt and compare to tck -> vtk -> tck -> txt
+tckconvert tracks.tck tmp.vtk -binary -force -quiet && tckconvert tmp.vtk tmp-roundtrip.tck -force && tckconvert tmp-roundtrip.tck tmp-roundtrip-streamlines-[].txt -force && cat tmp-roundtrip-streamlines-*.txt > tmp-roundtrip-all.txt && tckconvert tracks.tck tmp-mrtrix-streamlines-[].txt -force -quiet && cat tmp-mrtrix-streamlines-*.txt > tmp-mrtrix-all.txt && testing_diff_matrix tmp-roundtrip-all.txt tmp-mrtrix-all.txt
```
Lestropie, line 15: One lesson I've learned the hard way elsewhere is that if your own software is non-compliant, but the reader and writer functionalities have matching non-compliance, then round-trip tests do you no good.

If we could add an exemplar binary file into the test_data repository that demonstrably works in some other software, then we could ensure that that at least gets read correctly. Writing back to binary and then doing a binary diff *might* work for testing the write.

I don't think it's super critical here given that we presume you've implemented the writer for a purpose and it's being successfully read by something else, and also that the code is relatively independent of everything else. But as with another comment, for #411 we'd need to be more exhaustive.

jdtournier: Indeed. I'll add the outputs produced by the conversion to the test data repo and test against that. It'll catch regressions, and if @blezek can double-check that these work correctly in other packages, we should be good.

cmd/tckconvert.cpp
Lestropie: It'll be mostly inheritance, I suppose in retrospect it would depend on the amount of shared code between VTK ASCII vs. binary as to whether to have them as two templated versions of the same derived class or two different classes.

testing/binaries/tests/tckconvert
Lestropie: I don't have anything with which to visualise polylines.

Lestropie (approved):
Even if the binary VTK data in the test data repo hasn't been tested, I'll trust @blezek's prior work here. Happy to merge for 3.0.4 as-is.

blezek:
> Even if the binary VTK data in the test data repo hasn't been tested, I'll trust @blezek's prior work here. Happy to merge for 3.0.4 as-is.

@jdtournier @Lestropie sorry, this PR got away from me.  I verified the binary VTK file loads correctly in [ParaView](https://www.paraview.org/), so we are good to go!

PR #2490, opened 2022-06-24
Fixes incorrect face orientation in PLY output format

blezek:
The cap placed on the end of a streamline was incorrectly oriented
by the PLY writer.  Some 3D printing software will not correctly
recognize this condition and will ignore the endcaps.  Particularly
painful for 10's or 100's of thousands of streamlines!

A comment to the output PLY file is added clarifying the 
coordinate system and scale of the model.

This change is a bugfix and does not introduce new features, nor break
backwards compatibility.

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -496,12 +496,12 @@ class PLYWriter: public WriterInterface<float> { MEMALIGN(PLYWriter)
           isFirst = false;
         }
         if ( isLast ) {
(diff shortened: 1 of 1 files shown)


Lestropie (approved):
I've not played with this myself so will trust that you've diagnosed and rectified correctly.

I did initially have a concern that facet handedness could be violated by any spatial transformation with a negative determinant; but any such transformations are applied to the streamlines data prior to generation of the corresponding streamtubes, so should be no issue.

PR #2493, opened 2022-07-13
seeding from user-provided coordinates

dmitrishastin:
Follow up on [this MRtrix community post](https://community.mrtrix.org/t/seeding-from-pre-defined-coordinates/3276) @Lestropie.

All self-explanatory, the user-supplied file containing a matrix of coordinates is parsed, then seed coordinates (rows in the matrix) are selected at random until tckgen conditions are met (desired number of seeds / streamlines / etc reached) allowing repeated use of each coordinate.

_Technically, this mode is considered "Mechanism that provides random seed locations" in how it's handled so in the code it is defined by volume. As coordinates do not have a volume (volume is explicitly set to 0.0), the user is prevented from requesting any other seeding mechanisms in parallel as these would dominate._

### DIFF
--- a/src/dwi/tractography/seeding/basic.h
+++ b/src/dwi/tractography/seeding/basic.h
@@ -59,7 +59,29 @@ namespace MR
             Eigen::Vector3f pos;
             float rad;
 
+        };    
+
+
+
+        class Seed_coordinates : public Base
+        { MEMALIGN(Seed_coordinates)
+
+          public:
+            Seed_coordinates (const std::string& in) :
+              Base (in, "coordinate matrix", MAX_TRACKING_SEED_ATTEMPTS_RANDOM) {
+                seed_cds = load_matrix<float> (in);
+                if (seed_cds.cols() != 3)
+                  throw Exception ("Could not parse file \"" + in + "\" as a coordinate input; needs to have 3 columns"); 
+                volume = 0.0;               
+              }
+
+            virtual bool get_seed (Eigen::Vector3f& p) const override;
+
+          private:
+            Eigen::MatrixXf seed_cds;
+
         };
+
 
 
         class SeedMask : public Base
--- a/src/dwi/tractography/seeding/seeding.cpp
+++ b/src/dwi/tractography/seeding/seeding.cpp
@@ -38,6 +38,10 @@ namespace MR
       + Option ("seed_image", "seed streamlines entirely at random within a mask image ").allow_multiple()
         + Argument ("image").type_image_in()
 
+      + Option ("seed_coordinates", "seed from coordinates provided in a file as space-separated Mx3 matrix "
+                                    "(XYZ per row), must not provide any other seeding mechanism")
+        + Argument ("cds_path").type_text()
+
       + Option ("seed_sphere", "spherical seed as four comma-separated values (XYZ position and radius)").allow_multiple()
         + Argument ("spec").type_sequence_float()
 
@@ -119,6 +123,14 @@ namespace MR
         auto opt = get_options ("seed_image");
         for (size_t i = 0; i < opt.size(); ++i) {
           SeedMask* seed = new SeedMask (opt[i][0]);
+          list.add (seed);
+        }
+
+        opt = get_options ("seed_coordinates"); 
+        if (opt.size()) {
+          if (list.num_seeds())
+            throw Exception ("If seeding from coordinates, cannot specify any other type of seed!");
+          Seed_coordinates* seed = new Seed_coordinates (opt[0][0]);
           list.add (seed);
         }
 
--- a/docs/reference/commands/tckgen.rst
+++ b/docs/reference/commands/tckgen.rst
@@ -78,6 +78,8 @@ Tractography seeding mechanisms; at least one must be provided
 
 -  **-seed_image image** *(multiple uses permitted)* seed streamlines entirely at random within a mask image 
 
+-  **-seed_coordinates cds_path** seed from coordinates provided in a file as space-separated Mx3 matrix (XYZ per row), must not provide any other seeding mechanism.
+
 -  **-seed_sphere spec** *(multiple uses permitted)* spherical seed as four comma-separated values (XYZ position and radius)
 
 -  **-seed_random_per_voxel image num_per_voxel** *(multiple uses permitted)* seed a fixed number of streamlines per voxel in a mask image; random placement of seeds in each voxel
--- a/src/dwi/tractography/seeding/basic.cpp
+++ b/src/dwi/tractography/seeding/basic.cpp
@@ -42,6 +42,15 @@ namespace MR
 
 
 
+        bool Seed_coordinates::get_seed (Eigen::Vector3f& p) const
+        {
+          int coordinate_index = std::uniform_int_distribution<> (0, seed_cds.rows()-1)(rng);
+          p = seed_cds.row(coordinate_index);
+          return true;
+        }
+
+
+
 
         bool SeedMask::get_seed (Eigen::Vector3f& p) const
         {
@@ -56,8 +65,6 @@ namespace MR
           p = (*mask.voxel2scanner) * p;
           return true;
         }
-
-
 
 
 
@@ -94,10 +101,6 @@ namespace MR
           p = (*mask.voxel2scanner) * p;
           return true;
         }
-
-
-
-
 
 
 


Lestropie:
The default behaviour here may be different to what one may intuitively expect from such a mechanism.

Personally, if I were to provide a list of streamline seeds, I would expect that there would be exactly one streamline (or potentially exactly some number specified as part of the command-line option) seeded at each defined point. The duration of execution of the command would therefore be constrained by the number of rows in the provided file; a so-called "finite seeding mechanism" (which theoretically could be utilised in conjunction with other finite seeding mechanisms).

However what you're doing here is more akin to the myriad random "infinite" seeding mechanisms, only that instead of sampling randomly from eg. some volume, you are instead sampling randomly from a finite set of precise seed locations. So any number of seeds can be drawn, and the user can choose to terminate based on seed count and/or streamline count.

Is there a motivation for the latter over and above replication of published work? The former is more faithful to eg. FSL `probtrackx`, and makes more intuitive sense to me.

Theoretically these could be two separate mechanisms (assuming that sensible names can be determined for such)? @MRtrix3/mrtrix3-devs opinions?

Ideally there should be a test added, just to verify that the code executes; see [this file](https://github.com/MRtrix3/mrtrix3/blob/3.0.3/testing/binaries/tests/tckgen). You could either generate a seed text file as part of the command, or you could generate one and it could be added to the [test data](https://github.com/MRtrix3/test_data) (which I would need to do).

Lestropie:
I'd prefer two separate mechanisms and two separate command-line options. Having an optional argument following a command-line option breaks the flexibility of the command-line parsing: it becomes impossible to unambiguously determine which elements at the command-line are positional arguments and which are options and their associated arguments, whilst also permitting both to be placed in arbitrary order.

But they would ideally both derive from a common base class that loads and checks the seed location information, and the derived classes then sample from those in a different way. If you're not comfortable with the relevant C++ I can do that part myself.

dmitrishastin:
Thanks, that's helpful. After doing some reading I have come up with a solution but I don't know if this is remotely what you meant. My implementation has a multiple-inheritance structure whereby the Base class is joined by a coordinate parser class to produce the two algorithm-specific classes. It works, but I've no idea if it's a good arrangement.

I've called the two different behaviours "global" and "fixed" - how does sound? I feel "fixed" is more representative than "fixed". Suggestions?

As per your initial reply I have now added some strings to the ./run_tests file. Currently it is only testing that the algorithms execute. Would you suggest I generate some actual benchmarking data as well so that `testing_diff_tck` could be used? 

A little extra feature - with "global" seeding, it's possible to provide a fourth column which will be treated as weights. Rationale - for instance, it may be interesting to seed based on some function defined on a surface. Not sure if it is useful or even makes sense - I can remove it if you think it's excessive/unnecessary.

dmitrishastin:
Suppose I should add that (in my head) the main reason for not having a single Base class was the two different mechanisms, meaning its third input would have two options (`MAX_TRACKING_SEED_ATTEMPTS_RANDOM` or `MAX_TRACKING_SEED_ATTEMPTS_FIXED`) and I couldn't think of a way to establish which one to pass from seeding.cpp without changing the file more generally.

src/dwi/tractography/seeding/basic.h
```diff
@@ -160,6 +161,119 @@ namespace MR
...
+          public:
+
+            Coordinates_fixed (const std::string& in, const size_t n_streamlines) :
+              Base (in, "coordinate seeding fixed", MAX_TRACKING_SEED_ATTEMPTS_FIXED), 
+              Coordinate_parser(in) {
+
+                if (nc != 3)
+                  throw Exception ("Number of columns in \"" + in + "\" must equal 3!");   
+                if (n_streamlines == 0)
+                  throw Exception ("Number of streamlines to be generated per coordinate must be a positive integer!");
```
Lestropie, line 197: This can instead be enforced at the command-line by setting the minimum permissible value of the corresponding argument to 1.

src/dwi/tractography/seeding/basic.h
```diff
@@ -160,6 +161,119 @@ namespace MR
...
+
+                if (nc != 3)
+                  throw Exception ("Number of columns in \"" + in + "\" must equal 3!");   
+                if (n_streamlines == 0)
+                  throw Exception ("Number of streamlines to be generated per coordinate must be a positive integer!");             
+                current_coord = 0;
+                num_at_coord = 0;
+                expired = false;
+                nsl = n_streamlines;
+                count = nr * nsl;
```
Lestropie, line 202: These can be set as initialisers for the constructor, rather than within the code body of the constructor.

src/dwi/tractography/seeding/basic.h
```diff
@@ -160,6 +161,119 @@ namespace MR
...
+
+                if (nc == 4) {
+
+                  if (coords.col(3).minCoeff() < 0)
+                    throw Exception ("The seeding weights must be non-negative!");
+
+                  sorted_weights = coords.col(3);
+                  std::sort(sorted_weights.data(), sorted_weights.data() + nr);
+
+                  visited = Eigen::MatrixXf::Zero(nr, 1);
```
Lestropie, line 240: Constructor doesn't match variable type definition.

Lestropie: Also, appears to only be used in the constructor, not elsewhere; therefore it can be defined as a local variable rather than a class member.

Lestropie: Also `vector<bool>` would be better.

Lestropie: Also `bool` would be better than `float`.

src/dwi/tractography/seeding/basic.h
```diff
@@ -160,6 +161,119 @@ namespace MR
...
+                volume = 0.0;
+
+              }
+
+            virtual bool get_seed (Eigen::Vector3f& p) const override;
+            
+          private:
+
+            Eigen::MatrixXf sorted;
+            Eigen::VectorXf sorted_weights;
```
Lestropie, line 272: Can be temporaries in the constructor.

src/dwi/tractography/seeding/basic.h
```diff
@@ -160,6 +161,119 @@ namespace MR
...
+
+                  cumsum_weights = Eigen::MatrixXf::Zero(sorted_weights.size() + 1, 1);
+                  cumsum_weights.tail(sorted_weights.size()) = sorted_weights.array() / sorted_weights.sum();
+
+                  for(int j = 1; j < cumsum_weights.size(); ++j)
+                    cumsum_weights(j) += cumsum_weights(j - 1);
+
+                }
+
+                volume = 0.0;
```
Lestropie, line 263: Can set in constructor initialisation.

src/dwi/tractography/seeding/basic.h
Lestropie: Nevermind; it's already there.

src/dwi/tractography/seeding/seeding.cpp
```diff
@@ -157,6 +167,22 @@ namespace MR
...
+          Coordinates_fixed* seed = new Coordinates_fixed (str(opt[i][0]), opt[i][1]);
+          list.add (seed);
+        }
+
+        // those exclusive of other seeding mechanisms go to the end, with dynamic seeder last
+
+        opt = get_options ("seed_coordinates_global");
+        if (opt.size()) { 
+          if (list.num_seeds())
+            throw Exception ("If seeding from coordinates to a global target, cannot specify any other type of seed!");
```
Lestropie, line 181: "Global" might be misinterpreted here. Maybe:
"If seeding from pre-specified coordinates with no fixed number of streamlines per coordinate, ..."?

Lestropie:
> A little extra feature - with "global" seeding, it's possible to provide a fourth column which will be treated as weights.

Interesting idea. I can see how one might want to try to exploit this to eg. compensate for a gyral bias, albeit in an ad hoc fashion.

The constructor code is not the easiest to follow. My best guess is that you're sorting the weights prior to calculating the cumulative sum in order to minimise the loss of floating-point precision? Things like this benefit from code comments. There may also be a risk of the final cumulative sum entry being greater than 1.0, and therefore `get_seed()` overflowing. You'd need to run in debug mode with a few different seed sets to check.

The alternative strategy for implementation of something like this (which would be more consistent with other relevant MRtrix3 code) would be a rejection sampler: set the ceiling to the maximal weight, select a random value between 0 and that ceiling, and if that random value is less than the weight, select that seed. For a large number of candidate seed points and relatively homogeneous weights this would probably be faster too.

> Currently it is only testing that the algorithms execute.

That's probably fine; to verify precise operation requires addition of data to the `test_data` repository, which is clunky; plus there's existing mechanisms that don't yet have that level of detail of testing.

dmitrishastin:
> Interesting idea. I can see how one might want to try to exploit this to eg. compensate for a gyral bias, albeit in an ad hoc fashion.

Yes that was the first thought too. Suppose ideally the code could be expanded to also accept the mesh connectivity matrix such that points are sampled anywhere on the surface and not just at the vertices - but I probably can't commit to that right now unfortunately.

> The constructor code is not the easiest to follow. My best guess is that you're sorting the weights prior to calculating the cumulative sum in order to minimise the loss of floating-point precision? Things like this benefit from code comments. There may also be a risk of the final cumulative sum entry being greater than 1.0, and therefore `get_seed()` overflowing. You'd need to run in debug mode with a few different seed sets to check.

Now that you mention it, I didn't really need to sort the weights, you're right. Just doing the cumsum should be enough. Re: overflow, i suppose the alternative is to run the cumsum then divide all elements of the vector by its last entry?

> The alternative strategy for implementation of something like this (which would be more consistent with other relevant MRtrix3 code) would be a rejection sampler: set the ceiling to the maximal weight, select a random value between 0 and that ceiling, and if that random value is less than the weight, select that seed. For a large number of candidate seed points and relatively homogeneous weights this would probably be faster too.

Yes, happy to do that instead. Probably depends how much homogeneity you would expect in weights?

Let me know what you think, I'll then implement all the changes.

Lestropie:
> Suppose ideally the code could be expanded to also accept the mesh connectivity matrix such that points are sampled anywhere on the surface and not just at the vertices

@chunhungyeh was working on this. There might be functional code sitting somewhere, but it was intended to be included in a larger set of changes relating to surface handling.

> Re: overflow, i suppose the alternative is to run the cumsum then divide all elements of the vector by its last entry?

Yes, I think that would work.

> Probably depends how much homogeneity you would expect in weights?

No idea; it's entirely hypothetical at this point. But if random number generation is efficient (as it should be on modern CPUs that have hardware dedicated to such), and your surface contains > 10,000 vertices, you could have a couple of orders of magnitude variance and it'd likely still be faster than looping through the cumulative sums.

src/dwi/tractography/seeding/basic.h
dmitrishastin: Could not figure out a way to set count as initialiser - it belongs to Base class and I could not access it from here, am I missing something? The rest have been set as initialisers now.

src/dwi/tractography/seeding/basic.h
dmitrishastin: Were only required for sorting hence removed.

src/dwi/tractography/seeding/basic.h
dmitrishastin: As with count for the _fixed class, could not figure out a way to access this Base class's variable while initialising the constructor of a derived class.

dmitrishastin:
Updated as per suggestions above (had some problems with constructor initialisers for the derived classes). Adopted the rejection seeding approach. Additionally decided to generate unit weights when not provided with seeding to an overall target (_global). I could be wrong but I think it probably doesn't change performance (instead of checking the number of columns it now compares selector value against randomly selected weight which will always lead to loop exit), perhaps a tiny bit more overhead, but the code is easier to manage.

dmitrishastin:
> Additionally decided to generate unit weights when not provided with seeding to an overall target (_global). I could be wrong but I think it probably doesn't change performance (instead of checking the number of columns it now compares selector value against randomly selected weight which will always lead to loop exit), perhaps a tiny bit more overhead, but the code is easier to manage.

Scratch that, there is the additional random number generation which is unneccessary for unweighted mode, rolled back.

Lestropie:
Currently resolving this against current `dev` with vision of merging.
Need to think out loud for a moment regarding interface / naming / scope of capabilities.

Current seeding options (deliberately excluding outliers) and underlying mechanism:
1.  `-seed_image`: Randomly select a voxel, then randomly select a position within the voxel
2.  `-seed_random_per_voxel`: Fixed number per voxel, but random position within the voxel
3.  `-seed_grid_per_voxel`: Fixed number per voxel, arranged in a grid
4.  `-seed_rejection`: Randomly select some voxels more often than others, random position within the voxel

At its core, the proposal here is that a user-specified realspace coordinate can be used instead of a voxel.
This does not have an equivalent to the capability of seed voxels to sample from different spatial locations within it.

So how does this map?
1.  Closest analogue is randomly selecting a coordinate, currently implemented as `-seed_coordinates_global` with a 3-column input.
    A more direct match would be a seeding mechanism that chooses a voxel at random, but then always seeds from precisely the centre of that voxel
2.  No direct analogue; related to 1.
3.  Currently implemented here as `-seed_coordinates_fixed`
4.  Currently implemented here as `-seed_coordinates_global` with a 4-column input.

To make this interface more consistent I think needs something like:

| Option name                 | Input type    | Sub-element seed distribution | Count             |
| --------------------------- | ------------- | ----------------------------- | ----------------- |
| `-seed_voxels`              | Image         | Random                        | Global            |
| `-seed_random_per_voxel`    | Image         | Random                        | Fixed # per voxel |
| `-seed_grid_per_voxel`      | Image         | Grid                          | Fixed # per voxel |
| `-seed_rejection_per_voxel` | Image         | Random                        | Proportional      |
| `-seed_per_voxel_centre`    | Image         | Centre                        | Fixed # per voxel |
| `-seed_coords`              | Coordinates   | N/A                           | Global            |
| `-seed_per_coord`           | Coordinates   | N/A                           | Fixed # per voxel |
| `-seed_rejection_per_coord` | Coordinates   | N/A                           | Proportional      |

So:

-   Switching between choosing coordinates at random, and performing rejection sampling to select some coordinates more frequently than others, should be separate command-line options, given that for image-based seeding such an explicit split already exists.
-   `-seed_rejection` needs to be renamed to be explicitly based on an input image, since rejection seeding can be done for elements other than image voxels.
-   `-seed_image` would be better renamed to `-seed_voxels` to maintain consistency across the set of options.
-   `-seed_per_voxel_centre` isn't strictly necessary, but it would help to fill out the feature space.

Lestropie:
@dmitrishastin The rejection seeding implementation seemed to be incorrect. You were generating a single random sample in the range (0.0, 1.0), and then looking for a coordinate with a weight larger than that value. However in my testing this was not resulting in the relative proportions of coordinate selection matching what had been specified in those weights. I believe with rejection sampling in general, and it's certainly the case in the voxel-based rejection sampling, that each time you want to test a new candidate, you have to also sample a new threshold in the (0.0, 1.0) range against which to test. The CI test [here](https://github.com/MRtrix3/mrtrix3/pull/2493/commits/ca77f843dd371164a12e1f4ad8bdf30042662a40#diff-eb66cbb046083dc5553f880d30b7f7970d5e8d1f43584c4f6ad259f33eaa198cR16-R25) ensures that the proportions of the different seed candidates approximately matches what was specified in the input weights; that failed without the code change [here](https://github.com/MRtrix3/mrtrix3/pull/2493/commits/ca77f843dd371164a12e1f4ad8bdf30042662a40#diff-02a16f0c13f7385bdc21b4b082b58c39241018f9ab5ab1f9dc2584bf35b3ecbdR324-R328).

Lestropie:
@jdtournier Seeking your approval on specifically the renaming of long-standing command-line options. Historically we did this all the time, nowadays less so. Might be more problematic if we were doing more regular updates, but given the time between 3.0 and 3.1 maybe interface consistency is not of the utmost concern?

Lestropie:
Windows CI tests are failing here but succeeding on my own MSYS2 machine. Will need to modify the test to spit out the intermediate data to try to figure out what's going on.

Issue #2492, opened 2022-07-12
./build: progress indication misbehaves on a system with many cores


bjeurissen:
On a system with 128 "cores" (Ubuntu 20.04 LTS, 2 x AMD EPYC 7452 with hyper-threading), I noticed the total number of compilation jobs seems to change between threads:

```
...
(386/391) [LB] bin/warpinit
(384/396) [LB] bin/warpcorrect
(379/392) [LB] bin/tsfinfo
(380/391) [LB] bin/mrdegibbs
(382/392) [LB] bin/amp2response
(392/391) [LB] bin/dwidenoise
(393/396) [LB] bin/tckmap
(394/392) [LB] bin/mrmetric
(395/391) [LB] bin/mraverageheader
(396/396) [LB] bin/mrmath
(397/397) [LB] bin/mrtransform
```

jdtournier:
Yep, that's been the case for quite a while... Never found the time or motivation to fix it unfortunately. I might have a look again if I find the time, but as always, not very likely in the near future...

PR #2494, opened 2022-07-14
Fix build progress

jdtournier:
Attempt at fixing #2492. Definitely fixes the variability in the total number, but there is still scope for the numbering to be slightly out of order at the beginning, due to a race condition between when the index of the current job is generated and the point at which the status line is printed to the terminal. This could be fixed properly, but would require slightly deeper changes, and I'm not sure they're necessarily all that important.


bjeurissen (approved):
Works as advertised! I don't think being out of order is a huge deal. Having a sensible total number takes away a lot of the confusion + the worry that something might be going wrong.

PR #2497, opened 2022-08-01
mrview connectome tool: fix display of nodes when a node is empty

jdtournier:
See issue raised on forum:
https://community.mrtrix.org/t/connectome-init-not-working-for-fastsurfer-outputs/


Lestropie (approved):
Change looks safe.

Wouldn't be massively surprised if there are other little glitches that may arise from non-contiguous indexing; would think I'd have done some testing of such at the time, but you know how it can be with GUI code and weird use cases...

PR #2502, opened 2022-08-17
Calculate inter-quartile range in mrstats

dchristiaens:
Adds a means to calculate the inter-quartile range on an image, which can be a more robust estimate of variance. For example:
```
mrstats image.mif -output iqr
```

Alternatively, we could also output any percentile in addition to the median and IQR. I'm just not sure what would be the logical way to design the command interface. Any thoughts?

### DIFF
--- a/core/math/median.h
+++ b/core/math/median.h
@@ -68,6 +68,34 @@ namespace MR
         return med_val;
     }
 
+    template <class Container>
+    inline typename Container::value_type quantile (Container& list, default_type quantile)
+    {
+        size_t num = list.size();
+        // remove NaNs:
+        for (size_t n = 0; n < num; ++n) {
+          while (not_a_number (list[n]) && n < num) {
+            --num;
+            //std::swap (list[n], list[num]);
+            // Commented std::swap to provide bool compatibility
+            typename Container::value_type temp = list[num]; list[num] = list[n]; list[n] = temp;
+          }
+        }
+        if (!num)
+          return std::numeric_limits<typename Container::value_type>::quiet_NaN();
+
+        default_type f;
+        default_type t = std::modf(num * quantile, &f);
+        size_t loc = size_t(f);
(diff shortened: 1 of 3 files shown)


jdtournier:
Looks good to me. Not sure why the pylint tests are failing - maybe just a case of merging `master` back to `dev`...?

Otherwise, regarding percentiles, we could support a syntax like `-output p20` (for the 20th percentile), but that would require additional parsing, and couldn't be done using the [current `.type_choice` approach](https://github.com/MRtrix3/mrtrix3/blob/dev/core/stats.cpp#L36) for the argument handling. We'd need to resort to `.type_string` with explicit string matching, and parsing the requested percentile if requested. On the upside, that might allow us to simplify the process of requesting multiple outputs, which can currently be done using:
```
-output mean -output median -output std
```
to a simpler comma-separated list:
```
-output mean,median,std
```
Would anyone be interested in that?

Lestropie:
The multiple `-output` usage just to get small batches of text in the right order is indeed a little clunky. I'd be content with a comma-separated list, and then each string within that would either match a known "choice" or be 'p' followed by a floating-point number. Would need to go to `dev` given it's an interface change.

core/math/median.h
```diff
@@ -68,6 +68,34 @@ namespace MR
...
+        }
+        if (!num)
+          return std::numeric_limits<typename Container::value_type>::quiet_NaN();
+
+        default_type f;
+        default_type t = std::modf(num * quantile, &f);
+        size_t loc = size_t(f);
+        std::nth_element (list.begin(), list.begin()+loc, list.begin()+loc);
+        typename Container::value_type val0 = list[loc];
+        std::nth_element (list.begin(), list.begin()+loc, list.begin()+loc+1);
```
Lestropie: See important bug fix in d59c9101b57adb04a0e149a27ff116997d0775d9.

PR #2503, opened 2022-08-23
Correctly transform mask instead of last contrast

bjeurissen:
mrtransform was erroneously applied to the last contrast instead of the mask when initial alignment was turned off. Only happens when initial alignment is turned off, so most studies will not be affected by this.


Lestropie (approved):
Looks correct to me; can wait to see if @maxpietsch chimes in, but hard to envisage this update being a mistake.

Lestropie:
Also if merged make sure the changelog is updated.

maxpietsch (approved):

PR #2504, opened 2022-08-29
dwifslpreproc: Save bad slspec file

Lestropie:
Would have been convenient for my own experiment to be able to see the contents of this file. Not sure if there's a flaw with #2286...


Lestropie:
Doesn't work: replaces an `AssertionError` with a `MRtrixError` due to providing `matrix.save_matrix()` with an irregular matrix. Will soon superseded with a fix to both debugging and the problem being debugged.

PR #2505, opened 2022-08-29
dwifslpreproc: Refine slspec modification

Lestropie:
Currently awaiting test to see if this resolves the issue observed with a specific dataset.

With negative slice encoding direction, S2V correction, odd slice count, and EddyQc, current `master` results in an `AssertionError`, as the slspec file to be passed to EddyQC is irregular; the padding slice is not appropriately removed.


Lestropie:
Can confirm that this fixed processing for my `SliceEncodingDirection`: `k-` dataset.

Lestropie:
Should have moved out of draft after I had tested. Should be good to go.

jdtournier (approved):
Yes, I can see what this is trying to do, looks all good to me. :+1:

Issue #1458, opened 2018-09-17
dwi2tensor: Zero-weight non-physical values


Lestropie:
As discussed in [this thread](http://community.mrtrix.org/t/dwi2tensor-wrong-values/1860); listing explicitly in case anyone wants to give it a shot.

If image values are not strictly positive, leading to non-finite outputs following the log transform, zero the weights within the iterative reweighting model for such particular image volumes.

PR #2513, opened 2022-09-12
Outlier rejection in dwi2tensor

dchristiaens:
Requested by @jelleveraart. Ignores NaN values during tensor fitting (set weight to zero), as a means to reject outliers. Given an (external) outlier map (same dimensions as dMRI),
```
mrcalc outliers.mif nan dwi.mif -if dwi_or.mif
dwi2tensor dwi_or.mif tensor.mif
```
would ignore the outliers in the DTI fit. 

I have done this very quickly without testing -- waiting for Jelle to report back. I can work on a nicer interface when I have a bit more time.


Lestropie:
Relates to #1458.

NaN-ing values to be ignored is a lot more explicit than non-positive values, but it's nevertheless worth thinking about whether the behaviour should be different between the two.

bjeurissen:
I have not had the chance to test this, but will the zero-weights not be overwritten by non-zero ones in the subsequent iterations?

dchristiaens:
@bjeurissen is absolutely right, well spotted! 31c5a67 should fix that.

jdtournier:
> NaN-ing values to be ignored is a lot more explicit than non-positive values, but it's nevertheless worth thinking about whether the behaviour should be different between the two.

I agree these are two different cases, but I don't see how we can treat them differently... Did you have anything specific in mind here...?

dchristiaens:
> > NaN-ing values to be ignored is a lot more explicit than non-positive values, but it's nevertheless worth thinking about whether the behaviour should be different between the two.
> 
> I agree these are two different cases, but I don't see how we can treat them differently... Did you have anything specific in mind here...?

Actually, DWI = NaN and DWI < epsilon are now treated differently. For the former, the weight is set to zero. For the latter, the DWI value is clipped to epsilon, set to 1e-6 times the maximum DWI value. I think what Rob is suggesting is that we could consider also setting weight = 0 when DWI < epsilon.

Lestropie:
> I think what Rob is suggesting is that we could consider also setting weight = 0 when DWI < epsilon.

Yes, I wasn't advocating one way or the other, just stating that there's decisions to be made explicitly:

- Input NaN values:
  - Set weight to zero (no other choice)
- Negative values:
  - Set weight to zero
  - Set value to epsilon
- Zero values:
  - Set weight to zero
  - Set value to epsilon
- Values lower than epsilon:
  - Set weight to zero
  - Set value to epsilon

jdtournier:
OK, but in practice these are actually all the same, right? We'll have to set NaN (and non-finite) values to something to avoid getting NaNs back out anyway (I expect even if the weights are set to zero, these NaNs will propagate to the output). This applies to NaN, negative and zero values (due to log transform). 

Values lower than epsilon will probably already have tiny weights anyway, but we can also explicitly set their weights to zero, that won't hurt. Setting their values to epsilon won't have any effect on the fit though...?

jdtournier:
> Actually, DWI = NaN and DWI < epsilon are now treated differently. For the former, the weight is set to zero. For the latter, the DWI value is clipped to epsilon, set to 1e-6 times the maximum DWI value. I think what Rob is suggesting is that we could consider also setting weight = 0 when DWI < epsilon.

OK, ignore my last comment - you're right, I get what Rob is suggesting. 

I'm not sure what the right answer is to that one. I'm also not sure how much difference it'll make in practice - though we have seen quite a few issues with zero/negative values in the past following operations that involve interpolation (e.g. upsampling, or even just `dwibiascorrect`, IIRC). In these cases, I think it's fair to assume that the signal genuinely was low, and therefore clipping its value somewhere close to the noise floor makes sense. I'm not convinced the epsilon we're currently using (10<sup>-6</sup> of the b=0 signal) is necessarily justifiable (it's certainly _way_ above the noise floor for any realistic acquisition!), but as long as the _predicted_ signal is low, it'll be assigned a low weight in the fit anyway. 

My concern is that setting their weight to zero means we may end up ignoring quite a few data points in these situations. But then, it's fair to say that these data points are genuinely uninformative, so maybe it would be best just to ignore them...?

Issue #822, opened 2016-11-15
Implement kurtosis metrics


dchristiaens:
As we now have means to compute the diffusion kurtosis tensor in `dwi2tensor`, it would make sense to implement mean kurtosis, axial and radial kurtosis, and kurtosis anisotropy in `tensor2metric` too.

thijsdhollander:
Yes, makes sense.  I've actually looked into that once (when I added a couple of other common tensor metrics in there), but initially stopped pursuing it when I found out there's quite a few (different) definitions of some of those metrics, and didn't have the time to figure out which one(s) are the most "standard" or generally accepted ones. (e.g., see https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4378654/ , but there's many other articles out there listing definitions if I remember well)

Anyone who's into that knows if there's such a consensus or standard around?

Slightly related: if there's interest/demand for it, we could look into implementing Brian Hansen's method for fast MK computation from a minimal acquisition protocol too (this one works directly on the DWI data, doesn't obtain the kurtosis tensor in between).  They got interviewed for MRM highlights very recently: http://www.ismrm.org/qa-with-brian-hansen-sune-jespersen/

EmilieMcKinnon:
Hi Daan and Thijs,

I believe there has been a consensus for a while regarding MK and Kax.

As you know, MK is defined by the diffusional kurtosis averaged over all possible directions. Analytically we use equation 55 in the following paper:  _Jensen JH, Helpern JA. MRI Quantification of Non-Gaussian Water Diffusion by Kurtosis Analysis. NMR in biomedicine. 2010;23(7):698-710. doi:10.1002/nbm.1518._ Similarly, axial kurtosis is defined by the kurtosis along the diffusion eigenvector corresponding to the largest diffusion eigenvalue, or analytically we use equation 58 in the same paper as referenced above. 

There has been a little more discussion regarding the Krad and KFA metrics.

As far as I understand, a consensus has been made on how to calculate Krad. We calculate Krad by averaging the directional kurtosis perpendicular to the first diffusion eigenvector (equation 59).  Previously, people suggested to average the diffusional kurtosis along the second and third principal directions from the DT. Although it turned out that this was physically not optimal and did not result in a parameter that was rotationally invariant.

[...]

PR #2520, opened 2022-10-26
Constrained estimation of the diffusion (kurtosis) tensor and kurtosis metrics

bjeurissen:
This pull request adds support for constrained estimation of the diffusion (kurtosis) tensor (as requested on the forum, e.g. https://community.mrtrix.org/t/the-value-of-tensor-and-eigenvalue-is-negative/4208, https://community.mrtrix.org/t/tensor2metric-works-not-well-to-give-negative-rd-ad-and-rd-inside-the-brain/1333, https://community.mrtrix.org/t/tensor2metric-md-output-image-looks-weird/5519, https://community.mrtrix.org/t/difficulty-with-tensor2metric/3153). It uses the existing WLLS estimator already in place together with the CLLS solver that is used for MSMT-CSD. It enforces non-negative diffusivity, non-negative kurtosis, and monotonic signal decay.

This pull request also adds calculation of kurtosis metrics (addressing #822), since there is quite some overlap between the math for the constraints and that for the metrics. This has also been requested on the forum (e.g. https://community.mrtrix.org/t/patient-vs-control/6119, https://community.mrtrix.org/t/kurtosis-parameters-estimation/3101, https://community.mrtrix.org/t/dwi2tensor-command/213)


cmd/dwi2tensor.cpp
```diff
@@ -47,8 +49,12 @@ void usage ()
 
   SYNOPSIS = "Diffusion (kurtosis) tensor estimation";
 
+  AUTHOR = "Ben Jeurissen (ben.jeurissen@uantwerpen.be)";
+
+  SYNOPSIS = "Diffusion (kurtosis) tensor estimation";
+
```
Lestropie, line 55: File shows both of these fields being duplicated following merge

cmd/dwi2tensor.cpp
```diff
@@ -64,6 +70,20 @@ void usage ()
...
+    "or parameters that correspond to non-monotonic decay of the predicted signal. "
+    "By supplying the -constrain option, constrained optimization is performed instead "
+    "and such physically implausible parameters can be avoided. Depending on the presence "
+    " of the -dkt option, the -constrain option will enforce the following constraints:"
+
+  + "* Non-negative apparent diffusivity (always)."
+
+  + "* Non-negative apparent kurtosis (when the -dkt option is provided)."
+
+  + "* Non-monotonic signal decay in the b = [0 b_max] range (when the -dkt option is provided)."
```
Lestropie, line 85: Monotonic?

cmd/dwi2tensor.cpp
bjeurissen: Thanks for spotting this

cmd/dwi2tensor.cpp
bjeurissen: Yup

jdtournier:
@Lestropie, are we happy to merge this?

Lestropie (approved):
don't feel I'm in a position to review the algebra or the results to a greater degree than what @bjeurissen would have done internally. My only suggestion would be to add option groups to both `dwi2tensor` and `tensor2metric`. With or without that, happy to merge.

bjeurissen:
Currently, the constraints are evaluated along the 300 directions that are also used for CSD, but there is no option to override this. We could also make this an option as for the dwi2fod command.

bjeurissen:
Also note that I have reordered the internal parameter vector such that the S0 parameter comes first, then the DTI and DKI parameters. Had to modify the dwi2tensor _function_ to ensure that DTI tractography would continue to work.

jdtournier:
OK, I just a closer look through how the changes are implemented, I can see a lot of scope for optimisation and generally tidying things up. 

> My only suggestion would be to add option groups to both dwi2tensor and tensor2metric.

That's a good idea, we should do that. 

> Currently, the constraints are evaluated along the 300 directions that are also used for CSD, but there is no option to override this. We could also make this an option as for the dwi2fod command.

Shouldn't be too hard, though not all that essential. But if we're going to make further changes anyway, might as well do that at the same time.

I'll have a go at this when I find the time, but let's hold off on merging until that's done...

bjeurissen:
I have added the options in the meanwhile, will push an update soon.

jdtournier:
Thanks @bjeurissen. I've gone through the code and done my best to optimise it and clean it up. It's mostly about avoiding unnecessary heap allocations, it probably won't make a great deal of difference in terms of raw performance, but every little helps... 

Also added the option grouping and some tests to make sure all my changes weren't going to break anything...

jdtournier (approved):
@bjeurissen: it's all good to go from my point of view, but I think it would be good for you to run a few tests on your own data just to make sure I've not messed anything up that the tests might not have picked up...

bjeurissen:
screenshot0000

This is the absolute difference of MD, RD, AD, MK, RK, and AK, between the MRtrix version and my Matlab version. Note that diffusivities are expressed in milliseconds per micrometer squared. The Matlab version is using ``optimoptions('quadprog','Display','none','LinearSolver','sparse','MaxIterations',400,'OptimalityTolerance',1e-8,'StepTolerance',1e-12,'ConstraintTolerance',1e-8)`` to solve the constrained linear least squares problem.

jdtournier:
@bjeurissen: just to make sure I get this right, the differences you're showing are in the 10<sup>-3</sup> range with values expected to be in the 10<sup>3</sup>? So we're talking NRMSE of the order of 10<sup>-6</sup> - which is a good thing, right?

do you get these discrepancies between your MatLab version and your original C++ code, before I had a go at it (i.e. 66def77)? If so, then I guess the difference is down purely to the non-negativity least-squares solver. But if these differences are as small as I _think_ you're showing, then it's all good, right...?

jdtournier:
By the way, we still get failures in the CI due to ridiculously high values:
```
Error: _diff_image: [ERROR] images "tmp_ms_dkt_constrain.mif" and "dwi2tensor/ms_dkt_constrain.mif" do not match within absolute precision of 9.9999999999999998e-13 (3.92721e+15 vs 0)
```
Any idea what causes such clearly implausible values in the constrained DTI/DKI fit...? I think we see them in the unconstrained fit too (in fact, I think we see _more_ of them), but I would have hope that the constraints would have kept things in check, at least to some extent...

bjeurissen:
> By the way, we still get failures in the CI due to ridiculously high values:
> 
> ```
> Error: _diff_image: [ERROR] images "tmp_ms_dkt_constrain.mif" and "dwi2tensor/ms_dkt_constrain.mif" do not match within absolute precision of 9.9999999999999998e-13 (3.92721e+15 vs 0)
> ```
> 
> Any idea what causes such clearly implausible values in the constrained DTI/DKI fit...? I think we see them in the unconstrained fit too (in fact, I think we see _more_ of them), but I would have hoped that the constraints would have kept things in check, at least to some extent...

What is the S_0 parameter in those cases?

bjeurissen:
> Any idea what causes such clearly implausible values in the constrained DTI/DKI fit...? I think we see them in the unconstrained fit too (in fact, I think we see _more_ of them), but I would have hoped that the constraints would have kept things in check, at least to some extent...

Had a quick look and half of the voxels with large parameter values, have mean signals that look like this:

Screenshot 2022-11-21 at 12 39 17

bjeurissen:
> Will also run my Matlab code on the testing data to see if it does any better.

With the Matlab code on the test data, I get 0 voxels with parameter whose absolute value exceeds 1e10. With the MRtrix code, I get 38.

jdtournier:
OK, that doesn't look great... I'll have a look when I have a minute.

bjeurissen:
> OK, that doesn't look great... I'll have a look when I have a minute.

I think either I made a mistake in the Matlab to C++ port, or the QP solver needs to be fine-tuned.

jdtournier:
OK, quick look at one of the problematic voxels (index 8,8,8) with DKT values in the 10<sup>15<//sup> range, it looks like all elements of the solution vector from the fit are essentially zero, but things explode when [dividing by ADC² on line 207](https://github.com/MRtrix3/mrtrix3/blob/ae6768bafb0d2c8ee26517de0d52542586c8c274/cmd/dwi2tensor.cpp#L207) immediately prior to writing out the final values. Adding a few debug `VAR()` statements shows for that voxel:
```
dwi2tensor [cmd/dwi2tensor.cpp: 208]: x =      3.04758
 2.59274e-20
 1.12399e-19
 3.70399e-19
-2.02181e-20
-3.54659e-19
-2.33655e-19
 2.09404e-22
 1.53197e-23
 4.19291e-22
-3.72779e-22
   4.199e-23
 5.17176e-23
-2.13739e-22
 -1.3353e-22
-1.71754e-22
  3.0461e-22
 7.42286e-23
 6.44091e-23
-1.67706e-22
-5.25227e-23
 8.35545e-23
dwi2tensor [cmd/dwi2tensor.cpp: 209]: (Aneq*x).minCoeff() = -9.80718e-19
dwi2tensor [cmd/dwi2tensor.cpp: 210]: (Aneq*x).maxCoeff() = 1.87711e-19
dwi2tensor [cmd/dwi2tensor.cpp: 211]: adc_sq = 2.87557e-38
```
So essentially, we have a reasonable value for the _b_=0 signal, and everything else is essentially indistinguishable from zero. It's only that last divide by ADC² that messes things up. 

Not sure how to address this... Do you find similar issues in your MatLab implementation? 

One option is to add a small epsilon (e.g. 10<sup>-15</sup>) to the `adc_sq` value, which means the subsequent divide can't explode if the ADC actually computed is much smaller than that. What do you reckon?

Issue #2521, opened 2022-10-28
write_mrtrix_tsf: undefined/removed variable 'tracks'


diffusionmri:
Hi,
In the function [matlab/write_mrtrix_tsf.m], 'tracks.XXX' should be 'tsf.XXX', since the variable 'tracks' was removed. See lines 41-43 and line 29.

Lestropie:
Thanks for the report @diffusionmri.

Looks like the error was introduced in #2455, specifically in 42173769b, which erroneously duplicates the contents of 7cc090a49 from `write_mrtrix_tracks.m` into `write_mrtrix_tsf.m`.

Did you have any interest in committing the fix yourself? There's contribution instructions [here](https://github.com/MRtrix3/mrtrix3/blob/master/CONTRIBUTING.md).

Otherwise: @fionaEyoung?

fionaEyoung:
Woops my bad! 😬  @diffusionmri would you be able to verify that nothing else is broken after the changes in #2522? I don't work with TSF files myself so might be missing some other obvious issue...

PR #2522, opened 2022-11-02
matlab: Fix erroneous reference to non-existent "tracks" identifier

fionaEyoung:
Error introduced by overzealous copypasting in #2455, pointed out in #2521.
While I was there, also replaced other references to the "tracks" variable in the function description and error messages (hope that's appropriate?)


Lestropie (approved):
Absolving @jdtournier of review responsibility.

Certainly looks like it should resolve #2521, and hard to see it introducing any new problem, so I'm happy for it to go to `master` and see if any other issue is subsequently reported.

PR #2525, opened 2022-11-17
configure: fix Qt parameter parsing for MacOS

jdtournier:
Fix detection of parameters for compiling with Qt on macOS. Requires testing on other platforms...


bjeurissen (approved):
Works for me (on Mac)!

jdtournier:
Doesn't work for me on Linux...

jdtournier:
OK, works with that last commit. That will also need double-checking on macOS and MSYS2.

Lestropie:
Tested OK on MSYS2.

PR #2531, opened 2022-12-06
DICOM: fix parsing of DateTime entry

jdtournier:
I noticed a few issues when parsing this entry, which basically boiled down to assuming the fractional part was always 6 digits long - which doesn't hold, it can be any length _up to_ 6 digits... 

[Reference docs](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html) - look for entry `DT` DateTime

I'll run my DICOM tests before merging.

### DIFF
--- a/core/file/dicom/element.cpp
+++ b/core/file/dicom/element.cpp
@@ -423,11 +423,11 @@ namespace MR {
       std::pair<Date,Time> Element::get_datetime () const
       {
         assert (type() == DATETIME);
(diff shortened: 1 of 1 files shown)


Lestropie (approved):
Follows #2466. I don't think I've come across this one given you only implemented it 8 months ago, but it makes sense. 

Will require merge of #2532 to pass CI.

PR #2532, opened 2022-12-06
GitHub Actions: fix dependencies on Ubuntu 22.04

jdtournier:
Since we specify 'ubuntu-latest' for the runner on GitHub Actions, we will occasionally run into problems like this...

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -180,7 +180,7 @@ jobs:
       run: python3 -m sphinx -n -N -W -w sphinx.log docs/ tmp/
 
     - name: install build dependencies
-      run: sudo apt-get install clang-8 libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev
+      run: sudo apt-get install clang libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev
 
     - name: check configure with Python 2
       run: python2 ./configure || { cat configure.log; false; }
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -180,7 +180,7 @@ jobs:
       run: python3 -m sphinx -n -N -W -w sphinx.log docs/ tmp/
 
     - name: install build dependencies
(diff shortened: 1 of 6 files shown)


jdtournier:
@Lestropie: would be a good idea for you to have a good look through 03040fe (to fix "unused variable" warning) and the subsequent e5d327c (probably what was intended for that command, rather than just removing the warned-about code). This came up before in #2392, but hasn't been an issue till now because the GitHub runners were still using Ubuntu 20.04. Now that they've switched to 22.04, `clang-8` is no longer available, and the latest version throws that warning. 

I expect e5d327c is what we want to keep, but we can always revert/amend it if needed.

Lestropie:
RE 03040fe: The SIFT2 one was too long ago, I don't know where I would go looking to remind myself of the justification of the expression for the regularisation scaling. But what's happening currently (ie. `sum_PM` is not used) is consistent with the publication. So content with that being removed.

RE e5d327c: I'm sure this has been discussed before. The issue IIRC is what *should* be stored in `total_count` by `tckedit`. For instance, rather than being the sum of `total_count` across multiple input files, it could instead be the sum of `count` across multiple input files, since `tckedit` could be selecting only a subset of the input streamlines and this is the number that would be reflected in `count`. The definition of what those header entries represent is incomplete; they were added to the header by the 0.2 version of `tckgen`, but the downstream implications were maybe not considered.

RE a0a9a54: This will need to be reverted on `dev` since Python2 support is being dropped (#2215

jdtournier:
> RE e5d327c: I'm sure this has been discussed before. The issue IIRC is what _should_ be stored in total_count by tckedit. 

Yes, I can see this will require further discussion. I just had a look at the current approach, I'm guessing currently, this entry is just ignored and will just be stripped from the output of `tckedit` (worth double-checking...). 3040fe will preserve that behaviour; it may be simplest to stick to that, drop 5d327c for now and create a PR for it to `dev` with a view to 3.1.0 release (it is technically a behaviour modifying change - though we could argue it is also a minor bug) - we can then take the time to have the conversation around what that entry should contain. 

> RE a0a9a54: This will need to be reverted on dev since Python2 support is being dropped (#2215)

Good point. Though it'll take a bit more than reverting - we'll need to strip out all Python 2 testing as well. Will need to watch out for that when merging back to `dev`, thanks for the heads-up.

Lestropie (approved):

Issue #2179, opened 2020-09-28
mrcolour: png upside down


maxpietsch:
`mrcolour` produces pngs that are upside down. macOS, `PNG library: Header: 1.6.37; library: 1.6.37`

```
wget https://github.com/MRtrix3/script_test_data/raw/master/BIDS/sub-02/dwi/sub-02_dwi.nii.gz
mrconvert sub-02_dwi.nii.gz -coord 3 0 -coord 0 15  sag.mif
mrcolour sag.mif gray sag.png

mrconvert sub-02_dwi.nii.gz -coord 3 0 -coord 1 15  cor.mif
mrcolour cor.mif gray cor.png
```
![cor](https://user-images.githubusercontent.com/10046944/94407960-7246aa80-0174-11eb-9db9-b7e4e019d21e.png)
![sag](https://user-images.githubusercontent.com/10046944/94407962-72df4100-0174-11eb-8fc8-127c5dbfba8b.png)

Lestropie:
Applies to the PNG format generally, not specifically to `mrcolour`.

I spent some time getting this right, so not quite sure what happened at the tail end of #1541 to send it skewiff again.

Lestropie:
@jdtournier suggested that user stride request could be used to attribute image axes to 2D PNG axes. This *might* be possible. I need to essentially re-write the relevant section of code as it is to try to address this, so I'll try to keep this prospect in mind.

Lestropie:
Closing as issue became part of the expanded scope of #2713.

PR #2535, opened 2022-12-07
PNG export: Exception on too few axes

Lestropie:
Supersedes #2465.

This section of code is also relevant to #2179.

As described in [this comment](https://github.com/MRtrix3/mrtrix3/pull/2465#issuecomment-1097528572), the usage resulting in a segfault is likely not that intended. But further, attempting to save to `out-[].png` from a 1 x 128 x 128 image could (theoretically at least) yield either a single 128 x 128 image (effectively "ignoring" the square-bracket syntax), or, more faithfully, 128 different 1 x 128 images. Presence of unity axes is already used within the PNG handling to identify the slice orientation to use for PNG files.

For now, this comment instead catches the likely erroneous but also potentially ambiguous usage and errors out, rather than segfaulting. Later, hopefully for `3.1.0`, I'll try to restructure the relevant code in order to address #2179, and hopefully if the code is better structured the right solution in this instance will fall out.

### DIFF
--- a/core/formats/png.cpp
+++ b/core/formats/png.cpp
@@ -132,6 +132,16 @@ namespace MR
       if (H.ndim() == 4 && H.size(3) > 4)
         throw Exception ("A 4D image written to PNG must have between one and four volumes (requested: " + str(H.size(3)) + ")");
 
+      // TODO After looping over axes via square-bracket notation,
+      //   there needs to be at least two axes with size greater than one
+      size_t unity_axes = 0;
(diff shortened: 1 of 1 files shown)


Lestropie:
Should also note that unlike #2465 this is better guaranteed to not change behaviour on `master` under alternative use cases.

jdtournier (approved):
Seems to catch the problematic usage on my system. :+1:

Issue #2527, opened 2022-11-30
ENH New DWI pre-processing command


Lestropie:
Not sure what to call the issue as I'm not sure whether the approach should be implemented as an algorithm for an existing script or whether it should be standalone.

In MRtrix3_connectome I came up with the idea of combining brain masking, bias field correction, and intensity normalisation, into a single step. Relevant code is [here](https://github.com/bids-apps/MRtrix3_connectome/blob/0.5.3/mrtrix3_connectome.py#L1467-L1615). It's primarily a wrapper around `mtnormalise`, but:
-   Target output is bias field corrected and intensity normalized DWIs. Multi-tissue normalisation is the *mechanism* by which such is estimated, but unlike `mtnormalise` itself where the primary output is normalized ODFs, here it's the DWIs that are modulated.
-   The CSF *b*=0 intensity can be used as the reference image intensity.
-   The brain mask is estimated and refined based on `mtnormalise`, by selecting voxels where the sum of tissue signal fractions is at least 50%.

I would like to have this mechanism implemented as a stand-alone command that MRtrix3_connectome would simply invoke. Those who use *MRtrix3* but not MRtrix3_connectome would also be able to see if this approach works on their data. There may be some debate about whether such a script should be distributed with *MRtrix3* or whether it should be a stand-alone module.

Code placement and naming is difficult as it shares functionality with `dwi2mask`, `dwinormalise` / `mtnormalise`, and `dwibiascorrect` / `mtnormalize`.

bjeurissen:
If it works well, I think it would be good to distribute with MRtrix3. Brain masking, bias field correction, and intensity normalisation are exactly the things that routinely break, so it would be really great to have a robust solution in place. I also like that it modulates the DWI's instead of the ODFs as it makes the pipeline more generic.

Lestropie:
Closing as discussion can occur over at #2552.

PR #2552, opened 2022-12-28
dwibiasnormmask and mtnorm algorithms

Arshiyasan:
As described in #2527 this script first derives an initial brain mask using the legacy MRtrix3 dwi2mask heuristic (based on thresholded trace images), and then iteratively performs response function estimation, multi-tissue CSD (with a lower lmax than the dwi2fod default, for higher speed), and mtnormalise to remove any bias field, before the DWI brain mask is recalculated. These steps are performed iteratively, since each step may influence the others. The core algorithm has been adapted from https://github.com/BIDS-Apps/MRtrix3_connectome
It is possible to slightly modify the code to output bias field corrected DWI images as well, though I'm not sure it would then make sense as a "dwi2mask" algorithm.


Lestropie:
Thanks for giving this a go @Arshiyasan.

So my musing in #2527 echoes your final point: this approach provides multiple functionalities, and placing it exclusively in `dwi2mask` / `dwinormalise` / `dwibiascorrect` but then adding extra command-line options to export derived mask / corrected DWIs is unintuitive.

Where my thinking was leading was:

1.  Create a new command, with some appropriate name; let's for now call it `dwidothething`.
2.  Move the primary functionality of the approach out of the `bin/dwidothething` `execute()` function, and instead place it in some other named function; let's call it `thething()`.
3.  Add function parameters to function `thething()` that control what inputs are pre-defined vs. must be calculated, and what outputs are desired vs. not.
4.  Add a `dwi2mask` algorithm that executes `thething()` from `bin/dwidothething` but only yields the mask.
5.  Add a `dwibiascorrect` algorithm that executes `thething()` from `bin/dwidothething` but only applies bias field correction to the input DWI.
6.  Add a `dwinormalise` algorithm that executes `thething()` from `bin/dwidothething` but only applies global intensity normalisation to the input DWI.

This is a decently larger amount of effort, and its success is predicated on the determination of an appropriate name / acronym for the approach as a whole (since this would define not only the command name, but also the corresponding algorithm names for the other scripts from which the primary functionality can be invoked).

It's iterative, multi-tissue-based, bias-field-correcting, fluid-signal-normalising, tissue-density-based masking. Anyone capable of deriving a good acronym from that? 😬

Lestropie:
Been thinking about this a little. I've become less convinced that I want to homogenise the code itself. It's moreso that there's a proposed iterative algorithm for which there are individual concepts that may be applicable to certain sub-tasks in isolation.

So here's what I'm thinking currently:

-   `dwibiascorrect mtnorm`:
    -   Performs full iterative approach
    -   Primary output is bias-field-corrected DWI
    -   Option to control number of iterations
    -   Option to also export resulting estimated mask
    -   Option to control overall intensity normalisation, with the default being setting the *b*=0 CSF value to eg. 1000; but for instance another approach is to preserve the total spin density within the brain mask after bias field correction

-   `dwi2mask mtnorm`:
    -   Does *not* internally perform bias field correction
    -   Only derives a mask based on thresholding total density of multi-tissue decomposition
    -   Option to change tissue density threshold

-   `dwinormalise mtnorm`
    (note that "`dwinormalise individual`" would probably need to be renamed):
    -   Does *not* internally perform bias field correction
    -   Only performs global intensity normalisation
    -   Option to set global intensity normalisation reference; default is setting mean *b*=0 CSF to eg. 1000, but could for instance just use what `mtnormalise` does, which is to have a (geometric) mean tissue density of 1.0 throughout the mask

So `dwibiascorrect norm` would be the command that I would expect most people would use, whereas `dwi2mask mtnorm` and `dwinormalise mtnorm` extract specific capabilities of the overall concept.

bjeurissen:
Looks like a nice approach. Starts to bear resemblances to tools like the unified segmentation approach in SPM, where tissue segmentation, bias field correction, and masking are all merged into one.

You could also consider adding a simpler iterative version that I have found works quite well:

```
dwibiascorrect ants
dwi2mask on debiased images
dwibiascorrect ants with updated mask
dwi2mask on updated debiased images
....
```

In my experience this results in both better bias field correction and better masking.

Lestropie:
The reference approach that we're trying to integrate is *already* iterative between a number of steps. What I'm looking at doing is stripping out the iterative nature for *some* individual steps and providing them separately. However I'm not sure if I want to *completely* strip out the iterative nature and then provide an explicit wrapper that performs the iteration using the underlying components, because that would involve redundant calculations.

For reference, the original approach that we're trying to port across is:

1.  Generate initial brain mask using "default" approach
2.  For fixed 2 iterations:
    1.  Estimate 3 response functions
    2.  Perform MSMT CSD but only WM lmax=4
    3.  Run `mtnormalise`:
        1.  Using eroded mask
        2.  Capturing estimated bias field
        3.  Capturing global scaling factor
        4.  Capturing corrected ODFs
    5.  Apply bias field correction to DWIs
    6.  Scale DWIs so that mean *b*=0 CSF signal (based on estimated RF) is 1000.0
    7.  Estimate brain mask based on sum of ODFs > 0.5

So steps 2.4, 2.5 and 2.6 can theoretically be extracted and provided as stand-alone, non-iterative versions in `dwibiascorrect`, `dwinormalise` and `dwi2mask` respectively.

However providing an iterative algorithm that explicitly calls those three scripts would require redundant calculations, since:
-   2.4 depends on 2.3.2
-   2.5 depends on 2.1 & 2.3.3
-   2.6 depends on 2.3.4

So the question is, which of the following is best from an interface perspective:

-   One of `dwibiascorrect` / `dwinormalise` / `dwi2mask` internally implements the whole iterative approach, while the other two just implement subsets thereof;
-   All three of those scripts get a new algorithm that implement just the relevant portion of the whole iterative approach, and there is instead some other script (name pending) that does the whole iterative approach?
    -   And if so, should that script invoke the relevant components of `dwibiascorrect` / `dwinormalise` / `dwi2mask`, which would involve redundant calculations, or should it instead use the code implementation as in MRtrix3_connectome?

In retrospect, typing it out, I'm probably now leaning toward the latter, even if I'm not yet sure what to call it.

Arshiyasan:
The latter approach makes more sense to me too, we could have three non-iterative commands which are called by `unnamed_iterative_command`. To avoid duplicating calculations, what if each command could be optionally provided the files that corresponding steps in the iterative algorithm require? I.e. the three commands would at minimum require a dwi image, but:
- The `dwibiascorrect` command could also explicitly be provided with `-bias_field`
- The `dwinormalise` command could be provided `-csf_rf_bzero_lzero` and `-csf_balance_factor`
- The `dwi2mask` command could be provided `-normalised_odfs`

Within the `unnamed_iterative_command` wrapper the commands would be given these inputs explicitly and thus skip the relevant calculations. But when called individually, they first run steps 1 through 2.3.4 as implemented in the `unnamed_iterative_command` and use their required inputs (I imagine the steps could be bundled together and run non-iteratively when `unnamed_iterative_command` is called in this context).

Lestropie:
> To avoid duplicating calculations, what if each command could be optionally provided the files that corresponding steps in the iterative algorithm require?

That's a reasonable option, and would result in less code duplication. But it only makes sense IMO if the size of the calculations bypassed by providing pre-calculated data is smaller than the calculations that are still performed. If providing pre-calculated data means that the command is performing a minority of its intended operation, it starts to feel unintuitive.

-   Providing `dwibiascorrect` with a pre-calculated bias field is pretty redundant given it could be achieved with a single `mrcalc -div` call.
-   Providing `dwinormalise` with pre-calculated CSF RF *b*=0 *l*=0 and `mtnormalise` CSF balance factor is redundant given it's literally just an `mrcalc -mult` call at that point.
-   Providing `dwi2mask` with pre-calculated normalised ODFs is more data than it needs given it's only going to sum those and apply a 0.5 threshold, which again can be achieved with a single `mrcalc` or an `mrmath | mrthreshold`.

So I'm still leaning towards focusing on a single script---let's call it `dwibiasnormmask` in the absence of a better alternative---that does the whole iterative approach. The extra algorithms for `dwibiascorrect` / `dwinormalise` / `dwi2mask` would be primarily to demonstrate how those steps are done, and provide access to them to anyone who wants to attempt an alternative wrapper.

> in addition to dice_coefficient > 0.999, the loop stops if (the masks start to diverge) AND (dice_coefficient is at least 0.99) and a warning is issued

By "diverge", do you mean that the inter-iteration dice coefficient of one iteration is less than that of the previous iteration?

There's scope for having different termination criteria in addition to different parameters for each of those criteria. So for instance, you could choose from:

-   Having a fixed number of iterations
-   Detect when the dice coefficient is less than the previous iteration, and retain the bias field / normalisation factor / mask of that previous iteration as the output.

I suppose theoretically another thing that could be attempted to improve convergence would be to add a damping factor to the bias field estimates; ie. bring all values a little closer to 1.0 before applying to the ODFs and computing the mask. Might leave that out of scope for now, but it's an idea.

Lestropie:
Update based on private communication with @Arshiyasan:

Within the iterative estimation, the brain mask derivation (sum of `mtnormalise`-d ODFs > 0.5) can be a weak link. So option to progress may be:

1.  Stand-alone `dwi2mask` algorithm implements the `dwi2response | dwi2fod | mtnormalise | mrthreshold` approach.

3.  Command `dwibiasnormmask` allows for user selection of the algorithm that will be used internally for the brain mask estimation step.
    This functionality is already present for the sake of other scripts since #2197.
    What could be done is, if the user selects / the default is the `dwi2mask` algorithm in point 1, then rather than explicitly invoking `dwi2mask`, it would just do the same thresholding process internally using the pre-calculated ODFs, whereas for any other algorithm `dwi2mask` will be invoked explicitly.

This would potentially allow for combining the benefits of the masking / bias field correction iteration with superior brain masking algorithms; and for certain combinations would hopefully lead to fairly quick convergence across iterations.

Arshiyasan:
Okay, I have implemented both the three 'mtnorm' algorithms and the separate 'dwibiasnormmask' algorithm. This last algorithm first checks if SynthStrip is available and if it is, uses it to both create an initial meanbzero mask and to mask the balanced tissue sum images, both with the border from the brain surface set to zero (this configuration is what I am using for my own project).

Lestropie:
Cool, sounds like this should be at the point now where I can go through and make any desired tweaks.

The other thing that needs to be done is to add tests. For anything that invokes external tools, we just make sure that the script executes to completion, including if there are different command-line options specified. For anything that only uses *MRtrix3*, we take the output of the script for the current version of the code, make sure it's "correct" to the best of our ability, save it in the [`script_test_data` branch](https://github.com/MRtrix3/script_test_data), and then within the test suite explicitly compare the output of the script against that pre-calculated output; this tests for any unexpected changes in behaviour due to future code changes. The latter is a bit trickier and will require me to add the data, but you could have a go at adding the former if you wish.

Arshiyasan:
I had a go at the former and added a few tests to check dwibiasnormalise (when calling SynthStrip) runs with different options, which did not produce any errors.

Lestropie:
I've been playing with this on and off, but I need to focus on manuscripts for a little while so I've committed what I have thus far. I haven't looked at the other script algorithms yet, just `dwibiasnormmask`. Hopefully the changes in [58b2dc6](https://github.com/MRtrix3/mrtrix3/pull/2552/commits/58b2dc65b64a03e7fca022112fd13dc653077092) are somewhat self-explanatory. I haven't tested every possible use case, and probably need to add a reasonably long list of tests to verify. But it should still at least "work".

Lestropie:
- [ ] Remove erroneous addition of personal entries to `.gitignore`
- [ ] Remove unwanted additions from #2558 (or just merge that to `dev` first)
- [ ] @Lestropie to check `dwi2mask mtnorm` code
- [ ] @Lestropie to check `dwibiascorrect mtnorm` code
- [ ] @Lestropie to check `dwinormalise mtnorm` code
- [ ] Discuss naming of `dwinormalise` algorithms:
    Propose that existing algorithm "`individual`" be changed to "`manual`", since it's no longer the only algorithm that works on an individual basis
- [ ] Remove unwanted additions from #2565 (or just merge that to `dev` first)
- [ ] Add execution tests for:
    - [ ] `dwi2mask mtnorm`
    - [ ] `dwibiascorrect mtnorm`
    - [ ] `dwinormalise mtnorm`

Arshiyasan:
@Lestropie I **_think_** I have correctly reverted the two erroneous merges and edited .gitignore!

Lestropie:
[259318a](https://github.com/MRtrix3/mrtrix3/pull/2552/commits/259318ae38f6813787ff0ac10f7b3b10f39cd87c) is no good, it reverts all of the copyright headers; don't want multiple commits pushing them back and forth again. I might need to go through with some cherry-picking to separate this from #2565 and #2558 (while you yourself have been playing with all of them, it would be preferable to have them separate in terms of commit history & PRs).

Lestropie:
Closing due to being usurped by #2604.

PR #2558, opened 2023-01-09
5ttedit: Support for floating-point input images

Lestropie:
Started trying to document how I would like to see this modified, but eventually decided that it wouldn't take a whole lot more effort to just implement it.

@Arshiyasan Feel free to give a try if you'd like; might be preferable to thresholding at 0.5.

- [ ] Add capability to provide NaN values at input in order to force tissue volume fraction to zero
      (currently a zero in an input image means don't make any modification)
- [ ] Fully test all new features

### DIFF
--- a/cmd/5ttedit.cpp
+++ b/cmd/5ttedit.cpp
@@ -35,34 +35,42 @@ void usage ()
 
   AUTHOR = "Robert E. Smith (robert.smith@florey.edu.au)";
 
-  SYNOPSIS = "Manually set the partial volume fractions in an ACT five-tissue-type (5TT) image using mask images";
+  SYNOPSIS = "Manually set the partial volume fractions in an ACT five-tissue-type (5TT) image";
+
+  DESCRIPTION + "The user-provided images are interpreted as desired partial volume fractions "
+                "in the output image. For any voxel with a non-zero value in such an image, this "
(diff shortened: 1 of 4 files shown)


Arshiyasan:
Thanks for implementing this! The code seems to have an issue with re-normalizing tissue fractions. For example, trying to add pathological tissue to the brainstem (here with `-path 5tt_BS_capped.mif`) does add correct values to the last volume of the 5TT image, but creates these bands of others tissues (here CSF and WM).
path
csf
wm

Lestropie:
That's a particularly unusual effect given that the command operates entirely independently on each voxel. @Arshiyasan any chance of sharing some data so that I can attempt to reproduce?

Lestropie:
Should be fixed by [9f6dad5](https://github.com/MRtrix3/mrtrix3/pull/2558/commits/9f6dad54a2ae7f7939cdd6b4a3e7b05f456f93c6).

Arshiyasan:
Interesting... Reading the fix, I can't quite understand what the reason for those strange bands was. Did you find out what was causing that pattern?

Lestropie:
That was me failing to always move the thread-specific position in the input image before reading from it; therefore sometimes it would be reading input image data from a previously-processed voxel.

Issue #2564, opened 2023-01-16
ENH tckgen: Save anisotropy metric during propagation?


Lestropie:
Contemplating an implementation change to some of the internals of `tckgen`.

When performing truncation of a streamline due to its attempting to enter and then exit sub-cortical grey matter, it may be necessary to know the value of the algorithm-specific metric of anisotropy at each vertex to choose the vertex at which the truncation will take place. Currently, this is done by interpolating at the vertex location, estimating the streamline tangent at that location (which currently makes an assumption about being a first-order method), and sampling the relevant anisotropy metric at that location. This is a bit clumsy, and the code has to appear in a location other than where I'd like it to be. It's likely also a bit slow, but then again it's only 2-3% of streamline terminations for which this needs to be done.

The alternative would be to dynamically store the value of the anisotropy metric as computed during propagation alongside the vertex locations themselves. This would add some clock cycles for every streamline (though shouldn't be nearly as many as was necessary to compute that metric).

In addition to better facilitating that truncation, there would be a "free" enhancement of the ability to export a `.tsf` showing the value of this metric at every streamline vertex.

It might require some creativity to get it working with RK4.
iFOD2 should be calculating this internally already, it just needs to be buffered per arc.

Issue #2111, opened 2020-06-30
Back-end: Remove all char* / char** usages


Lestropie:
There's still quite a few places in the back-end where C-style strings, and null-terminated arrays of such, are used, probably having been carried over from 0.2 code; e.g. I'm thinking `MR::Argument::type_choice()`. Nowadays these irk me whenever I come across them, and would rather remove the whole lot in favour of `std::string` / `MR::vector` / `std::initialiser_list` etc..

Lestropie:
Giving this a bump due to having encountered an issue with such.

Replacing [this code](https://github.com/MRtrix3/mrtrix3/blob/3.0.2/cmd/mrfilter.cpp#L126):
```
DESCRIPTION
  + "The available filters are: fft, gradient, median, smooth, normalise, zclean."
```
with:
```
DESCRIPTION
  + "The available filters are: " + join(filters, ",").c_str() + "."
```
, in an attempt to not have to duplicate the list of available filters in the code, results in reading from freed memory (which `gdb` doesn't catch). `DESCRIPTION` only accepts `const char*`s, not `std::string`s, so a sensible string concatenation can't really be done there.

Can get around it by doing something like:
```
const std::string filter_list_description = "The available filters are: " + join (filters, ",") + ".";

...

DESCRIPTION
  + filter_list_description.c_str()
```
, but it doesn't prevent the risk of memory issues if a programmer isn't careful. Just abandoning unsafe pointers and null-terminated arrays altogether would feel more modern.

Lestropie:
It occurred to me during #3206 that #2911 does not fully resolve this. Now that `check_syntax` is working again on `dev`, it should be possible to check the `cpp/` directory for any explicit use of `char*`. There's only a handful left.

PR #2565, opened 2023-01-16
tckgen: SGM truncation changes

Lestropie:
Implementing in case it is of utility in a particular collaboration, but it may be preferable behaviour going forward also. Primary info in 94b9df76. @Arshiyasan feel free to try. Draft for now.

- [ ] Should the roulette wheel sampler set the probabilities for (non-)selection for vertex truncation be proportional to the anisotropy measure itself, or the difference between the empirical anisotropy measure and the cutoff threhsold?
      (though this could get slightly tricky due to these being measured from scratch---see #2564---such that it's possible that the measure could be *below* the cutoff threshold, at which point the math breaks down)

### DIFF
--- a/src/dwi/tractography/ACT/act.cpp
+++ b/src/dwi/tractography/ACT/act.cpp
@@ -31,6 +31,8 @@ namespace MR
 
         using namespace App;
 
+        const char* sgm_trunc_choices[] = { "entry", "exit", "minimum", "random", "roulette", nullptr };
+
         const OptionGroup ACTOption = OptionGroup ("Anatomically-Constrained Tractography options")
 
           + Option ("act", "use the Anatomically-Constrained Tractography framework during tracking; "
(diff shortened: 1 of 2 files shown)


Arshiyasan:
Thanks for implementing this algorithm! Unfortunately building on both my Ubuntu 20.04 system and the university's HPC cluster fails with the error `multiple definition of 'MR::DWI::Tractography::ACT::sgm_trunc_choices'`
This seems to have also happened in the checks (both linux-gcc and linux-clang builds). I couldn't figure out what this error stems from.

Lestropie:
[df8c33a](https://github.com/MRtrix3/mrtrix3/pull/2565/commits/df8c33a5ccfd9062d1859eee0f0219bd7dcc6a46) should fix it. Yet another case of #2111 giving me the irrits...

Issue #2500, opened 2022-08-16
ENH dwi2mask (dev): SynthStrip


Lestropie:
https://www.sciencedirect.com/science/article/pii/S1053811922005900
https://surfer.nmr.mgh.harvard.edu/docs/synthstrip/

Simply one to add to the list.
On initial inspection it appears to be single-volume input only, ie. can't use multiple shells.

bjeurissen:
Tried using it and seem to work as advertised

Lestropie:
- [ ] Initial familiarization
    - [ ] Install latest FreeSurfer version
    - [ ] Execute `mri_synthstrip` on exemplar data
    - [ ] Examine command-line interface and options available

- [ ] Generate template `dwi2mask` algorithm
    - [ ] Ensure that the `dev` branch of *MRtrix3* is installed and working
    - [ ] Create copy of an existing `dwi2mask` algorithm with a new name
    - [ ] Ensure that "new" algorithm is accessible from the command-line through the `dwi2mask` interface
    - [ ] Ensure that behaviour of new algorithm is identical to that copied

- [ ] Generate implementation of SynthStrip-based `dwi2mask` algorithm
    - [ ] Generate mean *b*=0 image from DWI input
    - [ ] Convert to NIfTI format as required for compatibility with FreeSurfer
    - [ ] Capture output of SynthStrip and convert back into *MRtrix3* `.mif` format
        - [ ] Check if the SynthStrip output file has the same orientation as the image that was input to `dwi2mask`; if it does not, modify strides of `.mif` file to match that of the user's input DIW series
        - [ ] Make output file of bitwise data type
    - [ ] Implement command-line options that echo SynthStrip options, and feed any user-specified options to the SynthStrip command
    - [ ] Integrate SynthStrip algorithm into the `dwi2mask consensus` algorithm
    - [ ] *(Optional)* When generating NIfTI for input to SynthStrip, modify strides such that image conforms to the orientations used in network training; this will obviate the need for SynthStrip to perform this step internally

- [ ] Add external dependencies for new `dwi2mask` algorithm to containers
    - [ ] Repeat instructions in #2529 in this context

- [ ] Generate Pull Request

PR #2575, opened 2023-01-25
dwi2mask new algorithm Synthstrip

Ruobing-Chen:
I created a file that adds Synthstrip algorithm, which uses the skull stripping method to produce a brain mask, into the options of dwi2mask algorithms list. However, in order to make use of this additional function, the user needs to have Synthstrip or FreeSurfer v7.3 or a later version installed. This is related to https://github.com/MRtrix3/mrtrix3/issues/2500


Lestropie:
- [ ] Requires execution of `docs/generate_user_docs.sh` and committing of the resulting additions / changes
- [ ] May require modifications following use of `run_pylint`
- [ ] Ideally would include addition of unit test, which will additionally require addition of new data to the `script_test_data` repository; I will work with @Ruobing-Chen on this
- [ ] I'll additionally make some suggests for code modifications in the PR diff

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+
+import shutil
+import argparse
+import os
+import warnings
+from mrtrix3 import MRtrixError
+from mrtrix3 import app, run
+
+
+DEFAULT_CLEAN_SCALE = 2
```
Lestropie, line 24: This is unused, as is the `-clean_scale` option. This is a component of the `legacy` algorithm, since it duplicates the existing behaviour of the `dwi2mask` C++ binary on `3.0.x`. But it should be omitted here. 

(Users can nevertheless use this functionality explicitly after the fact using `maskfilter clean` if they choose)

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+SYNTHSTRIP='mri_synthstrip'
```
Lestropie, line 25: Constant name should better reflect fact that this is a command name; eg. `SYNTHSTRIP_CMD`

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def usage(base_parser, subparsers): #pylint: disable=unused-variable
+  parser = subparsers.add_parser('synthstrip', parents=[base_parser])
+  parser.set_author('Ruobing Chen(chrc@student.unimelb.edu.au)')
```
Lestropie, line 30:
```suggestion
  parser.set_author('Ruobing Chen (chrc@student.unimelb.edu.au)')
```

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+  parser.set_synopsis('Use the Synthstrip heuristic (based on skull-stripping method)')
```
Lestropie, line 31:
```suggestion
  parser.set_synopsis('Use the FreeSurfer Synthstrip method')
```

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+  parser.add_argument('input',  help='The input DWI series')
+  parser.add_argument('output', help='The output mask image')
+  options=parser.add_argument_group('Options specific to the \' Synthstrip \'algorithm')
+  options.add_argument('-s', action='store_true', default=False, help='The output stripped image')
```
Lestropie, line 35: Rather than a "`store_true`" action (in MRtrix3 parlance we'd typically call this a command-line option "flag"), where it's just the "`-s`" that gets specified at the command-line, this should instead just be a "standard" command-line option (ie. no `action` or `default` kwargs). The user can then provide at the command-line, after the "`-s`", the location at which they wish for the (optional) stripped image to be saved.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+  options.add_argument('-h', action='store_true',default=False, help='The help messeage of Synthstrip algorithm')
```
Lestropie, line 36: I'd omit this. If someone wants to see the help page of the external command that gets invoked, they should instead interact with that external software directly.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def needs_mean_bzero(): #pylint: disable=unused-variable
+  return True
+
+
+
+def execute(): #pylint: disable=unused-variable
+  
+  run.command('dwiextract input.mif - -bzero | mrmath - mean mean_bzero.mif -axis 3')
```
Lestropie, line 62: Returning `True` in `needs_mean_bzero` should mean that this command is not required; the `dwi2mask` wrapper should pre-compute it for you.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+  
+  run.command('mrconvert mean_bzero.mif 3dbzero.mif -axes 0,1,2')
+  app.cleanup('mean_bzero.mif')
+  run.command('mrconvert 3dbzero.mif 3dbzero.nii')
+  app.cleanup('3dbzero.mif')
+  
+  
+  synthstrip_cmd = shutil.which("mri_synthstrip")
```
Lestropie, line 70: The name of the command is stored in the `SYNTHSTRIP` constant; so that should be used here.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+    cmd_string+=' -g'
+    
+  if app.ARGS.nocsf:
+    cmd_string+=' --no-csf'
+
+  if app.ARGS.b is not None:
+    cmd_string += ' -b' + ' '+str(app.ARGS.b)
+
+  if app.ARGS.mo:
+    cmd_string+= ' --model' + father_path + ' -model.nii'
```
Lestropie, line 97: Will need to look into what the intended functionality is here.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+
+  #try:
+  #  #run.command( 'mri_synthstip -i 3dbzero.nii -o' + str(app.ARGS.o)+ '-m' + str(app.ARGS.m)+'-no-csf'+str(app.ARGS.no-csf)+'-b'+str(app.ARGS.b))
+  #  run.command( 'mri_synthstrip -i 3dbzero.nii -o stripped.nii -m synthstrip_mask.nii')
+  #except run.MRtrixError:
+  #  app.warn('The input image may have more than 1 frames.')
```
Lestropie, line 103: Code that was written during development but not used in the final version shouldn't be left as comments, it should be removed.

lib/mrtrix3/dwi2mask/synthstrip.py
```diff
@@ -0,0 +1,114 @@
...
+  if app.ARGS.h:
+    run.command(cmd_string)
+    return 'input.mif'
+  else:
+    run.command(cmd_string)
+    run.command('mrconvert synthstrip_mask.nii synthstrip_mask.mif -datatype bit')
```
Lestropie, line 109: If variable `output_file` is created to store path "`synthstrip_mask.mif`", the code should make use of it. Otherwise, you could erroneously modify one but not the other, and the script would stop working; whereas if you use the path as stored in the variable throughout, you're protected against introducing that sort of problem.

Issue #2195, opened 2020-10-08
ENH dwi2adc: Support general DWI input


Lestropie:
If the input provided to the `dwi2adc` command is *not* pre-generated trace-weighted volumes for each unique *b*-value, but a 4D DWI series more generally, then the command should still be able to produce an ADC image from such; whether by invoking `DWI::Shells` and internally calculating those trace-weighted images, or if that's not possible, just feeding the individual DWI *b*-values & intensities directly into the solver.

Lestropie:
Promoted to `3.1.0` milestone as explaining how to use this command keeps annoying me.

Lestropie:
In addition to changing the interface of the command, it should be possible to change the estimation procedure as well. E.g.:

-   Do not invoke shells; instead, use the same estimation procedure as is used in `dwi2tensor` (default IWLS with 2 iterations), with each DWI volume used as an independent input.
-   Use the same approach as above, but *do* invoke the shelling algorithm, and use the *l*=0 term of the SH rather than the raw powder average in order to get an unbiased estimate of the mean DWI signal in each shell.

Lestropie:
As much as I want this change, it's a bit of work, and not strictly necessary for `3.1.0`, so I'm changing the milestone.

PR #2578, opened 2023-02-01
IVIM fitting

dchristiaens:
These changes extend `dwi2adc` with IVIM fitting, i.e., modeling S(b) = S(0) * ((1-f) * exp(-D * b) + f * exp(-D' * b)). The fitting strategy uses a 2-stage approach, where the first step estimates the ADC and A = S(0) * (1-f), and the second stage fits D' and calcuates the remaining parameters.

I've tested this on spinal cord data and the resuts are as I'd expect, but I would welcom (external) testing by users more experienced with IVIM.


dchristiaens:
The build fails in CI because I used a new matrix indexing feature introduced in Eigen 3.4. Should I rewrite the code to work against 3.3 or can we set 3.4 as a new minimum requirement?

jdtournier:
> Should I rewrite the code to work against 3.3 or can we set 3.4 as a new minimum requirement?

If it's on `dev`, I'm happy to set 3.4 as the minimum requirement. 

We would need to update the CI to match though... Will require changes to the [GitHub Actions file](https://github.com/MRtrix3/mrtrix3/blob/dev/.github/workflows/checks.yml), potentially involving an explicit clone of the appropriate version of Eigen (maybe pulled from [their git repo](https://gitlab.com/libeigen/eigen))?

Lestropie:
If, as appears to be the case, this modifies the non-IVIM `dwi2adc` functionality to take as input a DWI series rather than trace-weighted volumes, this closes #2195.

cmd/dwi2adc.cpp
```diff
@@ -88,25 +162,21 @@ void run () {
...
+  bool ivim = opt.size();
+  int bmin = get_option_value("cutoff", 120);
 
   Header header (dwi);
   header.datatype() = DataType::Float32;
   DWI::stash_DW_scheme (header, grad);
   PhaseEncoding::clear_scheme (header);
   header.ndim() = 4;
-  header.size(3) = 2;
+  header.size(3) = ivim ? 4 : 2;
```
Lestropie, line 174: So for non-IVIM, the output is still a 4D image with two volumes, being the predicted S0 and ADC? Personally I'd prefer the output in this scenario to be a 3D image containing ADC, and requesting the predicted S0 output should be a command-line option and a separate image. It's weird opening up the image that is the output of a command called "`dwi2adc`" and seeing what looks like a DWI.

Lestropie: #2804 is listing for dimensionality of output data for `dwi2adc`; did not realise it was not within scope of #2195. But this PR might nevertheless be the right context to contemplate.

dchristiaens: This PR (#2578) does not change the interface for dwi2adc: the output has always been a 4D image with 2 volumes as far as I can remember. I don't particularly like this either and we can think about changing the interface later, but the focus of this PR is on IVIM fitting.

dchristiaens:
As discussed in the meeting this morning: I agree we can think about the interface for `dwi2adc`, but this will require a bigger effort that may go into a later release. This PR is a direct extension of the current functionality, so I'd advocate to include it in the 3.1 release.

Unless there are any further objections, this should be ready to merge.

dchristiaens:
@jdtournier Are you happy to merge this PR?

cmd/dwi2adc.cpp
Lestropie: Omitting interface change for the sake of getting the PR though; already struggling with getting PRs resolved.

Lestropie (approved):
:+1: Let's merge as it stands now, and see if there are any revisions required once it hits the user community.

Issue #2529, opened 2022-12-05
dwi2mask: Add new prerequisites to containers


Lestropie:
Follows comment in #2370.

On `dev`, following #2197, there are new external dependencies for the new `dwi2mask` command.
These need to be integrated into the docker and singularity container recipes.
Note that if #2500 is implemented that will come with a new dependency also.

Lestropie:
- [ ] Need to fork *both* the *MRtrix3* repository *and* the "containers" repository

- [ ] "Containers" repository: Generation of new tarballed dependencies
    - [ ] In `Dockerfile`, install any new external software dependencies
    - [ ] In list of commands that are used to capture which files are accessed and which are not, add commands that will execute the new `dwi2mask` capabilities on the *MRtrix3* `dev` branch (what will become version `3.1.0`)
    - [ ] Follow instructions in "containers" repository README to generate new tarballs
    - [ ] Name new tarballs appropriately and upload to OSF
    - [ ] Create Pull Request to propose merge of changes

- [ ] *MRtrix3* repository: Import new tarballs
    - [ ] Update container recipe files to import new tarballs during container construction
    - [ ] Create Pull Request to propose merge of changes (to `dev` branch)

PR #2583, opened 2023-02-07
Dwi2mask synthstrip: New algorithm for dwi2mask

Ruobing-Chen:
Add a new script to execute the algorithm called Synthstrip which uses the skull-stripping method to produce a brain mask. The script can execute normally and should produce the desired output. However, the difference-comparing test is unnecessary because it makes use of additional software outside the MRtrix3. If the developer changes the source code of the software, it is possible for the script to produce a different output file while no changes are made to the script. In order to function the Synthstrip correctly, users need to have FreeSurfer v7.3.2 or a later version installed on their native systems. This is related to #2500

### DIFF
--- a/bin/dwi2mask
+++ b/bin/dwi2mask
@@ -94,7 +94,8 @@ def execute(): #pylint: disable=unused-variable
               + ' |'
               + ' mrconvert - '
               + path.from_user(app.ARGS.output)
-              + ' -strides ' + ','.join(str(value) for value in strides),
+              + ' -strides ' + ','.join(str(value) for value in strides)
+              + ' -datatype bit',
               mrconvert_keyval=path.from_user(app.ARGS.input, False),
               force=app.FORCE_OVERWRITE)
 
--- a/docs/reference/commands/dwi2mask.rst
+++ b/docs/reference/commands/dwi2mask.rst
@@ -870,7 +870,7 @@ dwi2mask synthstrip
 Synopsis
 --------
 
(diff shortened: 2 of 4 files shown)


Lestropie:
- May be worth looking into final outstanding requirements from #2500, which relate to image strides relative to input image / what SynthStrip requires internally. However not lumping this on @Ruobing-Chen, I'll need to look at it myself.

- Integration into containers is being deferred to #2529.

Lestropie:
In retrospect, the strides issue may not be worth the effort. The algorithm is making use of the `dwi2mask` wrapper to generate the mean *b*=0 image for SynthStrip input; further, the `dwi2mask` wrapper takes the result and applies to it the same strides as those possessed by the input DWI. So the only benefit would be in potentially providing SynthStrip with a NIfTI that it doesn't have to reorder, and that would involve bypassing the `dwi2mask` code that provides the mean *b*=0 image. And at that point it's only a 3D image that needs to have the strides changed. So I think we'll just omit that feature.

Lestropie:
If command `mri_synthstrip` is not available, the script should also look for `synthstrip-singularity`, and make use of that instead if it is available.

Lestropie:
The algorithm additionally needs to add the reference for the manuscript where the SynthStrip method is described.

Lestropie:
f910280f4821bfa536a2a00c88b5ce3bd5968987 claims that it is adding the relevant reference, but it does not; it only adds the capability to utilise the SynthStrip Singularity container.

PR #2589, opened 2023-02-20
tckgen: New algorithm PTT

### LINKED ISSUE #2160: ENH tckgen: Parallel Transport Tractography implementation
As discussed on [forum](https://community.mrtrix.org/t/how-does-ifod1-2-internally-handle-input-step-size-and-curvature-constraints/4000/3).

Lestropie:
Implements the Parallel Transport Tractography (PTT) algorithm.
Pinging @baranaydogan for assistance with verifying implementation.

Creating draft PR in order to produce notifications of updates.
Still a fair few things I'd like to attempt / change before merging; but perhaps here is a better place to list those than within the code itself...
There's also a debate to be had around having the implementation behave *exactly* as published / implemented elsewhere vs. homogenizing certain things with *MRtrix3* philosophy / design... hopefully what I mean by that will be clearer from the items below.

- [ ] Command-line option to switch calculation of path probability between FOD amplitude mean and geometric mean (PTT as published does the former, iFOD2 does the latter, may be interesting to contrast behaviour of both while switching this)
- [ ] Command-line interface to distinguish maximal number of trials for rejection sampling between seed vertex and other vertices
    (even though in the publication the suggested values for these two things is equivalent)
- [ ] Refine calculations regarding step size / maximal angle / radius of curvature / *kmax* / length calculations
    - [ ] Maximal angle per step doesn't make much sense when the step size is 1/40 voxel size, so setting radius of curvature from the command-line would be better (even though I think we explicitly removed this from 0.2 to 3.0)
    - [ ] Radius calculation is based on fixed chord length, whereas PTT uses fixed arc length
    - [ ] #1507 introduced calculations of streamline length that take into account resampling, but it also tries to only perform explicit calculation of length in cases where, based on the number of points before downsampling, it is feasible that the streamline following downsampling may lie outside of the permissible length range. That logic may need to be revised.
    - [ ] *kmax* needs to appropriate reflect the reciprocal of radius of curvature
- [ ] Try double-precision floats to see if it better preserves orthonormality of the PTF
- [ ] Attempt execution improvement by removing calculation of unused PTF rows
- [ ] Reconsider geometry of probe (need to discuss privately with @baranaydogan)
- [ ] Adapt calibrator already in use for iFOD# (this could remove arbitrary parameters regarding calibration sample count and subsequent scaling factor)
- [ ] Implement tracking of rejection sampling max value being exceeded within `Shared`
- [ ] *(optional)* Implement cylindrical probe

Closes #2160.


baranaydogan:
Hi Rob,

It is great to see the progress on this! Thanks a lot for working on it. I have not checked the code yet, and I am not sure if you still have further plans to add/remove things. But I thought I could still give it a shot to check the current state. 

Below is a comparison I ran to check the optical radiation and whole brain tractogram (only by seeding the left wm). (Full script at the end.) Overall, visually, streamlines look very similar in my opinion. 

Some differences are expected in any case due to how MRtrix/Trekker handle images/masks for applying anatomical constraints. Other reasons that come to my mind are: 
(i) default parameters (if any)
(iI) sampling strategy
(iii) SH expansion (minor difference maybe)

Otherwise, this is already looking great in my opinion! 

Quick note: It would be very useful to add a "-downsample" option for this method too since the step size is very small, which generates very big files.

mrtrixPTT_1

```bash

# Optical radiation

trekker track -f \
${FOD} \
--seed L_LGN.nii.gz \
--oneSided \
--seed_count 1000 \
-p require_entry L_V1.nii.gz \
--output opt_trekker_ptt.tck

tckgen -algorithm PTT --force \
${FOD} \
-seed_image L_LGN.nii.gz \
-seed_unidirectional \
-select 1000 \
-include L_V1.nii.gz \
opt_mrtrix_ptt.tck
```


```bash

# Whole brain (only left wm seeded)

trekker track -f \
${FOD} \
--seed lh_white_filled.nii.gz \
--seed_count 10000 \
--output lh_white_trekker_ptt.tck

tckgen -algorithm PTT --force \
-seed_image lh_white_filled.nii.gz \
-select 10000 \
lh_white_mrtrix_ptt.tck
```

Lestropie:
Hi @baranaydogan,

Actually no progress recently; you got a notification of an update because I borked the MRtrix3 repository and have had to do a bunch of force-pushing to clean up after myself :sweat_smile:
I do want to try to churn through the long list of incomplete PRs so that we can push a 3.1.0 update at some point, but no idea what the timeline is going to look like.

I didn't get a chance to validate this implementation much beyond basic sanity checking. Even prior to tracking on empirical data I want to compare the outcomes on very simple noise-less phantoms, just a straight bundle and a curved bundle. Those might help isolate where the implementations differ, or indeed whether there's some fundamental issue with my implementation as it currently stands.

> It would be very useful to add a "`-downsample`" option for this method too since the step size is very small, which generates very big files.

The `-downsample` option is available to all algorithms. The only thing that differs is the default value. Indeed it looks like I [tried](https://github.com/MRtrix3/mrtrix3/blob/e4a16aeebc2b3cd76d0b265f0a1d98f0c371f840/src/dwi/tractography/algorithms/ptt.h#L89-L91) to add some downsampling by default; maybe given you raised the concern I got the math wrong and so it's not currently doing as intended.

Issue #2595, opened 2023-03-04
5ttgen bug with recent version fsl_sub and Debian/Ubuntu Gridengine


glasserm:
In fsl.py --> check_first it uses SGE_ROOT to check for the sge queuing system; however, not all sge queuing systems use SGE_ROOT (e.g., Gridengine on Debian/Ubuntu).  If both SGE_ROOT or FSLPARALLEL were checked for, this would work on more folks' systems.  I am not a python coder, but on my system simply replacing SGE_ROOT with FSLPARALLEL resolved the issue.  I don't know if SGE_ROOT is still used by others, however.

Lestropie:
Thanks for the report Matt. Testing `FSLPARALLEL` in `fsl.check_first()` might resolve non-completion in some use cases, though it would potentially result in hanging rather than erroring out in some cases where one or more FIRST jobs fail and SGE is *not* being used.

It's not clear exactly what issue has been encountered and how:

- `fsl_sub` itself checks for `SGE_ROOT` so I don't think that an SGE environment could be being used in the absence of such;
- `fsl_sub` could use `FSLPARALLEL` > 1 without using SGE, and when that happens it calls `wait`, which means that all subprocesses *should* have completed by the time `run_first_all` completes. Absence of expected VTK files in such a scenario should therefore ideally result in an error message, rather than waiting indefinitely for files that will never appear.

Is there a discussion that's happened elsewhere that you can link me to?

glasserm:
Debian/Ubuntu gridengine and FSL function without SGE_ROOT to launch jobs on SGE.  On my system, swapping in FSLPARALLEL worked to allow the MRTrix code to wait to complete the jobs.  It failed immediately otherwise.  I don't see any other environment variables that would work, though I know there is a lot of ongoing modification of fsl_sub in recent versions of FSL and it is certainly possible that FSLPARALLEL is a hold over from older versions of FSL (it is being set in my .bashrc and unsetting it does not seem to prevent jobs from going to SGE through fsl_sub either).  Perhaps talking to some of the FSL developers would suggest a better solution for wrapping scripts that call fsl_sub in recent FSL versions?  fsl_sub (and first) still returns a job ID so perhaps monitoring the completion of that would be a solution.

Lestropie:
[...]

PR #2597, opened 2023-03-07
fsl.check_first(): Utilise PID

Lestropie:
Attempt at resolving #2595.

If possible, it will take the PID reported on stdout and wait for its completion.
Looking at the `run_first_all` code, I think that `fsl_sub` is used in such a way that each spawned processes inherits the previously executed processes as its children, as it's the PID of the last executed job that is sent to stdout.

This does as intended on my own system, in that the PID reported by `run_first_all` no longer exists and so the function can proceed as normal. However it really requires testing on a system with SGE configured. @glasserm is this something you're in a position to do easily?

Also, I've left the `path.wait_for()` call in place where SGE is easily detected just as a failsafe; I don't expect that code to actually be used anymore since it will only execute if the PID can't be queried.

### DIFF
--- a/bin/labelsgmfix
+++ b/bin/labelsgmfix
@@ -105,8 +105,8 @@ def execute(): #pylint: disable=unused-variable
   first_input_is_brain_extracted = ''
   if app.ARGS.premasked:
     first_input_is_brain_extracted = ' -b'
-  run.command(first_cmd + ' -m none -s ' + ','.join(structure_map.keys()) + ' -i T1.nii' + first_input_is_brain_extracted + ' -o first')
-  fsl.check_first('first', structure_map.keys())
+  first_stdout = run.command(first_cmd + ' -m none -s ' + ','.join(structure_map.keys()) + ' -i T1.nii' + first_input_is_brain_extracted + ' -o first').stdout
+  fsl.check_first('first', structures=structure_map.keys(), first_stdout=first_stdout)
 
   # Generate an empty image that will be used to construct the new SGM nodes
   run.command('mrcalc parc.mif 0 -min sgm.mif')
--- a/lib/mrtrix3/_5ttgen/fsl.py
+++ b/lib/mrtrix3/_5ttgen/fsl.py
@@ -188,8 +188,8 @@ def execute(): #pylint: disable=unused-variable
   first_verbosity_option = ''
   if app.VERBOSITY == 3:
     first_verbosity_option = ' -v'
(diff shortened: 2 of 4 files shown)


Lestropie:
Note that this solution uses the `psutil` module, which is seemingly not a "default" and needs to be explicitly installed in some environments. This would need to be addressed not only for CI, but also in package generation, containers, and installation-from-source documentation. It may not be an absolute *requisite* (I can have the code skip those steps if the module is not available), but we should try to propagate the dependency nevertheless.

glasserm:
Happy to help if I can easily.  Currently I have this as a conda install, but I am pretty illiterate with conda and python.  What is the quickest way to get my current setup in a state to test this patch?

Lestropie:
Because it's not part of a tagged release, you'd need to clone the feature branch called `check_first` and then [build from source](https://mrtrix.readthedocs.io/en/latest/installation/build_from_source.html).

glasserm:
I'll see if I can get to this this weekend...

glasserm:
Sorry for the delay.  I realized you just changed python files so I applied the patch manually.  Unfortunately it doesn't work.  SGE does not spit out a process ID, rather it spits out a job ID.  You could check to see if it is still in the queue list like this qstat | grep <job ID>.  If it returns the job ID, it hasn't finished running, if it returns nothing, it has.

Lestropie:
Ah OK, it has to be queried much like a SLURM job.
I have no experience with SGE and no ability to test, so I'm taking shots a little blindly here.

Unfortunately there doesn't seem to be any established / mature Python package for querying SGE jobs. So I would indeed be forced to `subprocess` regular `qstat` calls. I don't think that just querying the presence of the `qstat` command in `PATH` as a proxy for potential use of SGE works.

Alternatively I could change strategy entirely, and directly access & query the various FIRST log files, looking for any errors. I didn't want to go this way as it could theoretically change easily between FSL versions, and would be entirely inapplicable to any other context (as opposed to `path.wait_for()`). But it might turn out to be simpler...

glasserm:
Could you just fsl_sub -j <JobID> something and then check for it to finish?  Perhaps that would be agnostic to SGE vs SLURM (with fsl_sub handling the queuing system).  Your fsl_sub -j <JobID> command would not run until all of the FIRST jobs had completed (or failed) telling you when it is time to check the outputs.

Lestropie:
Closed in favour of #2609.

PR #2604, opened 2023-03-15
dwibiasnormmask and mtnorm algorithms (take 2)

Lestropie:
This is a cleanup of #2552, since the history had become a little corrupted by code from other PRs / version and copyright date changes etc.. Also resolves better against #2583.

@Arshiasan: While my name now appears alongside your own on some of these commits, that's only because I generated the commits; the contribution statistics are still attributed exclusively to yourself as author.

Outstanding tasks carried over from #2552:

- [x] Add tests for:
    - [x] `dwi2mask mtnorm`
    - [x] `dwibiascorrect mtnorm`
    - [x] `dwinormalise mtnorm`
- [x] Discuss changing name of `dwinormalise individual`
    Personally I like `dwinormalise manual` since it acts in a similar way to `dwi2response manual`
- [x] Append new dedicated docs page on DWI masking


Lestropie:
Review open to anyone willing to have a scroll through and look for anything dumb.
Otherwise we'll just let it break in production. :upside_down_face: 
Reproducibility tests are in place and passing.

Lestropie:
Have additionally force-pushed `script_test_data` such that the tip pointed to here is now named `dev` on that repo.

PR #2607, opened 2023-03-21
path.wait_for(): Fix symbol aliasing

Lestropie:
Discovered while toying with #2597, but isn't a solution to such.

I'm sure that I had tested this code path when first written, perhaps the aliasing got introduced at a later date with variable name changes? Will try to re-test before merging.


Lestropie:
Yep, problem introduced in fc8f3b4 as part of #1449.

Lestropie:
Closed in favour of #2608.

PR #2608, opened 2023-03-21
path.wait_for(): Fix symbol aliasing

Lestropie:
Replacement for #2607 as problem exists on `master`.

Will still try to test this before triggering a review request.


Lestropie:
Seems to do as claimed: added a dummy `path.wait_for()` call for a file that doesn't exist, then after some time created that file from another terminal.

bjeurissen (dismissed):

Issue #2612, opened 2023-03-27
Address Sanitiser


Lestropie:
[ASan](https://github.com/google/sanitizers/wiki/AddressSanitizer) is sold as the modern preferable alternative to `valgrind` for detection of various memory errors.

While it claims to be substantially faster than `valgrind`, we will nevertheless continue to use release mode with assertions enabled for CI builds. This is purely a capability for developers to use themselves.

Either:

1.  Whenever a debug build is requested in `configure`, automatically add the relevant compile / link flags.
2.  Add a command-line option to `configure` that will add the requisite compile / link flags.

The latter may be preferable given that it can require further modification of compile flags to get it to work (I myself had to add "`-gdwarf-4`" to `CPP_FLAGS`).

While this can be achieved with the environment variables that `configure` reads from, it might nevertheless be preferable to embed the capability in order to encourage its use.

PR #2613, opened 2023-03-27
App::get_matches(): Fix memleak

Lestropie:
Discovered this while playing with ASan (#2612) to diagnose a fault with my own code. Seems that `std::string::compare()` is buffer-overrunning the C-style string. And because it's a C-style string, we can't query the size. So this code goes one character at a time. But it would be preferable for all of the back-end, including command-line parsing, to be using `std::string` and safe pointers (#2111).

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -128,7 +128,14 @@ namespace MR
       inline void get_matches (vector<const Option*>& candidates, const OptionGroup& group, const std::string& stub)
       {
         for (size_t i = 0; i < group.size(); ++i) {
-          if (stub.compare (0, stub.size(), group[i].id, stub.size()) == 0)
+          bool match = true;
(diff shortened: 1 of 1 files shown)


Lestropie:
@jdtournier Can I get an OK on this?

jdtournier:
Yes, that should be OK as a stop-gap, until we do things properly as you've already suggested in #2111...

That said, a cleaner option might be just to cast the `const char*` `group[i].id` variable to a `std::string`, like this:
```
        for (size_t i = 0; i < group.size(); ++i) {
          if (stub.compare (0, stub.size(), std::string (group[i].id)) == 0)
            candidates.push_back (&group[i]);
        }
```
I reckon that would also work as expected, with minimal change to the code?

Lestropie:
Yep, that's probably better; closing in favor of #2629.

Issue #2477, opened 2022-05-23
mrconvert: fslgrad and RealignTransform: 0


bjeurissen:
I noticed that running ``mrconvert data.nii.gz -fslgrad bvecs bvals dwi.mif`` with ``RealignTransform: 0`` in the config file can lead to a mif file with incorrectly oriented gradient directions (e.g. on the HCP data).

Not sure how big a deal this is, but at least it was unexpected.

Lestropie:
This will be the use case where my hypothetical in #2416 actually causes a demonstrable problem.

diagiraldo:
Something similar is happening vice versa, when I have `RealignTransform: 0` in `.mrtrix.conf` and convert from mif to nifti with `-strides -1,2,3,4` the resulting gradients are not correct, here the example:
```
mrconvert DWI.mif -strides -1,2,3,4 -export_grad_fsl bvecs.txt bvals.txt DWI.nii.gz 
dwiextract DWI.nii.gz -singleshell - -fslgrad bvecs.txt bvals.txt | amp2sh - sh_DWI.mif -force 
```
![screenshot0000](https://user-images.githubusercontent.com/6461378/196713958-3c5a943f-c855-4db1-8a9e-cb77e216f74a.png)

I should say it doesn't happen with `-strides 1,2,3,4`:
```
mrconvert DWI.mif -strides 1,2,3,4 -export_grad_fsl bvecs.txt bvals.txt DWI.nii.gz 
dwiextract DWI.nii.gz -singleshell - -fslgrad bvecs.txt bvals.txt | amp2sh - sh_DWI.mif -force 
```
![screenshot0001](https://user-images.githubusercontent.com/6461378/196715830-f1f1899c-66d6-45f7-a395-be913fc8e660.png)

PR #2617, opened 2023-04-06
Changes to enable meshconvert to warp mesh files.

Arshiyasan:
As suggested by @Lestropie, this enhancement enables users to specify a new mode for `meshconvert` through setting `-transform warp <warp image>` to apply a deformation field to mesh files. The approach is similar to `tcktransform`. It works well on my data!

### DIFF
--- a/cmd/meshconvert.cpp
+++ b/cmd/meshconvert.cpp
@@ -16,6 +16,7 @@
 
 #include "command.h"
 #include "header.h"
+#include "image.h"
 #include "surface/mesh.h"
 #include "surface/mesh_multi.h"
 #include "surface/filter/vertex_transform.h"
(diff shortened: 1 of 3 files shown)


cmd/meshconvert.cpp
```diff
@@ -81,6 +85,10 @@ void run ()
       case 2: transform->set_voxel2real(); break;
       case 3: transform->set_real2voxel(); break;
       case 4: transform->set_fs2real   (); break;
+      case 5:
+        transform->set_warp();
+        transform->set_warp_image(Image<float>::open(opt[0][1]));
```
Lestropie, line 90:
```suggestion
        transform->set_warp_image(H.get_image<float>());
```

Also `transform->set_warp()` shouldn't be necessary; it should be possible to set that internal flag when `set_warp_image()` is called.

Lestropie:
-   I think I would prefer to have an explicit `meshtransform` command.
    Once one starts providing non-linear deformation fields, it's no longer just "a conversion between mesh formats".
    So just as for `mrtransform` and `tcktransform`, `meshtransform` could use linear or non-linear transforms.
    `meshconvert` would then likely be exceptionally simple, since the linear transformation options would be moved to `meshtransform`, but it should nevertheless remain there for discoverability; and any more advanced controls regarding export (eg. switching between VTK binary and ASCII) would belong there.

-   Would be nice to have `meshtransform` accept a "full" warp file, and control the target and destination spaces of the transformation. However this is not strictly requisite for the functionality to be added.

-   Wherever more than a sentence is required to explain the operation of a command-line option, I prefer to move the primary description out of the option's help string and into the command description / example usages. Eg. Particularly if an option to support full warps is not implemented, an example usage where the output of `warpconvert` is piped to `meshtransform` could be specified and described verbosely.

-   Currently, if a vertex lies outside of the warp image FoV, an uninitialised vertex is pushed. This will corrupt any transformed mesh. Given that we are exclusively dealing with closed surfaces, I think the only solution is to yield an error if no valid warp information can be accessed for just one vertex.

-   The vertex sanity check currently only checks the warp image FoV. There are however other instances where no valid warp information is present: NaN-filling (easy to detect), and zero-filling (harder to detect, but there should be a wrapper class somewhere that detects such). 

-   Not strictly related to the proposed change, but it comes to mind looking at the code:
    Here function `File::NIfTI::adjust_transform()` is being used in the "`fs2real`" function. However as occurs elsewhere (most recently #2477), this may instead actually be dependent on the transform realignment that occurs during image load, not what would hypothetically occur on NIfTI write.

Issue #2468, opened 2022-04-21
ENH fod2fixel: Output segmented FODs


fionaEyoung:
When I want to run an analysis on individual fibre populations of an FOD image represented as SH distributions, I currently go through a very longwinded sequence: `fod2fixel` -> `fixel2peaks` -> separate peaks (e.g. `mrconvert afd_scaled_peaks.mif -coord 3 0:2 f1.mif`) -> `peaks2fixel` -> `fixel2sh`
Additionally, the resulting "fixel ODFs" are essentially delta functions (scaled by AFD or peak amp or whatever) at the fixel peak directions, without any of the detail (e.g. dispersion) present in the original FOD lobe (like in the white boxed voxels in the images below).

Left: Original WM FOD, Right: First fixel converted to SH
<img width="200" alt="Screenshot 2022-04-21 at 18 17 03" src="https://user-images.githubusercontent.com/33662704/164515559-886476e0-e158-42fd-98aa-d1bf23d3b8d4.png">  <img width="200" alt="Screenshot 2022-04-21 at 18 17 46" src="https://user-images.githubusercontent.com/33662704/164515555-a87a0a4a-d39e-428a-884b-c7ce526e7b54.png"> 

It would be super useful to be able to directly acces the results of the FMLS FOD segmenter, before only the peak directions, AFD etc are retained and written out to fixel.

I've taken a look at the code to see if I have any hope of implementing this myself. Where I get lost is in [`fmls.cpp`](https://github.com/MRtrix3/mrtrix3/blob/master/src/dwi/fmls.cpp) - I can't tell if the amplitude values of each direction sample  are retained anywhere, which could be used to (approximately) reconstruct the individual lobes with something like `amp2sh`.

I appreciate the wishlist is very long 😉 I'd be game to try this myself but I'll probably need some *pointers, as my C++ knowledge is not advanced enough to handle the amount of abstraction in MRtrix's code.

Lestropie:
Hi Fiona,

I would need to know exactly what you're trying to do with the output of that sequence of steps to know whether that particular sequence makes sense or whether there is some alternative. But my intentions for development in this context separately to your proposal may help to carve a path forward.

[...]

PR #2621, opened 2023-04-25
ENH fod2fixel: Add option to output lookup table from FMLS process

fionaEyoung:
An implementation of the suggestions discussed in #2468. 

This catches the [`FOD_lobes` lookup table](https://github.com/MRtrix3/mrtrix3/blob/3.0.3/src/dwi/fmls.h#L167) from the FMLS process and adds the option to output this data as an XxYxZx1281 image with directions embedded in the header.

It assumes that the index values used in the lookup table correspond to the order of the corresponding `FOD_lobe`s within the `FOD_lobes` vector. (i.e. that the `i`th direction is indeed assigned to the fixel at `offset + vox_fixels.lut[i]`; I don't see how this could _not_ be the case but thought it worth mentioning anyway). 

Feedback welcome! Not sure the cartesian to spherical conversion is particularly optimal.

I've also added a change to `sh2amp` to allow the use of the `lut` output as a directions file (i.e. additionally looking for  a "directions" header entry if no valid DW scheme entry is found). This change is more specific to my use case and also feels a bit hacky, so I'm happy to re-push without bec52de if that would be preferable.

- [x] `./check_syntax` 
- [ ] `./docs/generate_user_docs.sh`: had issues building due to version conflict (the 3.0.03->3.0.4 change doesn't seem to be reflected in my local tags, but changing (and committing) `version.h` back to 3.0.3 didn't seem like the right option either?).

### DIFF
--- a/cmd/fod2fixel.cpp
+++ b/cmd/fod2fixel.cpp
@@ -23,6 +23,7 @@
 #include "fixel/helpers.h"
 
 #include "math/SH.h"
+#include "math/sphere.h"
 
 #include "thread_queue.h"
 
@@ -258,9 +259,13 @@ void Segmented_FOD_receiver::commit ()
     lut_header.datatype().set_byte_order_native();
     // Add directions set to lut image header.
     std::stringstream dir_stream;
-    for (ssize_t d = 0; d < directions.size() - 1; ++d)
-      dir_stream << directions[d](0) << "," << directions[d](1) << "," << directions[d](2) << "\n";
-    dir_stream << directions[directions.size() - 1](0) << "," << directions[directions.size() - 1](1) << "," << directions[directions.size() - 1](2);
+    Eigen::Vector3d az_el_r;
+    for (ssize_t d = 0; d < directions.size() - 1; ++d) {
+      Math::Sphere::cartesian2spherical(directions[d], az_el_r);
+      dir_stream << az_el_r(0) << "," << az_el_r(1) << "\n";
+    }
+    Math::Sphere::cartesian2spherical(directions[directions.size() - 1], az_el_r);
+    dir_stream << az_el_r(0) << "," << az_el_r(1);
     lut_header.keyval()["directions"] = dir_stream.str();
     lut_image = make_unique<LutImage> (LutImage::create (Path::join(fixel_directory_path, lut_path), lut_header));
   }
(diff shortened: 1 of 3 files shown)


Lestropie:
🤦‍♂️ 

I did literally this fairly recently, since it's requisite for a local project:
https://github.com/MRtrix3/mrtrix3/tree/fod2fixel_lookup
And indeed was intending to soon implement something more immediate for #2468...
Sorry, I should have posted something earlier.
I will have to cross-check my commit with your proposals here and figure out how to resolve.

> It assumes that the index values used in the lookup table correspond to the order of the corresponding FOD_lobes within the FOD_lobes vector.

Mine does the same; but need to ensure that any dixels that are not assigned to any *retained* fixel are properly identified as such. From memory I had to do a bit of code refactoring to ensure that this was the case. For instance `fod2fixel` could restrict the number of fixels per voxel, but any fixels discarded at that stage would still be present in the LUT.

> Not sure the cartesian to spherical conversion is particularly optimal.

Only had a quick look, but it should be possible to do using a function that operates on the full matrix and then Eigen string formatting.

> I've also added a change to sh2amp to allow the use of the lut output as a directions file (i.e. additionally looking for a "directions" header entry if no valid DW scheme entry is found).

Makes sense. There's a few functionalities in the software where there are various potential sources of an antipodally symmetric direction set. Ideally all such instances should be checking for all such sources. Indeed it should potentially be integrated into a back-end function, with eg. consistent hierarchy of what is selected first, issuing a warning if there are multiple such sources that are not identical.

fionaEyoung:
👀 awkward!

> I will have to cross-check my commit with your proposals here and figure out how to resolve.

As far as I can tell, the only functional difference in this implementation compared to 6e1d231 is the conversion to spherical directions and the `-lut` output as an option (rather than on by default). 
Other than that, its the same behaviour but less efficient and without the proper checks, so from my point of view, this doesn't need merging, unless anyone else needs the same functionality before the more wider changes are ready to be merged.

Lestropie:
What I might end up doing is rebasing 6e1d231155309f3bc66c2dc8f10971a1b2cb24d7 onto your changes. That way your contributions still get integrated into the history, and I get to keep both the requisite code re-arrangement and the various other tidbits that got integrated into that commit while I was doing it. But I'll do a visual side-by-side when I get back to that code.

I'm intending to not only make the LUT export by default, but also make its path a reserved filename in the fixel directory format, so that when present, downstream apps will detect its presence and make use of it without user intervention. Eg. When assigning streamlines to fixels it's better to have that information than to have just the mean direction per fixel, and it would be nice for that to happen automatically.

fionaEyoung:
Replaced by #2625

PR #2622, opened 2023-04-26
link to FFTW3 on all platforms

bjeurissen:
This fixes FFTW3 linking issues on macOS

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -6,14 +6,12 @@ add_library(mrtrix::core ALIAS mrtrix-core)
 find_package(Eigen3 REQUIRED)
 find_package(ZLIB REQUIRED)
 
-if(WIN32)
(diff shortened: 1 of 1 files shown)


daljit46:
I have now added a CMake module (taken from Eigen) that should automatically find fftw and link against it.

bjeurissen:
Thanks!

It really sucks that this is necessary and that you can't do a simple ``find_package(fftw3 REQUIRED)``. 

One should be able to do this, but apparently cmake support for fftw3 has been broken for 5 years now.... (see https://github.com/FFTW/fftw3/issues/130) and distributions are not going to fix it (see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1729652).

As a result of the fatal error this produces, one can not even use the pkg-config fallback that find_package normally uses, and one has to resort to a custom module... All to add a few compiler and linker flags ...

PR #2624, opened 2023-05-02
Add compiler caching support for CI checks

daljit46:
This pull request add supports for compiler caching in the CI checks workflow. It uses [sccache](https://github.com/mozilla/sccache) which is a compiler cache that speeds up recompilation by caching previous compilations by automatically detecting whether the same compilation is already cached. This allows us to leverage the free cache (up to 10GB) provided by Github Actions. At the moment, it only works with Clang and GCC builds on Linux.

In our current workflow, all the checks are done by compiling from scratch and for example, a typical CI build with Clang on Linux takes about 35-40 minutes. My preliminary tests show that sccache can complete a cached build in about 5-10 minutes. 

I've decided to use sccache instead of the more popular [ccache](https://ccache.dev/) because in my experience it has been more reliable on Windows (particularly with MSVC) and because they provide a pre-made [Github Action](https://github.com/marketplace/actions/sccache-action) to facilitate integration. However, this is not necessary and we could use ccache if wanted (integrating it should be [fairly straightforward](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/)).

Any feedback welcome!

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -18,6 +18,8 @@ jobs:
     env:
       CFLAGS: -Werror
       QT_SELECT: qt5
+      SCCACHE_GHA_ENABLED: "true"
+      SCCACHE_CACHE_SIZE: "2G"
 
     steps:
     - uses: actions/checkout@v1
@@ -27,8 +29,14 @@ jobs:
          sudo apt-get update
          sudo apt-get install clang libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev 
 
+    - name: Run sccache-cache
+      uses: mozilla-actions/sccache-action@v0.0.3
+    
+    - name: Run sccache-stats
+      run: sccache --show-stats
+        
     - name: configure
-      run: ./configure -assert || { cat configure.log; false; }
+      run: COMPILER_PREFIX=sccache ./configure -assert || { cat configure.log; false; }
 
     - name: build
       run: ./build -nowarnings -persistent -nopaginate || { cat build.log; false; }
@@ -52,6 +60,8 @@ jobs:
       CXX: g++-9
       CFLAGS: -Werror
       QT_SELECT: qt5
+      SCCACHE_GHA_ENABLED: "true"
+      SCCACHE_CACHE_SIZE: "2G"
 
     steps:
     - uses: actions/checkout@v1
@@ -61,12 +71,17 @@ jobs:
          sudo apt-get update
          sudo apt-get install g++-9 libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev 
 
+    - name: Run sccache-cache
+      uses: mozilla-actions/sccache-action@v0.0.3
+    
+    - name: Run sccache-stats
+      run: sccache --show-stats
+      
     - name: configure
-      run: ./configure -nooptim || { cat configure.log; false; }
+      run: COMPILER_PREFIX=sccache ./configure -nooptim || { cat configure.log; false; }
 
     - name: build
       run: ./build -nowarnings -persistent -nopaginate || { cat build.log; false; }
-
 
 
 
--- a/configure
+++ b/configure
@@ -88,6 +88,9 @@ ENVIRONMENT VARIABLES
     Multiple environment variables can be set this way as needed.
     The following environment variables are available:
 
+    COMPILER_PREFIX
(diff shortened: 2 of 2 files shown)


daljit46:
Ok so I think most things are working now for all platforms. The improvements in build times are significant and should considerably improve our build and test cycle. 
The only problem that remains now is that on MSYS2, the meshconvert test is failing for some reason and also gcc 12/13 triggers warning during compilation. Our current CI workflow was all tested against gcc 9 and thus these warnings were suppressed. I think we should fix all the warnings before we merge this in.

jdtournier:
I can replicate on my Win10 laptop - though I also get failures for `mrhistmatch` & `tckconvert`. Go figure...
I'll investigate when I have a minute.

jdtournier:
OK, `mrhistmatch` & `tckconvert` failures were simply due to the absence of `diff` & `bc` on my system... Easy fix. 

For `meshconvert`, the issue is due to opening the file in text mode, which only differs from binary mode on Windows systems. Because these VTK files can be written as text or as binary, I'm not sure how to handle this. If we change to always opening these files in binary mode, the other tests fail (for the VTK text files). But since the binary files start as text and switch to binary halfway through, it's going to require switching modes half-ay through or something... I'll have a go at that in a minute.

daljit46:
> For `meshconvert`, the issue is due to opening the file in text mode, which only differs from binary mode on Windows systems.

But why didn't the check fail on the previous Windows CI workflow?

jdtournier:
Ok, having looked into this, I'm starting to think this might be a compiler bug... The offset reported by input stream's `tellg()` when the file is opened in non-binary mode is completely off, which is what is causing the issue when the file re-opened in binary mode. I'll investigate further when I get the chance, but it might be a while...

jdtournier:
Yet more context on this issue... Here's another test cpp file that demonstrates the issue. This time, it writes a few lines of text (with a file open in binary mode), then dumps binary data, with single bytes incrementing in value and wrapping around for a long as requested. 

---

**test_tellg.cpp**

```
#include <fstream>
#include <iostream>
#include <string>

int main ()
{
  using namespace std;

  {
    ofstream out ("testfile.txt", ios_base::binary);
    out << "first line of text\nsecond line of text\nthird line of text\nsome binary after this line\n";

    vector<char> data (1000);
    char n = 0;
    for (auto& c : data) {
      //if (n==26) n++;    <-- uncomment and it works as expected
      c = n++;
    }
    out.write (data.data(), data.size());
  }


  ifstream in ("testfile.txt", ios_base::in);
  string line;
  cerr << "offset = " << in.tellg() << "\n";

  for (int n = 1; n <= 5; ++n) {
    getline (in, line);
    if (!in) break;
    cerr << "line " << n << ": \"" << line << "\" (size = " << line.size() << "), current offset = " << in.tellg() << "\n";
  }
}
```
---

It turns out that things behave as expected as long as the file does not contain the ASCII 26 (substitute) character. Output from code above:
```
offset = 0
line 1: "first line of text" (size = 18), current offset = 993
line 2: "ïîìÄÅÉæÆôöòûùÿÖÜø£Ø×ƒáíóúñÑªº¿®¬½¼¡«»░▒▓│┤ÁÂÀ©╣║╗╝¢¥┐└┴┬├─┼ãÃ╚╔╩╦╠═╬¤ðÐÊËÈıÍÎÏ┘┌█▄¦Ì▀ÓßÔÒõÕµþ" (size = 93), current offset = -1
```

and output of same command with the commented line enabled (i.e. skipping the character with ASCII code 26):
```
offset = 0
line 1: "first line of text" (size = 18), current offset = 19
line 2: "second line of text" (size = 19), current offset = 39
line 3: "third line of text" (size = 18), current offset = 58
line 4: "some binary after this line" (size = 27), current offset = 86
line 5: "☺☻♥♦♣♠ " (size = 10), current offset = 97
```

---

[According to Wikipedia](https://en.wikipedia.org/wiki/Substitute_character), this ASCII 26 character used to mark the end-of-file, so maybe that has something to do with it. But regardless, the problematic character is encountered much later in the file, and yet it affects the file offsets reported way earlier. Feels like there's a glitch somewhere in MSYS2...

jdtournier:
OK, one solution that might work is to always open the file in binary mode, and strip any trailing carriage return `'\r'` characters from the end of each line after every `getline()` call. Not pretty, but it seems to work on that toy example above...

Lestropie:
Can we move the MSYS2 `meshconvert` discussion to a separate issue? It's happening on `master` and is unrelated to the contents of this PR.

jdtournier (approved):

PR #2625, opened 2023-05-04
fod2fixel: Export lookup table

Lestropie:
This PR resolves the conflict between the developments I had made in 6e1d231155309f3bc66c2dc8f10971a1b2cb24d7 (but not broadcast) and the contributions from @fionaEyoung in #2621 in achieving the desired functionality in #2468.

I have explicitly merged the two in 435c6aacbad02e92f58d4dda2e7362b587bdd2cb so that both appear in the history, though in its final form the state of the code will more closely conform to my own changes originally in 6e1d231155309f3bc66c2dc8f10971a1b2cb24d7 since:
-   It performed a requisite move of enforcing maximal number of fixels per voxel from the export step to the FMLS segmenter itself (since the lookup needs to reflect such)
-   It embeds the lookup information as a key part of the fixel directory format, rather than being just a 4D image that might appear sometimes, since downstream applications will need to know where to look for it and how to use it.

While I have subsequent changes I wish to make based on this, it might be reasonable to merge this into `dev` rather than waiting for all of those subsequent changes to get stacked up.

### DIFF
--- a/cmd/dwi2adc.cpp
+++ b/cmd/dwi2adc.cpp
@@ -47,13 +47,18 @@ using value_type = float;
 
 
 
-class DWI2ADC { 
+class DWI2ADC {
   public:
     DWI2ADC (const Eigen::MatrixXd& binv, size_t dwi_axis) :
       dwi (binv.cols()),
       adc (2),
       binv (binv),
       dwi_axis (dwi_axis) { }
+    DWI2ADC (const DWI2ADC& that) :
+      dwi (that.dwi.size()),
+      adc (that.adc.size()),
+      binv (that.binv),
+      dwi_axis (that.dwi_axis) {}
 
     template <class DWIType, class ADCType>
       void operator() (DWIType& dwi_image, ADCType& adc_image) {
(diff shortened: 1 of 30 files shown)


Lestropie:
Might be further effort / cleanup required depending on whether the encoding of header key-value "`directions`" as cartesian vs spherical causes problems across the multiple commands where it may be written / read.

Lestropie:
New binary tests, including modification of existing data and addition of new data, needs to be done here (I'll merge #2626 to here and then deal with tests, since there are conflicting required changes across both PRs).

-   Some fixel quantitative values are fractionally different due to the basis direction set being replaced with the double-precision equivalent (differences are imperceptible in `mrview` but result in test failure).
-   `fod2fixel` has additional outputs that need to be stored in the test data for `testing_diff_fixel` to succeed.

Lestropie:
1.  If your destination fixel directory results in images being stored in RAM for delayed write-back, then this functionality will chew up your RAM reserves very quickly.
2.  I only verified on a small test dataset; I'll try on something bigger and see if I can reproduce.
3.  Might be some expected delay for the OS to synchronize with the filesystem before the command can close. I'll throw some timing in there and see if I can find anything major.

Lestropie:
2 should be fixed.
On a full image dataset adding the `-dixel_images` option for resulted in ~ 25% longer execution time, but no unexpected delays.

PR #2626, opened 2023-05-04
Multiple changes relating to direction set handling

Lestropie:
On contemplating the issue [commented](https://github.com/MRtrix3/mrtrix3/pull/2625#issuecomment-1533978145) in #2625, I decided to have a look at how to better standardise the handling of basis direction sets across the code base. On looking across a few different commands it was clear that this should have been a unified functionality some time ago. Eg. Some code necessitates spherical coordinates as input, elsewhere cartesian coordinates are necessary; some places can read from `directions` from an image header, others can't. Even writing matrix data to an image header; there was custom code scattered in multiple locations.

Given I did it in the context of #2625, that's the base branch. 

Some description in a864b82fc3f59589f98d08a04c1cead38db7bf7c, but I will maintain a list below of stuff that still needs to be looked at, and further external comments very welcome.

- [ ] Code placement. Much of the functionality is not actually DWI-specific, and arguably could go into `MR::Math::Sphere`.
    - [x] Where this gets slightly tricky is the fact that in some contexts, a basis direction set could be extracted from the DW gradient scheme (mostly still with a single-shell requirement, but one or two commands can operate on multi-shell dixel data).
    - [x] This would include contemplation of `DWI::Directions::Set` (and by implication `FastLookupSet`). One can have and use a basis direction set without invoking that class, and the functionalities of that class are not DWI-specific.
        **Edit**: `Set` is currently serving two distinct purposes. First is construction of the convex hull and resulting determination of dixel adjacency. Second is construction of static information that relates to the use of binary data to represent dixel masks, to facilitate byte-wise operations. The latter would ideally be not only separated, but removed and the relevant functionality deferred to `vector<bool>`.
- [x] As per code comment, will eventually add `-directions` command-line option to `fod2fixel` so that different basis sets can be used for segmentation if desired. But I need to change the construction of what is currently called `DWI::Directions::Set` to facilitate that.
- [x] I added a paragraph that can be added to the command help page Discussion section that explains the different content that can be provided to that option. But I think its inclusion would be overkill in some commands where the use of that direction set is kind of "secondary"; eg. non-negativity constraints as opposed to something like `sh2amp`. Might require some contemplation or revision.
    **Edit**: If retaining current strategy of adding the Description paragraph to only commands where the functionality is primary, need to revise the function that generates the option so that it does not reference the Description section if it is not present.
- [ ] While this passes existing binaries testing, I likely need to add a few more tests to look for anything stupid happening.
- [x] I've added some basic data checks, but might want to make sure that code is being executed wherever relevant.
- [x] Might want to change a couple of function names from American to British spelling.
- [x] Consider terminology, particularly whether "dixel" or "basis dixels" or "dixel set" should be used rather than just "directions".
- [ ] Change implementation of fast dixel assignment class to base its lookup on cartesian rather than spherical coordinates in order to avoid expensive trigonometric functions.
    **Edit**: Deferred to #2635.
- [x] Move `MR::DWI::gen_direction_matrix()`: This is just a cartesian -> spherical transform and a slicing operation, so doesn't need to be in the DWI namespace.
    **Edit**: Same for `condition_number_for_lmax()`.


Lestropie:
@jdtournier Any thoughts on having a new directory `core/math/sphere/`, which would contain:
-   `core/math/sphere.h`
-   `core/math/SH[.cpp|.h]`
-   `core/math/ZSH.h`
-   Contents of `src/dwi/directions/` currently on this branch as a sub-directory, maybe something like `core/math/sphere/set/`?

Or any other alternative variation?

With the exception of loading sets from `dw_scheme`, and needing to be aware of contexts in which multi-shell schemes are vs. are not permissible, the rest of this is independent of DWI...

jdtournier:
Sure, sounds good to me, go for it!

Lestropie:
In retrospect, alternative would be to leave `core/math/SH.*` / `core/math/sphere.h` / `core/math/ZSH.h` untouched, but move what's currently `src/dwi/directions/` into `core/dixel/` to live alongside `core/fixel/`?

Lestropie:
After brief discussion `core/dixel/` may not be the best placement given that "dixel" typically refers to an observation or quantitative value that is ascribed to a specific direction (one of a dense basis set) *within a specific voxel*.

Conversely, regarding original suggestion of creating directory / namespace `core/math/sphere/`, there are elements within the set of functionalities here for which it is arguable whether they warrant inclusion in the `MR::Math` namespace. For instance the class that assigns an arbitrary direction on the unit sphere to the nearest of a set of basis directions on the unit sphere. Or arguably the definition of the predefined basis sets.

Nevertheless I'm not seeing a third alternative yet so I might implement the first approach and see what it looks like.

Lestropie:
@jdtournier see what you think of [aece364](https://github.com/MRtrix3/mrtrix3/pull/2626/commits/aece36443106801970352aa3974c61ab3e9b0bd1). It does mean moving `MR::Math::SH` to `MR::Math::Sphere::SH`, which is fairly widely used and so appears all over the place. And it's not great to be moving such a core API piece without a corresponding major version update, though I think we'd agreed that given the software is not primarily used as an API for external apps we're not too concerned about it. But otherwise it seems a reasonably sensible grouping to me.

But is "set on the sphere" sufficiently accurate about what's being handled here, or is there a better alternative?

fionaEyoung:
The increase in precision from [d8f0a9f](https://github.com/MRtrix3/mrtrix3/pull/2626/commits/d8f0a9f6a8896d296b3372b14a4bfd40f51072bc) seems to blow up the time (10-30min) spent on [`Assigner::initialise()`](https://github.com/MRtrix3/mrtrix3/blob/1b320ca24b30931fe3fd40c28cea434de995b7b7/core/math/sphere/set/assigner.cpp#L63), at least on my end. If I understand correctly this is separate to the speedup mentioned in #2635 which is about the assignment of dixels using lookup grid, not the creation of the grid itself?

PR #2629, opened 2023-05-10
App::get_matches(): Fix buffer overrun

Lestropie:
Supersedes #2613.

### DIFF
--- a/core/app.cpp
+++ b/core/app.cpp
@@ -128,7 +128,7 @@ namespace MR
       inline void get_matches (vector<const Option*>& candidates, const OptionGroup& group, const std::string& stub)
       {
         for (size_t i = 0; i < group.size(); ++i) {
-          if (stub.compare (0, stub.size(), group[i].id, stub.size()) == 0)
+          if (stub.compare (0, stub.size(), std::string(group[i].id), 0, stub.size()) == 0)
             candidates.push_back (&group[i]);
         }
       }


jdtournier (dismissed):

core/app.cpp
```diff
@@ -128,7 +128,7 @@ namespace MR
       inline void get_matches (vector<const Option*>& candidates, const OptionGroup& group, const std::string& stub)
       {
         for (size_t i = 0; i < group.size(); ++i) {
-          if (stub.compare (0, stub.size(), group[i].id, stub.size()) == 0)
+          if (stub.compare (0, stub.size(), std::string(group[i].id), 0, stub.size()) == 0)
             candidates.push_back (&group[i]);
```
jdtournier, line 132: Was expecting this to be simply:
```
if (stub.compare (0, stub.size(), std::string(group[i].id)) == 0)
```
I don't think it would actually change the outcome, it just seems unnecessary to specify the size the option name we'll be matching against: either the option name is longer than the stub, and the stub needs to match up to that point, or the stub is longer than the option name in which case it can't match anyway. If the stub is longer than the option name, the size of the second string will be reduced down to the size of the option name anyway, [according to the docs](https://en.cppreference.com/w/cpp/string/basic_string/compare), so it shouldn't make any difference anyway, right?

Lestropie: In your proposal, the entire stub would be compared against the entire option string. It would therefore preclude the usage where the user provides a stub that is a subset of the option string. Which is the raison d'etre of the relevant code.

jdtournier: So it does... Sorry, my mistake!

jdtournier (approved):

PR #2632, opened 2023-05-10
Fix compilation warnings on GCC13

Lestropie:
While I was not able to reproduce the MSYS2 CI compilation issue currently manifesting in #2624, upgrading to GCC13 in my MSYS2 resulted in other compilation warnings. This fixes some but not all of them: some files relating to registration are failing to compile but do not produce any error message.

Pending observations here and in #2631, Will want to either rebase or cherry-pick to `master`.

### DIFF
--- a/cmd/dwi2adc.cpp
+++ b/cmd/dwi2adc.cpp
@@ -47,13 +47,18 @@ using value_type = float;
 
 
 
-class DWI2ADC { 
+class DWI2ADC {
   public:
     DWI2ADC (const Eigen::MatrixXd& binv, size_t dwi_axis) :
       dwi (binv.cols()),
       adc (2),
       binv (binv),
       dwi_axis (dwi_axis) { }
+    DWI2ADC (const DWI2ADC& that) :
+      dwi (that.dwi.size()),
+      adc (that.adc.size()),
+      binv (that.binv),
+      dwi_axis (that.dwi_axis) {}
 
     template <class DWIType, class ADCType>
       void operator() (DWIType& dwi_image, ADCType& adc_image) {
--- a/cmd/fod2fixel.cpp
+++ b/cmd/fod2fixel.cpp
@@ -240,7 +240,7 @@ void Segmented_FOD_receiver::commit ()
     Fixel::check_fixel_size (*index_image, *disp_image);
   }
 
-  size_t offset (0), lobe_index (0);
+  size_t offset = 0;
 
 
   for (const auto& vox_fixels : lobes) {
@@ -283,7 +283,6 @@ void Segmented_FOD_receiver::commit ()
     }
 
     offset += n_vox_fixels;
-    lobe_index ++;
   }
 
   assert (offset == fixel_count);
--- a/cmd/shbasis.cpp
+++ b/cmd/shbasis.cpp
@@ -116,13 +116,11 @@ void check_and_update (Header& H, const conv_t conversion)
   header_mask.ndim() = 3;
   header_mask.datatype() = DataType::Bit;
   auto mask = Image<bool>::scratch (header_mask);
-  size_t voxel_count = 0;
   {
     for (auto i = Loop ("Masking image based on DC term", image, 0, 3) (image, mask); i; ++i) {
       const value_type value = image.value();
       if (value && std::isfinite (value)) {
         mask.value() = true;
-        ++voxel_count;
       } else {
         mask.value() = false;
       }
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -392,26 +392,27 @@ class PLYWriter: public WriterInterface<float> {
     }
 
     void computeNormals ( const Streamline<float>& tck, Streamline<float>& normals) {
-      Eigen::Vector3f sPrev, sNext, pt1, pt2, n, normal;
-      sPrev = (tck[1] - tck[0]).normalized();
-
+      Eigen::Vector3f sPrev = (tck[1] - tck[0]).normalized();
+      Eigen::Vector3f normal = Eigen::Vector3f::Zero();
+      
       // Find a good starting normal
-      for (size_t idx = 1; idx < tck.size(); idx++) {
-        pt1 = tck[idx];
-        pt2 = tck[idx+1];
-        sNext = (pt2 - pt1).normalized();
-        n = sPrev.cross(sNext);
+      for (size_t idx = 1; idx < tck.size() - 1; idx++) {
+        const auto& pt1 = tck[idx];
+        const auto& pt2 = tck[idx+1];
+        const auto sNext = (pt2 - pt1).normalized();
+        const auto n = sPrev.cross(sNext);
         if ( n.norm() > 1.0E-3 ) {
           normal = n;
           sPrev = sNext;
           break;
         }
       }
+
       normal.normalize();  // vtkPolyLine.cxx:170
-      for (size_t idx = 0; idx < tck.size(); idx++) {
-        pt1 = tck[idx];
-        pt2 = tck[idx+1];
-        sNext = (pt2 - pt1).normalized();
+      for (size_t idx = 0; idx < tck.size() - 1; idx++) {
+        const auto& pt1 = tck[idx];
+        const auto& pt2 = tck[idx+1];
+        const auto sNext = (pt2 - pt1).normalized();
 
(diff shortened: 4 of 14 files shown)


Lestropie:
Note that the above comment RE making changes on `master` also applies to the change in #2631.

daljit46:
I've pushed some further changes to fix some may-uninitialized warnings. In [SH.h](https://github.com/MRtrix3/mrtrix3/blob/gcc13_warnings/core/math/SH.h), `lmax` is used as a positive integer value but it's type is defined as an `int`. I changed the type to `unsigned int` inside the derivatives calculation function to tell the compiler that the value is always positive. I'm not sure this is the best thing to do, especially in light of the fact that throughout the file `lmax` is inconsistently defined with and passed around with types `size_t`, `int` or `const int`.

Secondly, I pushed another change for [InternalEnergyComputer::pickNeighbour()](https://github.com/MRtrix3/mrtrix3/blob/992ebeff8a1388eae2237809c3c14ed03958234d/src/dwi/tractography/GT/internalenergy.cpp#L92). The current behaviour, it's technically possible for the function to return an empty `ParticleEnd` instance. GCC is complaining that this isn't a good idea since none of its member variables have suitable default values. My solution was to just throw an exception if this was the case, but I'm not sure. @bjeurissen would you mind taking a look at [this](45b582f4a53b757527339ee83d88ee1dc062ec40)

Lestropie:
> throughout the file `lmax` is inconsistently defined

Could be super explicit and do something like `using MR::Math::SH::lmax_t = uint32_t` and then grep for any variables named `lmax`?

daljit46:
> Could be super explicit and do something like `using MR::Math::SH::lmax_t = uint32_t` and then grep for any variables named `lmax`?

Yes, that would certainly be an improvement over the current code.

dchristiaens:
> Secondly, I pushed another change for [InternalEnergyComputer::pickNeighbour()](https://github.com/MRtrix3/mrtrix3/blob/992ebeff8a1388eae2237809c3c14ed03958234d/src/dwi/tractography/GT/internalenergy.cpp#L92). In the current behaviour, it's technically possible for the function to return an empty `ParticleEnd` instance. GCC is complaining that this isn't a good idea since none of its member variables have suitable default values (although IMO they really should). My solution was to just throw an exception if this was the case, but I'm not sure. @dchristiaens would you mind taking a look at this [45b582f](https://github.com/MRtrix3/mrtrix3/commit/45b582f4a53b757527339ee83d88ee1dc062ec40)?

I've had a look and I agree with GCC that this wasn't pretty. However, in practice `pickNeighbour()` should never return an uninitialized `ParticleEnd` because the neighborhood collection is guaranteed to contain at least one element, initialized in the constructor of [`InternalEnergyComputer`(https://github.com/MRtrix3/mrtrix3/blob/a5e5ae85c4a2f16efb720139b589523c6c755321/src/dwi/tractography/GT/internalenergy.h#L45-L49). Unless I'm missing anything, the exception you put in should therefore never be returned and the behavior should be unchanged. I'll test this to be sure.

daljit46:
> Yes, that would certainly be an improvement over the current code.

Upon reflection and discussion, I think introducing the use of unsigned integers in this branch is not appropriate. Mixing unsigned and signed integers is always tricky business and thus I think it would be a rather big change (it will force us to deal with casting issues and unnecessarily make the code verbose). If needed, we can discuss this separately, but for now I have reverted the changes introducing unsigned integers.
To deal with the maybe-uninitialised warning I have added an assert for safety.

Now the code compiles fine with gcc 13.1.0 (I tested on Mac OS) with no warnings, so I think this should be ready for merging. @Lestropie

dchristiaens:
> Unless I'm missing anything, the exception you put in should therefore never be returned and the behavior should be unchanged. I'll test this to be sure.

Test has finished. No exceptions thrown; results are good.

Lestropie:
Still get this on MSYS2 13.1.0 Rev6:

```
In file included from src/dwi/tractography/mapping/mapper_plugins.h:24,
                 from src/dwi/tractography/mapping/mapper.h:31,
                 from src/dwi/tractography/tracking/exec.h:32,
                 from cmd/tckgen.cpp:23:
./core/math/SH.h: In function 'void MR::Math::SH::derivatives(const VectorType&, int, typename VectorType::Scalar, typename VectorType::Scalar, typename VectorType::Scalar&, typename VectorType::Scalar&, typename VectorType::Scalar&, typename VectorType::Scalar&, typename VectorType::Scalar&, typename VectorType::Scalar&, PrecomputedAL<typename VectorType::Scalar>*) [with VectorType = Eigen::Matrix<float, -1, 1>]':
./core/math/SH.h:608:44: warning: '*AL[0]' may be used uninitialized [-Wmaybe-uninitialized]
  608 |           amplitude = sh[index (0, 0)] * AL[index_mpos (0, 0)];
```

This needs [a8ffe19](https://github.com/MRtrix3/mrtrix3/pull/2632/commits/a8ffe199cefdeb5170d9473037379759b2e04598) but that was reverted in [f4f7461](https://github.com/MRtrix3/mrtrix3/pull/2632/commits/f4f746155a16f588b7d11ca93b85e89edd3e9fa1). Problem is that the assert statement in [3af5328](https://github.com/MRtrix3/mrtrix3/pull/2632/commits/3af5328bcec8238dd5ad0945efb7fa73eac763e4) prevents the warning when compiling in debug mode, but the warning still appears when compiling in release mode as the assertion is removed by the precompiler.

With CI caching (#2624) it should be possible to add compiler checks for different compilation configurations, such as with & without assertions / debugging symbols. From memory we had to exclude compilation with debugging symbols from testing because it timed out?

Might be preferable to clean up the history by removing close immediate commit / reversion pairs and cherry-pick everything else.

daljit46:
Ok, so is it our aim to eliminate warnings even when asserts are disabled? My understanding was that we only compile with warnings when doing the CI checks, but the release versions aren't compiled with -Werror.

Lestropie:
> the release versions aren't compiled with `-Werror`

Moreso beyond CI it would be nice to not be met with compiler warnings if compiling the software in release mode during development, for eg. testing computation speeds / test data that takes too long to process if running in debug mode / external users utilising development branches for specific features.

daljit46:
If we want to get rid of that warning without asserts and the use of unsigned integers, we can either do something like:
```
if(lmax < 0)
    // return or __builtin_unreachable() or throw an exception;
}
```
This should get rid of the warning. The `__builtin_unreachable` macro is supported on Clang and GCC (but not MSVC) so it should be ok (C++23 will introduce a [std::unreachable](https://en.cppreference.com/w/cpp/utility/unreachable) utility that is compiler agnostic.

daljit46:
Ok this should now compile without warnings even  without asserts enabled.

Lestropie:
Getting `lmax < 0` would be a programmatic mistake. Anywhere it's specified by the user there should be a lower limit set in the command-line interface, and anywhere it's generated by code it should be mathematically impossible to get negative. Personally in such scenarios I tend to place assertions rather than throwing Exceptions. Admittedly our testing doesn't make an exhaustive attempt to break it. But personally I reserve Exceptions for feedback that would be useful to a user, and it shouldn't be possible for a user to create such a scenario.

I suppose more robust would be not just a typedef for lmax data, but a class that forces explicit rather than implicit casting between signed and unsigned? If that were appropriately implemented in `SH.h` it would necessitate adoption in all dependent code at compile time.

daljit46:
Asserts save us when debugging, but for releases we need a different solution. The `__builtin_unreachable` macros gives liberty to the compiler to completely optimise out the case where the variable is negative, thus essentially making the assert useless.
`lmax` being non-negative is a precondition for the `derivatives` function to operate correctly.  The [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e2-throw-an-exception-to-signal-that-a-function-cant-perform-its-assigned-task) mention that exceptions are appropriate if this is the case. In light of this, I've changed the code to throw `std::logic_error` if lmax is negative inside the function.

I agree that a more robust solution is needed since `lmax` is being used in many other places. However, we should probably that discussion outside of this pull request as that would imply a significant effort in deciding what would be the appropriate course of action.

Lestropie:
Voting for 1, with suggestion of creation of a separate thread to discuss prospects of transition to using built-in exceptions rather than `MR::Exception` class.

PR #2639, opened 2023-05-21
FOD FMLS: Masks instead of lookup tables

### LINKED ISSUE #2637: Fixel lookup data: Reducing storage requirements
With #2625, a lookup table for each voxel is stored. For FOD segmentation using an N-direction set (1281 by default), this lookup table is an X x Y x Z x N image, where the N values in each voxel correspond to the fixel indices to which each of the N directions is ascribed.

The current `DWI::Fixel_map<>` implementation (used in eg. `tcksift(2)`) dynamically allocates this N-vector only in non-empty voxels, which minimises the total RAM usage. That class is in the process of being deprecated as part of larger changes standardising representation of fixel data between filesystem storage and commands that operate on fixel data in RAM.

For #2625, I chose to not go for clever storage and just write it as a 4D image, sacrificing storage space for simplicity. **But its presence increases the storage space of a fixel dataset by about an order of magnitude.** And in extreme circumstances it can get rather large, especially for high-resolution data: I was shown a 0.75mm template where the size of the lookup table alone was > 32GB. Gzipping will be effective for storage due to to the amount of zero-filling, but that's not much help if uncompressing into RAM causes problems.

Possibilities to consider:

1.  Rather than a dense X x Y x Z x N image, store *a bitwise N-vector per fixel*, being a mask of those directions that are attributed to that fixel. So it would be a bitwise fixel data file of size F x N x 1, where F is the total number of fixels. 
    This would reduce storage requirements as long as the average number of fixels per voxel is less than 8.
    It would however mean that the typically requested operation---from a given direction, choose the corresponding fixel---would not be quite as simple.
    It would also not preclude the case where a single dixel can be attributed to multiple fixels. This could be seen as good or bad. In the FMLS segmentation there are in fact dixels that could theoretically be ascribed to multiple fixels, though currently they are just assigned to whichever is largest; here such a dixel could be explicitly ascribed to both fixels. But any downstream application detecting such would need to be able to do something with that information. Something like SIFT could for instance contribute half of its intersection length in that voxel to each of the two fixels.

2.  Utilise a *4-bit integer data type*. The values contained in the image are offsets from the first fixel in the voxel. Therefore with a 4-bit data type it would still be possible to support up to 16 fixels per voxel, which I think is sufficient (though I would not go as low as 2-bit = 4), and would halve the total filesystem space required for these data.

3.  Use an *indexing operation*, similar to what is done for the fixel data (indeed it's the same as what is done in `mrclusterstats`).
    So there would be an integer X x Y x Z image, where the value in any given voxel provides a unique index for that voxel, indexing into a V x N x 1 image (where V is the total number of non-empty voxels) where the lookup table for that voxel is found.

5.  *Accept* the large size.    

Writing it out now, I'm tempted to pursue option 1.
But external opinions welcome.
Particularly given that it's intended that the format of this lookup information will be a part of the fixel directory format specification (not requisite to the same extent as index and directions, but nevertheless standardised).

Lestropie:
Likely resolves #2637.

The storage size problem was worse than I thought.
For the BATMAN dataset, `fod2fixel` produces ~ 100,000 fixels and requires ~ 5.5MB.
Writing the lookup table as currently done in #2625 requires an additional ~ 700MB.
The solution here requires an additional ~ 16MB.

In addition to the reduction in storage space, it actually provides a little bit more information.
Part of the FOD FMLS segmentation process is identifying dixels where the FOD lobes that are segmented as two different fixels "touch".
By representing the association between dixels and fixels as a binary dixel mask per fixel rather than a fixel index per dixel, it's possible to attribute such dixels to *both* fixels, which is more faithful, even if it potentially means more complex calculations downstream.

Given the magnitude of the change in storage, I think I need to commit to this solution.
But in addition to the consequences to the stuff that I'm implementing for which this is all requisite, it also modulates the proposed change to the fixel directory format.

#### Question:

People need to be content with both the proposed structure and the image name (I'm not sold on "`dixelmasks.mif`", open to suggestions).

### DIFF
--- a/cmd/fod2fixel.cpp
+++ b/cmd/fod2fixel.cpp
@@ -33,7 +33,7 @@
 
 #define DEFAULT_DIRECTION_SET 1281
 
-#define DIXEL_IMAGE_PREFIX "fixel_dixels_"
+#define AMPLITUDES_IMAGE_PREFIX "amplitudes_"
 
 
 const char* fixel_direction_choices[] { "mean", "peak", "lsq", nullptr };
@@ -89,9 +89,7 @@ const OptionGroup OutputOptions = OptionGroup ("Options to modulate outputs of f
                           "options are: " + join(fixel_direction_choices, ","))
     + Argument ("choice").type_choice (fixel_direction_choices)
 
-  + Option ("nolookup", "do not export the lookup table image")
-
-  + Option ("dixel_images", "write a set of dixel images, where each encodes the FOD amplitudes for one fixel in each voxel");
+  + Option ("amplitude_images", "write a set of dixel images, where each encodes the FOD amplitudes for one fixel in each voxel");
 
 
 
@@ -154,16 +152,15 @@ class Segmented_FOD_receiver {
   public:
 
     using DataImage = Image<float>;
-    using DixelImage = Image<float>;
-    using IndexImage = Image<index_type>;
-    using LookupImage = Image<uint8_t>;
+    using AmplitudeImage = Image<float>;
+    using IndexImage = Image<uint64_t>;
+    using DixelMaskImage = Image<bool>;
 
     Segmented_FOD_receiver (const Header& header,
                             const DWI::Directions::FastLookupSet& dirs,
                             const std::string& directory,
                             const fixel_dir_t fixel_directions,
                             bool save_as_nifti = false,
-                            bool write_lookup = true,
                             bool write_dixel = false);
 
     ~Segmented_FOD_receiver();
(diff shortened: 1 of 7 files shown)


Lestropie:
Also note that storing the relevant information in this way would *preclude* subsequent generation of fixel amplitude images as [shown](https://github.com/MRtrix3/mrtrix3/issues/2468#issuecomment-1521903935) in #2468, which explicitly requires the data to be stored in a lookup table format. However as shown in #2625 (particularly [8e354768](https://github.com/MRtrix3/mrtrix3/pull/2625/commits/8e35476845b034ef9e5cdca0fe8f59933563fdec)), this is better addressed through addition of a command-line option to `fod2fixel` to generate the desired images at segmentation time when dixel data are already at hand.

PR #2650, opened 2023-06-05
Replace MR::vector with std::vector

daljit46:
As the title suggests, this contains changes to replace all instances of `MR::vector` with `std::vector` in light of our coming switch to C++17.
This was one of the objectives identified in #2585.


Lestropie:
There's also a corresponding test in `check_syntax` that will need to be removed.

Issue #2643, opened 2023-05-26
Use clang-format for C++ code


daljit46:
I would like to suggest to use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to format our C++ code. This will provide consistency and make the formatting of our C++ code uniform for all developers. We can enforce this by a simple CI check that ensures that new codes adhere to our chosen style. clang-format is available on all platforms and is supported by all major IDEs.

As for what formatting style we should use, clang-format provides 6 default styles: LLVM, Google, Chromium, Mozilla, WebKit or Microsoft. We should just pick one and stick with it (you can experiment with the different styles and options [here](https://clang-format-configurator.site/)). If we want to customise specific formatting rules we can, but we should try to stick with one of the defaults.

Lestropie:
Personally I'm leaning toward adopting this along with the other software engineering changes, especially post-production-release and given that git blame can be kept clean (#2592) and that changes can be attributed to @MRtrixBot.

Regarding styles:
- Agree that it'd be preferable to avoid custom rules if possible; but depending on how my assessment below goes I might not be happy with any one out of the box...
- Would prefer to avoid due to distinguishing features:
    - Mozilla: Really dislike newline between function return type and name
    - WebKit: Same as Mozilla but also dislike quad spacing
    - Microsoft: Dislike massive vertical spacing caused by dedicated lines for opening brackets
- Regarding the other three: I ran an exemplar .h / .cpp file pair and looked at what it did to my code, and generated the table below. Looks like currently my preference would be LLVM but with a couple of additional constraints. Not guaranteed to be an exhaustive test, would want to re-evaluate with more exemplar code if committing.
- Not a massive fan of column width limit across the board. If I need my code in a narrow column I'll turn on word wrapping... yet looks like all styles default to 80 characters...

--------------------------

| Feature                                                                                | LLVM                   | Chromium               | Google               |
| -------------------------------------------------------------------------------------- | ---------------------- | ---------------------- | -------------------- |
| Forces newline between macro invocation and inputs                                     | No (good)              | Yes (bad)              | Yes (bad)            |
| Short if-statement + operation                                                         | Force two lines (good) | Force two lines (good) | Force one line (bad) |
| Forbids multiple class member variable initialisations on one line                     | No (bad)               | Yes (good)             | Yes (good)           |
| Forbids multiple function arguments on one line if multiple lines                      | No (bad)               | Yes (good)             | No (bad)             |
[...]

Issue #2579, opened 2023-02-01
Automated determination of #include's


Lestropie:
I think this had been discussed at some point in the past, but isn't explicitly listed on here.

When adding new code files / augmenting existing code files / moving code between `.h` and `.cpp`, it can be difficult to get the `#include` directives correct. Typically one adds enough to get compilation working, but sometimes it relies on second-order identification of dependencies (ie. a `#include` within one of the `#include`'d files). What would be nice is an automated system that applies a similar interpretation to the compiler, identifying the minimal set of `#include`'s necessary for compilation / recognition of dependencies, that would additionally be executed and compared for differences as part of CI.

Previously when I looked into this I didn't find a solution I thought was adequate. But yesterday I came across [Include What You Use](https://include-what-you-use.org/), and it looks like it could work. It can provide explicit justifications for each `#include`, it can replace such with forward declarations in `.h` files where possible, and it can edit files in-place.

Lestropie:
- [ ] Install IWYU tool and verify that it works when executed on a single source code file
- [ ] Write Python script that:
    - [ ] Loops through MRtrix code base and identifies all files for which it is appropriate to execute IWYU (C++ source / header files only, may need to explicitly exclude some source files that did not originate from MRtrix)
    - [ ] Executes IWYU on that list, saving the results somewhere
    - [ ] Applies the recommended changes to all source code files
- [ ] Add new capabilities to Continuous Integration (CI):
    - [ ] Install IWYU
    - [ ] Execute Python script above
    - [ ] Yield test failure if the recommended changes are non-zero

PR #2652, opened 2023-06-09
Apply formatting for C++ using clang-format

daljit46:
This pull request will introduce `clang-format` as the formatting tool for our C++ code. As discussed in #2643, this will provide consistency and a uniform formatting way for all developers. 
It contains a new CI workflow that checks whether new code is formatted correctly.  It also contains a [pre-commit](https://pre-commit.com/) configuration file to install a local git hook, which developers can use to ensure that their code is properly formatted before pushing.
Formatting has already been applied to all C++ files using the LLVM style with some custom settings (still needs tweaking).

Any suggestions or feedback is very welcome!


daljit46:
OK, I think this looks quite good now, and I think it's not a bad idea to merge this in. There are few things that need to be considered. Since we are enforcing the 100 words character limit, the scripts for generating the documentation and updating the copyright notice were broken. For the former, I directly updated the related rst files and adjusted the regular expressions needed to parse the configuration options and the environment variables in the code. For the latter, I adjusted the `update_copyright` script to update the copyright notices correctly. It'd be good to review these changes (last three commits in the pull request).

Other than that, everything looks good.

Lestropie:
🤦‍♂️ I'd written a bunch of stuff on this PR at some point, but clearly didn't hit the send button, and now can't find it unsubmitted on either system... I'll have another go, I can probably remember some of the code snippets I didn't like.

Regarding destination branch / merge timing: I'd commented on Teams and in #2655 that there's a bunch of proposed modifications that aren't C++17 and therefore aren't #2585 / `cpp17` branch, but nevertheless fall into some kind of category of "back-end cleanup"; eg. #2111, arguably #2579 due to its interaction with this PR. An alternative proposal would be that just like a central `cpp17` branch was created to act as the target for all C++17-related PRs to eventually all be merged to `dev` in one go, there could be considered to be another set of non-C++17 software eng modifications with its own target branch. Not necessarily pushing for it as it's a fairly loose grouping, just throwing it out there.

daljit46:
I think out of the c++17 changes, the only one that makes sense to deliver together with this PR is #2641. The way I see it is that this is (mostly) a formatting change, thus it shouldn't modify the behaviour of the code, while for example #2111 is a change that will affect the generated code (and most likely cannot really be automated because each macro usage will have to be evaluated before switching to `constexpr`).
#2579 could also be a good candidate if we decide to create a relevant super PR that includes this one, although I think we need to test that change (in the past, I've had quite a few issues with `include-what-you-use` when dealing with system headers).

Lestropie:
1.  Commit changing formatting needs to go into `.git-blame-ignore-revs`.

1.  Still debate to be had about the character limit. Personally I'd rather have code that is a little wide, maybe doesn't display quite right on some people's systems in some contexts, and maybe draws an explicit request to split, rather than a hard limit less than a quarter the width of my screen that I have to sacrifice readability of code logic or variable names to conform to.
    (I'd joked to @daljit46 that maybe we should use a limit of 132 characters.... let's see if @jdtournier gets the reference)

1.  Many text strings are poorly split when done automatically, and we may want to go through and manually clean a few up.

1.  In `usage()` of `cmd/*.cpp`, the positioning of operators & spaces around them is really not great. This might be due to being an unconventional use of classes or something. But I'd like to try to modify that without breaking formatting elsewhere.

1.  Ternary operators seem to appear at line start rather than line end, which is alien to me; but I'd get used to it if need be.

1.  In `core/adapter/extract.h` I see:

    ```
    Extract(const ImageType &original, const vector<vector<uint32_t>> &indices)
          : base_type(original), current_pos(ndim()), indices(indices), trans(original.transform()) {
    ```

    Here I think I'd rather `PackConstructorInitializers: Never`

    See also:

    ```
    Gaussian1D(const ImageType &parent,
                 default_type stdev_in = 1.0,
                 size_t axis_in = 0,
                 size_t extent = 0,
                 bool zero_boundary = false)
          : base_type(parent), stdev(stdev_in), axis(axis_in), zero_boundary(zero_boundary) {
    ```

    Initializers often shadow constructor parameters, so having the latter split but the former not just looks weird.

1.  In function `paragraph()` defined in `core/app.*`, it splits the return type onto a newline rather than splitting the function parameters. Does it treat naked functions different to class member functions?

(as far as I'm getting this evening; I might continue looking through more exemplar source files at a later date)

Lestropie:
Some examples of where stringently enforced line width causes ugliness:

Old:
```
template <typename X>
    struct max_digits<X, typename std::enable_if<std::is_fundamental<typename X::Scalar>::value, int>::type> { 
      static constexpr int value () { return std::numeric_limits<typename X::Scalar>::max_digits10; }
    };
```

New:
```
template <typename X>
struct max_digits<
    X,
    typename std::enable_if<std::is_fundamental<typename X::Scalar>::value, int>::type> {
  static constexpr int value() { return std::numeric_limits<typename X::Scalar>::max_digits10; }
};
```

-----

Old:
```
  //! check whether type is compatible with MRtrix3's file IO backend:
  template <class ValueType>
    struct is_data_type :
      std::integral_constant<bool, std::is_arithmetic<ValueType>::value || is_complex<ValueType>::value> {  };
```

New:
```
//! check whether type is compatible with MRtrix3's file IO backend:
template <class ValueType>
struct is_data_type
    : std::integral_constant<bool,
                             std::is_arithmetic<ValueType>::value || is_complex<ValueType>::value> {
};
```

daljit46:
I've added a new script called `clang-format-all` in the top level directory, which formats all C++ code in the source directory (excluding 3rdparty headers). The script also allows you to specify a custom `clang-format` binary (e.g. `./clang-format-all -e custom_clang_format`) so you can always a specific version of `clang-format`. This should ease the process of migrating existing PRs to our new formatting rules.

I think this can now be merged in. This will require all existing and future PRs to format their code with `clang-format` and pass our CI checks.

daljit46:
@MRtrix3/mrtrix3-devs unless anyone have any reservations, I will merge this into dev soon.

PR #2654, opened 2023-06-13
Mesh::load_vtk() always open files in binary mode

### LINKED ISSUE #2653: meshconvert: binary tests failing on Windows
Binary test for `meshconvert` fails on dev branch because VTK files are opened in text mode on Windows. This seems to result in `std::istream::tellg` reporting nonsensical offset values (`tellg` on Windows can only be relied upon if files are opened in binary mode). For further details see #2624

To reproduce the bug you can run the binary tests after building MRtrix with gcc 13 or later on MSYS2 under Windows 10/11 (non Windows platforms are unaffected by these bugs because they don't distinguish between text or binary modes). 

An error log of the test failure can be found [here](https://github.com/MRtrix3/mrtrix3/actions/runs/5013945044/jobs/8989008895).

jdtournier:
Fix #2653.

### DIFF
--- a/src/surface/mesh.cpp
+++ b/src/surface/mesh.cpp
@@ -96,14 +96,14 @@ namespace MR
     void Mesh::load_vtk (const std::string& path)
     {
 
-      std::ifstream in (path.c_str(), std::ios_base::in);
+      std::ifstream in (path.c_str(), std::ios_base::binary);
       if (!in)
         throw Exception ("Error opening input file!");
 
       std::string line;
 
       // First line: VTK version ID
-      std::getline (in, line);
+      MR::getline (in, line);
       // Strip the version numbers
       line[23] = line[25] = 'x';
       // Verify that the line is correct
@@ -111,10 +111,10 @@ namespace MR
         throw Exception ("Incorrect first line of .vtk file");
 
       // Second line: identifier
-      std::getline (in, line);
+      MR::getline (in, line);
 
       // Third line: format of data
-      std::getline (in, line);
+      MR::getline (in, line);
       bool is_ascii = false;
       if (line == "ASCII")
         is_ascii = true;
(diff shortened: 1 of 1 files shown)


src/surface/mesh.cpp
```diff
@@ -96,52 +96,45 @@ namespace MR
     void Mesh::load_vtk (const std::string& path)
     {
 
-      std::ifstream in (path.c_str(), std::ios_base::in);
+      std::ifstream in (path.c_str(), std::ios_base::binary);
```
Lestropie, line 99:
```suggestion
      std::ifstream in (path.c_str(), std::ios_base::in | std::ios_base::binary);
```
Not sure what the behaviour will be if the path doesn't exist and only the `binary` flag is set; it might create an empty file?

Lestropie (approved):
Makes sense, and passes MSYS2 tests.
One [small suggestion](https://github.com/MRtrix3/mrtrix3/pull/2654#pullrequestreview-1478517272).
Should this instead (/ also) be going to `master`?

src/surface/mesh.cpp
jdtournier: It took me a while to find out any info on what the flags end up being set to when non-default, but I finally found confirmation of the obvious [on the cplusplus website](https://cplusplus.com/reference/fstream/ifstream/ifstream/):

> * `in` is always set for [ifstream](https://cplusplus.com/ifstream) objects (even if explicitly not set in argument _mode_).

So there's no scope for that using a `std::ifstream`... :sweat_smile:

jdtournier: Sorry, hadn't realised this was a suggested change - must be a new feature I hadn't come across. Happy to commit that change too if you think it's important to be explicit, though I personally feel this is redundant - up to you.

jdtournier:
> Should this instead (/ also) be going to master?

Yes, I think it will need to. I've only worried about `dev` so far since this is what's blocking the compiler caching changes to the CI (#2624), but there's no doubt a recent build of `master` on a recent version of MSYS2 will also be affected. What I'm not sure about is whether the precompiled binaries we provide for MSYS2 would also be affected when running on a modern version of MSYS2 - it depends on whether the bug is in the system C/C++ dynamic libraries that MRtrix links to at runtime, or in the MRtrix package code generated by the compiler...

src/surface/mesh.cpp
Lestropie: Lol... ignore me.

PR #2656, opened 2023-06-15
Mesh::load_vtk() always open files in binary mode

jdtournier:
Duplicate of #2654 for `master`. Might be worth also adding CI compiler caching (#2624) while we're at it, but that's less likely to be much of an issue given the (hopefully!) much lower rate of PRs on `master`. Would really only be worth doing to make sure any subsequent packages get built with a more modern version of MSYS2.


Lestropie (approved):
:+1:

PR #2611, opened 2023-03-24
pull request for Automated determination of #includes #2579

988158:
This pull request contains the script that executes IWYU on each .cpp and .h file within the mrtrix3 directory.
The script is able to:

- load comilpation flags from config file.
- find every .cpp and .h file to be analysed.
- execute IWYU on those files.

The script is not able to:

- execute iwyu on gui-related files.
- save and apply the result of iwyu

PR #2661, opened 2023-06-22
Experimental include-what-you-use support

daljit46:
This is a preliminary attempt at incorporating [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use) in our workflow. The scope is to figure out whether it makes sense to follow the strategy outlined in #2579. 
It remains to be seen how IWYU should be incorporated, whether as a simple hinting mechanism for contributors or as an enforcing CI check that blocks PRs (possibly after running the [automatic fixing script](https://github.com/include-what-you-use/include-what-you-use#how-to-correct-iwyu-mistakes) on the entire codebase). 

The benefit of IWYU are the removal of unnecessary includes and an avoidance of relying on transitive includes.

### DIFF
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -1,0 +1,43 @@
+name: include-what-you-use
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    env:
+      CFLAGS: -Werror
+      QT_SELECT: qt5
+      
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install dependencies
+      # Note that we need to install clang-13, since iwyu on latest Ubuntu is built with clang-13.
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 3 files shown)


Lestropie:
If having a proper go at IWYU, if possible I'd like to have https://github.com/MRtrix3/mrtrix3/pull/2611/commits/d5b191e1adf74ed1c5f44a22f463c7d104a2560a from the `iwyu` branch included in the history as per #2611. But we'll need to look at exactly how it's being proposed to integrate as to whether that's possible / appropriate.

Lestropie:
Note for if/when this gets finished:
In #3368 I've added to `check_syntax` a check to make sure that the MRtrix custom Eigen plugins are specified before any Eigen headers are loaded. Any mechanism for modifying presence / order of include headers would need to preserve this.

daljit46:
A note on this: a major blockage when I pursued this is described [here](https://github.com/include-what-you-use/include-what-you-use/issues/1265). I can't remember what my conclusion was, but that probably needs further investigation. I also believe that Eigen 4 has added support for IWYU in their header files.

Issue #2605, opened 2023-03-21
CLI: Export Pydra interfaces


Lestropie:
https://github.com/nipype/pydra

Similar to existing `__print_usage_*__` hidden command-line functions, it should be possible to auto-export---from both C++ and Python---the requisite interface information for Pydra.

Pinging @tclose.

Lestropie:
Code links:

-  C++:
    -   Catching special options: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/app.cpp#L1035
    -   `__print_full_usage__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/app.cpp#L659
    -   `__print_usage_markdown__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/app.cpp#L687
    -   `__print_usage_rst__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/app.cpp#L798

-   Python:
    -   Catching special options: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/lib/mrtrix3/app.py#L134
    -   `__print_full_usage__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/lib/mrtrix3/app.py#L874
    -   `__print_usage_markdown__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/lib/mrtrix3/app.py#L919
    -   `__print_usage_rst__` implementation: https://github.com/MRtrix3/mrtrix3/blob/3.0.4/lib/mrtrix3/app.py#L992
    -   Note that type information of Python command-line arguments / options is currently limited; will be enhanced by: https://github.com/MRtrix3/mrtrix3/issues/1392 (currently under development; see https://github.com/ankitasanil/mrtrix3/pull/1)

Lestropie:
Correct.

For eg. online documentation, a script generates that documentation, and that same script is re-run in CI to check for differences, so that software interface and online documentation are guaranteed to have correspondence. So that content is encapsulated within the repository itself, including the versions that correspond to prior tagged releases. For Pydra, is there a preference for having that interface information embedded within the relevant repository vs. stored somewhere externally?

tclose:
> For Pydra, is there a preference for having that interface information embedded within the relevant repository vs. stored somewhere externally?

[...]

Issue #2580, opened 2023-02-03
Reduce generalization of `.type_various()`


Lestropie:
There are now a number of commands where the user's input for a particular argument could be interpreted by the command in multiple ways; eg. a transform could be a text file with an affine transformation or a warp image, or a set of directions on the unit hemisphere could come from a text file or an image header. To support these cases, we're added the "`.type_various()`" classification, such that any special handling within the command-line parsing is effectively bypassed. This will also appear in some cases where the command-line interface is exported to text.

What would be a more targeted solution is to allow each argument to receive multiple classifications. So eg. in the examples above, the argument would be classified as being compatible with either an input file or an input image, but not anything else. Any command-line interpretation cleverness would then be compatible with the user providing either of those cases. And exported command-line interface information would more faithfully communicate what is and is not permitted for that argument.

Not sure how much benefit could actually be gained from such---probably not much---but I wanted to document it in case it may be recognised as the solution to some issue.

Lestropie:
Another aspect that is probably closely enough related to this to warrant grouping them together rather than listing a separate Issue:

Currently, `.type_choice()` is treated as a distinct entry in the list of possible types, with the elements to choose from being a list of strings. This assumes that the only situation in which one might want to restrict the user to selecting one from a distinct set of possibilities is if that set of possibilities are of text form. Hypothetically, the presence or absence of a finite set of permissible inputs is not limited to arguments of text form. Instead, these current usages could be considered as *both* being of text form *and* having a finite set of permissible values. In conceiving of the operation in this way, arguments of type other than text could also have a finite set of permissible values, where the elements of that set have the same type as the argument. So eg. some integer input may not just be restricted to minimum and maximum values, but be forced to be one from some set of permissible integer values; but the type of that argument should still be integer, whereas currently achieving this using `.type_choice()` kind of forces them to be of text type, with subsequent explicit conversion to integer after command-line parsing.

PR #2665, opened 2023-06-29
Implement Pydra code-generators

tclose:
Implementation of #2605

### Tasks:

- [x] `__print_pydra_code__` for C++ commands. 
- [x] `__print_pydra_code__` for Python commands
- [x] Script to loop through all cmds and print their "pydra usage" to *.py files
- [x] Fix up package TOML (maybe switch to `hatchling`?)
- [x] GitHub action to be run on new release to upload `pydra-mrtrix3` package to PyPI
- [x] Tests to check generated code can print cmdlines properly (@Lestropie, do you have any ideas for this? I was thinking if there was a hidden "dry-run" option then 

### Blocked by: 
- [x] Complete #1392 so that Python commands type file-types


Lestropie:
As mentioned in #2605, Python will require #1392 via https://github.com/ankitasanil/mrtrix3/pull/1 to have typed command-line parameters.
@ankitasanil: Still interest in completing the PR, or will @MRtrix3/mrtrix3-devs need to finish it off?

tclose:
> As mentioned in #2605, Python will require #1392 via [ankitasanil#1](https://github.com/ankitasanil/mrtrix3/pull/1) to have typed command-line parameters. @ankitasanil: Still interest in completing the PR, or will @MRtrix3/mrtrix3-devs need to finish it off?

I have written the Python version as best as I can with the current level of typing, but you are right that it will need more to get it to work in a guaranteed way.

On typing of files, how many different formats does MRtrix interact with (e.g. bvec, bval, response.txt, etc...) and would it be too much to ask to be able to give each format its own ArgType (as opposed to just Image* and Track* currently). As I have a PR that refactors Pydra's type-checking to be aware of file-formats (nipype/pydra#662), and it would be great to be able to specify them exactly instead of defaulting to a generic `FileSet` class.

Lestropie:
> how many different (file) formats does MRtrix interact with

Assuming upstream exclusion of images & tractograms from that set, here's what comes to mind:

-   1D & 2D numerical data input & output (currently ASCII only but also NPY after #2437)
-   TSFs
-   JSONs (in multiple contexts)
-   `.dcm` files
-   Other text files:
    -   Lists of paths to other files (eg. stats command inputs)
    -   Generic tabulated data (eg. parcellation lookup tables but others also)

As far as further custom types in user interfacing, I would argue that:

-   That could potentially be the case for TSFs / JSONs / `.dcm`s, since those represent singular unambiguous cases of file extension / content of data / requisite parsing code / interpretation of data.
-   The specific examples you provide (bvec / bval / response functions) I'm more sceptical of, as they are just numerical data, albeit sometimes with expectations of dimensionality / size; they're just subsets of "Nd numerical data" to which certain interpretations are applied *after the data are parsed*.

Curious to hear others' thoughts.

tclose:
> > how many different (file) formats does MRtrix interact with
> 
> Assuming upstream exclusion of images & tractograms from that set, here's what comes to mind:
> 
> * 1D & 2D numerical data input & output (currently ASCII only but also NPY after [Npy file format support #2437](https://github.com/MRtrix3/mrtrix3/pull/2437))
> * TSFs
> * JSONs (in multiple contexts)
> * `.dcm` files
> * Other text files:
>   
>   * Lists of paths to other files (eg. stats command inputs)
>   * Generic tabulated data (eg. parcellation lookup tables but others also)
> 
> As far as further custom types in user interfacing, I would argue that:
> 
> * That could potentially be the case for TSFs / JSONs / `.dcm`s, since those represent singular unambiguous cases of file extension / content of data / requisite parsing code / interpretation of data.
> * The specific examples you provide (bvec / bval / response functions) I'm more sceptical of, as they are just numerical data, albeit sometimes with expectations of dimensionality / size; they're just subsets of "Nd numerical data" to which certain interpretations are applied _after the data are parsed_.
> 
> Curious to hear others' thoughts.

Yes, maybe file formats is a bit of a misnomer, data types would be more accurate/actually what is captured in Pydra. In any case, it would be very helpful on the Pydra code-gen front if these file-data-types could be specified at bit finer granularity

tclose:
Ok, so this just needs some unit tests to test out the generated Pydra interfaces and to wait for https://github.com/MRtrix3/mrtrix3/issues/1392.

@Lestropie, on the unit tests, do you think it would be an issue if I added a `--dry-run` option to the general app options, which basically tests to see whether the provided options parse and then exits?

tclose:
> Ok, so this just needs some unit tests to test out the generated Pydra interfaces and to wait for #1392.
> 
> @Lestropie, on the unit tests, do you think it would be an issue if I added a `--dry-run` option to the general app options, which basically tests to see whether the provided options parse and then exits?

If you don't want to clutter the help menu further, maybe there could be special configure arg to insert it in?

Lestropie:
One note I might throw in here just because I encountered it while working on the code and it's of some limited relevance.

For a command-line option, in C++, you specify the option, and then you can specify some number of compulsory command-line arguments that must follow that option, and get consumed from the command-line by the parser immediately following the appearance of the option. Each of those arguments may have its own type.

In Python, this can't currently be done. In `argparse`, the option itself is assigned a type, as well as a number of arguments. Therefore you can't have >1 arguments following a command-line option where those arguments have different types to one another.

daljit46:
Please note that this PR will need to conform to the formatting requirements introduced in #2652, thus all C++ files need to formatted using clang-format (version 16). See [here](https://github.com/MRtrix3/mrtrix3/blob/dev/CONTRIBUTING.md#coding-conventions).

tclose:
> In Python, this can't currently be done. In `argparse`, the option itself is assigned a type, as well as a number of arguments. Therefore you can't have >1 arguments following a command-line option where those arguments have different types to one another.

I think the typical way to represent this is to use the `typing.Tuple[MyType1, MyType2]` form

Lestropie:
> I think the typical way to represent this is to use the typing.Tuple[MyType1, MyType2] form

Having had a small taste of Pydra now, this might be better than doing the most permissible thing of just accepting any string in such a scenario, but doesn't fully address the issue to the extent that the MRtrix3 C++ interface does, at least to the best of my understanding.

Say for instance a command-line option takes as input a string and then an integer. Yes you could make that command-line option a `typing.Tuple[str, int]`, but that would mean that the user could erroneously specify an integer and then a string, and that wouldn't be caught by the command-line parser.

PS. Relevance to #2580: making the argument types in C++ a bitwise flag indicating what set of types is permissible would be faithful to the `typing.Tuple[MyType1, MyType2]` usage I've seen.

Lestropie:
@tclose: After discussion with @jdtournier, there is trepidation about embedding the Pydra interfaces within the repository.

On the plus side, it would enforce consistency, PR changeset would show interface changes that might otherwise be invisible to the CI checks of the online documentation, and be convenient from the perspective of workflow managers grabbing both the software and the interface to it.

On the minus side, it's implicitly taking on management responsibility for interfaces to an environment for which none of the core development team have experience; even if the interfaces were to be committed in a functional state, and the MRtrix3 commands were to not change, it would nevertheless be possible for there to be changes to Pydra that would necessitate revisions here.

What would it look like from the Pydra perspective if the MRtrix3 Pydra interface were to be provided as a stand-alone repository? Ie. What would be the magnitude of inconvenience? There's obviously the issue that on a tagged release of MRtrix3 you would need to update that external repository and generate a corresponding tag.

tclose:
@Lestropie @jdtournier just to clarify that the interfaces are not being stored, instead generated from the MRtrix C++ & Python command declarations (see `__print_usage_pydra__`). Therefore, the interfaces should evolve with changes to the MRtrix commands as long as the command definition syntax stays stable.

It would be great if the __print_usage_pydra__ special command (or whatever it is called) can live in the MRtrix source, as I'm not sure it will be easy to implement/maintain that in a separate code-base (if possible at all). However, the generation code and Python package can be stored in a separate repository if required. This is what has been typically done for other toolboxes.

We are hoping to move to a model where the Pydra interfaces could be integrated into the tool source repos to make it easier for them them stay in step with each other. Given it is my code, I'm happy to put my hand up to deal with any issues that might arise, as I would need to do this if it were stored in a separate repository. But in any case, totally up to you guys, I can integrate the changes into the existing nipype/pydra-mrtrix3 repo instead.

jdtournier:
Hey @tclose!
Just a few points/questions:

No problem at all from our point of view to include the code responsible for _producing_ the interface definitions. That would indeed be really tricky to maintain outside of the core repo, but it's not something that I expect to change much once implemented, and it makes sense for it to be there. I would opt for a name such as `__print_usage_json__` (assuming Pydra does expect JSON...) just to avoid giving the impression that this can only be used in Pydra - if it's JSON, there's no reason it can't be used in other contexts as well .

The bigger question from my point of view is whether to _store_ the generated definitions in the core repo. It's trivial to generate them at build-time, and maybe make it part of the CMake install or something. But that doesn't require the generated interface definitions themselves to also be stored on the main repo. What I was trying to figure out was how you envisage these files to be used in practice. If they're stored alongside a full MRtrix install, it would be possible for an application like Pydra to query them at run-time and offer a version of the interface that is always guaranteed to be in lock-step with the version of MRtrix installed. If they're stored elsewhere, then there is always the risk of a mismatch between the version installed and the interface files, which could lead to odd errors. I expect with that kind of approach, you would need to host interface definitions for each version of MRtrix, along with the ability to determine the current version of MRtrix, and would only guarantee correct operation for official release versions. But the main reason I'm discussing this is because in the former case, I don't see a need to host these interface definitions on the main repo, while I might see a call for that with the latter approach (though I still think it would be cleaner from our point of view to have them hosted elsewhere, presumably somewhere more obviously related to the Pydra project). 

What do you reckon?

Lestropie:
It might be the case that we need to consider separately:

1.  The *implementation* of `__print_usage_pydra__`
    (which without a doubt should be implemented within the main repository, even if it may one day need to be changed in response to Pydra changes)

2.  The *storage of the outcomes* of `__print_usage_pydra__`, which would extend #2808, and potentially not require any more developer effort (eg. have a single script that generates both the interfaces and the online documentation), and CI / code review would potentially highlight CLI changes not exposed by help page changes.

3.  The *additional content* currently present in this PR.
    Current discussion with @jdtournier may have overlooked the fact that there's a lot of additional content here over and above the direct outputs from `__print_usage_pydra__`. Which if your concern is developer maintenance burden, this would exacerbate that quite a lot.

I wondered about an intermediate solution where we do 1 and 2 but not 3. However I don't think that that solves @jdtournier's concerns, and it might not make @tclose's handling any cleaner.

cmd/amp2response.cpp
```diff
@@ -52,7 +52,7 @@ void usage() {
   ARGUMENTS
   +Argument("amps", "the amplitudes image").type_image_in() +
       Argument("mask", "the mask containing the voxels from which to estimate the response function").type_image_in() +
-      Argument("directions", "a 4D image containing the estimated fibre directions").type_image_in() +
+      Argument("directions_image", "a 4D image containing the estimated fibre directions").type_image_in() +
```
tclose, line 55: This change is to avoid a clash with one of the option names

tclose:
Ooops, I just realised that in between switching between the old build script and the cmake version I inadvertently committed my old pre-cmake build directory to the branch, which might have been confusing things. After speaking to Rob today we decided to strip everything out of this PR except to the `print_usage_pydra` special methods in favour of storing the package stub in https://github.com/nipype/pydra-mrtrix3. However, if you don't mind, I have kept in the GH action workflow which generates the interfaces and runs the tests with the PARSE_ONLY env var set.

The Pydra export is actually Python code rather than JSON so it will need to be specific to Pydra if that is ok. Also, the interface syntax is likely/hopefully going to change a fair bit in the near future if my proposal is accepted, so we can probably just leave this PR as a WIP until then (I can build and release the package on PyPI with this dev branch in the mean time) and the first cmake release.

cmd/amp2response.cpp
```diff
@@ -57,7 +57,7 @@ void usage() {
   ARGUMENTS
     + Argument ("amps", "the amplitudes image").type_image_in()
     + Argument ("mask", "the mask containing the voxels from which to estimate the response function").type_image_in()
-    + Argument ("directions", "a 4D image containing the estimated fibre directions").type_image_in()
+    + Argument ("directions_image", "a 4D image containing the estimated fibre directions").type_image_in()
```
tclose, line 60: I changed this to avoid a clash with an option of the same name

Lestropie: (Hid / resolved prior post about the same issue)

Issue here is that the *argument* "`directions`" and the *command-line option* "`directions`" encode two very different pieces of information; so disambiguating by calling one an "image" may not be the optimal change. I'm hesitant to change the command-line option, as there are many commands relating to "`amp`" images that have an option "`-directions`" (even if it's not currently centralised with a `const Option` in `core/` / `src/`). Conversely, the functionality of the positional argument here also appears in other commands; so renaming it to something different only here would also be slightly unusual.

A few possibilities come to mind:

1.  Across various commands (thinking `dwi2response` in particular), where an image encoding a unit 3-vector per voxel is used, adopt "`orientations`" rather than "`directions`".

2.  While it's technically applicable to the data type in terms of command naming convention, "`peaks`" would likely not be a good choice here, since can't have multiple vectors per voxel, and the data may have been derived using some approach other than ODF peak-finding.

3.  Rename here (and preferably in other comparable instances also) to "`fibre_directions`".

4.  Create a function that returns an `Option`, creating the current "`-directions`" command-line option with a custom help string per use (I'm pretty sure from experience it's the considerable difference in these help strings that has caused there to *not* be centralisation of this command-line option). Once this works, change the command-line option name to something like "`-amp_directions`".
    (We might need to think carefully about our use of the header field `"Directions"` also...)

Lestropie:
- [ ] Add test data and tests for new interface generation capability (both C++ and Python)

Lestropie:
#3074 may have implications here. Because of the use of `argparse`, it's not possible to define "multiple types per argument", but it is possible to define custom argument types, which in turn could be some form of type union or the like.

tclose:
@Lestropie do you know why I would be getting

```
github.GithubException.GithubException: Resource not accessible by integration: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/pulls/reviews#create-a-review-for-a-pull-request", "status": "403"}
```

in the clang-tidy-review and 

```
Error: _diff_image: [ERROR] images "tmp/cfe.mif" and "fixelcfestats/default/cfe.mif" do not match within absolute precision of 9.9999999999999995e-07 (5114.81 vs 5114.81)
Error: _diff_image: [ERROR] exception thrown from one or more threads "loop threads"
CMake Error at /Users/runner/work/mrtrix3/mrtrix3/cmake/RunTest.cmake:17 (message):
  Test
  /Users/runner/work/mrtrix3/mrtrix3/testing/binaries/tests/fixelcfestats/default
  failed!
```

in the macos test?

Lestropie:
First one I've not seen before. I'm not familiar with the API permissions necessary for allowing the clang tidy review action to post to PRs. Hopefully @daljit46 can resolve.

Second one is due to #3149 not yet having been merged back to `dev`. I need to first check #3101 and merge that before dealing with the post-3.0.5 `master` changes.

cmd/amp2response.cpp
tclose: I have gone with Option 3 and changed all "directions" arguments to "fibre_directions"

tclose:
> * [ ]  Add test data and tests for new interface generation capability (both C++ and Python)

I have set up the pydra.yaml GH action to clone the [https://github.com/nipype/pydra-tasks-mrtrix3](https://github.com/nipype/pydra-tasks-mrtrix3) run the generated pytest tests. Currently these are all just marked as XFAIL but it will test whether the generated code is valid Pydra at least

tclose:
> #3074

For complicated corner cases, I think manual post-generation edits in https://github.com/nipype/pydra will end up being necessary

cmd/amp2response.cpp
Lestropie: For cross-ref:
49d4788540a8387a5eac3e69ffafddb4471c49a8
91a62f6338fa036be2fb8f745cac38bf984f0f95
I'm OK with this choice, I don't think there's any correct answer; up to any other @MRtrix3/mrtrix3-devs to interject.

tclose:
Looks like the clang-tidy-review error is a 403 permissions issue

tclose:
Just to confirm that the new `types` member of `ArgTypesFlag` represents an `or` operation

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
   return s;
 }
 
+std::string pydra_code() {
+
+  std::string CMD_PREFIXES[] = {"Fivett", "Afd",    "Amp",       "Connectome", "Dcm",   "Dir",    "Dwi",   "Fixel",
```
daljit46: Can you please use `std::array<std::string, N>` here? We're trying to move away from C-style arrays in the codebase.

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
...
+    if (!prefix.compare("5tt")) {
+      name_string = "Fivett" + name_string.substr(3, name_string.length());
+    }
+  }
+
+  std::string base_indent("    ");
+  std::string indent = base_indent + "    ";
+
+  std::string PYTHON_KEYWORDS[] = {
+      "and",    "as",     "assert", "break",   "class",    "continue", "def",       "del",   "elif",
```
daljit46, line 855: Can you please use std::array<std::string, N> here? We're trying to move away from C-style arrays in the codebase.

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
...
+    bool is_keyword = std::any_of(std::begin(PYTHON_KEYWORDS),
+                                  std::end(PYTHON_KEYWORDS),
+                                  [&id](const std::string_view kword) { return kword == id; });
+    if (is_keyword)
+      escaped += "_";
+
+    return escaped;
+  };
+
+  auto format_type = [&](const ArgTypeFlags &types, bool optional = false) {
```
daljit46: The `optional` argument is unused, so can we get rid of it?

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
...
+    if (types[ArgTypeFlags::TracksOut])
+      type += " | Tracks";
+    if (type.empty())
+      type = "Any";
+    else
+      type = type.substr(3); // drop the preceding " | "
+    return type;
+  };
+
+  auto format_option_type = [&](const Option &opt, bool for_output = false) {
```
daljit46: The`for_output` argument is unused, so can we get rid of it?

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
...
+    std::string f;
+    bool is_multi = (opt.flags.allow_multiple()) && (!opt.size() || !opt[0].types[ArgTypeFlags::FileOut]);
+    if (is_multi) {
+      f += "MultiInputObj[";
+    }
+    if (!opt.size()) {
```
daljit46: Here, I think it'd be better to use `if(opt.empty())` for better readability.

cpp/core/app.cpp
```diff
@@ -782,6 +782,403 @@ std::string restructured_text_usage() {
...
+  // Create actual class
+  s += "\n\n@shell.define\nclass " + name_string + "(shell.Task[\"" + name_string + ".Outputs\"]):\n";
+  s += "    \"\"\"";
+  // Add description
+  if (DESCRIPTION.size()) {
+    for (size_t i = 0; i < DESCRIPTION.size(); ++i)
+      s += base_indent + std::string(DESCRIPTION[i]) + "\n\n";
+  }
+
+  if (EXAMPLES.size()) {
```
daljit46: Here, I think it'd be better to use `if(!opt.empty())` for better readability.

cpp/core/app.cpp
Lestropie: I might be able to extend the work in #3212 to catch use of C-style arrays across the board so we're less dependent on code review to catch their re-insertion.

cpp/core/app.cpp
tclose: Thanks for the pick up

cpp/core/app.cpp
tclose: Actually, that is needed I just inadvertently dropped the functionality it was meant to trigger

tclose:
@Lestropie Is it possible to get this merged in now that I've got the tests to pass again :) ?

PR #2667, opened 2023-06-29
mrview: fix tsf custom colour

### LINKED ISSUE #2662: mrview: [ERROR] GLSL log [vertex shader]: ERROR: 0:28: Use of undeclared identifier 'colourmap_colour'
When changing the color of a tck-file to File and then requisition the colormap 'Color', I get a GLSL crash because of an undeclared identifier 'colourmap_colour':

Screenshot 2023-06-23 at 13 19 01

~~Not sure if this is specific to macOS.~~

jdtournier:
Fixes #2662


bjeurissen (approved):
Works without issues! Thanks!

Issue #2481, opened 2022-06-16
ENH labelconvert: Multiple inputs


Lestropie:
Idea for how to combine data from different parcellation sources into a single parcellation image for connectome construction. Example is something like combining default FreeSurfer cortical segmentation with a brain stem nuclei atlas.

`labelconvert` could theoretically take *multiple pairs* of `< parcellation_image, lookup_table >` as input. The output LUT would contain the indices and names of parcels that could be pulled from any one of the input images.

- Would need to make sure that there are no node names that appear in multiple input LUTs.

- If the input images are defined on different voxel grids, would need to either throw an exception, or internally do the re-gridding. For label images one would want to use whichever voxel grid is of the highest spatial resolution. But if there's any ambiguity about which image should be used as the template, then an exception would need to be thrown.

PR #2669, opened 2023-07-03
Schaefer2018 7/17 Networks LUT files

ppruc:
Adds LUT files to share/mrtrix3/labelconvert. 

LUTs can be used to extract the relevant cortical grey matter parcellations from the Schaefer2018 atlas (https://github.com/ThomasYeoLab/CBIG/tree/master/stable_projects/brain_parcellation/Schaefer2018_LocalGlobal) and relevant subcortical grey matter parcellations from the default FreeSurfer segmentation (desikan_killiany).

As of 07/2023 the left accumbens area is incorrectly labeled as "Left-Accumbens-a" instead of "Left-Accumbens-area" in the official LUTs (see https://github.com/ThomasYeoLab/CBIG/issues/49). Since labelconvert uses string recognition, I provided both strings ("Left-Accumbens-a"/"Left-Accumbens-area") and matched them with the corresponding label integer. The additional label can be omitted should the issue be resolved in the future.


Lestropie:
1.  First commit on the branch, [0fef112](https://github.com/MRtrix3/mrtrix3/pull/2669/commits/0fef112d0c5a568a8f48b8db1de3ea4260bddc5c), only modifies test data submodule pointers, which is likely an unwanted regression. Rather than doing and then undoing this change, it would likely be preferable to start the changeset from scratch.

2.  Looking around their repository a bit, I found [this directory](https://github.com/ThomasYeoLab/CBIG/tree/master/stable_projects/brain_parcellation/Schaefer2018_LocalGlobal/Parcellations/MNI/freeview_lut), which contains lookup tables with indices incrementing sequentially from 1. So it might be that the data proposed here is somewhat redundant.
    
    One difference between the two is that the files proposed here include not only cortical but also sub-cortical structures. This however assumes that one intends during connectome construction to use each discrete macroscopic sub-cortical grey matter nucleus as a single parcel. Particularly for higher resolution parcellations, this need not necessarily be the case; see for example the [Melbourne Subcortical Atlas](https://github.com/yetianmed/subcortex), which we used in the [UK Biobank connectome](https://github.com/sina-mansour/UKB-connectomics) project. For a hierarchical cortical parcellation such as Schaefer, it might make more sense to leave this up to the researcher. Long-term the goal is to allow researchers to specify their own lookup table that combines cortical and subcortical (and hypothetically multiple of such) parcellations, and feed multiple parcellation image - LUT input pairs to `labelconvert`, and have that command combine them appropriately; see #2481.

3.  While there's been external contribution of lookup tables in the past, and the line counts involved have been fine, here because of the high resolution / hierarchical / multi-label nature of the data, the ratio of line count to manual labour investment might be a little too gratuitous. I would probably be more comfortable pulling some git gymnastics to attribute the bulk of the additions to @MRtrixBot but still a non-zero contribution from @ppruc.
 
    But as per point 2, I'm not sure it needs to be debated extensively as I don't know that the duplication of external data is justified in this case.

Lestropie:
Closing in favour of #3043.

PR #2671, opened 2023-07-03
tcksift2 weights: move save operation outside of allocation try block

### LINKED ISSUE #2668: Misleading Memory Warning in tckfactor.cpp (SIFT2)
Hi everyone,

when running a standard tcksift2 command, I encountered the following warning, suggesting a problem with memory assignment: 
tcksift2: [WARNING] Unable to assign memory for output factor file: "weights.txt" not created 

I initially expected the problem to be RAM limitations, so I downsampled the FOD image, however, still encountered the same warning. It was only then that I realised the output directory was incorrectly specified (should've been "results/sub-005/weights/weights.txt"). 

Potentially it would be helpful for other users to generate a warning like the "No such file or directory" in other commands. Given the computational expense of tcksift2, a check would probably be best before running the command. 

Hope this helps!

```
tcksift2 -act subs/sub-005/5tt final.mif.gz subs/sub-005/tracks_10M.tck subs/sub-005/wmfod_norm.mif.gz results/weights/weights.txt
tcksift2: [100%] uncompressing image"subs/sub-005/sub-005_preop/wmfod_norm.mif.gz"
tcksift2: [100%] uncompressing image "subs/sub-005/sub-005 _preop/5tt_final.mif.gz" 
tcksift2: [100%] segmenting FODs tcksift2: [100%] mapping tracks to image
...
tcksift2: [done]
tcksift2: [WARNING] Unable to assign memory for output factor file: "weights.txt" not created
```

https://github.com/MRtrix3/mrtrix3/blob/32d17043ec75babbaed65b31d62c063b85ddfa24/src/dwi/tractography/SIFT2/tckfactor.cpp#L354C24-L354C24

jdtournier:
Fix #2668

Issue what that allocation and writing of weights were both contained within the same `try{}` block, and the `catch()` block produces an error message related to allocation only. Simplest fix is to move the write stage outside that `try{}` block, and rely on its own error handling, which should be clear enough here (should produce a message like `error opening file "XYZ": reason` in this case).

### DIFF
--- a/src/dwi/tractography/SIFT2/tckfactor.cpp
+++ b/src/dwi/tractography/SIFT2/tckfactor.cpp
@@ -350,10 +350,10 @@ namespace MR {
             weights[i] = (coefficients[i] == min_coeff || !std::isfinite(coefficients[i])) ?
                          0.0 :
                          std::exp (coefficients[i]);
-          save_vector (weights, path);
         } catch (...) {
           WARN ("Unable to assign memory for output factor file: \"" + Path::basename(path) + "\" not created");
         }
+        save_vector (weights, path);
       }
 
 


Lestropie:
This doesn't compile, as the `save_vector()` call outside of the `try` block is attempting to read from a variable that exists only within the `try` block.

jdtournier:
oops.

PR #2678, opened 2023-07-12
Python CLI changes

### LINKED ISSUE #1392: `run.command()`: Support for command-line piping

Lestropie:
Supersedes https://github.com/ankitasanil/mrtrix3/pull/1.

Somewhat requisite for #2665.

Closes #1392.

Many outstanding requirements as per checkbox list in #1392.


bin/dwi2response
```diff
@@ -62,7 +62,7 @@ def execute(): #pylint: disable=unused-variable
   # Sanitise some inputs, and get ready for data import
   if app.ARGS.lmax:
     try:
-      lmax = [ int(x) for x in app.ARGS.lmax.split(',') ]
+      lmax = app.ARGS.lmax
```
Lestropie, line 65: May be multiple instances of unnecessary try clauses like this because the type conversion is no longer happening

bin/dwishellmath
```diff
@@ -63,6 +65,9 @@ def execute(): #pylint: disable=unused-variable
     # make a 4D image with one volume
     app.warn('Only one unique b-value present in DWI data; command mrmath with -axis 3 option may be preferable')
     run.command('mrconvert ' + files[0] + ' ' + path.from_user(app.ARGS.output) + ' -axes 0,1,2,-1', mrconvert_keyval=path.from_user(app.ARGS.input, False), force=app.FORCE_OVERWRITE)
+  
+  if('mrtrix-tmp-' in app.ARGS.output):
+    sys.stdout.write(app.ARGS.output)
```
Lestropie, line 70: This is leftover from initial attempts to get output piping working; a more general solution will be required.

cmd/mrregister.cpp
```diff
@@ -445,8 +445,11 @@ void run () {
       case 2:
         rigid_estimator = Registration::LP;
         break;
-      default:
```
Lestropie, line 448: Will have to back-trace where this difference has arisen from and eradicate it from this PR.

core/signal_handler.cpp
```diff
@@ -135,9 +135,25 @@ namespace MR
...
-      flag.clear();
+      //ENVVAR name: MRTRIX_DELETE_TMPFILE
+      //ENVVAR This variable decides whether the temporary piped image
+      //ENVVAR should be deleted or retained for further processing.
+      //ENVVAR For example, in case of piped commands from Python API, 
+      //ENVVAR it is necessary to retain the temp files until all
+      //ENVVAR the piped commands are executed.
+      char* MRTRIX_DELETE_TMPFILE = getenv("MRTRIX_DELETE_TMPFILE");
+      if (MRTRIX_DELETE_TMPFILE != NULL) {
+        if(strcmp(MRTRIX_DELETE_TMPFILE,"no") != 0) {
```
Lestropie, line 146: This should perhaps be a more generic boolean expression evaluation

lib/mrtrix3/dwi2mask/fslbet.py
```diff
@@ -24,12 +24,12 @@ def usage(base_parser, subparsers): #pylint: disable=unused-variable
...
   parser.add_citation('Smith, S. M. Fast robust automated brain extraction. Human Brain Mapping, 2002, 17, 143-155', is_external=True)
-  parser.add_argument('input',  help='The input DWI series')
-  parser.add_argument('output', help='The output mask image')
+  parser.add_argument('input', type=app.Parser.ImageIn(), help='The input DWI series')
+  parser.add_argument('output', type=app.Parser.ImageOut(), help='The output mask image')
   options = parser.add_argument_group('Options specific to the \'fslbet\' algorithm')
   options.add_argument('-bet_f', type=float, help='Fractional intensity threshold (0->1); smaller values give larger brain outline estimates')
   options.add_argument('-bet_g', type=float, help='Vertical gradient in fractional intensity threshold (-1->1); positive values give larger brain outline at bottom, smaller at top')
-  options.add_argument('-bet_c', nargs=3, metavar='<x y z>', help='Centre-of-gravity (voxels not mm) of initial mesh surface')
+  options.add_argument('-bet_c', type=app.Parser.SequenceFloat(), nargs=3, metavar='<x y z>', help='Centre-of-gravity (voxels not mm) of initial mesh surface')
```
Lestropie, line 32: If a voxel position rather than real-space position, metavar should instead be ijk.
Also needs to be either 3 float-type arguments, or one floating-point sequence-type argument.

lib/mrtrix3/dwi2response/dhollander.py
```diff
@@ -31,10 +31,10 @@ def usage(base_parser, subparsers): #pylint: disable=unused-variable
...
   parser.add_citation('Dhollander, T.; Mito, R.; Raffelt, D. & Connelly, A. Improved white matter response function estimation for 3-tissue constrained spherical deconvolution. Proc Intl Soc Mag Reson Med, 2019, 555',
                       condition='If -wm_algo option is not used')
-  parser.add_argument('input', help='Input DWI dataset')
-  parser.add_argument('out_sfwm', help='Output single-fibre WM response function text file')
-  parser.add_argument('out_gm', help='Output GM response function text file')
-  parser.add_argument('out_csf', help='Output CSF response function text file')
+  parser.add_argument('input', type=app.Parser.ImageIn(), help='Input DWI dataset')
+  parser.add_argument('out_sfwm', type=app.Parser.FileOut(), help='Output single-fibre WM response function text file')
+  parser.add_argument('out_gm', type=app.Parser.ImageOut(), help='Output GM response function text file')
+  parser.add_argument('out_csf', type=app.Parser.ImageOut(), help='Output CSF response function text file')
```
Lestropie, line 37: Incorrect types here: Should be files, not images

core/signal_handler.cpp
```diff
@@ -135,9 +135,25 @@ namespace MR
 
     void mark_file_for_deletion (const std::string& filename)
```
Lestropie, line 136: Upstream of this, there has already been added in #2125 a `static bool` that controls whether or not piped images are deleted. It would IMO make more sense to give the ability to modulate the value of that boolean based on an environment variable, rather than having two sequential if statements in two different locations in the code controlling whether or not any given image is marked for deletion.

Lestropie: Addressed in [2894b1d](https://github.com/MRtrix3/mrtrix3/pull/2678/commits/2894b1dc392dd56aa068b359544ce199cde0422c).

Lestropie:
On disconnected pipes:

1.  The terminal error message generated when a user nominates an output image pipe, but no command is connected to the pipe (and so `stdout` is a TTY). on C++ this is currently:
    "attempt to pipe image to standard output (this will leave temporary files behind)"
    This is a bit weird because that supposed "will" *won't* happen.
    For Python I've currently made the message as of 913ce00:
    "Cannot pipe output images as no command connected to stdout"
    Whatever is decided, I suggest that the same error message should be used for both languages.

2.  Currently on C++, if an input piped image is specified but there is no connected command, the command hangs waiting for input from stdin. Does anyone have any recollection of how much we have investigated into detection of when stdin might be waiting for input from some other command vs. connected to user input? It may not be possible, just curious.

cmd/mrregister.cpp
Lestropie: Extracted and proposed in #2700; should merge that to `dev` before PR draft status is removed.

Lestropie:
Transitioned from draft to full PR to start triggering CI. Should be close to completion.

- [ ] Verify that new tests for Python CLI are being executed and tested in CI.
    Key difference here is that there's now a file in `testing/bin`, which needs to find module `mrtrix3.app`. We're potentially changing how this is achieved in build / installed software (#2741); for this testing executable perhaps a different solution is warranted. If the directory structure of the build directory is known, then likely just at the header of that file manually insert to `sys.path` arelative path from the source file. But I might look at #2834 first so as to hopefully not be deceived.
- [ ] Manually verify that appearances of interface for CLI test command are all as expected.
- [ ] Consider incorporating #2821 here.

Lestropie:
@daljit46: New file `testing/tools/python_cli` needs to find the `mrtrix3` module at runtime. This should be the contents of `lib/` within the build directory. Given how the testing now works, how would this be best handled?

1.  I could imagine adding a `PYTHONPATH=<build_dir>/lib:$PYTHONPATH` to the test call.
2.  Adding `sys.path.insert(0, <build_dir>/lib)` to the head of the source file would rely on there being a known relative path between `testing/tools/python_cli` and the build directory; so I think that can't be done?

daljit46:
> I could imagine adding a PYTHONPATH=<build_dir>/lib:$PYTHONPATH to the test call.

This seems sensible to me. 
CMake does provide a way to set an environment property per test as follows:

```
set_test_properties(my_test PROPERTIES ENVIRONMENT "PYTHONPATH=some_dir")
```
So perhaps we could add a new function argument to `add_bash_test` in `cmake/BashTests.cmake` to make this general.

Lestropie:
- [ ] Add missing tests for utilisation of piped images
- [ ] Add tests to ensure that any files can be specified via paths that include whitespace

Lestropie:
Documenting here a gotcha I encountered during the process of merge conflict resolution:

Here `argparse` returns `pathlib` objects for user-specified filesystem paths. Thesee may include whitespace characters, and will have double quotation marks inserted if written to an f-string. However, if one uses such information as an input to `os.path.join()`---for instance, to access a file within a user-specified input directory---then the result of that will be a string, which could then cause problems if invoking `run.command()` with a string rather than list-of-strings input. Currently I'm falling back to using list inputs to `run.command()` where necessary. An alternative would be to take relevant instances of `os.path.join()` and either do that operation with, or cast the result to, the custom derivative of `pathlib.Path` that quote-escapes when formatted to an f-string; which I'm not a massive fan of. I'm at least setting the precedent here to be testing operation of scripts where filesystem paths may contain whitespace.

Lestropie:
Annoying CI failures that I don't think have anything to do with this PR.

## Linux Clang

```
: && /usr/bin/clang++  -O3 -DNDEBUG   cmd/CMakeFiles/warpinit.dir/warpinit.cpp.o  -o bin/warpinit  -Wl,-rpath,/home/runner/work/mrtrix3/mrtrix3/build/src:/home/runner/work/mrtrix3/mrtrix3/build/core:  src/libmrtrix-headless.so  src/libmrtrix-exec-version-lib.a  core/libmrtrix-core.so  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libtiff.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libfftw3.so  /usr/lib/x86_64-linux-gnu/libfftw3f.so  /usr/lib/x86_64-linux-gnu/libfftw3l.so  /usr/lib/x86_64-linux-gnu/libfftw3.so  /usr/lib/x86_64-linux-gnu/libfftw3f.so  /usr/lib/x86_64-linux-gnu/libfftw3l.so  core/libmrtrix-core-version-lib.a && :
/usr/bin/ld: src/libmrtrix-headless.so: undefined reference to `std::ios_base_library_init()'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

(occurs for multiple targets)

## MSYS2

```
FAILED: src/CMakeFiles/mrtrix-gui.dir/gui/opengl/gl_core_3_3.cpp.obj 
C:\hostedtoolcache\windows\sccache\0.8.0\x64\sccache C:\msys64\ucrt64\bin\c++.exe -DMINGW_HAS_SECURE_API=1 -DMRTRIX_BASE_VERSION=\"3.0.4\" -DMRTRIX_BUILD_TYPE=\"Release\" -DMRTRIX_PNG_SUPPORT -DMRTRIX_TIFF_SUPPORT -DMRTRIX_WINDOWS -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_OPENGLWIDGETS_LIB -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DUNICODE -DWIN32 -DWIN64 -DWINVER=0x0A00 -D_ENABLE_EXTENDED_ALIGNED_STORAGE -D_UNICODE -D_WIN32_WINNT=0x0A00 -D_WIN64 -Dmrtrix_gui_EXPORTS -ID:/a/mrtrix3/mrtrix3/build/src/mrtrix-gui_autogen/include -ID:/a/mrtrix3/mrtrix3/src -ID:/a/mrtrix3/mrtrix3/core -isystem C:/msys64/ucrt64/include/eigen3 -isystem C:/msys64/ucrt64/include/qt6/QtCore -isystem C:/msys64/ucrt64/include/qt6 -isystem C:/msys64/ucrt64/share/qt6/mkspecs/win32-g++ -isystem C:/msys64/ucrt64/include/qt6/QtGui -isystem C:/msys64/ucrt64/include/qt6/QtWidgets -isystem C:/msys64/ucrt64/include/qt6/QtOpenGL -isystem C:/msys64/ucrt64/include/qt6/QtNetwork -isystem C:/msys64/ucrt64/include/qt6/QtOpenGLWidgets -O3 -DNDEBUG -std=gnu++17 -Werror -Winvalid-pch -include D:/a/mrtrix3/mrtrix3/build/src/CMakeFiles/mrtrix-gui.dir/cmake_pch.hxx -MD -MT src/CMakeFiles/mrtrix-gui.dir/gui/opengl/gl_core_3_3.cpp.obj -MF src\CMakeFiles\mrtrix-gui.dir\gui\opengl\gl_core_3_3.cpp.obj.d -o src/CMakeFiles/mrtrix-gui.dir/gui/opengl/gl_core_3_3.cpp.obj -c D:/a/mrtrix3/mrtrix3/src/gui/opengl/gl_core_3_3.cpp
D:/a/mrtrix3/mrtrix3/src/gui/opengl/gl_core_3_3.cpp:64:30: error: 'PROC' does not name a type
   64 | static int TestPointer(const PROC pTest) {
```

daljit46:
> : && /usr/bin/clang++  -O3 -DNDEBUG   cmd/CMakeFiles/warpinit.dir/warpinit.cpp.o  -o bin/warpinit  -Wl,-rpath,/home/runner/work/mrtrix3/mrtrix3/build/src:/home/runner/work/mrtrix3/mrtrix3/build/core:  src/libmrtrix-headless.so  src/libmrtrix-exec-version-lib.a  core/libmrtrix-core.so  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libtiff.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libfftw3.so  /usr/lib/x86_64-linux-gnu/libfftw3f.so  /usr/lib/x86_64-linux-gnu/libfftw3l.so  /usr/lib/x86_64-linux-gnu/libfftw3.so  /usr/lib/x86_64-linux-gnu/libfftw3f.so  /usr/lib/x86_64-linux-gnu/libfftw3l.so  core/libmrtrix-core-version-lib.a && :
/usr/bin/ld: src/libmrtrix-headless.so: undefined reference to `std::ios_base_library_init()'
clang: error: linker command failed with exit code 1 (use -v to see invocation)

After troubleshooting for this quite a while (it was really puzzling), I thought that this was probably caused by some mismatch between `libstdc++` and `clang`.  I tried to reproduce this locally in Docker without success, this directed to me think that the problem is likely with caching artefacts by `sccache`. Clearing the cache manually (using Github CLI) fixed the problem. It's likely there was some update on Ubuntu 22.04 which changed something in libstdc++, but the cached artefacts were incompatible with the change. 

For the Windows failure, I'll need to have a further look.

Lestropie:
Awesome, thanks for the effort.

PR #2681, opened 2023-07-14
FOD FMLS: fix assertion failure in retrospective assignment

fionaEyoung:
Was getting segfaults from `fod2fixel` after [2704f7e](https://github.com/MRtrix3/mrtrix3/pull/2644/commits/2704f7eb06ed0b405ba465509520510e96b1469f)

A misplaced brace in the restrospective assignments loop seems to be the issue, unless I've really misunderstood the assignment process.

### DIFF
--- a/src/dwi/fmls.cpp
+++ b/src/dwi/fmls.cpp
@@ -209,9 +209,9 @@ namespace MR {
                 lobe = l;
               }
             }
-            assert (lobe < out.size());
-            out[lobe].add (dirs, i, amplitude, (*weights)[i]);
           }
+          assert (lobe < out.size());
+          out[lobe].add (dirs, i, amplitude, (*weights)[i]);
         }
 
         for (auto i = out.begin(); i != out.end();) { // Empty increment


Lestropie:
In all of the modifications I've bee making of late, I've been back and forth a bit with the whole retrospective assignments mechanism, so it's maybe unsurprising that I've made a mistake in there. Surprising though that it would not have come up during my own testing.

When I did 84707f1079b2aefead8cace6b03a460a126645bd I probably should have added a test within a `#ifndef NDEBUG` clause to actually check for mutual exclusivity, which would have picked this up. This error looks like it will assign dixels on the bridge between multiple fixels to all such fixels---and without modulating its contribution to each correspondingly as was done back with e856b3e2ef3ff7e1af6bac32a05e3782fb15f163---which will both break mutual exclusivity of the masks and erroneously over-estimate the voxel-wise integral (so the issue is a little more than an assertion failure in that regard).

Lestropie (approved):
Thanks for the fix!

PR #2689, opened 2023-07-27
Change build system to CMake

daljit46:
This PR supersedes #2620 (which you should check out for the motivations for this proposal). It adds support for building MRtrix3 using CMake.
A major change is that we now build all the code (except for the commands) as a shared library that links to each executable. 
As far as I can tell, this is branch should be feature complete and shouldn't miss any major functionalities compared to its base branch. Instructions to build and test can be found [here](https://github.com/MRtrix3/mrtrix3/blob/cmake/cmake-instructions.md). 

What's missing?

- [ ] Addition of secondary checks in the CI to run Pylint and update copyright headers.
- [ ] Handling of the files inside the `share` directory.
- [ ] Removal of possible redundant files that are closely tied to the old build system.
- [ ] Other minor details in the CMake scripts.

All of the above would need to be discussed before this is ready for merging. 
@MRtrix3/mrtrix3-devs if there is anything needs to be added or changed or something I missed, please let me know.


daljit46:
@MRtrix3/mrtrix3-devs For the full documentation of the new build system, shall I just create a new document that details how to build the project? Or is better if I wait until we actually start using the branch in practice to obtain more feedback?
 The [cmake-instructions](https://github.com/MRtrix3/mrtrix3/blob/cmake/cmake-instructions.md) provides the basics but a more thorough explanation is probably warranted.

Furthermore, we currently have two documents (one in the user documentation directory and the other in the developer documentation) that both documents the build process with significant overlap. I propose that the we should get rid of the any documentation regarding building MRtrix3 from the user documentation and instead provide a single page where the CMake build process is documented.

daljit46:
Following our discussion last week, I have now made some changes to this PR:

- Lowered the minimum CMake version to 3.16. This is the most up-to-date version of CMake on Ubuntu 20.04 LTS. User of older distros can always download CMake’s prebuilt binaries or use `pip install cmake`. The CI checks now run using 3.16 to ensure that we don’t introduce any backwards incompatible changes.
- The README document has been update to include instructions on how to build the project using CMake. The instructions set is bare bones, the [cmake-instructions](https://github.com/MRtrix3/mrtrix3/blob/cmake/cmake-instructions.md) document contains more information, but I’m not sure if we want to retain that.
- `mrview` required custom handling of [QFileOpenEvent](https://doc.qt.io/qt-6/qfileopenevent.html), hence we needed to override `bool QApplication::event(QEvent*)`. Since linking a shared library needs to contain the definition of this function, and we need to provide two definitions, one for `shview` and the other for `mrview` which leads to an ODR violation, I opted for using an event handler using `std::function` for `mrview`. This is not the most elegant of approaches, but it works.
- The `docs/generate_user_docs.sh` script now requires a `—build-dir` argument. This is because we have now a clean separation between source and build directories.
- Added some missing checks for the CI.
- Minor adjustments to the CMake scripts.

I think this is now ready for review and to be merged.

Lestropie:
- [ ] It is intended to be possible to execute both compiled C++ binaries and Python scripts from the build location, without going through the installation process. There is however some chance that when doing so, the `run` module will not be successful in identifying the version-matched binary *MRtrix3* commands and ensuring that those are invoked rather than relying on the shell to choose the binary based on `PATH`. When running scripts both from the build location and an installed location, executing with `-debug` should show that when invoking *MRtrix3* commands, version-matched executables are being found.

- [ ] Suggest that where an optional inclusion (eg. PNG) fails to get configured at the configure stage, if there is a dedicated environment variable for that library, the warning message should include that detail.

- [ ] @Lestropie currently having issues with FSL content getting erroneously included (first invoking their compiler rather than the system one, then ghosting between their included libraries and the system ones).
    - [ ] If possible, cmake should check that compiler is not related to Conda in any way
    - [ ] Document use of `CMAKE_CXX_COMPILER` when necessary
    - [ ] Figure out why `CMAKE_IGNORE_PATH` is currently not successful in omitting FSL paths when finding shared libraries
    - [ ] Try re-installing FSL using latest version of their installer
    - [ ] If it turns out to be necessary to manually scrape FSL from `PATH` before running `cmake`, this should be documented.

Lestropie:
- [ ] Python executable scripts do not have execution privilege after installation step.

- [ ] Discuss location of Python algorithms.
    Had discussed previously but don't recall final outcome. Files at `python/lib/mrtrix3/*/*.py` don't really belong in `lib/`, not only in the repo but also following installation. Now would be the right time to change it.
    Possible options:
    -   `share/<scriptname>/*.py`: Not great, don't think this is typical content for "share".
    -   `bin/<scriptname>/*.py`, with executable Python scripts receiving the `.py` suffix: Arguably a better place for them. But would involve a change of interface and breaking of backward compatibility with external projects.
        Would require updating of documentation, and `run.command()` calls within any scripts that themselves invoke other scripts.
    -   `bin/<scriptname>_algos/*.py: Like above, but doesn't require executable renaming. But it's unconventional.
    -   Others?

daljit46:
OK, so I have fixed some of the issues that were raised:
- Symbolic links for Python files should now be created successfully. As far as I am aware, there shouldn't be any problem with the `run` module. The [code](https://github.com/MRtrix3/mrtrix3/blob/e757e07081a619bbdd36e58648dc8dbfade56467/lib/mrtrix3/__init__.py#L47) seems to be using `os.path.abspath` for locating the binary directory and to my knowledge, this doesn't resolve symlinks (one needs to use `os.path.realpath` if that is wanted).
- Installed Python scripts now use the same permissions as the source.
- There is a new option called `MRTRIX_STRIP_FSL` that is turned on by default. When enabled, this adds `$FSLDIR/bin $FSLDIR/lib $FSLDIR/include $FSLDIR/share/fsl/bin` to `CMAKE_IGNORE_PATH`. I think the reason why it didn't work in our meeting might have been that it doesn't automatically add any subdirectories (CMake 3.23 introduced [`CMAKE_IGNORE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_IGNORE_PREFIX_PATH.html#variable:CMAKE_IGNORE_PREFIX_PATH) which does that, but we can't rely on it yet).
I cannot test this atm, because I'm struggling to install FSL on my system (trying to run their script nuked my WSL installation on Windows), so @Lestropie it'd be great if you could test this on your machine.

As for the documentation, I'd be happy to add more instructions on how to use CMake, however I think we should consider this further. Most projects do not provide extensive documentation on CMake because it is fairly, and it's usually assumed that people are familiar with its basic feature set. Additionally, any documentation that we may provide will inevitably end up with duplicating CMake's official doc pages and other resources. Having said that, personally I'm a fan of comprehensive documentation. I will expand the `cmake-instructions` document, but we should probably move its contents to somewhere more appropriate.

For the location of Python algorithms, is there an issue that I can check out?

Lestropie:
-   RE FSL stripping: I get the "FSL detected in PATH, adding FSL directories to CMAKE_IGNORE_PATH to avoid conflicts" message, but I then nevertheless get "Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")", and have to manually strip FSL from `PATH`. This happens for both 3.22 and 3.27. I found the `CMAKE_IGNORE_PREFIX_PATH` possibility during meeting but noted the 3.23 requirement against my 3.22 install; so I separately installed 3.27 but `CMAKE_IGNORE_PREFIX_PATH` does not resolve either. Looking at the trace it seems that `list(APPEND CMAKE_IGNORE_PATH ${mrtrix_ignore_path_fsl} )` appears after all the damage has been done.
    [zlib_3.27.txt](https://github.com/MRtrix3/mrtrix3/files/12825596/zlib_3.27.txt)

-   RE executable version matching:
    `$ ~/src/cmake/build/bin/dwishellmath dwi_den_unr_preproc_unbiased_reg.mif mean test.mif -force -debug`
    `dwishellmath: [DEBUG] run.version_match() (from run.py:343): Version-matched executable for mrconvert: /home/unimelb.edu.au/robertes/src/cmake/build/bin/mrconvert`
    So that matches without issue. However the Python executables in build/bin/ seem to be duplicates, not softlinks? Meaning that I can edit `python/bin/dwishellmath` and there is no effect. Indeed I can edit `python/bin/dwishellmath` and then re-run `cmake --build build/` and there is still no effect.

-   RE cmake documentation: Agree generally. Not looking for extensive documentation. But what would potentially be of benefit is:
    -   Quick explanation of specifically translation of older *MRtrix3* workflow to new one
    -   Any esoterics that are specific to *MRtrix3*
    -   Solutions to common issues (eg. depending on how the FSL confict rectification works out)

daljit46:
Could you try with `-DCMAKE_IGNORE_PATH=/opt/fsl/bin;/opt/fsl/lib;/opt/fsl/include;/opt/fsl/share/fsl/bin`? This should preload the variable before any call to locate libraries and binaries. 

For the Python files, currently they're copied at configure time (but it probably makes more sense to copy the at build time). For creating symbolic links we need a more robust approach, a possible way to do this is add logic to the python module search code to distinguish between build directories and non-build directories (e.g. by checking the presence of a `CMakeCache.txt` in the parent directory). Not very elegant, but should work.

daljit46:
Regarding the FSL issue, I was finally able to reproduce this and at least on my setup, using `-DCMAKE_IGNORE_PATH` seems to avoid selecting the compiler they ship in their directories. 
Now the question is how do we want to deal with this? Do we want to prompt the user to manually do this if we detect FSL in `PATH` or do we want to automate this in the configure scripts? The latter clearly runs into problems since the compiler selection happens earlier than we can direct CMake to ignore FSL's paths. 
One thing we could do is to create a `PreLoad.cmake` file in the root directory to handle the logic for ignoring Anaconda and FSL paths, this file gets loaded before CMake looks for library and compilers so it's exactly what we need. The trouble is that this feature of CMake is [undocumented](https://gitlab.kitware.com/cmake/cmake/-/issues/17231) (you can see it in their [code](https://github.com/Kitware/CMake/blob/110b6940174018163a0ee34239ca7c6ee83d7fb4/Source/cmake.cxx#L2682) though) and CMake devs recommend to not rely on this for this reason. 
It's very unlikely that CMake devs will ever remove this feature because of their strict stand on backwards compatibility, so I think we should consider this option, even thought it's unorthodox and slightly ugly.

Lestropie:
`Preload.cmake` seems an appealing option. As long as it works, I think the absence of documentation is OK, though it would be preferable that a warning be issued at the terminal to state that it's happened, and also to be a note within our own documentation for compiling from source.

The alternative would be, if there is any kind of configuration issue, whether it's possible to perform a check on the compiler absolute path / contents of `PATH` / etc., and concatenate the error message to say that this is a potential culprit.

Lestropie:
If g++ compiler is selected at configure time, ensuring that the system version of such is used after references to FSL have been scrubbed, then the FSL version of ld nevertheless gets invoked during build time, because g++ sets up a system call to invoke ld, and that system call checks PATH for the first executable it finds called "ld". `CMAKE_LINKER` has no effect, since ld is not invoked directly. We have not found any simple mechanism by which to prevent g++ from finding the FSL version of ld:
-   `DCMAKE_CXX_FLAGS="-fuse-ld=/usr/bin/ld"` doesn't work; it seems that g++ doesn't accept absolute paths for that command-line option. 
-   No simple identified way to manipulate the contents of `PATH` specifically during invocation of compile jobs, including the fact that ideally any manipulation would be preserved between compile and build regardless of other system changes.

It may be the case that the best we can do is a warning + link to online document, and rely on devs to clean their PATH before attempting to build.

jdtournier (approved):

Lestropie:
@daljit46 Do you want to have a go at updating [this wiki page](https://community.mrtrix.org/t/testing-different-code-branches/2570) to also reflect this change?

daljit46:
Sure, I'll make a PR (not to be merged) to ease collaboration.

Issue #2679, opened 2023-07-13
Runtime test of command-line parsing


Lestropie:
Branching from question in #2665; giving its own thread as there's a few points to discuss that will diverge some distance from the code changes currently proposed there.

If I read correctly, the fundamental request is the capability to flag to a command that it should terminate at completion of command-line parsing, presumably for the purpose of verification of validity of a pipeline before commencing execution of the pipeline as a whole.

1.  In *MRtrix3*-land, that would essentially equate to aborting after [`app.Parser.parse_args()` in Python](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/lib/mrtrix3/app.py#L152) or [`MR::App::parse()` in C++](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/command.h#L100C1-L100C1).

2.  Important to recognize that this would not guarantee that the command is capable of executing; only that there was not an error that was detectable at the point of command-line parsing. For instance we do not check for the existence of input / output images, since the user could be doing something like [this](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#multi-file-numbered-image-support) and just checking the filesystem for a file corresponding to that string won't work. There would be no check that things like text files expected to contain numerical matrix data actually contain such.

3.  The proposal in #2665 of "`--dry-run`" is potentially an issue in several ways:
    1.  *MRtrix3* uses single-dash rather than double-dash for options.
    2.  For many commands there may be a reasonable interpretation of what an option called "dry run" would do that would be quite different to the proposed functionality.
    3. See point 4.2 below.

4.  Regarding how one might engage such a functionality, there are multiple possibilities:
    1.  We already have "hidden" command-line options for exporting command usages that are not included in the command help pages in any format. For functionality such as that proposed, which would only be of utility to workflow management engine developers, I think it would be reasonable that it not be shown to users.
    2.  Conversely, it seems fragile to me to use a command-line option to test parsing of command-line options. Theoretically such a command-line option could be inserted at any point; but what if its insertion were to be what broke the command-line parsing? This would also prevent differential handling of command-line parsing errors in the case where this test is being performed, as there is no guarantee that the fact that such a test is being performed will be detected.
    3.  Flagging in the "configuration" as suggested (which could perhaps be interpreted in a couple of ways) might not be the best way, given that I'd generally consider this a per-invocation toggle, whereas the point of a "configuration" is to set parameters that are common to most if not all invocations.
[...]

PR #2694, opened 2023-08-11
Quit after command-line parses env var

tclose:
This PR addresses #2679 by defining an environment variable ("MRTRIX_PARSE_ONLY"), which if set to "1", will cause all commands to terminate after successful completion of the command-line parsing with a "warning" stating that the CL args have been parsed successfully.

@Lestropie @jdtournier


tclose:
@Lestropie @jdtournier this PR is ready for review if you have a moment

tclose:
NB: Just rebased it onto dev instead of master

Lestropie (changes requested):
Switched PR target branch to match rebasing to `dev`.

1.  Currently early exit does not apply when running in an R environment. I don't know the extent to which that is used beyond the original implementer... nevertheless, given the definition and claimed implementation of the environment variable, it should probably be doing the same thing there also.
2.  Not sure if "parse only" is adequately descriptive. In the context of CLI / workflow engine, sure, but in the context of the software as a whole, maybe something more verbose like "CLI parse only"? @jdtournier?

jdtournier:
> Currently early exit does not apply when running in an R environment. I don't know the extent to which that is used beyond the original implementer... 

Not at all used as far as I know... I don't think the original implementer is even still working in the field! I've been meaning to remove support for R integration, it always felt like a bit of a hack. I'll work on that when I have a minute. But in the meantime, I really wouldn't worry a great deal about this!

Lestropie:
> Which commands are R based?

All C++ commands; see parallel functions in `core/command.h`.

tclose:
> > Which commands are R based?
> 
> All C++ commands; see parallel functions in `core/command.h`.

Ah, I misunderstood. You mean all mrtrix commands can be called from R, rather than R-based mrtrix commands. If it is an issue, we could just add a disclaimer that it won't work when calling from R, but it sounds like it will be dropped in any case.

Is it good to merge now? Did you request a change somewhere in the code @Lestropie? Because I can't see where that is if so

core/command.h
```diff
@@ -98,6 +99,18 @@ int main (int cmdline_argc, char** cmdline_argv)
     ::MR::GUI::App app (cmdline_argc, cmdline_argv);
 #endif
     ::MR::App::parse ();
+
+    //ENVVAR name: MRTRIX_CLI_PARSE_ONLY
+    //ENVVAR Set the command to parse the provided inputs and then quit
+    //ENVVAR if it is 1. This can be used in the CI of wrapping code,
+    //ENVVAR such as the automatically generated Pydra interfaces
+    char* parse_only = std::getenv("MRTRIX_CLI_PARSE_ONLY");
+    if (parse_only && ::MR::to<int>(parse_only)) {
```
Lestropie, line 108:
```suggestion
    if (parse_only && ::MR::to<bool>(parse_only)) {
```

https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/mrtrix.h#L285-L293

core/command.h
```diff
@@ -98,6 +99,18 @@ int main (int cmdline_argc, char** cmdline_argv)
...
+      WARN(
+        "Quitting after parsing command-line arguments successfully because environment variable 'MRTRIX_CLI_PARSE_ONLY' is set to '1'"
+      );
```
Lestropie, line 111:
```suggestion
      INFO(
        "Quitting after parsing command-line arguments successfully due to environment variable 'MRTRIX_CLI_PARSE_ONLY'"
      );
```

-   Warning-level is for when there is some kind of ambiguity or uncertainty regarding how to proceed, that the user needs to be alerted to, but the command is going to attempt to proceed nevertheless.

-   Environment variable value shouldn't necessarily need to be "`1`".

lib/mrtrix3/app.py
```diff
@@ -192,6 +192,15 @@ def _execute(module): #pylint: disable=unused-variable
   CMDLINE.print_citation_warning()
 
   return_code = 0
+
+  cli_parse_only = os.getenv("MRTRIX_CLI_PARSE_ONLY")
+  if cli_parse_only == "1":
+    warn(
+      "Quitting after parsing command-line arguments successfully because "
+      "environment variable 'MRTRIX_CLI_PARSE_ONLY' is set to '1'"
+    )
```
Lestropie, line 201:
```suggestion
  cli_parse_only = os.getenv('MRTRIX_CLI_PARSE_ONLY')
  if cli_parse_only:
    try:
      if cli_parse_only.lower() in ['yes', 'true'] or int(cli_parse_only):
        info(
          'Quitting after parsing command-line arguments successfully due to '
          'environment variable "MRTRIX_CLI_PARSE_ONLY"'
        )
        sys.exit(return_code)
    except ValueError:
      warn('Potentially corrupt environment variable "MRTRIX_CLI_PARSE_ONLY" '
           '= " + cli_parse_only + '; ignoring')
```

Needs testing.

-   The environment variable should be interpreted as a boolean value (ideally the *MRtrix3* interpretation of strings as bools should be functionalised somewhere for Python as it is for C++).

-   Should be INFO level rather than WARN.

pydra/src/pydra/tasks/mrtrix3/_version.py
```diff
@@ -0,0 +1,4 @@
+# file generated by setuptools_scm
+# don't change, don't track in version control
+__version__ = version = '3.0.5.dev589+g9e2806609.d20230817'
+__version_tuple__ = version_tuple = (3, 0, 5, 'dev589', 'g9e2806609.d20230817')
```
Lestropie, line 4: Are these files intended to be included in the PR? Certainly the header comment suggests not.

pydra/src/pydra/tasks/mrtrix3/v3_0/fivetissuetype2gmwmi.py
```diff
@@ -0,0 +1,161 @@
+import typing as ty
```
Lestropie, line 1: This file also looks like an accidental addition not relevant to this specific PR.

Lestropie:
Given the absence of interest in the R interface, perhaps just clarify very briefly in the description for the new environment variable that it only applies to the shell interface and not to R?

I've left a bunch of comments in the code, and you will need to resolve conflicts with latest `dev`; I'll take another look after that.

tclose:
@Lestropie Ok, I have made those changes now so I believe it should be good to re-review. Note that there is a difference in behaviour in the message printing between the c++ and python, i.e. it shows up by default in python commands but only when the -info flag is passed to c++ commands, but this seems to be the case with other "info" messages as well.

(NB: I can't see the "1 change requested", do you know what that is referring to)

Lestropie:
> it shows up by default in python commands but only when the -info flag is passed to c++ commands, but this seems to be the case with other "info" messages as well.

Uh, I suggested calling function `info()` [here](https://github.com/MRtrix3/mrtrix3/pull/2694/files/d5f3dd8e140d43493f31304ba0c3bca0bb5caa4e#r1305093728) rather than `warn()`, thinking it was an analogue to C++ `INFO()` macro, but function `app.info()` doesn't actually exist. So I'm honestly not sure what that is even calling...

My intention was for nothing to be printed to `stderr` unless `-info` was specified. I think that if the environment variable has been set, the command is being run in a setting where they don't need to read a user-friendly message to know that parsing completed successfully and the command is exiting prematurely. There's an argument to be made that perhaps this message should be `CONSOLE()` / `app.console()`-level in case a user somehow has this envvar set unexpectedly and needs to be notified of such. And that would also make it easier to get consistent behaviour between the two languages. I'm not feeling terribly strongly either way.

> I can't see the "1 change requested", do you know what that is referring to

[This comment](https://github.com/MRtrix3/mrtrix3/pull/2694#pullrequestreview-1585989384) was made as a review. It'll resolve once I re-review.

lib/mrtrix3/app.py
```diff
@@ -192,6 +192,21 @@ def _execute(module): #pylint: disable=unused-variable
...
+    try:
+      if cli_parse_only.lower() in ['yes', 'true'] or int(cli_parse_only):
+        console(
+          'Quitting after parsing command-line arguments successfully due to '
+          'environment variable "MRTRIX_CLI_PARSE_ONLY"'
+        )
+        sys.exit(return_code)
+    except ValueError:
+      warn('Potentially corrupt environment variable "MRTRIX_CLI_PARSE_ONLY" '
+           '= "' + cli_parse_only + '; ignoring')
```
Lestropie, line 207:
```suggestion
           '= "' + cli_parse_only + '"; ignoring')
```

My bad.

core/command.h
```diff
@@ -92,6 +93,18 @@ int main(int cmdline_argc, char **cmdline_argv) {
...
+
+    // ENVVAR name: MRTRIX_CLI_PARSE_ONLY
+    // ENVVAR Set the command to parse the provided inputs and then quit
+    // ENVVAR if it is set. This can be used in the CI of wrapping code,
+    // ENVVAR such as the automatically generated Pydra interfaces.
+    // ENVVAR Note that it will have no effect for R interfaces
+    char *parse_only = std::getenv("MRTRIX_CLI_PARSE_ONLY");
+    if (parse_only && ::MR::to<bool>(parse_only)) {
+      INFO("Quitting after parsing command-line arguments successfully due to environment variable "
+           "'MRTRIX_CLI_PARSE_ONLY'");
```
Lestropie, line 105:
```suggestion
      CONSOLE("Quitting after parsing command-line arguments successfully due to environment variable "
              "'MRTRIX_CLI_PARSE_ONLY'");
```

tclose:
> > it shows up by default in python commands but only when the -info flag is passed to c++ commands, but this seems to be the case with other "info" messages as well.
> 
> Uh, I suggested calling function `info()` [here](https://github.com/MRtrix3/mrtrix3/pull/2694/files/d5f3dd8e140d43493f31304ba0c3bca0bb5caa4e#r1305093728) rather than `warn()`, thinking it was an analogue to C++ `INFO()` macro, but function `app.info()` doesn't actually exist. So I'm honestly not sure what that is even calling...
> 
> My intention was for nothing to be printed to `stderr` unless `-info` was specified. I think that if the environment variable has been set, the command is being run in a setting where they don't need to read a user-friendly message to know that parsing completed successfully and the command is exiting prematurely.

Maybe the code could be refactored to use Python's std logger, which would give you the behaviour you are after, but that seems like a fair bit of work for not much gain.

> There's an argument to be made that perhaps this message should be `CONSOLE()` / `app.console()`-level in case a user somehow has this envvar set unexpectedly and needs to be notified of such. And that would also make it easier to get consistent behaviour between the two languages. I'm not feeling terribly strongly either way.

This was my thinking with the warning, but in either case it probably isn't worth worrying about too much

> > I can't see the "1 change requested", do you know what that is referring to
> 
> [This comment](https://github.com/MRtrix3/mrtrix3/pull/2694#pullrequestreview-1585989384) was made as a review. It'll resolve once I re-review.

Lestropie (approved):
Only thing that still catches my eye is that if the contents of the envvar can't be converted to bool, the exception won't be handled neatly in C++. But realistically it's not an issue, this is a feature purely for developers anyway. So should be good to go from here.

Issue #2697, opened 2023-08-29
mraverageheader: Decreased mean voxel size


Lestropie:
Had observed quite some time in the past, but never got around to listing.

Consider the use case where you run `mraverageheader` with two input images. Those two images have identical voxel spacing and size, and differ only in the header affine transformation. In this usage scenario, one would expect the output of this command to be a header where the centre of the FoV is the average of the two images, the axis orientations are the averages of the best corresponding axes, and the voxel spacings should be identical to the input images. However the voxel spacings are increased in the output image for axes where there's a difference in orientation between the two input images.

Diagnosing and rectifying this without breaking something else might be a bit tricky. Don't know if it's a code problem or a math problem.

Lestropie:
One prospect here is that, rather than choosing output image voxel spacings based on the voxel spacing as projected on the input image along the average header axis (which is I think what the current code is doing), the code should instead choose for each of the three output header axes the input axis that most closely corresponds, and include the input image voxel spacing along that input image axis in the relevant calculation (ie. whether taking the mean or the minimum across images for that output image axis). Any thoughts?

Lestropie:
Had a thought that there's potentially additional ways to determine voxel sizes in the average header. Not sure yet if any come out exactly the same as the pre-existing "projection" calculation or one another.

1.  Each axis in the average header image has an absolute cosine distance to each of the three axes of any given input image. In the "nearest" mechanisms implemented in #2702, the maximal of these values is found, and that input image axis contributes to the aggregate measure for that average header axis across input images. Instead of just choosing the nearest input image axis, it could instead take a weighted average of the three input voxel spacings, with weights determined by those cosine distances.

2.  The three input image axes and corresponding spacings define an ellipsoid. The spacing for a givein average header axis is determined by the amplitude of the ellipsoid along that direction.

Issue #1741, opened 2019-09-25
dwicat: Potential enhancements


Lestropie:
1. Would prefer to scale all inputs to a common central intensity, rather than using the first input as a reference.

1. Currently the script *assumes* that the only data common to each input DWI series is the *b*=0 data, and therefore utilises only that data in the determination of scaling factors between acquisitions. But if two series contain data from the same *b*-value, then theoretically those data could additionally be used to inform determination of the scaling factor.

1. Rather than just calculating a single scaling factor per input series, the appropriate scaling factor for each image volume couuld be modeled as a continuously varying function, with discontinuities permitted between separate input series. In this case the command would be more appropriately named `dwidedrift`, and the ability to provide multiple inputs and have them concatenated (in addition to being scaled appropriately` would be a secondary defining attribute.

1. Eventually, no reason why the SHARD basis could not be used for generating the image predictions for scaling factor estimation.

Lestropie:
5. If you wanted to get *really* intense, you could theoretically look for temporally-smooth intensity fluctuations on a per-slice-stack basis... Probably overkill through.

Lestropie:
As mentioned in https://github.com/bids-apps/MRtrix3_connectome/issues/123, one issue that can occur when concatenating multiple DWI series is that the voxel grids on which the data are defined may not be precisely identical. Currently in that App this results in immediate abortion of execution (though it is not currently using `dwicat`).

Something for which it would be reasonable to defer to `dwicat` would be to identify when voxel grid mismatches occur, and when necessary, generate the average header, and transform the DWI data onto that header (ideally without resampling if it is only a rigid body difference), rotating gradients accordingly.

@dchristiaens Is there any interaction here with your recon software workflow? Or does that need to bypass this command entirely by necessity? I could imagine for instance that it could be an issue if modelling subject motion as temporally locally smooth. Indeed I suppose that something like EddyQC might report huge motion that is actually induced by the processing forcing everything onto a common grid without resampling...

Lestropie:
Requirements for supporting data on different voxel grids:

- [ ] Analyse set of input headers, and assign input data to one of three classifications:
    -   *Equal*: Input voxel grids are identical; basic concatenation operations can be performed throughout.
    -   *Rigid-body*: Voxel spacings and axis sizes are identical, but header transformations are not.
    -   *Inequal*: Data will need to be interpolated onto the final output voxel grid.

[...]

PR #2702, opened 2023-08-31
dwicat: Support different voxel grids

Lestropie:
Follows discussion in #1741.

Relates to question in #2697. Got around the uncertainty there by essentially preserving the current behaviour and implementing alternatives alongside them.

Needed to implement this to facilitate processing of a minority of session data coming through the lab, where series with opposing phase encoding directions aren't always on precisely the same voxel grid, but voxel spacings and axis dimensions are identical.

Plan is to use this to resolve https://github.com/bids-apps/MRtrix3_connectome/issues/123.

1203d745dcc19be811e28b0a4f53fa16566204f0 has an extensive description of changes.

I can't say that for DWI series with different voxel sizes this is 100% the best way to proceed, but I can at least say that it *would* proceed.


Lestropie:
Something I failed to previously report explicitly here:

Because the default behaviour of `mraverageheader` is being altered (mean of the voxel spacings of the nearest axes, rather than the mean projected spacing), the behaviour of `population_template` will also change in the scenario where the desired voxel spacing of the template image is *not* specified.
I have a suspicion that many have been explicitly specifying their desired voxel size specifically *because* `mraverageheader` was producing template images with voxel spacings not quite identical to their input data. Therefore I would view this as a beneficial change, even if it wasn't the original target. It is nevertheless a change that needs to be reported.

Lestropie:
- [ ] Add test with input data on different voxel grids that causes `dwicat` to fail, but succeeds with PR changes.
- [ ] Investigate determination of centre of FoV
    (A quick test just now suggests that the centre of the output FoV is dominated by the location of the first image, rather than the mean of all images)

Lestropie:
There was a trap here that is resolved in 9be53371faa52f1019813d0b58e3a809cea150a7.

Before that, I had altered the default behaviour of average header calculation, such that the outcomes of `mraverageheader` would change in both `dwicat` and `population_template`. However the same code is also used by `mrregister`, and so that change resulted in a lot of `mrregister` tests failing. In trying to finally tie this one off, I have reverted the default behaviour of that average header calculation, which will be reflected in both `mrregister` operation and `mraverageheader` default behaviour; `dwicat` and `population_template` instead explicitly request the mechanism that is IMO more appropriate in those scenarios.

There's an argument to be made about the generation of unbiased midway spaces in `mrregister` using this new spacing calculation rather than the legacy one. But I'm at the point now where I'd prefer to defer that. I don't have the background in registration to be making an informed call. The changes for `dwicat` and `population_template` are out of necessity in the former case, and to produce user-intuitive outcomes in the latter case (ie. generating a template from all 2mm isotropic input images should produce a 2mm isotropic template).

Lestropie:
Both test data repositories have been updated such that the `dev` branch points to the same SHA1 as the `dev` branch of this repository now points to, and the corresponding feature branches in the test data repositories have been deleted.

Issue #2703, opened 2023-09-01
CI division following clang-format addition


Lestropie:
Follows #2652 & #2624.

Previously, CI actions fell into two categories: "compilation / testing", where there's one per OS / compiler, and "secondary checks", which are bundled into a single Action.

Noted on looking at PRs to `dev` following merge of #2652 that `clang-format` wasn't bundled with the other "secondary checks", it's its own Action.

This might make sense from a division of computation perspective (which we've discussed extensively some time in the past), but maybe isn't the most intuitive when looking at the results of CI testing, since there's now "a secondary check" that isn't bundled in with the other "secondary checks".

Would it make more sense downstream of these two PRs to now separate out the "secondary checks" into multiple Actions?

PR #2706, opened 2023-09-01
Use faster CI action for clang-format checking

daljit46:
In #2652, we introduced clang-format to enforce that all of our C++ code is formatted consistently across the entire codebase. To enforce this, a separate CI job was needed that uses [this](https://github.com/jidicula/clang-format-action) GitHub Action to check the formatting of all C++ files when a pull request is made. On average, this takes about 6-7 minutes to run, which is acceptable but still quite slow.
This PR introduces the use of a new GitHub Action (see [here](https://github.com/DoozyX/clang-format-lint-action)), that does exactly the same thing but much faster (on average it takes about 30 seconds). 

Additionally, this should also address #2703, since the clang-format check now is part of our "secondary checks", as a separate job (which previously was added to parallelise the format checking) for clang-format is no longer needed.


Lestropie:
Is it known why the `DoozyX` action is so much faster than the `jidicula` one? I can't help noticing the word "lint" in there and wondering if it's somehow doing something less exhaustive.

Otherwise, seems good. Though question in #2703 about whether to split all of these tasks across different Actions still stands. Might be nicer for identifying causes of CI test failures if each is its own Action; but presumably it's more overhead for spinning each one up.

daljit46:
I think the reason the `jidicula` action is slower is firstly because it effectively runs clang-format on each file, but without any kind of parallelism (see [here](https://github.com/jidicula/clang-format-action/blob/ecfca62efc4fd3e8265b8cf4061cd6ff3847d3a7/check.sh#L81). For each file, it invokes docker in interactive mode (mounting the current directory). This is rather slow and thus results in long processing time.
The `DoozyX` action instead processes files by spawning `clang-format` processes in parallel. This also the approach I took in the [clang-format-all](https://github.com/MRtrix3/mrtrix3/blob/dev/clang-format-all), which runs at a similar speed.

Lestropie (approved):
OK awesome, was hoping it was just an efficiency gain.

PR #2713, opened 2023-09-11
PNG: Fix check for too few axes, memory, orientations

Lestropie:
Fixes erroneous check put in place in #2535 (3acf6cc).

@jdtournier #2535 looks wrong just from the code. Look at the check vs. the error message. This I expect to still catch the erroneous use case before segfaulting, but no longer block normal operation as seems to be happening with 3.0.4. Would ideally have some unit tests...

### DIFF
--- a/core/image_io/png.cpp
+++ b/core/image_io/png.cpp
@@ -30,14 +30,13 @@ namespace MR
     void PNG::load (const Header& header, size_t)
     {
       DEBUG (std::string("loading PNG image") + (files.size() > 1 ? "s" : "") + " \"" + header.name() + "\"");
+      addresses.resize (1);
+      segsize = (header.datatype().bits() * voxel_count (header) + 7) / 8;
+      addresses[0].reset (new uint8_t[segsize]);
       if (is_new) {
-        segsize = (header.datatype().bits() * voxel_count (header) + 7) / 8;
-        addresses.resize (1);
-        addresses[0].reset (new uint8_t[segsize]);
         memset (addresses[0].get(), 0x00, segsize);
       } else {
-        segsize = (header.datatype().bits() * header.size(0) * header.size(1) * (header.ndim() == 4 ? header.size(3) : 1) + 7) / 8;
-        addresses.resize (files.size());
+        const size_t slice_bytes = (header.datatype().bits() * header.size(0) * header.size(1) * (header.ndim() == 4 ? header.size(3) : 1) + 7) / 8;
         for (size_t i = 0; i != files.size(); ++i) {
           File::PNG::Reader png (files[i].name);
           if (png.get_width() != header.size(0) ||
@@ -49,8 +48,7 @@ namespace MR
             e.push_back ("File \"" + files[i].name + ": " + str(png.get_width()) + "x" + str(png.get_height()) + " x " + str(png.get_bitdepth()) + "(->" + str(png.get_output_bitdepth()) + ") bits, " + str(png.get_channels()) + " channels");
             throw e;
           }
-          addresses[i].reset (new uint8_t[segsize]);
-          png.load (addresses[i].get());
+          png.load (addresses[0].get() + (i * slice_bytes));
         }
       }
     }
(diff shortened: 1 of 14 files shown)


Lestropie:
While I believe the fix in de6d0c37a339799240c9f3c8543925023eefb6f2 to be correct, couple of other things came up when running the relevant code in debugging mode.

- [ ] `PNG::load()` seems to be allocating way more memory than is actually necessary (which lead to memory allocation fault in one instance). It's taking a product of `voxel_count (header)` and `files.size()`. It should be either enough memory for the image, or enough memory for one slice times the number of slices.

- [ ] Better would be for `PNG::load()` to use a separate address per input file.

Lestropie:
I ended up expanding the scope of this PR, as I discovered multiple other issues in the process of diagnosing and rectifying the original observation.

This PR includes adding [new test data](https://github.com/MRtrix3/test_data/compare/e5646b6b5a5311df287610c1b0ea4e13388d6740...68869a7fe0780c7b1e9577ba68466d5b407a3fe8), which demonstrates the capability to generate PNG images in different slice directions with the correct orientation (proving rectification of #2179), and new `mrconvert` tests to make sure that these are generated and loaded correctly. These tests have also been confirmed to pass with using address sanitiser.

jdtournier (approved):
Seems all good. Hit a slight snag when writing to a non-existent folder (segfault when I was expecting a more informative error message), but that last commit seems to sort that out. Otherwise, things seem to work with the very limited testing I've done on my system. But your battery of tests seems way more comprehensive, so I expect that's plenty enough validation to merge. :+1:

Lestropie:
I am unable to merge this PR: 

> New changes require approval from someone other than jdtournier because they were the last pusher.

I cannot re-review it myself, even after removing myself as assignee. So presumably it is checking for first commit / most commits.

While I have the ability to use admin rights to bypass branch protections, we should maybe think about conforming to what's being semi-enforced as best practise.

Lestropie:
I *really* hope this finally addresses all issues now.

You can look at files `testing/binaries/data/mrconvert/png*` and see that they are all sensible and of the correct orientation. The updated tests then make sure that:
1.  The generated PNGs match those pre-calculated;
2.  Reading the pre-calculated PNGs and comparing to `.mif`-based data (where the voxel size and transform have been set to default, but the orientations are otherwise correct) yields no differences.

Lestropie:
Seeking review on this as I want to subsequently merge back to `dev` and continue on #2806.
Notably, in #2792, where I added a new test for single input image support:
`mrcat dwi.mif - | testing_diff_image - dwi.mif`
, I immediately noticed that the piped image has different strides, which should not need to be the case.

core/image_io/png.cpp
```diff
@@ -59,19 +56,15 @@ namespace MR
...
+      if (writable) {
+        const size_t slice_bytes = (header.datatype().bits() * header.size(0) * header.size(1) * (header.ndim() == 4 ? header.size(3) : 1) + 7) / 8;
+        for (size_t i = 0; i != files.size(); i++) {
+          File::PNG::Writer png (header, files[i].name);
+          png.save (addresses[0].get() + (i * slice_bytes));
         }
-        DEBUG ("deleting buffer for PNG image \"" + header.name() + "\"...");
-        addresses[0].release();
       }
+      delete[] addresses[0].release();
```
Lestropie, line 67: This is necessary because `MR::ImageIO::Base::addresses` is of type `vector<std::unique_ptr<uint8_t[]>>. However there are other classes in `MR::ImageIO` that do *not* invoke `delete[]`. My suspicion is that in this PR I caught a memory leak when diagnosing other PNG issues, but this is actually a more general memory leak / issue across may image formats:

-   `Default` just calls `addresses[n].release()`
-   `GZ` does nothing
-   `Mosaic` does nothing
-   `Pipe` just calls `addresses[0].release()`
-   `RAM` calls `delete [] addresses[0]`
-   `Scratch` calls `addresses[0].reset()`
-   `TIFF` calls `addresses[0].release()`
-   `VariableScaling` does nothing

jdtournier: I think it's enough for that line to just disappear - the `addresses` vector will be cleared immediately afterwards anyway, and the default destructor for `std::unique_ptr<uint8_t[]>` will free the associated memory anyway.

jdtournier (approved):
Had a look through the code and all the tests, and everything looks good. :+1:

PR #2722, opened 2023-09-25
dwifslpreproc: Fix for PE direction, but not readout time, in header

Lestropie:
Error encountered by user attempting to process their own data during workshop. 

Description in faf0788a21c80498d431ed017661c0c179de2437.


Lestropie:
Fixed & added test as proof. Seeking approval review.

daljit46 (approved):

PR #2723, opened 2023-09-27
Barzilai-Borwein gradient descent: Step size options

Lestropie:
Hoping to draw comment from @maxpietsch.

Been looking at [Barzilai-Borwein gradient descent](https://en.wikipedia.org/wiki/Barzilai-Borwein_method) as a possible numerical solver for a problem I'm currently facing. This is already in use for the purpose of affine image registration.

On looking at the code I immediately spotted a difference between the [current implementation](https://github.com/MRtrix3/mrtrix3/blame/3.0.4/core/math/gradient_descent_bb.h#L228) for determining step size:
`dt = (x2-x1).norm()/(g2-g1).norm();`
and the [Wikipedia page](https://en.wikipedia.org/wiki/Barzilai-Borwein_method):
BBGD_step_size

What I'm not sure about is whether this is an outright bug in implementation, or if there's something else going on. I'll try to at some point run rigid-body registration with the verbosity increased and see if the alternative expressions provide faster or more stable convergence. But external expertise would be welcome also.

### DIFF
--- a/core/math/gradient_descent_bb.h
+++ b/core/math/gradient_descent_bb.h
@@ -43,14 +43,20 @@ public:
   }
 };
 
+
+// "LEGACY" corresponds to implementation prior to 3.1.x, which may be erroneous
+enum class bbgd_step_size_t {SHORT, LONG, GEOMMEAN, LEGACY};
+
+
 //! Computes the minimum of a function using a Barzilai Borwein gradient descent approach. ENH: implement stabilised
 //! version https://arxiv.org/abs/1907.06409
 template <class Function, class UpdateFunctor = LinearUpdateBB> class GradientDescentBB {
 public:
   using value_type = typename Function::value_type;
 
-  GradientDescentBB(Function &function, UpdateFunctor update_functor = LinearUpdateBB(), bool verbose = false)
+  GradientDescentBB(Function &function, bbgd_step_size_t step_type, UpdateFunctor update_functor = LinearUpdateBB(), bool verbose = false)
       : func(function),
+        step_type (step_type),
         update_func(update_functor),
         x1(func.size()),
         x2(func.size()),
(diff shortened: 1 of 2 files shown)


bjeurissen:
To me it looks like a bug. The code appears to be using:

``dot(x_delta,x_delta)/dot(grad_delta,grad_delta)``

rather than the expected

``dot(x_delta,x_delta)/dot(x_delta,grad_delta)``  or ``dot(x_delta,grad_delta)/dot(grad_delta,grad_delta)``

core/math/gradient_descent_bb.h
```diff
@@ -230,14 +239,33 @@ template <class Function, class UpdateFunctor = LinearUpdateBB> class GradientDe
...
+  assert(dg.norm() > 0.0);
+  switch (step_type) {
+    case bbgd_step_size_t::SHORT:
+      dt = dx.dot(dg) / dg.squaredNorm();
+      break;
+    case bbgd_step_size_t::LONG:
+      dt = dx.squaredNorm() / dx.dot(dg);
+      break;
+    case bbgd_step_size_t::GEOMMEAN:
+      dt = std::sqrt(Math::pow2 (dx.dot(dg) / dg.squaredNorm()) + Math::pow2 (dx.squaredNorm() / dx.dot(dg)));
```
Lestropie, line 258:
```suggestion
      dt = std::sqrt((dx.dot(dg) / dg.squaredNorm()) * (dx.squaredNorm() / dx.dot(dg)));
```

Note that this has been assessed as being identical in magnitude to the legacy expression (or, alternatively phrased, the seemingly erroneous legacy expression may have, deliberately or inadvertently, been computing the geometric mean of the two BB expressions all along). Unlike this computation however, the legacy expression *loses the sign of the inner product*, which may contribute to divergence.

Issue #2596, opened 2023-03-07
ENH tckconvert: Support newer VTK format


Lestropie:
Newer VTK files used to store tractography data may encode, rather than a list of vertex indices for each streamline, a set of offsets to the first (/ last?) index of each streamline. Currently the code residing within `tckconvert` only supports the former for both read and write, but the latter may be generated by other DWI softwares.

Will definitely need to support if attempting #411, but might not be too hard to integrate into `tckconvert`.

Lestropie:
https://protect-au.mimecast.com/s/mIUICjZroMFjl5EBPSRYTFN?domain=discourse.vtk.org

https://protect-au.mimecast.com/s/xVX1Ck8vpKsnYWlK1hQwwVb?domain=discourse.vtk.org

PR #2724, opened 2023-09-27
tckconvert: Support read from offsets-based VTK

Lestropie:
Newer VTK format for storing streamlines data that reduces the storage size. Instead of a list of vertex indices per streamline, the file instead stores the index of just the first vertex in each streamline.

First reported in #2596. For now I have only implemented a reader function.

@ppruc: Hopefully this gets you by; let me know.

### DIFF
--- a/cmd/tckconvert.cpp
+++ b/cmd/tckconvert.cpp
@@ -190,23 +190,25 @@ private:
   size_t current_index = 0;
 };
 
-template <class T> void loadLines(vector<int64_t> &lines, std::ifstream &input, int number_of_line_indices) {
-  vector<T> buffer(number_of_line_indices);
-  input.read((char *)&buffer[0], number_of_line_indices * sizeof(T));
-  lines.resize(number_of_line_indices);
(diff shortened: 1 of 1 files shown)


ppruc:
Hi Rob, 

can confirm that the command runs now without error when converting vtk to tck, however, the output file does not contain anystreamlines unfortunately!

Philip

ppruc:
Hi Rob, 

now tckinfo shows that the file contains the expected number of streamlines. However, when trying to load it up in mrview, they are not visible. Also they are not assigned to parcels when passing them to tck2connectome. 

When using tckmap to produce an image this is the result: 
image

Hope this helps!

Best,
Philip

Lestropie:
Clearly the streamlines are there in the file somewhere, as otherwise `tckmap` wouldn't be able to produce a result. My best guess is that the vertices have been defined with respect to image space rather than scanner space, such that they don't have spatial correspondence with the corresponding image data when MRtrix3 interprets them as being defined in scanner space. If you load an image and your tractogram, disable "crop to slice" in the Tractography tool, and zoom out, you'll probably find them. If I'm correct, `tckconvert` has command-line options to do the conversion.

PR #2725, opened 2023-10-04
Fix online documentation

Lestropie:
The ReadTheDocs online documentation is currently not displaying correctly.

[Here it is](https://mrtrix.readthedocs.io/en/3.0.4/) at tag `3.0.4`, appearing as intended:
Screenshot from 2023-10-04 13-27-14

[Here](https://mrtrix.readthedocs.io/en/latest/) is the current `latest`:
Screenshot from 2023-10-04 13-25-37

Now I found [here](https://blog.readthedocs.com/use-build-os-config/) instructions for config modification that needs to be done by October 16; so I did that, but that [produces](https://mrtrix.readthedocs.io/en/readthedocs_build_os/) a different incorrect answer: 
Screenshot from 2023-10-04 13-26-21

So this PR does not resolve the issue, but probably the fix for such should be build atop it.

I don't have experience with ReadTheDocs itself, so hoping somebody might recognise the issue from some precedent.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.7"
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
@@ -20,6 +25,5 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-   version: 3.7
    install:
       - requirements: docs/requirements.txt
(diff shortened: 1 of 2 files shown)


Lestropie:
As a sanity check, I took the changes on this branch and applied them to `3.0.4`, and the result is the same as the third figure above. That is to say, the regression in documentation appearance has nothing to do with changes in the *MRtrix3* repository between `3.0.4` and `master`.

Lestropie:
Superseded by #2727.

jdtournier:
Oops, sorry I didn't spot you'd already opened this PR...

PR #2727, opened 2023-10-04
ReadTheDocs: updates to fix render

jdtournier:
Resorted to re-generating a default Sphinx config, and adding back the bits we specifically needed. Seems to fix the render issues we were having. I'm not sure what caused the problem exactly, but the config files are now considerably simpler, which is no bad thing... 

See rendered version of this branch on ReadThedocs: 
https://mrtrix.readthedocs.io/en/fix_readthedocs_render/


Lestropie (approved):
Looks to be rendering correctly.

What's not evaluated here is compilation of non-HTML versions of the documentation. We'd previously discussed bundling these with pre-compiled binary downloads. But not sure that this capability is currently in use anywhere. Would rather fix the online docs now, and if there's issues elsewhere that can be raised standalone.

Issue #2731, opened 2023-10-22
Python imp module is now deprecated


jdtournier:
According to the [documentation for the `imp` module](https://docs.python.org/3.11/library/imp.html), it is now deprecated:

>Deprecated since version 3.4, will be removed in version 3.12: The [imp](https://docs.python.org/3.11/library/imp.html#module-imp) module is deprecated in favor of [importlib](https://docs.python.org/3.11/library/importlib.html#module-importlib).

This is used in the `mrtrix3.py` module to locate and load the main `mrtrix3` module. Already [causing trouble on the forum](https://community.mrtrix.org/t/batman-tutorial-step-2-4-zsh-command-not-found-dwipreproc/7035/5?u=jdtournier). We need to fix this asap...

Lestropie:
[Looks](https://docs.python.org/2.7/library/importlib.html) to me like `importlib` does not have the requisite functionality in Python 2, which is still supported on `master`. So need to choose between:

-   Embed logic in `bin/mrtrix3.py` to use `importlib` in Python3 and `imp` in Python2
-   Drop Python2 support on `master` and tag `3.0.5`.

Can change separately on `dev` since Python2 support has already been dropped there (#2215 pending check in #2534).

jdtournier:
I reckon we'll need to support both versions for some time. I've had a crack at doing this in #2735, and it's too ugly. I would vastly prefer it if they didn't keep changing things, but there we are...

I'll close this for now, we can discuss on #2735

PR #2735, opened 2023-10-24
add support for python importlib to locate MRtrix3 python libraries

jdtournier:
First attempt at addressing #2731...

Seems to work from my limited testing. Main idea is to use `importlib` (following [suggestions from StackOverflow](https://stackoverflow.com/a/50395128)) as long as Python version is 3.5 or newer and `importlib` module can be loaded successfully. Otherwise fall back to the previous `imp` approach. Seems to work on my system using either Python 3.10 (using `importlib`) or Python 2.7 (using `imp`).

### DIFF
--- a/bin/mrtrix3.py
+++ b/bin/mrtrix3.py
@@ -14,22 +14,48 @@
 #
 # For more details, see http://www.mrtrix.org/.
 
-import imp, os, sys
+import os, sys
 from distutils.spawn import find_executable
 
-def imported(lib_path):
-  success = False
-  fp = None
+try:
+  # since importlib code below only works on Python 3.5+
+  # https://stackoverflow.com/a/50395128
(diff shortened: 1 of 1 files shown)


Lestropie (approved):
Confirmed working for me on both Python 2 and 3 also.

Lestropie:
Agree with comment in #2735 that it's ugly. But as long as it works I'm okay with it. With 3.1.0 this will reduce in two ways:
1.  Python2 support is revoked (#2215), so can use `importlib` across the board
2.  Function `imported()` will either disappear or have reduced utilisations. That's there to handle potential discovery of the API through multiple possible mechanisms, at least one of which (looking at the `build` softlink) will disappear.
    (PR for proposing Python changes following `cmake` adoption is pending)

PR #2739, opened 2023-10-25
Support for macOS app bundles for GUI apps

bjeurissen:
This can absorb some of the logic around this currently found in https://github.com/MRtrix3/macos-installer/ and/or https://github.com/MRtrix3/mrtrix3/tree/app_bundles


bjeurissen:
In the current binary installer for macos, we add wrapper scripts in the bin folder that point to the GUI apps (mrview and shview) inside their app bundle folders, e.g. https://github.com/MRtrix3/macos-installer/blob/master/mrview, to ensure seamless operation of mrview and shview from both the macOS command-line and macOS GUI.

We probably have to do something similar in CMake or come up with a more elegant solution.

cmd/CMakeLists.txt
```diff
@@ -14,11 +14,13 @@ function(add_cmd CMD_SRC IS_GUI)
         $<IF:$<BOOL:${IS_GUI}>,mrtrix::gui,mrtrix::headless>
         mrtrix::exec-version-lib
     )
-    set_target_properties(${CMD_NAME} PROPERTIES 
+    set_target_properties(${CMD_NAME} PROPERTIES
```
daljit46, line 17: We probably should version information as well here. Something like `MACOSX_BUNDLE_BUNDLE_VERSION ${PROJECT_VERSION}` should work. Additionally, we may want to add `MACOSX_BUNDLE_GUI_IDENTIFIER` too (which should be done automatically only after Qt 6.1).

cmd/CMakeLists.txt
```diff
@@ -14,11 +14,13 @@ function(add_cmd CMD_SRC IS_GUI)
...
+        MACOSX_BUNDLE ${IS_GUI}
+        MACOSX_BUNDLE_ICON_FILE mrview.icns
         LINK_DEPENDS_NO_SHARED true
         RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin
     )
-    install(TARGETS ${CMD_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+    install(TARGETS ${CMD_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} BUNDLE DESTINATION ${CMAKE_INSTALL_BINDIR})
```
daljit46, line 23: Can we indent this on multiple lines for better readability?

bjeurissen:
I have cleaned up the generation of and completed the contents of the plist files by using custom Info.plist template support provided by CMake (this is the only way one can generate plist files for app bundles to liking).

What is currently still missing is:

- [ ] copying the icons in icons/macos into the appropriate Resource folders of the application bundles
- [ ] copying the wrappers that point mrview to mrview.app/Contents/MacOS/mrview (e.g. https://github.com/MRtrix3/macos-installer/blob/master/mrview). Note that this can't be a simple symlink, but needs to be a wrapper script, otherwise CLI and GUI integration breaks.

I don't know CMake well enough to do this in the appropriate way...

cmd/CMakeLists.txt
bjeurissen: Done

cmd/CMakeLists.txt
bjeurissen: Done

daljit46:
Unfortunately, CMake's documentation is rather lacking on this front. For GUI use cases, I always recommend checking [Qt's documentation](https://doc.qt.io/qt-6/appicon.html), which is often excellent.

For setting the icon, something like this should work:

```
    set(mrtrix_icon_macos "mrtrix.icns")
    set_source_files_properties(${mrtrix_icon_macos} PROPERTIES 
        MACOSX_PACKAGE_LOCATION "Resources"
    )
    target_sources(mrview PRIVATE ${mrtrix_icon_macos})

    set_target_properties(mrview PROPERTIES
        MACOSX_BUNDLE_ICON_FILE "mrtrix.icns"
    )
```

bjeurissen:
Noticed that while the app bundles work fine in the build directory, they break when installed to the installation directory. Cmake correctly replaces the absolute library paths to relative ones on installation, but appears to apply the exact same logic to both the regular binaries and the ones in a bundle.

As a result, in the current form we need to run:

```
install_name_tool -rpath @loader_path/../lib @loader_path/../../../../lib ~/mrtrix3_installation/bin/mrview.app/Contents/MacOS/mrview
install_name_tool -rpath @loader_path/../lib @loader_path/../../../../lib ~/mrtrix3_installation/bin/shview.app/Contents/MacOS/shview
```

daljit46:
OK, so I have had a look at this in more detail. I think it'd be good to move `cmake/bundle` to `packaging/macos/bundle`. Also, to handle the logic in `cmd/CMakeLists.txt`, it'd be sensible to create a new CMake module `cmake/MacOSBundle.cmake` which provides some helper functions to achieve what we need. 

To solve the issue with the `RPATH` for the installation directory, it should be sufficient to set the `INSTALL_RPATH` property to something like `@executable_path/../../../../lib` instead of relying on the global `CMAKE_INSTALL_RPATH` which is by default set to `@loader_path/../lib`.

bjeurissen:
Currently nothing seems to trigger the creation of the app bundle (at least not on my machine).

daljit46:
> Currently nothing seems to trigger the creation of the app bundle (at least not on my machine).

It should work now.

bjeurissen:
> > Currently nothing seems to trigger the creation of the app bundle (at least not on my machine).
> 
> It should work now.

It does! Thanks!

A few outstanding things that need to be fixed:
- [ ] ``<key>LSMinimumSystemVersion</key>         <string>${CMAKE_OSX_DEPLOYMENT_TARGET}</string>`` does not seem to be working, we need to either set this variable or use an alternative.
- [ ]  copying the wrappers that point mrview to mrview.app/Contents/MacOS/mrview (e.g. https://github.com/MRtrix3/macos-installer/blob/master/mrview). This can't be a simple symlink, but needs to be a wrapper script, otherwise CLI and GUI integration breaks.

daljit46:
>  When the app bundles are used, the CI binary tests for mrview and shview are skipped.

What do you mean by this? We don't have any binary tests for those commands.

bjeurissen:
> > When the app bundles are used, the CI binary tests for mrview and shview are skipped.
> 
> What do you mean by this? We don't have any binary tests for those commands.

Never mind. It was the docs generation that was failing in the CI before when the mrview/shview wrappers were not present, not the binary tests (which are indeed not present).

cmake/MacOSBundle.cmake
```diff
@@ -0,0 +1,29 @@
+function(set_bundle_properties executable_name)
+    if(${executable_name} STREQUAL "mrview")
```
bjeurissen, line 2: mrview should not only include the document-icon, but also the icon for the application itself (like shview does)

bjeurissen:
One more thing about the CLI-wrappers: they are currently only added in the "installation" phase. Is it possible to already do this in the "build" stage? So that a developer can use the build without installing?

daljit46:
> One more thing about the CLI-wrappers: they are currently only added in the "installation" phase. Is it possible to already do this in the "build" stage? So that a developer can use the build without installing?

Done.

cmake/MacOSBundle.cmake
daljit46: Done.

PR #2745, opened 2023-10-30
Integration of clang-tidy in CI

daljit46:
One of the primary aims of #2689 was to facilitate the integration of our codebase with IDEs and other third party tools. This PR adds support for integrating [clang-tidy](https://clang.llvm.org/extra/clang-tidy/), a linter tool for C++ source files, in our CI workflow. Static analysers are very useful to catch bugs and enforce stricter guidelines than your compiler's default warnings, and clang-tidy in particular is arguably the most used and well-known.
Now, we cannot enforce any particular set of guidelines in one go, as that would be unfeasible (and probably not a good idea). For this reason, this PR includes the integration of the [clang-tidy-review](https://github.com/ZedThree/clang-tidy-review) GitHub action, which creates a pull request review based on clang-tidy warnings. Note that these checks won't be a point of failure (I think this is the most reasonable thing to do since our codebase isn't clang-tidy clean).
These warnings then, hopefully, will be useful to contributors to improve the quality of their code.

One important thing to discuss would be: what checks should be included in the warnings? Clang-tidy has an extensive list of [available checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html) and allows defining a custom configuration file (see `.clang-tidy` file in this PR).  I have created a starting configuration file, which I think is reasonable, but we should discuss this further (especially on checks that are intended to improve readability).

You can run clang-tidy locally on file as follows:

```
cmake -B build
clang-tidy -p build /path/to/cppsourcefile
```
Also many IDEs (e.g. Qt Creator) comes with built-in support for clang-tidy or provide an easy way to integrate it.


daljit46:
@MRtrix3/mrtrix3-devs it'd be great if anyone could try out to run clang-tidy in your IDE and see the warnings it produces. There may some warnings that are either too noisy or superfluous. 
So unless anyone has objections, I would like to merge this is as it is, since we can always improve the clang-tidy configuration in future PRs.

PR #2746, opened 2023-10-30
Minimize superfluous dynamic library dependencies

bjeurissen:
This partly addresses the issues listed in #2744. Best looked at by someone who better understands CMake.

### DIFF
--- a/cmd/CMakeLists.txt
+++ b/cmd/CMakeLists.txt
@@ -3,7 +3,7 @@ find_package(FFTW REQUIRED)
 set(GUI_CMD_SRCS mrview.cpp shview.cpp)
 set(FFTW_CMD_SRCS mrfilter.cpp mrdegibbs.cpp)
 
-file(GLOB HEADLESS_CMD_SRCS *.cpp)
+file(GLOB HEADLESS_CMD_SRCS CONFIGURE_DEPENDS *.cpp)
 foreach(CMD ${GUI_CMD_SRCS})
     list(REMOVE_ITEM HEADLESS_CMD_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/${CMD})
 endforeach(CMD)


daljit46:
In my opinion, this is not the right approach to take to deal with #2744. When migrating to CMake, we deliberately made the effort to ensure that we don't deviate significantly from the previous code structure and build workflow. One of the key discussions was about globbing files using regex expressions, which isn't good practice in the CMake world (see [here](https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil)). The recommended approach is to be *explicit* and list all targets in CMake scripts. 
This is also the approach I would like to see to address the goals of this PR. Every command should be explicitly listed in `cmd/CMakeLists.txt` and thus we would automatically obtain the granularity of controlling the specific compilation flags for each command. If we keep relying on the globbing of files, we seriously run the risk of creating an if-else spaghetti that is hard to read and maintain.
Giving up on globbing files does mean that a developer will need to edit the CMake scripts whenever they add or remove commands, however in my opinion, this cost is trifling, especially when compared to the alternative.

As an aside, the use of `CONFIGURE_DEPENDS` is not supported on all generators, and thus we didn't make use of it, but perhaps it makes to use it everywhere in the CMake scripts since at least Ninja does.

bjeurissen:
The main point of this PR was to avoid linking binaries against the dynamic libraries that are needed by the core library but not by the binaries themselves. So all these minor changes do is prevent a cascade of external dynamic library dependencies as one move from core library to the final binaries. The way I see it, this has nothing to do with globbing. I am not talking about the decision to link fftw3 or not depending on the specific binary, that is just something that I added based on the existing logic for Qt for completeness.

I agree that the CONFIGURE_DEPENDS addition has nothing to do with the goal of the PR. I just added it because it would restore the behaviour we had in the previous build system and because it appears to be harmless (the handful of developers can use a supported generator if they want this feature, and packagers using a generator that does not support it, will not care about or be harmed by this feature).

Whether you want to keep relying on globbing, that is an entirely different discussion.

daljit46:
Perhaps I explained myself unclearly, but the point I'm trying to make is that globbing makes controlling the granularity of each command difficult and ugly (which is what we need to do this in this PR). For example, this

```
foreach(CMD ${GUI_CMD_SRCS})
    list(REMOVE_ITEM HEADLESS_CMD_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/${CMD})
endforeach(CMD)
foreach(CMD ${FFTW_CMD_SRCS})
    list(REMOVE_ITEM HEADLESS_CMD_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/${CMD})
endforeach(CMD)
```
in my opinion, is starting to look quite ugly.
If we didn't have globbing, this kind of conditional branches and lists would not be needed. So I think #2744 provides a good enough motivation to get rid of globbing.
I agree that this PR addresses a separate concern, my point is just that we can address that much better if we stop globbing files.

bjeurissen:
> Perhaps I explained myself unclearly, but the point I'm trying to make is that globbing makes controlling the granularity of each command difficult and ugly (which is what we need to do this in this PR). For example, this
> 
> ```
> foreach(CMD ${GUI_CMD_SRCS})
>     list(REMOVE_ITEM HEADLESS_CMD_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/${CMD})
> endforeach(CMD)
> foreach(CMD ${FFTW_CMD_SRCS})
>     list(REMOVE_ITEM HEADLESS_CMD_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/${CMD})
> endforeach(CMD)
> ```
> 
> in my opinion, is starting to look quite ugly.

I agree. The fundamental issue this PR was trying to address is that external libraries are currently linked to all our internal dynamic libraries as well the final binaries, when they have to be linked only to our own dynamic libraries (e.g. by changing the scope of some target_link_libraries to PRIVATE and by moving some target_link_libraries from the library to the cmd directory). As a result, a typical command only depends on lib core.

PR #2748, opened 2023-11-15
Apodized disc function

Lestropie:
Just a little tidbit I thought of as part of a collaboration project.

Replicated the same optimisation process as was used for the refined aPSF RH coefficients in the TODI manuscript; but instead of being 1.0 at [0.0, 0.0, 1.0] and 0.0 everywhere else, it is instead 1.0 at [*, *, 0.0] and 0.0 everywhere else. Produces flat discs that have minimal ringing / zero values.

It is primarily just a new class available in the API. I am however proposing one behavioural change: that this is a better choice compared to a flat disc (with severe ringing) for the first iteration of the `dwi2response tournier` algorithm. Don't know the extent of testing that people may demand to see to accept such a change?


Lestropie:
`shview` lmax=8 disc vs. aDF:

shview_disc_small
shview_adf_small

Lestropie:
RE `dwi2response tournier`:

Here's evolution of the RF across iterations with and without the proposed change.

`dev`:
```
1 -1 1
376.805270226616 -203.56229064848 45.5421871436722
359.500137671528 -216.29936852886 99.0743208764644 -33.4955757075114 9.94302754062162 -2.38438636634449
331.844538720627 -208.310031184849 104.821566205136 -39.5574643776443 12.2196311655877 -2.88128103683975
327.971397615805 -206.538127745527 105.003148698615 -40.2532944437056 12.6716479108733 -2.90993542723148
327.263069981557 -206.205653747887 104.889338353721 -40.3200504483873 12.6847679392149 -2.83404732985545
327.263071592273 -206.205443027769 104.889117154467 -40.3206038734074 12.6852861629654 -2.83443562969652
```

PR:
```
0.282095 -0.208024 0.064202
344.759065741468 -220.308803495748 49.288818310649
351.801388539209 -213.140131360375 99.9861730784289 -34.5854393524409 10.1117479036674 -2.39355618982337
331.687270647411 -208.33017614012 104.882028109163 -39.5595116573054 12.2655345875388 -2.89785207501518
327.97139773695 -206.538015070824 105.003041246285 -40.2534893024122 12.6719800508235 -2.91020244463503
327.263069983523 -206.205653439194 104.889338037535 -40.3200514194019 12.6847700924034 -2.83404849422592
327.263071592383 -206.205443020872 104.889117153052 -40.3206038849667 12.6852861869665 -2.83443562466666
```

In this test subject they converge to selection of exactly the same set of voxels.

In terms of RF magnitude, there's a strictly monotonic decrease on `dev`, whereas with initialisation using the aDF it does not. No idea if any will consider this a potential issue.

Subjectively, it seems to me that the set of single-fibre voxels selected following spherical deconvolution using the aDF is closer to the final selection than that chosen based on the current `dev` code.
Iteration 1 when using `[1 -1 1]`:
disk_iter1
Iteration 1 when using apodised disc function:
aDF_iter1
Final selection (either method):
final

So personally I'm convinced enough to apply the change to `dwi2response tournier`, but am happy to hear contrary opinions.

Lestropie:
Given absence of feedback to the contrary, and the fact that there's already supra-threshold differences in other `dwi2response` results on `dev` that will need to be hunted down in #2811, I'm going to proceed with this.

PR #2755, opened 2023-11-22
dev threading fixes

jdtournier:
Having some fun with Thread Sanitizer & clang-tidy...

### DIFF
--- a/core/algo/threaded_loop.h
+++ b/core/algo/threaded_loop.h
@@ -342,7 +342,7 @@ template <class OuterLoopType> struct ThreadedLoopRunOuter {
     struct PerThread {
       MutexProtected<Shared> &shared;
       PerThread(const PerThread &) = default;
-      PerThread(PerThread &&) = default;
+      PerThread(PerThread &&) noexcept = default;
       PerThread &operator=(const PerThread &) = delete;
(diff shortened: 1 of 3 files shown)


daljit46:
I've now introduced a `MutexProtected` class as part of this PR, following #2778. This can be used to better protect resources across threads.

core/algo/threaded_loop.h
```diff
@@ -315,114 +316,123 @@ template <class OuterLoopType> struct ThreadedLoopRunOuter {
...
+          void execute() {
+            auto pos = shared.lock()->iterator;
+            while (shared.lock()->next(pos))
+              func(pos);
+          }
+        } loop_thread = {shared, functor};
+
+        auto threads = Thread::run(Thread::multi(loop_thread), "loop threads");
+
+        __manage_progress(&shared.lock()->loop, &threads);
```
daljit46, line 358: This part I'm not sure about. It seems that we are looking to a shared resource (the `loop`) outside the scope of this file to the `ProgressBar`. What guarantees that this is safe? @jdtournier

daljit46: Ok, so after discussing this should be safe, but we should revisit the logic of `ProgressBar` in future PRs.

PR #2763, opened 2023-11-30
Add periodic checks with sanitizers

daljit46:
Sanitizers are very valuable tools to detect problems by instrumenting code for runtime bug detection. They have already proven useful for us as demonstrated in #2760, #2761 and #2755. 
This PR adds a new GitHub Actions workflow that runs our unit and binary tests compiled with sanitizers provided by [LLVM](https://github.com/google/sanitizers) each night. It'd be rather computationally expensive to run these checks on pull requests, but my tests show that they are fast enough to be run once a day (with caching enabled).

Note that this PR must be merged into `master` since scheduled GitHub Actions only run on the default or base branch.

### DIFF
--- a/.github/workflows/nightly-sanitizers.yml
+++ b/.github/workflows/nightly-sanitizers.yml
@@ -1,0 +1,62 @@
+name: Nightly checks with sanitizers
+
+on:
+  workflow_dispatch:
+  # Run this every day at midnight
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  linux-clang-build:
+    runs-on: ubuntu-latest
+
+    env:
+      SCCACHE_GHA_ENABLED: "true"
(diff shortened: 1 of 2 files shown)


daljit46:
After the discussion today, I think it may make sense to wait for this PR to be merged until most issues spotted by the sanitizers have been fixed on dev.
Also, I think it may make sense to schedule this weekly, instead of daily (although we should try to make sure that CI cache isn't invalidated between consecutive runs).

daljit46:
I think this is mostly ready to be merged. All of issues mentioned in #2772, seems to have fixed apart from #2779. I think that's still waiting a review from @jdtournier.
Once that's merged, I don't see any problem with merging to master.

daljit46:
All tests are now passing, so I think this is now mergeable.

Lestropie (approved):

PR #2764, opened 2023-11-30
Use std containers and std::make_unique/shared

daljit46:
This PR supersedes #2650 and #2651. It replaces the use of custom `MR::vector` and `MR::deque` with `std::vector` and `std::deque`. It also replaces `MR::make_shared` and `MR::make_unique` with `std::make_shared` and `std::make_unique`.


daljit46:
Additionally to changes outlined above, this branch now also contains ca36d7bde7bfd7c75b07d485d4448858a3139543 which was necessary for the CI tests to pass with GCC 13.2. As the compiler, issued a warning (likely a false positive) on using `std::vector`. I replaced this with a `std::array<bool, 3>` as that's the most appropriate type to use there regardless.

Issue #2675, opened 2023-07-09
[mrview]: [View compressed dicom]


Ulquoirra-rn:
Hey. Will mrview ever support compressed dicom transfer syntax?? I know it's possible to uncompress before viewing but some people are not very willing to do that...

Lestropie:
There's the prospect of supporting it if there is adequate demand for it. Often enough new DICOM features are unsupported because the developers have not yet encountered them in the wild and therefore have had no justification for implementing support for them. Can you provide any insight on where the data have come from? Is there any chance of sharing data that would assist developers in implementing support?

(I have zero knowledge of how this aspect of the transfer syntax operates, so don't have a sense of how much effort might be involved; @jdtournier might have better insight)

Lestropie:
Dummy creation can be slightly dangerous if the way in which it's done becomes in any way unrepresentative of real-world data. What can work if file size is a problem is to explicitly acquire a low-resolution dataset from the scanner for the purpose of sharing. Also ideally we'd like something that can be integrated into our private DICOM test data suite. I just checked and found out that the new scanner at my own facility has the ability to save compressed DICOMs, so I should be able to generate some sample data for myself and take a peek at what might be involved.

Ulquoirra-rn:
> What can work if file size is a problem is to explicitly acquire a low-resolution dataset from the scanner for the purpose of sharing. 

The main reason for the large file size in my case is because I read one of your comments on the forum to use vox for improving resolution in tckmap command. I'm all ears if there's a better way to improve conversion of tck without increasing resolution. You are correct in that many scanners save compressed dicoms. I have received such datasets and they are obviously not viewable in mrview.

>  so I should be able to generate some sample data for myself and take a peek at what might be involved.

I think a decompression algorithm would be required based on the transfer syntax to uncompress the pixel data.

Also can you explain a bit more on why dummy creation could be harmful?
Thanks.

Lestropie:
> I'm all ears if there's a better way to improve conversion of tck without increasing resolution.

The question's a little ill-posed as I wouldn't classify this as a "conversion"; it's fundamentally a different type of data. You could look at the Karawun software since that does a more faithful embedding of streamline data in DICOM, but it depends on what those data are being fed into.

> can you explain a bit more on why dummy creation could be harmful?

[...]

PR #2767, opened 2023-12-04
DICOM: add null imageIO handler to allow parsing of DICOM data even with unsupported transfer syntax

jdtournier:
This at least allow mrinfo to run. Any operation that attempts to access the voxel data will of course fail.


core/file/dicom/mapper.cpp
```diff
@@ -82,13 +85,9 @@ namespace MR {
...
           for (auto image_it : *series_it) {
-            if (!image_it->transfer_syntax_supported) {
-              Exception E ("unsupported transfer syntax found in DICOM data");
-              E.push_back ("consider using third-party tools to convert your data to standard uncompressed encoding");
-              E.push_back ("See the MRtrix3 documentation on DICOM handling for details:");
-              E.push_back ("   http://mrtrix.readthedocs.io/en/latest/tips_and_tricks/dicom_handling.html#error-unsupported-transfer-syntax");
-              throw E;
-            }
+            if (!image_it->transfer_syntax_supported)
+            transfer_syntax_supported = false;
```
Lestropie, line 89:
```suggestion
              transfer_syntax_supported = false;
```

jdtournier:
For info: the motivation for this is to at least allow `mrinfo` to report the contents of a DICOM dataset even if the voxel data are stored in a compressed format that we can't currently handle (e.g. JPEG)

Lestropie (dismissed):
I had wanted to test this on some local data with unsupported transfer syntax, but am now having trouble finding any. I do plan on generating some at some point using the local scanner in order to evaluate scope of current support and prospect of better support (eg. #2675) and to augment the testing suite, but have not yet found the time to do it. 

I nevertheless checked the code at the time and was happy with it (over and above aef44bb83a9caf62ffa6d41237a7702ea89b2dc2). Only other thing to consider would be version-matching the documentation hyperlink, but that's not requisite for the PR.

Lestropie:
Actually, in retrospect, the [documentation](https://mrtrix.readthedocs.io/en/latest/tips_and_tricks/dicom_handling.html#error-unsupported-transfer-syntax) shows `mrinfo` failing due to unsupported transfer syntax, whereas with this change, `mrinfo` will succeed but others will fail. So that example should ideally be updated.

Lestropie:
Found some unsupported data. Works as intended at my end, though the user feedback is perhaps sub-optimal:
-   `mrinfo` issues the full warning, but then reports header information anyway, without any specific message about being able to read the header but not the intensity data;
-   Any other command first issues the same warning, then reports as error "No suitable handler to access data". This is because the same `Exception` class is thrown regardless of whether a handler is incapable of accessing an image vs. there was some issue in the appropriate handler reading the image. Transitioning to STL standard exceptions (which we've discussed, but is not a C++17 thing, and doesn't yet have its own Issue) would potentially facilitate disambiguation here. Not a dealbreaker though.

Lestropie (dismissed):
57c0579d63a41c660e6d35926fe33b872be16d6d resolves 2 of the 3 concerns I raised about documentation & terminal messages. I think that improves self-consistency and maybe decodes what's going on a little bit for any users who manage to encounter it.

Lestropie (approved):

Lestropie:
This requires review by someone other than myself or @jdtournier as we can't approve our own changes to `master`.

daljit46 (approved):

Lestropie (approved):

daljit46 (approved):

PR #2771, opened 2024-01-09
Update header copyright to 2024

Lestropie:
Necessary update for successful merging of any new changes to `master`.

### DIFF
--- a/bin/5ttgen
+++ b/bin/5ttgen
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright (c) 2008-2023 the MRtrix3 contributors.
+# Copyright (c) 2008-2024 the MRtrix3 contributors.
 #
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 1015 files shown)


jdtournier (approved):

daljit46:
Can we merge this into dev too?

Lestropie:
> Can we merge this into dev too?

Normally I'd opt to do this as a regular merge of `master` into `dev`. However with `cmake` & `clang-format` being on `dev` now this is a very unclean merge. So I think maybe we'll just redo the copyright update on `dev` separately, and any fixes on `master` that need to make their way to `dev` (which is not all of them) can be individually merged / cherry-picked.

Lestropie:
It's possible that for something like `update_copyright`, running separately on each branch prior to merging might give `git merge` a bit of a helping hand in terms of reducing the scope of conflicts. I'll have to check the other PRs that have since gone to `master` to see the scope of those. A naive `git merge` produced a much larger conflict than what I would have expected, probably due to differences in indentation. Manually cherry-picking each PR onto `dev` and then accepting local in the merge conflict might be the way to go, to make sure that no desired changes are lost in the merge.

PR #2774, opened 2024-01-11
Update copyright to 2024 on dev

Lestropie:
Replicates #2771 for `dev`.

### DIFF
--- a/check_syntax
+++ b/check_syntax
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2008-2023 the MRtrix3 contributors.
+# Copyright (c) 2008-2024 the MRtrix3 contributors.
 #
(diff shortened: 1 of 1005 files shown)


Lestropie:
@daljit46: The [CI failed test](https://github.com/MRtrix3/mrtrix3/actions/runs/7483343156/job/20368428429?pr=2774) only states which test failed, it fails to quote the error message produced by that test. Is that just a matter of adding `--output-on-failure` to the `ctest` call within the GitHub Action?

Lestropie:
Re-running failed jobs:

-   `clang-tidy` was aborted by a keyboard interrupt, perhaps due to timing out (since it's checking every modified file, which for this PR is all of them)
-   `tckgen` test 3 failed, which is `-seed_random_per_voxel`, which has been known to be off by one every once in a while. The thread sanitiser has seemingly not yet caught this one?

PR #2779, opened 2024-01-16
Fix buffer overflow in DWI bootstrapping

Lestropie:
Two components to change:

-   Removal of `current_chunk` should have no effect. It looks to me like vestigial code from trying different approaches to keep track of buffer chunks, but the use of `next_voxel` & `last_voxel` satisfies the requirement.

-   While `get_voxel()` looks like it's doing exactly the same thing, though perhaps less elegantly, it nevertheless resolves a buffer overflow. Though it looks to me like an invalid or overwritten address rather than a conventional "off-by-one"-type overflow.
    On `dev`, within an individual thread, the `value_type*` yielded by the map for a previously sampled voxel is occasionally different to that used when buffer data were allocated for that voxel. Maybe one in every 1000 times that a pointer for a previously sampled voxel is requested.
    I can't report exactly what is wrong with the `dev` code. If it were possible for the `std::map` implementation to perform an internal tree rebalancing after it has done its search as part of `insert()`, but the yielded pointer address pointer were that from before rather than after the rebalancing, that would seem to me like a pretty large STL implementation bug.

### DIFF
--- a/src/dwi/bootstrap.h
+++ b/src/dwi/bootstrap.h
@@ -69,7 +69,6 @@ public:
       voxel_buffer.push_back(vector<value_type>(NUM_VOX_PER_CHUNK * size(3)));
     next_voxel = &voxel_buffer[0][0];
     last_voxel = next_voxel + NUM_VOX_PER_CHUNK * size(3);
-    current_chunk = 0;
   }
 
 protected:
@@ -78,14 +77,10 @@ protected:
   vector<vector<value_type>> voxel_buffer;
   value_type *next_voxel;
   value_type *last_voxel;
-  size_t current_chunk;
 
   value_type *allocate_voxel() {
     if (next_voxel == last_voxel) {
-      ++current_chunk;
-      if (current_chunk >= voxel_buffer.size())
-        voxel_buffer.push_back(vector<value_type>(NUM_VOX_PER_CHUNK * size(3)));
-      assert(current_chunk < voxel_buffer.size());
+      voxel_buffer.push_back(vector<value_type>(NUM_VOX_PER_CHUNK * size(3)));
       next_voxel = &voxel_buffer.back()[0];
       last_voxel = next_voxel + NUM_VOX_PER_CHUNK * size(3);
     }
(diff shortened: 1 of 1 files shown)


Lestropie:
`clang-tidy`:

1.   is complaining about `if` statements without braces, which I think we're permitting in `clang-format`? Do we want to disable that check?

2.   doesn't like single-character variable names, but we use these quite extensively. Do we want to disable this check, or agree to not do this with any new code going forward?

daljit46:
As mentioned in #2745, we should discuss which clang-tidy checks we should enable as some of them may be either undesirable or superfluous in the context of our codebase. However, it's strange that it's complaining about `[readability-identifier-length]` since it should be disabled according to our [current settings](https://github.com/MRtrix3/mrtrix3/blob/808adc3f6411d925ceec5e4e538f0013a726103e/.clang-tidy).

Lestropie:
After fixing merge conflicts to get this to `dev` following #2764, this proposed fix is no longer passing tests with address sanitiser; so not yet ready to merge.

src/dwi/bootstrap.h
```diff
@@ -82,11 +82,10 @@ class Bootstrap : public Adapter::Base<Bootstrap<ImageType, Functor, NUM_VOX_PER
 
   value_type *allocate_voxel() {
     if (next_voxel == last_voxel) {
-      ++current_chunk;
-      if (current_chunk >= voxel_buffer.size())
+      if (++current_chunk >= voxel_buffer.size())
         voxel_buffer.push_back(std::vector<value_type>(NUM_VOX_PER_CHUNK * size(3)));
       assert(current_chunk < voxel_buffer.size());
-      next_voxel = &voxel_buffer.back()[0];
+      next_voxel = &voxel_buffer[current_chunk][0];
```
Lestropie, line 88: @jdtournier Note bug here resolved in b2fa527dbca273096bfae407b706739856b16ec0.

Lestropie:
OK, this now passes tests with address sanitiser. 
Turns out there was a bug in the implementation that resulted in my being deceived regarding its operation. It looks good to me now, but I'd prefer to have an explicit review from @jdtournier before merging.
Also: IMO this class belongs in `src/dwi/tractography/` rather than `src/dwi/`. It specifically deals with the case of storing results of bootstrapping only for some subset of voxels, for which I don't foresee any application other than tractography.

jdtournier (approved):

PR #2791, opened 2024-01-30
Implement MR::apply using std::apply

daljit46:
I noticed some potential use after move in `core/apply.h`. I think this code can be fully replaced by [std::apply](https://en.cppreference.com/w/cpp/utility/apply) introduced in C++17. This PR replaces `MR::unpack` by directly calling `std::apply` instead. `MR::apply` doesn't have a strict equivalent in the STL, but can be easily emulated using fold expressions and lambdas as below:

```cpp
namespace MR {
//! invoke \c Function f for each entry in \c Tuple t
template <typename Function, typename Tuple> constexpr void apply(Function &&f, Tuple &&t) {
  std::apply([&f](auto &&...x) { (f(x), ...); }, std::forward<Tuple>(t));
}

} // namespace MR
```
We could of course replace every instance of `MR::apply` with the definition of the function above, but that would be verbose and tedious. So I think we should still keep the `core/apply.h` header with this wrapper function, however perhaps we should rename `MR::apply` to something else?

### DIFF
--- a/cmd/fixel2voxel.cpp
+++ b/cmd/fixel2voxel.cpp
@@ -130,12 +130,12 @@ struct LoopFixelsInVoxelWithMax {
                      const index_type offset,
                      const std::tuple<DataType &...> &data)
         : num_fixels(num_fixels), max_fixels(max_fixels), offset(offset), fixel_index(0), data(data) {
-      MR::apply(set_offset(offset), data);
+      MR::apply_for_each(set_offset(offset), data);
     }
     FORCE_INLINE operator bool() const { return max_fixels ? (fixel_index < max_fixels) : (fixel_index < num_fixels); }
     FORCE_INLINE void operator++() {
       if (!padding())
-        MR::apply(inc_fixel(), data);
+        MR::apply_for_each(inc_fixel(), data);
       ++fixel_index;
     }
     FORCE_INLINE void operator++(int) { operator++(); }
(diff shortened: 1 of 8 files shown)


daljit46:
I've renamed `MR::apply` to `MR::apply_for_each`. This seems sensible to me, but we could even call it `invoke_for_each` but I think `apply_for_each` is closer to the STL naming thus it easier to understand its purpose.

jdtournier:
I don't really mind, but I think naming it close to the STL `std::apply` might actually cause _more_ confusion given that it behaves differently. I note there's also a `std::invoke`, which weirdly seems to do pretty much the same `std:: apply`, so that's probably not a great choice for the same reason. I note there's also [a `tuple_for_each` in boost](https://codereview.stackexchange.com/a/193536), might be worth using a similar name?

daljit46:
My reasoning for `apply_for_each` is that a programmer familiar with `std::apply` will hopefully think that "this is probably like `apply` but applied to each argument".
`tuple_for_each` is also quite good I guess, the only issue being that the function is not just applicable to tuples, but also to arrays and initialiser lists. 
However, this is really a minor issue. Whatever name we choose is going to better than just `MR::apply`. Shall I change it to `tuple_for_each`?

jdtournier:
> My reasoning for apply_for_each is that a programmer familiar with std::apply will hopefully think that "this is probably like apply but applied to each argument".

OK, I'm sold. Go for `apply_for_each` :+1:

PR #2792, opened 2024-01-31
mrcat: Change positional argument structure

Lestropie:
Currently, `mrcat` enforces at the point of command-line parsing that there are *at least two* input images. It does this by having `Argument` `image`, then `Argument` `image2` with `.allow_multiple()`.

I have two issues with this design:

1.  Consider #2665. Ideally, the utilisation of `mrcat` in a Pydra workflow would look something like:
    ```
        images = ['image1.mif', 'image2.mif']
        workflow.add(mrcat(inputs=images, ...))
    ```

    However with an automatically-generated Pydra interface, what would actually be required 
    would end up being a rather clunky:
    ```
        workflow.add(mrcat(image1=images[0], image2=images[1:], ...))
    ```

2.  This mechanism is absent from other commands where it is equally applicable; eg. `mraverageheader`, `transformcompose`. So the software is not internally consistent in this regard.

For the first point in particular, I think it makes sense to have a single positional input argument with `.allow_multiple()`. But I am open to a counter-argument.

-----

Hypothetically, we could introduce some other `Argument` modifier that would set the minimum number of occurrences. This would allow catching command mis-use at the point of command-line parsing. But we would then need to consider how to reflect this information in the various `__print_usage__` interface exports. Any thoughts?


Lestropie:
After further consideration in the context of Pydra workflow construction, it's possible that changing the `mrcat` operation with a single input from an `Exception` to a `WARN()` may be preferable.

jdtournier:
> After further consideration in the context of Pydra workflow construction, it's possible that changing the `mrcat` operation with a single input from an `Exception` to a `WARN()` may be preferable.

Personally, I don't think this should even raise a warning. There's nothing theoretically wrong with passing a single input image to `mrcat`, and I can see situations where some script may perfectly legitimately need to concatenate one or more inputs, and raising a warning every time the user happens to provide a single input would just cause needless confusion. Just my 2¢...

PR #2802, opened 2024-02-13
population_template: Fix linear drift correction

Lestropie:
Found this one while porting code for #2678.

Problem appears to have been introduced [here](https://github.com/MRtrix3/mrtrix3/commit/5ad072f7721063b1d91c0a786dc23891aa11a3e0#diff-052a1ef62be05c312d0356207bf56c9896d258d58eec06da679f174ae9a3921dR1242-R1245) in 5ad072f as part of #2471. 

It does *not* affect `master`, as the drift correction functionality does not yet exist there.

I'm posting as a draft for now, as currently the average transformation is recomputed at every linear registration iteration, whereas it should only need to be done once; but I'll want to test a little more before committing that.

@maxpietsch: Do you have any good exemplar data with a clear drift?

### DIFF
--- a/python/bin/population_template
+++ b/python/bin/population_template
@@ -1241,9 +1241,9 @@ def execute(): #pylint: disable=unused-variable
       #   - If one subject's registration fails, this will affect the average and therefore the template which could result in instable behaviour.
       #   - The template appearance changes slightly over levels, but the template and trafos are affected in the same way so should not affect template convergence.
       if not app.ARGS.linear_no_drift_correction:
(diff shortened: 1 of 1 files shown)


Lestropie:
I have ensured that `population_template` at least executes for both combinations of `-linear_no_drift_correction` (present vs. absent) and `-initial_alignment none` vs. default (`mass`). Script tests fail due to divergence with previously generated data, but that is due either to the drift correction itself, or something else upstream. I think that I will defer testing to #2811 once `3.1.0` is closer; for each failed test I can do `git bisect`s to figure out what introduces changes, and regenerate reference data if required.

PR #2805, opened 2024-02-14
Switch to Qt 6 as the default 

daljit46:
We are currently building our GUI commands using Qt5 by default. This PR switches the build (and hence also the CI workflows) to use Qt 6 instead. Once merged, an explicit `-DMRTRIX_USE_QT6=OFF` will be required to use Qt 5.
On a related note, should we remove support for building with Qt 5?


daljit46:
It seems that Qt 6 requires a CMake version >= 3.21 on Apple platforms (see [here](https://doc.qt.io/qt-6/cmake-get-started.html), so we will need to change this on the CI workflow for MacOS.

daljit46:
Unless anyone has any objections, I will merge this soon into dev.

bjeurissen:
I think it makes sense to make Qt6 the default.

bjeurissen (approved):

Lestropie:
> Once merged, an explicit -DMRTRIX_USE_QT6=OFF will be required to use Qt 5.

This may require revision of online documentation for building from source.

Don't know if there's external precedent with respect to setting boolean values in environment variables; for command-line options we support "yes/no" and "true/false". Alternatively with `MRTRIX_USE_QT5` you wouldn't even need to check its contents, just its presence; we use that a lot with precompiler directives. 

(Or, as can happen with some other aspects of building, it could make an attempt to build with Qt6, and then re-try with Qt5 if the former fails; though maybe with the `cmake` changes it'd be preferable to be more explicit.)

> On a related note, should we remove support for building with Qt 5?

Depends on the magnitude of maintenance burden. Right now I don't know that there'd be any major benefit to removing support; perhaps if there were larger developments happening in the visualisation space and compatibility was getting in the way it would make sense to deprecate then. But I've not tracked the Qt6 issue so I may be overlooking something.

daljit46:
> This may require revision of online documentation for building from source.

Good point, I will update the instructions we have. 

> Don't know if there's external precedent with respect to setting boolean values in environment variables; for command-line options we support "yes/no" and "true/false". Alternatively with MRTRIX_USE_QT5 you wouldn't even need to check its contents, just its presence; we use that a lot with precompiler directives.

I think specifying boolean values (or "options" in CMake world) is idiomatic when using CMake, but I agree it would indeed make sense to have `MRTRIX_USE_QT5=ON` instead of `MRTRIX_USE_QT6=OFF`, as the latter feels passive. So, I will implement this change.

> Depends on the magnitude of maintenance burden. Right now I don't know that there'd be any major benefit to removing support; perhaps if there were larger developments happening in the visualisation space and compatibility was getting in the way it would make sense to deprecate then. But I've not tracked the Qt6 issue so I may be overlooking something.

Generally, I agree. One downside I considered was that if Qt 6 becomes the default, we will no longer test our builds with Qt 5 so essentially this will become an untested path.

jdtournier:
Personally, I'm happy to remove support for Qt5 for `3.1.0` as long as we provide solid precompiled builds for all platforms, and ideally also provide weekly builds for those who need specific bug fixes.

PR #2806, opened 2024-02-15
dwifslpreproc & mrcat: Strides fix

Lestropie:
Traced issue back to [here](https://github.com/MRtrix3/mrtrix3/commit/e165a0184e046ba7091670b1b0d07de8a779ed01#diff-0b2ffce33b736ea9db78d1705c8baac2d647d2bdcdeb3d87918caec89dcd54c4R769), which ties all the way back to #1541.

Was trying to diagnose an issue with improper correction of susceptibility distortions, and this is the first issue I came across (which unfortunately for me is likely not the only issue...). `applytopup` was doing absolutely nothing to the input data, meaning that the `eddy` brain mask was aligned with the distorted data, whereas the whole intention of using `applytopup` is to produce a processing mask that approximates where the brain will be at completion of processing. It turns out that this is because `dwifslpreproc` was telling `applytopup` that the phase encoding directions were "`k`" and "`k-`", and presumably internally `applytopup` is deeming this unacceptable and either zeroing out that component of the phase encoding vectors or just code branching to a straight copy-input-to-output. Unfortunately it doesn't provide any indication of this at the terminal.

The stride manipulation within `concatenate<vector<Header>>()` seems to have been outright incorrect, so I don't think that it should cause any problems, but before merging I might make the same change on `dev` and run the additional PNG tests in #2713 just to make sure it doesn't have any unintended consequences (existing CI should cover anything else).


Lestropie:
As best as I can ascertain without spending an inordinate amount of time on it, I believe that the bug in 41c02e9 was cancelling out a bug in PNG handling, such that rectification of the former exposes the latter. Plan is to wait for merge of #2713 to `master` before proceeding here.

jdtournier (approved):

PR #2807, opened 2024-02-15
Add workflow to create tarball for Linux

daljit46:
This PR adds a new GitHub Actions workflow to generate self-contained tarballs for Linux releases. To achieve this, mrtrix3 is built on Ubuntu 20.04 and dependencies are packaged with the help of [linuxdeploy](https://github.com/linuxdeploy/linuxdeploy). 
In theory, this will allow users to run MRtrix3 on any distro with a version of `glibc` >= 2.31 (users may still need to install essential packages such as OpenGL). 

I've tested this on fresh installs of (using Gnome Boxes):

- Ubuntu 20.04 LTS
- Ubuntu 22.04 LTS
- Fedora 39
- openSUSE Leap 15.5

If there are distros that you think should be tested, please let me know.

There are still some things to work out:

- [x] Bundle script to install `mrview` icons and desktop files.
- [x] Bundle script to add `./bin` to `PATH`.

We already have these scripts in our codebase, but they may require modifications to work as expected.


daljit46:
The build workflow has now been simplified by adding a new bash script `package-linux-tarball.sh`. This allows an easier testing of the workflow on a local Ubuntu system.

daljit46:
This can now be merged. May need to do some changes to the packagiing script after dealing with #2827, but can handle that later.

Issue #2798, opened 2024-02-05
Save software interfaces in multiple formats


Lestropie:
Have been working a bit on #2678 in order to facilitate #2605 / #2665.

One topic we had discussed in the context of the latter (though I'm not finding a conversation right now to cross-reference) is the prospect of having the Pydra interfaces exported and bundled *within the repository itself*, such that version-matching between software and the parsing of its interface would be intrinsically linked, and consistency of such could be checked by CI.

While this had initially been discussed for the sake of Pydra, it occurs to me that:

1.  This is quite similar to the generation and self-consistency of the online command documentation for those interfaces.
2.  Pydra is not necessarily the only interface for which encapsulation of interface and enforcement of self-consistency is applicable. Eg.:
    1.   `__print_full_usage__` has been around forever, even if nothing has come along to really exploit it;
    2.   While I never finished it, I did [commence](https://github.com/MRtrix3/mrtrix3/tree/porcupine) export of interfaces for [Porcupine](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1006064).
    3.   Interfaces for Nipype could potentially be generated automatically from the internal CLI configuration.
3.  If there are changes or additions to the command-line interface, it might make sense that over and above the code itself, the sensibility of those command-line changes should be checked during code review, and something like the output of `__print_full_usage__` would serve that purpose.

Therefore my "proposal" of sorts here is something along the lines of having a directory somewhere within the repository (maybe not within `docs/`; not sure if there's a good precedent for where to put such things), where the CLI is exported *to multiple formats*, and checking the self-consistency of all of those is integrated into the CI.

Any thoughts?

PR #2808, opened 2024-02-19
Generate interfaces

Lestropie:
Proposed extension of #2678 (PR uses that as the base branch).
As discussed in #2798.
Relevant for #2665, which will expand upon the content here to store Pydra interfaces alongside the legacy outputs of `__print_full_usage__`.

As per #2798, would like to know people's thoughts on incorporating this into the repository. I'm proposing it as part of #2678 as in the absence of that changeset the Python scripts fail to conform to the precedent of `__print_full_usage__` due to the absence of more complex custom command-line argument types. The additional documentation here should be one mechanism by which the validity of those changes is manually checked.


Lestropie:
Seeking feedback from any @MRtrix3/mrtrix3-devs who may have opinions on this; would like to proceed with merging into #2678.

Lestropie:
Following discussion in #2665 and a bit more thought on the topic:

We are not going to generate and store the software interfaces within the main repository.
I do however want to make sure that the interface generation processes are robust.

In #2678, 80658031b3e41a4806d5a6ab9afaa9cfb3eb8805 adds unit tests for the C++ CLI that mirror those added for the evaluation of typed Python CLI. The interface of this dummy command as generated in multiple formats can be manually reviewed. Any changes that may affect CLI parsing, eg. addition of new types, should be verified using these tests (or potentially more).

PR #2810, opened 2024-02-19
Allow filename in write_mrtrix.m to be string or char array

lukeje:
Currently, `write_mrtrix.m` crashes if the output filename is a `string`:
```matlab
% char array
write_mrtrix(struct('data',rand(3,3,3)), 'test.mif')
% works without problem

% string
write_mrtrix(struct('data',rand(3,3,3)), "test.mif")
% Gives error: 
% > Array indices must be positive integers or logical values.
% >
% > Error in write_mrtrix (line 134)
% > if strcmp(filename(end-3:end), '.mif')
```

As the error message indicates, this is because array indexing does not work for `string`s. This pull request uses the Matlab functions `fileparts` and `fullfile` to avoid explicit indexing. 

I also took the opportunity to make the error message if the extension is not matched a little more informative.

### DIFF
--- a/matlab/write_mrtrix.m
+++ b/matlab/write_mrtrix.m
@@ -131,7 +131,8 @@ end
 
 fprintf (fid, '\nfile: ');
 
-if strcmp(filename(end-3:end), '.mif')
+[filepath,basename,ext] = fileparts(filename);
+if strcmp(ext, '.mif')
   dataoffset = ftell (fid) + 18;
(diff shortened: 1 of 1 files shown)


jdtournier:
Wow, I had no idea there was a difference between a character array and string in matlab... Guess you learn something new every day!

All looks good and sensible to me.

jdtournier (approved):

Issue #2138, opened 2020-07-27
tcksift(2): Change handling of mu for outputs


Lestropie:
From [forum](https://community.mrtrix.org/t/regarding-tdi-endpoint-maps-interpretation-thresholding-and-smoothing/3887/6).

Much easier than requiring use of the `-out_mu` option, and won't change main command output content.

Lestropie:
Preserving in `3.1.0` milestone as it should only require a small amount of effort. It's something that I regret not having addressed every time I have to deal with such data, but I also want to expand the scope of the solution.

There has always been a tension between:
-   My desire for this scaling to be understood and treated as a discrete optional step
-   The fact that in some usage scenarios, one may want to perform more scaling factors than what would otherwise be captured by simply multiplying the weights by mu in `tcksift2`;
-   The inconvenience of having to do two matrix loads & a scalar multiplication externally to *MRtrix3* to achieve what is recommended to be the most appropriate scaling;
-   The risk of imperfect scaling if a user relies exclusively on the scaling factor being applied here but not properly dealing with other confounds.

So what I think I need to do here:

- [ ] Always write value of mu to output header
    - [ ] This should mean that `-out_mu` should no longer be necessary; but it might be preferable to leave that command-line option in place for backward compatibility, at least initially.
- [ ] Add ability to toggle whether or not the output weights will or will not be scaled by some global scaling factor
    - [ ] Decide whether the default should be to scale or not to scale
    - [ ] Decide what factor should be used to determine that scaling
        In retrospect, I think the most sensible thing is to scale not by just mu directly, but also by the input image voxel resolution, so that the data are of units AFD / mm<sup>2</sup>.
    - [ ] Add command-line option to toggle scaling behaviour from the default to the alternative
    - [ ] Indicate in file header whether or not the values contained within have been subjected to that scaling
    - [ ] Add config file entry to modify default behaviour
- [ ] Add information to command documentation explaining the difference in behaviour, and linking to online documentation
- [ ] Update online documentation page dedicated to SIFT, adding information about SIFT2, link to the Aperture manuscript, and explaining the circumstances in which one might do it one way vs another

Lestropie:
My latest thinking here is that for SIFT2 I want to add a command-line option and config file entry that specifies the "units" of the output streamline weights:

-   If "units" == "NOS / Number Of Streamlines", then the behaviour will be the same as that currently; internal scaling by mu is performed such that across the tractogram the streamline weights are approximately distributed around unity
[...]

PR #2815, opened 2024-02-21
Disable clang format for CLI code

Lestropie:
Partial reversion of #2652; though it's not "reverted" so much as replaced with manually formatted code.

Upon attempting to commence #2138 I found the reformatted `usage()` code so unbearable that I committed to refactoring all of the affected code.

I've tried to follow something resembling standards in terms of code formatting, including reducing text width even if column size isn't going to be enforced within these blocks of code. I also took the opportunity to tweak certain command documentation as I was going; shouldn't be anything controversial.

Could do with a fresh set of eyes over the manifest changes to the online command documentation just to be sure I've not broken anything.


Lestropie:
Line count is a little gratuitous; will need to refactor and put the bulk of changes under the MRtrixBot account.

daljit46:
LGTM, modulo the fact that sometimes you've missed a space in `// clang-format off`. I also recommend installing `pre-commit` locally and then run `pre-commit install` to install a git hook that will catch if you have code that is not formatted.

Lestropie:
I've put off installing the pre-commit hook / format on save because there's a lot of branches that I still need to explicitly transition from pre-clang-format code. I do however need to get into the habit of running it explicitly. Not so concerned on commits like this one; it's moreso not wanting 200 clang-format-related commits on 100 different branches all making their way into the history.

Lestropie:
Replaced by #2818.

Issue #2821, opened 2024-02-26
Test Python 3.7 compatibility


Lestropie:
With `3.1.0`, support for Python 2 is being removed (#2215).

With #2678. the minimum version will go up to 3.7 due to the use of f-strings.
However I can't guarantee off the top of my head with 100% assurance that `3.1.0` will work on Python 3.7.

I'd like to add a CI test, probably in conjunction with #2154, that builds a Python 3.7 environment and does some degree of testing of Python scripts. This would ensure that the documented lower bound of support is in fact accurate.

daljit46:
The [setup-python](https://github.com/actions/setup-python) action may be useful for implementing this. IMO, it's a good idea to run a small set of tests with the minimum supported version (similarly to how we test the CI builds with CMake 3.16).

Lestropie:
Closed by #3329.

Issue #2789, opened 2024-01-28
Testing: Split test string


Lestropie:
Currently, there are quite a few tests where a single test string actually contains:
-   One or more commands that generate intermediate data prior to running the test
    (partly to reduce storage space of the test data, partly to make the nature of the test data self-explanatory)
-   The command invocation itself
-   Evaluation of the outputs from the command, typically by comparing to a pre-generated output from an earlier version

Questions primarily for @daljit46:
1.  Can `ctest` "classify" lines based on which of the three categories each belongs to?
2.  Is there an "industry accepted" way of arranging such?

daljit46:
1. I think the answer here is no. `CTest` is just a test launcher, it's not aware of what those tests are doing.
2. I don't know any big software projects that implement tests the way we do, although there may be projects like terminal shells that use a similar strategy for part of their testing. Looking for "approval testing" on Google, I came across this project https://github.com/DannyBen/approvals.bash.

Lestropie:
Given: 

> I don't know any big software projects that implement tests the way we do

, is the complexity I'm pointing out a problem that arises *because* of such non-conformity, and adopting a more conventional approach would naturally resolve it? Or does it appear to you that our "unconventional" testing approach is out of necessity?

Alternatively, I'm wondering if your comment is moreso an observation on our extensive use of regression testing rather than unit testing?

Just picking at your brain to see if there's a logical & efficient way to change our tests across the board.

daljit46:
For our usecase, I think that our tests are fairly suitable especially since they emulate how an end user would use the tools that we develop. 

I think the "peculiarity" of our testing framework is mainly due to the use of bash scripts. I was just pointing out that because of this, I don't know about the existence of guidelines and best practices that are commonly used in this scenario. 

> Alternatively, I'm wondering if your comment is moreso an observation on our extensive use of regression testing rather than unit testing?

On this point, I don't think that unit testing and integration/approval testing are mutually exclusive. In an ideal scenario, we would have both. For example, an obvious limitation of our test suite is that we don't test any GUI code. This is because there is not obvious way of running `mrview` and verify an "output". While GUI testing is extremely hard, unit testing would at least allow you to test the UI logic without running the entire executable. 
[...]

PR #2825, opened 2024-02-26
"Quick" testing of Python scripts

Lestropie:
Creating a new PR to deprecate #2154 given the incompatibility of some of those differences with a `cmake`-based `dev`.

At initial posting, this just includes a single text file.

- [ ] Determine appropriate filesystem location for this file.
    #2154 kind of hacked its way into providing this test in terms of setting the data location / working directory. I doubt that just dumping this file in its current location will work, since it needs to (obtain and) set its location to the script data directory.
- [ ] Add this file to the list of tests to be performed as part of CI.
    Obviously will need to look at its total runtime to see if it's prohibitive to include as part of CI.
    **Edit**: Evaluating this set of tests within a Python 3.7 environment could kill two  birds with one stone (#2821).


Lestropie:
@daljit46 Open to feedback on how to best achieve this, given I'd like to combine it with #2821 before merging #2678.

Previously, this was slated to be achieved by adding an extra text file, which could be specified as the input argument to `run_tests`. With `cmake` this makes less sense: the test suite wasn't built from the ground up with `ctest` in mind, and so the direct translation from what we had to `cmake` is a little hacky given that origin (#2789 could hopefully help improve that a little), and the proposal here feels like exacerbating that. What's proposed here isn't a new set of tests; it's just a chosen subset of the existing tests. So I think what would better suit would be to create a new test label? This would be reasonably easy if the list of tests was explicit, but harder when it's grepped. #2789 would at least give each a more meaningful name for identification, but there'd still need to be a way to cross-check tests against this list as they are registered.

(On reading for the first time I also wonder if using labels to separate between C++ command / Python command / unit tests would be better than grepping on test names?)

daljit46:
If this suite of tests is under the category of "scripts", then I think it would make sense to place this under `testing/scripts`. 

If we want to create a subset of an existing set of tests, then I think we need a way to parse a "label" for a test from a text file. Something like a comment or another syntax that tells us the specific label of the test. Then, we can easily run a specific set of test matching a label.
CMake does support [labels](https://cmake.org/cmake/help/latest/prop_test/LABELS.html) as part of a test's properties, so we could make use of that (BTW I can't remember why I didn't add labels to our current tests, possibly because I intended to do it at a later point, but we definitely should do that regardless of this PR).

Lestropie:
Given dc3566f1efa2d0dbdd39effd4f95c74a47338879 in #2865 lists each test explicitly for `cmake`, it should be possible there to manually add a label to those tests of Python commands that are manually chosen to execute as part of CI, rather than relying on the interrogation of test shell script content. Since that is already listed as one of the tasks for #2865, I'm content to close this PR, though it will continue to serve as a reference for the set of tests that have previously been flagged as suitable for inclusion in CI.

Issue #2828, opened 2024-02-27
Universal revisions based on clang-tidy 


Lestropie:
Since #2745, `clang-tidy` was introduced.
It is run only on proposed changes within PRs, not across the whole code base, since the latter would be overly burdensome.
We can however elect to take any recommendations that it makes based on isolated code, and apply the same changes across the code base, eg. in cases where instances of such can be found through a regex search or where it's known that a particular design pattern is used in a finite number of places.
This issue will serve to facilitate listing `clang-tidy` proposals that may be worthwhile revising the code base.
Hopefully it will doubly serve to show `clang-tidy` proposals that we are aware of but may not necessarily resolve in any unrelated PR, so that those less experienced will not panic if they see them.

-----

- [ ] `cppcoreguidelines-pro-bounds-array-to-pointer-decay`
    Exemplified in #2818.
    Really does not like the way that strings and lists-of-strings are handled in the CLI; particularly for `Argument::type_choice()`, but also for eg. `App::DESCRIPTION`.
    Would be best resolved comprehensively with #2111.

- [ ] `cppcoreguidelines-macro-usage`
    Exemplified in #2818 for `cmd/mrview.cpp`, but macro functions are likely used in a number of other places.

daljit46:
`readability-container-size-empty` can be applied using `clang-tidy -fix`, but the other two would too risky to change using automation. In particular the first would most likely need manual supervision on all instances of `const char*` as we would need to think about the lifetimes of strings. 
I'm in favour of this though, so I think a gradual adoption of `std::string` and perhaps `std::string_view` would be ideal. Perhaps, we could start doing this with all application level code (i.e. all code in `cmd`).

Lestropie:
I wouldn't want to accept the automated changes for `readability-container-size-empty` either; it suggests changing:
`value = opt.size() ? A : B;`
to:
`value = !opt.empty() ? A : B;`
, whereas preferable would be:
`value = opt.empty() ? B : A;`

Lestropie:
On #2702 I'm currently getting a lot of `bugprone-narrowing-conversions` going from unsigned to signed array indices. `clang-tidy` seems to be even more aggressive in this respect than the current state of the compilers (which themselves have progressively flagged more code in this respect). Not sure if this is one we might actually want to *disable* in `clang-tidy`, as it could produce a lot of noise resulting in people ignoring the meaningful suggestions.

daljit46:
I agree on that example.

[...]

Issue #2822, opened 2024-02-26
mrview & shview version mismatch on dev


Lestropie:
On the current `dev` branch, I did a full delete of my `cmake` build directory, a `ccache -C`, and re-built in a new directory. Commands `mrview` and `shview` nevertheless fail to execute due to mismatch between `git` versions reported for the executable vs. the shared library.

```
robertes@9520l-003953-l:~/src/mrtrix3$ ldd ./build_cmake/bin/mrview
	libmrtrix-gui.so => /home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/src/libmrtrix-gui.so (0x00007fe8e9400000)
	libmrtrix-core.so => /home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/core/libmrtrix-core.so (0x00007fe8e9000000)
	libmrtrix-headless.so => /home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/src/libmrtrix-headless.so (0x00007fe8e7600000)
	...
robertes@9520l-003953-l:~/src/mrtrix3$ ./build_cmake/bin/mrview
mrview: [ERROR] executable was compiled for a different version of the MRtrix3 library!
mrview: [ERROR]   mrview version: 3.0.4-708-g895628e0
mrview: [ERROR]   library version: 3.0.4-623-g4f3836b9-dirty
mrview: [ERROR] Running ./build again may correct error
```

In addition to resolving whatever is causing the conflict, the error message also needs to be updated to reflect the `cmake` transition.

daljit46:
My guess is that this is caused by something wrong in your `PATH` that's somehow preventing a correct selection of the libraries at runtime. I remember @dchristiaens having a similar issue.

Lestropie:
I've wiped my `PATH` of all traces of MRtrix3, ensured `LD_LIBRARY_PATH` is empty, wiped compiler cache, deleted build directory, configured and built:

```
robertes@9520l-003953-l:~/src/mrtrix3$ ./build_cmake/bin/mrinfo -version
== mrinfo 3.0.4-623-g4f3836b9-dirty ==
...
robertes@9520l-003953-l:~/src/mrtrix3$ ldd ./build_cmake/bin/mrinfo
	libmrtrix-headless.so => /home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/src/libmrtrix-headless.so (0x00007fc9fd800000)
[...]

PR #2829, opened 2024-02-27
Check emptiness of containers using empty()

daljit46:
As suggested by clang-tidy (see #2828), emptiness of containers should not be checked by using `size()` but we should  use `empty()`. Other than readability, this also has potential performance benefits as size() may not be constant-time.

This PR replaces all such instances of `size()` with `empty()` using `clang-tidy`.


Lestropie:
@daljit46: The executable - library version mismatch issue reported in #2822 is now occurring within the CI; eg.:
```
bitset: [ERROR] executable was compiled for a different version of the MRtrix3 library!
bitset: [ERROR]   bitset version: 3.0.4-869-g3a2db299
bitset: [ERROR]   library version: 3.0.4-623-g4f3836b9-dirty
bitset: [ERROR] Running ./build again may correct error
```

Lestropie:
If I had to guess, I'd say that because I'm dealing with a lot of pre-`cmake` PRs that I have to explicitly merge to `dev`, if I run the old `build` script, which generates `core/version.cpp`, then use `cmake`, the old `core/version.cpp` is getting used by the compiler rather than the version of that file that `cmake` is supposed to be generating.

Lestropie:
Unfortunately `clang-tidy` doesn't like some of my rearrangement of string formation...
Some of them I'm a little surprised by; wondering if there's a way to instruct `clang-tidy` to be content with a given structure as long as it doesn't violate general rules, rather than forcing it to be *exactly* as `clang-tidy` determines? I don't want to disable `clang-format` for little things like this.

@daljit46:

1.  Can spreading string concatenation across multiple files be triggered to happen *only* if maximal columns is violated? I've been trying to newline string formation based on punctuation as it's more human-readable.
    eg.:
    ```
    --- ./src/dwi/tractography/SIFT/model.h	(original)
    +++ ./src/dwi/tractography/SIFT/model.h	(reformatted)
    @@ -157,8 +157,10 @@
             ++num_tracks;
             max_index = std::max(max_index, i);
           }
    -      WARN("Only " + str(num_tracks) + " tracks read from input track file;"
    -           " expected " + str(contributions.size()));
    +      WARN("Only " + str(num_tracks) +
    +           " tracks read from input track file;"
    +           " expected " +
    +           str(contributions.size()));
           contributions.resize(max_index + 1);
         }
       }
    ```

2.  If function parameters are spread across multiple lines and aligned using indentation so as not to exceed maximal width, is it possible to have `clang-tidy` accept that rather than newlining on the parenthesis?
    eg.:
    ```
    --- ./core/file/dicom/patient.cpp	(original)
    +++ ./core/file/dicom/patient.cpp	(reformatted)
    @@ -58,10 +58,8 @@
     }
     
     std::ostream &operator<<(std::ostream &stream, const Patient &item) {
    -  stream << MR::printf("  %-30s %-16s %10s\n",
    -                       item.name.c_str(),
    -                       format_ID(item.ID).c_str(),
    -                       format_date(item.DOB).c_str());
    +  stream << MR::printf(
    +      "  %-30s %-16s %10s\n", item.name.c_str(), format_ID(item.ID).c_str(), format_date(item.DOB).c_str());
     
    ```

3.  If composing a long boolean expression, and it can't fit on one line, I'd prefer to have one line per item rather than sometimes putting multiple on a line just because they fit:
    ```
    --- ./cmd/mrtransform.cpp	(original)
    +++ ./cmd/mrtransform.cpp	(reformatted)
    @@ -424,8 +423,7 @@
 
       // Detect FOD image
       const bool is_possible_fod_image =
    -      input_header.ndim() == 4 &&
    -      input_header.size(3) >= 6 &&
    +      input_header.ndim() == 4 && input_header.size(3) >= 6 &&
           input_header.size(3) == (int)Math::SH::NforL(Math::SH::LforN(input_header.size(3)));
```

Lestropie:
@jdtournier: Given I tweaked some of the DICOM code (just collapsing unnecessary nested `if` statements), it might be worth running the private DICOM tests.

jdtournier:
Going through the DICOM tests (which needed some tweaks to work from `dev`), there are a few failures, I'll look into them when I have a minute this afternoon.

jdtournier:
OK, seems issues are already present in `dev` anyway. Investigations ongoing...

jdtournier:
OK, that'll be due to #2691 being on `master`, but not on `dev`. Means it's time to merge `master` back to `dev`... Not looking forward to it, there's going to be a mountain of conflicts in there with all the cmake changes that have happened since the last time... :weary:

Lestropie:
Merging so that this larger set of changes can be resolved against other PRs. If it turns out that the minor changes to the DICOM handling implemented in here do in fact cause problems (previously couldn't be verified because was awaiting #2893), that can be addressed in its own PR.

PR #2834, opened 2024-02-28
Regenerate CMake tests when test file is modified

### LINKED ISSUE #2832: Rebuild not pulling test updates
In trying to finally resolve #2702 I ran into a trap with the current `cmake` configuration.

I first configure, build, then run a test. The test fails. I determine that the problem is not with the implementation of the thing I'm testing, but with the test itself. So I change the test, and re-run `cmake --build`. However the tests still fail as though I had not made any change.

daljit46:
Fixes #2832. Test files are parsed by CMake at configure time and added using `add_test`. This PR ensures that when a bash test file is modified CMake automatically reconfigures to ensure that the latest version of the tests are deployed.
@Lestropie let me know if this addresses the problem.

### DIFF
--- a/cmake/BashTests.cmake
+++ b/cmake/BashTests.cmake
@@ -15,6 +15,9 @@ function(add_bash_tests)
     set(working_directory ${ARG_WORKING_DIRECTORY})
     set(exec_directories ${ARG_EXEC_DIRECTORIES})
 
+    # Regenerate tests when the test script changes
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${file_path})
+
     # In MINGW, we need to replace paths starting with drive:/ with /drive/
     # when invoking a bash command (e.g. using bash -c "command")
     if(MINGW AND WIN32)


Lestropie:
Seems to work with updates to existing files. What it doesn't seem to catch is the scenario where I add a new script test file to the source tree that was absent at configure time; even after rebuilding, it doesn't find that new test.

(Might not seem like a huge problem now, but if we were to do #2789, which I've been tempted to do given difficulties / confusion I've had with testing all these Python-related PRs, it would be a more substantial limitation)

daljit46:
> What it doesn't seem to catch is the scenario where I add a new script test file to the source tree that was absent at configure time; even after rebuilding, it doesn't find that new test.

This is by design and intendended behaviour. The current workflow to add tests is different than the one we had previously. In order to add a new suite of tests for. a given command, you will need also to update the corresponding CMake file in `testing`. 

The motivation is that we don't want to rely on [globbing](https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil) as it is not idiomatic in the CMake world and generally discouraged. 

We do use globbing for cpp files, but that was done mostly to minimise changes to the current file structure. 
However even for C++ files you will need to reconfigure the project if you add a new file.
We may want to revisit this too, see for example #2746.

I think it's worth having a discussion whether automatic reconfiguration of files is desirable or not. However, I feel like that would be beyond the scope of this PR.

Lestropie:
Two points on the addition of new tests:
-   Given the `cmake` transition, there is some risk that someone (ahem... :raised_hand_with_fingers_splayed:) may add files attempting to run new tests, but unless the `cmake` files are updated accordingly, they won't be run, and the test suite (whether run locally or in CI) will just succeed. We may simply have to be diligent about it at code review time.
-   If I were to do #2789 (which I'm considering), what is currently one file per command would become one directory per command, with one file per test to be performed. If each of these would need to be explicitly listed for `cmake`, it might increase the probability of omission errors, since individual additions of new test files would happen more frequently.

I am however content to accept this fix as-is, and we'll deal with these confounds separately.

Lestropie (approved):

PR #2835, opened 2024-02-29
Add periodic releases workflow

daljit46:
This PR adds support for weekly releases. It provides a single workflow that builds and creates appropriate packages for each OS. Naming convention has been set to `mrtrix3-OS-weekly-DATE-COMMIT.extension`. Packages will be built for Linux, MacOS and Windows and released every Sunday at 12AM UTC time. 
To enable this we need to create a new release tagged `weekly` which will be updated every week replacing the assets built the week before. 

Tasks to be completed:
- [x] Merge #2830 
- [ ] Create release tagged `weekly` and change `RELEASE_ID` and `UPLOAD_URL` variable inside the workflow.
- [x] Ensure that releases are built correctly on a fork of MRtrix3

Needs to merged onto `master` because the workflow needs to run periodically, but it will the use `dev` branch to build the packages.

Feedback welcome.

### DIFF
--- a/.github/workflows/weekly-releases.yml
+++ b/.github/workflows/weekly-releases.yml
@@ -1,0 +1,150 @@
+name: Weekly Releases
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+  workflow_dispatch:
+
+env:
+  RELEASE_ID: 143178005
+  UPLOAD_URL: https://uploads.github.com/repos/daljit46/mrtrix3-ci-test/releases/143178005/assets{?name,label}
+
+jobs:
+  linux-weekly-release:
+    runs-on: ubuntu-20.04
+    env:
+      SCCACHE_GHA_ENABLED: "true"
+      SCCACHE_SIZE: "2G"
+
+    steps:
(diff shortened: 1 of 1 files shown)


daljit46:
The workflow now supports a "branch" input that a developer can specify to build a release from a specific branch. 
Still need to ensure that when a custom branch is used, the output is not a standard release but an artifact that can be downloaded via Github's web interface for actions.

daljit46:
After thinking about this, I think it'd be better if we had one single release workflow that can handle:
1. tagged releases
2. weekly releases
3. manually triggered releases that upload a GitHub Actions artifact rather than a release artifact

This will prevent code duplication in the CI code. One problem is that we want to test the workflows on master and dev. However, master is still on the old build system rather than CMake. To get around this, I think it's best to wait until master is also on CMake and incrementally implement the three features above. 
What I would like to do is to first create a workflow on master that solely implements 3, so that we can test that build scripts are working as expected. Once, we have master on CMake we can implement and test 1 and 2.
Thus, I will be closing this PR and instead submit a new PR for 3.

Issue #2836, opened 2024-03-01
cmake: tests that are not .cpp standalones


Lestropie:
@daljit46 I'll need your input on this one.

I believe that this was a failure on my part to fully resolve #2437 against #2689.

Whatever is the solution here, will also be applicable to #2678, where I'm trying to add a test of the Python CLI changes.

Relates to #2834, but is sufficiently different to justify a separate thread.

-----

Previously, the "unit tests", as distinct from the "binaries" (ie. C++) and "scripts" (ie. Python) tests, were all just `.cpp` files, that could be compiled by invoking the `build` script in the `testing/` directory. A set of text files then defined the tests, each of which literally just invoked the compiled binary with no arguments.

I think in #2437 (and now coming in #2678), I was exploiting this design to achieve testing in a different way, and now with the move to `cmake` perhaps it no longer works.

There, I added a more complex test structure:
-   `testing/bin/testing_*_npy` are written in Python (though do not use the MRtrix3 Python API), and are used to generate and check data for evaluating the C++ NPY support;
-   `testing/tools/testing_unit_tests_npy*.cpp` result in compilation of binaries that perform functionalities to be evaluated by the Python commands above;
-   `testing/tests/npy` executes the C++ and Python functions above in sequence as necessary to evaluate the proper function of the C++ NPY support.

This is however being overlooked by the new `cmake` test suite construction, which I am naive to.

On naive inspection, I think what's happening now on `dev` is that:
-   The set of unit tests is defined explicitly in `testing/unit_tests/CMakeLists.txt`
-   For each test name:
    -   The `.cpp` file with that name is compiled;
    -   A test with that name that executes the compiled binary is constructed.

Trouble here is that my NPY tests don't conform to this structure. 

Files `testing/tools/testing_unit_tests_npy*.cpp` are I think in the correct location, though "`unit_tests_`" should probably be stripped from their names. The problem is that I want to be able to define a unit test (ie. it's test of neither a specific binary nor a Python command), where the commands to be invoked to complete that test is more complex than just the name of a compiled binary. That will I think require more complex `cmake` directives than what currently exists on `dev`.

I'm hoping that if `dev` is modified such that the NPY tests are performed as intended, then #2678 will re-use that structure.

daljit46:
[...]

PR #2842, opened 2024-03-04
Fix the execution of NPY tests

daljit46:
As mentioned in #2836, when #2437 was merged, NPY tests were mistakenly added to the codebase without accounting for the new testing structure post CMake.
This PR aims to address that following the discussion in #2836.
In particular, the CMake logic for our "unit tests" and any executables (under `testing/tools`) used by a test has now been modified to account for Python files.


Lestropie:
Did just a little bit of re-arrangement.

-   Renamed from "Python tests" to "Bash tests". These are not Python tests. They are tests that make use of a combination of C++ and Python tools to evaluate a particular C++ capability. The test invocations themselves are just Bash. They differ from the other "unit tests" in that there is no longer a direct correspondence between tool name (as a `.cpp` file) and execution of the binary compiled from that file as the test itself.

-   The tests initially failed. This happened because function `add_bash_tests()` defined in `cmake/BashTests.cmake` is utilised. This function treats each individual line of the input file as an individual test, and explicitly inserts "cleanup" jobs at the head and tail of this list; essentially replicating the behaviour of the old `run_tests` script. It does this by erasing `tmp*` and `-tmp-` *relative to the data directory*. But for the unit tests, no such data directory existed. For now I've just set it to use the source directory as the data directory; but in #2678 there *will* be a new directory for storing data relevant to "unit tests".

Lestropie (approved):
Confirmed working locally & executing in CI.

PR #2845, opened 2024-03-05
Add argument to specify environment for bash tests

daljit46:
This adds a new function argument to add_bash_tests() to specify environment variables that should be defined for running a test.
Usage example:

```cmake
add_bash_tests(
    FILE_PATH file
    PREFIX script
    WORKING_DIRECTORY ${DATA_DIR}
    ENVIRONMENT "PYTHONPATH=${PROJECT_BINARY_DIR}/lib"
)
```

In #2678, the possibility of specifying `PYTHONPATH` for a given test seems to be useful. This PR addresses this need more generally, by allowing the specification of any environment variables when running a collection of tests specified in a text file.

@Lestropie this should be enough to address the issue raised, but let me know if this actually solves the problem.

### DIFF
--- a/cmake/BashTests.cmake
+++ b/cmake/BashTests.cmake
@@ -1,6 +1,6 @@
 # A function that adds a bash test for each line in a given file
 function(add_bash_tests)
-    set(singleValueArgs FILE_PATH PREFIX WORKING_DIRECTORY)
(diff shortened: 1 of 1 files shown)


Lestropie:
Included in merge of #2678.

PR #2846, opened 2024-03-05
Fix python CLI CI test

Lestropie:
Testing whether the proposed change in #2845 facilitates executing the Python CLI unit tests in CI for #2678.

I do realise in retrospect that this solution might not actually live that long (even if it's a useful capability to leave in place): with #2741, the Python CLI test would need to have the same mechanism of executable binary generation.

### DIFF
--- a/cmake/BashTests.cmake
+++ b/cmake/BashTests.cmake
@@ -1,6 +1,6 @@
 # A function that adds a bash test for each line in a given file
 function(add_bash_tests)
-    set(singleValueArgs FILE_PATH PREFIX WORKING_DIRECTORY)
+    set(singleValueArgs FILE_PATH PREFIX WORKING_DIRECTORY ENVIRONMENT)
     set(multiValueArgs EXEC_DIRECTORIES)
     cmake_parse_arguments(
         ARG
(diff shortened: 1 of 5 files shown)


Lestropie:
d9587e650a19c5fc4a54bfac8c654a356d387b13 got the new Python CLI test working on both Unix and MacOS, but the Windows CI test fails, with command `testing_python_cli` unable to find the `mrtrix3` Python module.

Lestropie:
On Linux:

```python
sys.stderr.write(repr(os.environ))
[..., 'PYTHONPATH': '/home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/lib/', ...]
sys.stder.write(repr(sys.path))
[..., '/home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/lib', ...]
```

On MSYS2:

```python
sys.stderr.write(repr(os.environ))
[..., 'PYTHONPATH': 'C:/msys64/home/rob/src/mrtrix3_cmake/build_cmake/lib', ...]
sys.stder.write(repr(sys.path))
[..., '/home/rob/src/mrtrix3_cmake/testing/data/C', '/msys64/home/unimelb.edu.au/robertes/src/mrtrix3/build_cmake/lib', ...]
```

Setting `PYTHONPATH` via `ENVIRONMENT` is being done with `${PROJECT_BINARY_DIR}`; this seems to be yielding an absolute path from the drive root with Windows-style drive separator; this gets corrupted when `python` reads it and appends `sys.path`.

PR #2847, opened 2024-03-05
Fix python CLI CI (test PR only)

Lestropie:
#2846 is the actual proposed merge; creating a fresh PR here to `dev` just to trigger CI tests (changing the base branch of #2846 just left them required but not executing).


Lestropie:
CI checks passed. `clang-tidy` seems to be sometimes hanging and eventually exceeding runtime (6h); but that shouldn't be relevant here.

Issue #2826, opened 2024-02-26
Update testing/README.md


Lestropie:
File `testing/README.md` on `dev` does not yet reflect the transition to `cmake`.

Lestropie:
Closed by #2855.

PR #2855, opened 2024-03-08
Update testing README

daljit46:
Addresses #2826.
Testing tips and instructions have been updated to reflect the transition to CMake. Sections on the handling of testing data assume #2848 has been merged. 
Some instructions may not be clear or need removing due to the fact that some of the concerns/workflows no longer apply with the transition to CMake. 

It's still WIP.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -86,6 +86,15 @@ ctest -L dwi2response # Runs all tests of the Python command dwi2response
     ctest -R 5ttgen_hsvs # Run only the tests of specifically the "HSVS" algorithm of the 5ttgen command
     ```
 
+-   Environment variables "`MRTRIX_BINARIES_DATA_DIR`" and "`MRTRIX_SCRIPTS_DATA_DIR`"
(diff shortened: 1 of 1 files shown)


daljit46:
I've updated the testing instructions to reflect the latest changes in testing code. @Lestropie could you please have a look and see if there's anything that's missing and/or incorrect?

Lestropie:
-   There are now labels both for language and for each command name, so no need to use regexes.
-   There were still some out-of-date references to being one test per line of a test script file, predating #2865.
-   The later text read like it was written when we first added the testing. I've replaced it with something new describing how I typically go about it these days.

daljit46:
I think you've removed the section on using a local clone to download the test data. Shall we omit that in the testing readme? I think it'd be for that mechanism to be documented *somewhere* at least.

Lestropie:
Good catch. Think I failed to move it partway through, then erroneously deleted it entirely when I had to do an explicit rebase.

daljit46:
This seems to be good to go IMO, so will merge soon.

Issue #2862, opened 2024-03-14
Avoid using reserved identifiers by the C++ standard


daljit46:
There are many instances in our codebase (e.g. pretty much all include guards) where variables are declared beginning with a double underscore. As pointed out by `clang-tidy`([bugprone-reserved-identifier](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone/reserved-identifier.html), this results in undefined behaviour since the C++ standard reserves all identifiers beginning with a double underscore.

daljit46:
My preference would to use `#pragma once` because I think header guards are ugly. While `#pragma once` is not strictly in the standard, it's virtually supported by all compilers in existence.

Lestropie:
Closed by #2871.

daljit46:
Note that the use of `#pragma once` is not enough to consider this issue solved. There are many instances in mrtrix3 where `__` is used in C++ code (e.g. [here](https://github.com/MRtrix3/mrtrix3/blob/b4363744723c856ff79f1c542c4ffa2ddb22e024/cpp/core/mrtrix.cpp#L103)).

Lestropie:
Closed by #3368.

PR #2871, opened 2024-04-05
Replace include guards with #pragma once

daljit46:
Proposal to replace header include guards with `#pragma once`. This means less typing and lines of code, less distractions and less chances of symbol collisions (caused by same values used by guard values). 
It would also address most of the issue outlined #2862. 

Migration from include guards was done using [guardonce](https://github.com/cgmb/guardonce). The only manual change needed was to address the following in `core/fixel/legacy/image.h`:

```
#error File that #includes "fixel/legacy/image.h" must explicitly #include "image.h" beforehand
```


Lestropie (approved):
LGTM.

Tried restarting the `clang-tidy` job, but I've a suspicion it'll time out again because of the number of files to check. Feel free to merge without it.

daljit46:
> Tried restarting the `clang-tidy` job, but I've a suspicion it'll time out again because of the number of files to check. Feel free to merge without it.

The log message says:

>   CMake Warning at core/CMakeLists.txt:79 (message):
    libpng not found, disabling PNG support
  CMake Warning at core/CMakeLists.txt:86 (message):
    libtiff not found, disabling TIFF support
  fatal: detected dubious ownership in repository at '/github/workspace'
  To add an exception for this directory, call:
  	git config --global --add safe.directory /github/workspace
  fatal: detected dubious ownership in repository at '/github/workspace'
  To add an exception for this directory, call:
  	git config --global --add safe.directory /github/workspace
  CMake Error at /github/workspace/cmake/FindVersion.cmake:21 (message):
    MRtrix3 base version does not match the git tag!

However, I think this issue is independent of the changes in this PR so it should be safe to merge regardless.

Issue #2868, opened 2024-03-26
GLM: Fixing race condition in Z-transform


Lestropie:
Branching train of thought from #2857 as much of the discussion there relates specifically to the `std::lgamma()` function, whereas the generation of lookup tables for computing Z-scores is an issue in its own right.

From specifically the [latest comment](https://github.com/MRtrix3/mrtrix3/issues/2857#issuecomment-2017121164), I'm considering point 5: doing an up-front generation of the necessary lookup tables before multi-threaded processing, such that those tables can be shared across threads as `const`.

I also think that the heteroscedastic case may require different handling. I need to do some testing to confirm, but I suspect that the Welch-Satterthwaite degrees of freedom are regularly going to not attain precise floating-point equivalence, and therefore generating lookup tables for specific values doesn't make sense. It probably needs a 2D lookup table.

So my current logic is:

- [ ] For no NaNs, homoscedastic errors:
    - [ ] For t-tests:
        - [ ] Generate 1D lookup table: t->Z
    - [ ] For F-tests:
        - [ ] For each unique hypothesis rank:
            - [ ] Generate 1D lookup table: F->Z
- [ ] For NaNs present, homoscedastic errors:
    - [ ] For t-tests:
        - [ ] For every possible DoF following data scrubbing:
            - [ ] Generate 1D lookup table: t->Z
    - [ ] For F-tests:
        - [ ] For every possible DoF following data scrubbing:
            - [ ] Generate 1D lookup table: F->Z
- [ ] For heteroscedastic errors:
    - [ ] For t-tests:
        - [ ] Generate 2D lookup table: (t, dof)->Z
    - [ ] For F-tests:
        - [ ] Generate 2D lookup table: (F, dof)->Z

- [ ] Ensure `vectorstats` tests cover all use cases

- [ ] Ensure pre-generated tables always cover the requirements of subsequent processing

Lestropie:
- [ ] Consider whether F-test lookup tables should be done in logarithmic domain
   F < 1.0 I'm not particularly concerned about loss of precision since statistical enhancement will zero out anything with Z < 0.0; but it might be a better basis nevertheless given the requisite lookup range is quadratically larger than that of t-tests.

Lestropie:
Also of note: If utilisation of `std::lgamma()` is fixed in #2857, could comment `MRTRIX_USE_ZSTATISTIC_LOOKUP` on `dev` until there's greater confidence in operation of the lookup tables from here.

[...]

PR #2269, opened 2021-01-27
Refactoring of statistical inference code

Lestropie:
Posting draft PR as a reminder to myself to redo the requisite tests and merge at some point. Also provides a nice link to have third parties test the code on their data.

Result of hunting for performance issues in `fixelcfestats` trying to do actual experiments of my own. Don't think it's actually yielded anything, but it's nevertheless a good cleanup of #1543. Main changes are to pre-allocate Eigen matrix variables and re-use them across permutations (and also across elements in the case of the variable-GLM-per-element classes), and to have a better classification of what information is shared across threads and what is local.

Includes fix in #2260.

Was what spawned #2259, but as can be seen in the `src/stats/permtest.cpp` changes, those changes aren't actually necessary to pull off the changes here, specifically because the multi-threading back-end copy-constructs the `Stats::PermTest::Processor` / `Preprocessor` classes, not `Math::Stats::GLMBase`.

PR #2874, opened 2024-04-08
GLM: Temporarily disable Z-statistic lookup table use

Lestropie:
Until I can find the time to implement the more exhaustive fix in #2868, I believe that doing this change in conjunction with the merge of #2866 should be adequate to address the threading race condition reported in #2857.

I don't recall exactly how much time efficiency is gained from use of lookup tables for this operation; I'll see if I can quickly rig up a timing test.


Lestropie:
Closing in favour of solution in #2269.

PR #2875, opened 2024-04-10
Use LLVM linker on Linux if available

daljit46:
Enables the use of the LLVM linker by default on Linux platforms. It is much faster than the standard GNU linker and can provide a significant improvement in linking time, especially in incremental builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ file(RELATIVE_PATH relDir
 set(CMAKE_INSTALL_RPATH ${base} ${base}/${relDir})
 
 include(BuildType)
+include(LinkerSetup)
 include(FindFFTW)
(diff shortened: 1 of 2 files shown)


Lestropie:
Does this require explicit test with `cmake` pre & post 3.29, or is that already done / duplicated code from elsewhere that you're confident to merge immediately?
(I'm happy for you to merge if you are, just posing the question)

daljit46:
I think this set of changes is fairly standard (I've seen other projects using something similar). 
For CMake 3.29 (which is already available using `pip install cmake`), I've tested this and it works as intended. Ideally, at some point (probably years ahead I guess) the code for prior versions will be removed.

Lestropie (approved):
Cool; in that case feel free to go ahead.

Issue #2876, opened 2024-04-11
Improving build times


daljit46:
MRtrix3 can take a long time to build. I often like to use a Windows laptop (i5 10th generation quad-core processor) to test changes on MSYS2 and compile times are in the 20-25-minute ballpark. The situation is much better on a Macbook M2 PRO, where I can compile the project in just under 2 minutes. 
Nonetheless, I think it would be worth to trying to improve the compilation time.

Possibly the most taxing factors on build times are:

- Heavy use of C++ templates throughout all the codebase. 
- Use of Eigen.
- Heavy header files that include definitions of functions and classes (at times not necessary).
- Unnecessary includes.

To examine this more carefully, I carried out a [build analysis ](https://github.com/daljit46/mrtrix-build-analysis/blob/main/.github/workflows/build.yml) using [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) and here's the output:

```
**** Time summary:
Compilation (725 times):
  Parsing (frontend):         3927.2 s
  Codegen & opts (backend):   2658.8 s

**** Files that took longest to parse (compiler frontend):
 56624 ms: mrtrix3/build/cmd/CMakeFiles/mrregister.dir/mrregister.cpp.o
 46102 ms: mrtrix3/build/cmd/CMakeFiles/mrtransform.dir/mrtransform.cpp.o
 39887 ms: mrtrix3/build/src/CMakeFiles/mrtrix-headless.dir/registration/transform/initialiser_helpers.cpp.o
 35432 ms: mrtrix3/build/cmd/CMakeFiles/dwidenoise.dir/dwidenoise.cpp.o
 34612 ms: mrtrix3/build/src/CMakeFiles/mrtrix-headless.dir/registration/transform/rigid.cpp.o
 33769 ms: mrtrix3/build/core/CMakeFiles/mrtrix-core.dir/math/average_space.cpp.o
 33294 ms: mrtrix3/build/src/CMakeFiles/mrtrix-headless.dir/registration/transform/affine.cpp.o
 32190 ms: mrtrix3/build/cmd/CMakeFiles/mrmetric.dir/mrmetric.cpp.o
 30567 ms: mrtrix3/build/src/CMakeFiles/mrtrix-headless.dir/registration/nonlinear.cpp.o
 29932 ms: mrtrix3/build/cmd/CMakeFiles/transformcalc.dir/transformcalc.cpp.o

**** Files that took longest to codegen (compiler backend):
[...]

PR #2877, opened 2024-04-15
Optimisation of compilation speed with CMake/Ninja

daljit46:
This PR aims to significantly reduce the time to compile MRtrix3 and address the issue raised in #2876.

I have tested the changes locally on my laptop on Fedora Asahi and Mac OS 14 and, roughly speaking, I see an improvement of almost 30% compared to `dev` (without using `ccache`).

Here is a table showing indicative build times (in seconds) on my Macbook M2 Pro 2023 (I took the best of 3 successive builds).

|                         | dev  | new  |
|-------------------------|------|------|
| MacOS 14 (AppleClang)   | 124s | 89s  |
| Fedora Asahi (clang 17) | 139s | 91s  |
| Fedora Asahi (gcc 13)   | 154s | 110s |

This was achieved with a lot of trial and error and the help of `ClangBuildAnalyzer`. @MRtrix3/mrtrix3-devs it'd be great if you could test this on your machines to see if you can reproduce similar results. One caveat is that to achieve the most optimal results, you must use the latest version of `Ninja`(v 1.12), released last week (see first point below). On MacOS, it's just a matter of running `brew upgrade`. On Linux you will probably need to download it directly from their [Github](https://github.com/ninja-build/ninja/releases) releases page and then run `export PATH=/ninja_dir/:$PATH` (it's unlikely that your distro has packaged the latest version). Before rebuilding, you obviously need to clear your build directory and compiler cache (`ccache -C`).

More in detail, there are three changes in this PR:

- Increased build parallelism. During the initial stages of our CMake effort, it was pointed out by @jdtournier that the CPU load during the build phase didn't stay 100% throughout its entirety. So looking more into this, I found that when building `dev`, the Ninja build graph looks like this:
Screenshot 2024-04-15 at 01 48 55 If you look at the end of the graph, you can see that `mrregister.cpp.o` takes a long time to compile (about 45s) and in fact the rest of the object files finish compiling much earlier. So there is a period when the compiler is effectively operating using a single thread. Ideally, the compilation of `mrregister` should start at the beginning of the compilation phase, however, there is effectively no way to control the order of compilation in CMake (or in Ninja). To get around this, I used a simple trick (see `cmd/CMakeLists.txt`): create a dummy executable and then add `mrregister` as a dependency of this dummy ([this](https://github.com/ninja-build/ninja/pull/2177) is the relevant change in Ninja 1.12 that makes this possible). This is enough to trick Ninja and change the build graph to the following:
Screenshot 2024-04-15 at 01 53 52 Now, you see that we are operating multi-threaded throughout the entire compilation phase. 

- Precompiled headers. Precompiled headers speed up compilation by storing parsed header information in a special file. During subsequent compilations, the compiler reuses this precompiled data instead of re-parsing headers. They are particularly suited for "stable" header files, i.e. files that do not change very often (e.g. STL headers). If files listed in a precompiled header file change, then a target using the precompiled header will need to be recompiled from scratch. In this PR, I only included the following headers (other than STL and third parties) in precompiled headers:  `gl.h` and `exception.h`. The first one is only used for `mrview/shview` and the second one is for all the code. If a dev needs to frequently edit these files, it'd be best to turn off `MRTRIX_USE_PCH` as otherwise, any change in those files will trigger a rebuild of the whole project.

- Explicit template instantiations. C++11 introduced the `extern template` syntax, which lets you say, "Hey, the template's definition? It's elsewhere!". So if we use this in a header file, we can tell the compiler to avoid automatically instantiating a specific template in *every* .cpp file that includes the header (of course we have to make sure that the template is instantiated somewhere, e.g. in the corresponding .cpp for the header).
After looking at the report by `ClangBuildAnalyzer`, I used this feature to explicitly instantiate some class and function templates to reduce compilation time. I did this fairly conservatively (I really don't like dealing with the complexity that templates bring), but the few changes I made still made a significant difference in build time.


daljit46:
Managed to obtain further improvements by enabling PCHs for executables too (previously they were only used for the shared libraries.
The new timings look like below

|                         | dev  | new  |
|-------------------------|------|------|
| MacOS 14 (AppleClang)   | 124s | 84s  |
| Fedora Asahi (clang 17) | 139s | 94s  |
| Fedora Asahi (gcc 13)   | 154s | 103s |

cmd/CMakeLists.txt
```diff
@@ -18,6 +26,10 @@ function(add_cmd CMD_SRC IS_GUI)
         LINK_DEPENDS_NO_SHARED true
         RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin
     )
+    # Don't use PCHs for heavy commands, as we want start their compilation as soon as possible
+    if(MRTRIX_USE_PCH AND NOT ${CMD_NAME} IN_LIST HEAVY_CMDS AND NOT ${IS_GUI})
+	target_precompile_headers(${CMD_NAME} REUSE_FROM pch_cmd)
```
Lestropie, line 31:
```suggestion
        target_precompile_headers(${CMD_NAME} REUSE_FROM pch_cmd)
```

core/core_pch.h
```diff
@@ -0,0 +1,3 @@
+#include "exception.h"
```
Lestropie, line 1: Should decide whether inclusion of MRtrix header files here is to be based on pragmatism (ie. what provides the greatest speedup vs. is least likely to be modified) vs. some sort of principle (ie. what are "the most fundamental" core header files). There might be an argument to be made for `core/*.h` (ie. root of `core/` but not its sub-directories). But I'm not currently able to form a robust opinion on this.

core/image.h
```diff
@@ -477,4 +477,6 @@ template <class ImageType> typename enable_if_image_type<ImageType, void>::type
     WARN(std::string("error invoking viewer: ") + strerror(errno));
 }
 
+extern template MR::Image<bool>::~Image();
+extern template MR::Image<float>::~Image();
```
Lestropie, line 481: These are probably the two most frequent, but would it make more sense to be exhaustive here?

I'm also thinking about the prospect of external projects compiling against core. Previously, straight use of templates would allow such projects to instantiate whatever alternative template usages they wished. We've not yet fleshed out exactly how external projects will work under `cmake`, but should be wary to not break such. With the syntax I've historically used myself for explicit instantiation of templates, that list of instantiations would need to be exhaustive; I'm inferring from this syntax that a similar problem *doesn't* occur here? Ie. The `extern template` says that that specific instantiation appears in the external file, but any other template instantiation can occur within any compilation job?

src/dwi/tractography/roi.cpp
```diff
@@ -118,3 +118,5 @@ Image<bool> Mask::__get_mask(const std::string &name) {
 }
 
 } // namespace MR::DWI::Tractography
+
+template MR::Transform::Transform(const MR::DWI::Tractography::Mask &);
```
Lestropie, line 122: I think this one should be possible to avoid. `MR::DWI::Tractography::Mask` inherits from `Image<bool>`, which has a `transform()` member; it's just trying to pre-compute the inverse transform and share it across threads (which probably isn't worth the expense of de-referencing a shared pointer to save memory given it's just a 4x4 matrix). Having `Mask` derive from *both* `Image<bool>` *and* `MR::Transform` might make more sense?

Lestropie:
Are you able to glean insight into what it is that makes the `mrregister` / `mrtransform` compilation so long? That would potentially be the place to be looking into precompiled headers / precompiled template instantiations.

core/core_pch.h
daljit46: I think a nice balance between stability and speed gain should prioritised. Additionally, we should probably include header files that are likely to affect most of the codebase.  For example, this was the reasoning to add `#include "exception.h", since it showed up in `ClangBuilderAnalyzer`'s report and I think there are little chances for a developer to modify this header. Furthermore, this header is indirectly included in the majority of our codebase and thus even without precompiled headers, its modification will trigger a recompilation of a substantial chunk of the project (on this point, it's worth noting that you can always turn off PCHs by setting `cmake -DMRTRIX_USE_PCH=OFF`.
Overall, I do think that we should be very conservative with the inclusion of user headers in a PCH.

core/image.h
daljit46: > These are probably the two most frequent, but would it make more sense to be exhaustive here?

Yes, I guess that would be sensible although I'm not sure whether it would provide any build time benefits. Perhaps it makes sense to add one for `double` or are there others that you were thinking about?

> any other template instantiation can occur within any compilation job?

Yes, that's correct.

daljit46:
> Are you able to glean insight into what it is that makes the `mrregister` / `mrtransform` compilation so long? That would potentially be the place to be looking into precompiled headers / precompiled template instantiations.

Looking at the output of `ClangBuildAnalyzer`, for `mrregister` it shows that only 20% of the time is spent in the compiler's front-end (parsing and instantiation templates), while the rest of the time the compiler is busy optimising the code.
Hence I think the best course of action would be to split this file into multiple files (or at least move parts of its content somewhere else) so that the compiler can operate multithreaded.

src/dwi/tractography/roi.cpp
daljit46: > Having Mask derive from both `Image<bool>` and `MR::Transform` might make more sense?

That would involve replacing all instances of `MR::Transform<MR::DWI::Tractography::Mask>::Transform(const MR::DWI::Tractography::Mask&)` with just `MR::DWI::Tractography::Mask()` right?

daljit46:
Further finding on `mrregister`: on my desk PC (a 20-core Intel(R) Xeon(R) CPU E5-2687W v3 @ 3.10GHz), the compilation of `mrregister.cpp.o` takes more than the time required to build all other the object files in the project!

core/image.h
Lestropie: - The fixel format uses `Image<uint32_t>` for the index image; `label`-related commands will also use `Image<uint32_t>`.
- `Image<double>` is probably used somewhere but nothing's coming to mind.
  (Arguably should be used in some locations where scratch images are used for intermediate results; and ideally in my mind, for many workhorse commands, if any input data are double precision then the main processing task should be templated to operate in double precision)
- I see `Image<uint16_t>` used in `dwidenoise`.
- `mrcalc` uses `Image<cfloat>`; I see `mrfilter` using `Image<cdouble>`.

Possibly what would make sense would be to instantiate all possibilities that are reflected in `MR::DataType`. That would intrinsically communicate the *intended* scope of template types (which has caused problems in the couple of instances where it's been violated), and would hopefully mean that if any external project were to make use of the shared library than any reasonably expected image type should already be there.

daljit46: I don't think that `MR::DataType` would be a suitable place to instantiate these templates. Note that we are not instantiating the types `MR::Image<float>` or `MR::Image<bool>`, but the *destructors* of those classes (odd I know, but `ClangBuildAnalyzer` showed that they took a long time to instantiate). I think it would look very weird if the destructor of a class appeared to be in a completely unrelated source file from its header.

Lestropie: Not suggesting that the relevant code should appear in `MR::DataType`, rather that the set of templates instantiated for `MR::Image<>` should echo the data type possibilities enumerated in `MR::DataType` (omitting endianness). Both are then exhaustive of the possible data types that could be used to store image data.

So:
- `bool`
- `int8_t`
- `uint8_t`
- `int16_t`
- `uint16_t`
- `int32_t`
- `uint32_t`
- `int64_t`
- `uint64_t`
- `half`
- `float`
- `double`
- `cfloat`
- `cdouble`

src/dwi/tractography/roi.cpp
Lestropie: I'll try to modify this myself on `dev` with a separate PR. Just won't happen this week.

cmd/CMakeLists.txt
```diff
@@ -5,6 +5,22 @@ foreach(CMD ${GUI_CMD_SRCS})
...
+    target_include_directories(pch_cmd PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../core)
+    find_package(Eigen3 REQUIRED)
+    target_link_libraries(pch_cmd PRIVATE Eigen3::Eigen)
+    target_precompile_headers(pch_cmd PRIVATE
+        [["app.h"]]
+	[["image.h"]]
+	<complex>
+	<string>
+	<vector>
+    )
```
Lestropie, line 20:
```suggestion
    target_precompile_headers(pch_cmd PRIVATE
        [["app.h"]]
        [["image.h"]]
        <complex>
        <string>
        <vector>
    )
```

core/CMakeLists.txt
```diff
@@ -63,6 +63,13 @@ target_include_directories(mrtrix-core-version-lib PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
+if(MRTRIX_USE_PCH)
+    target_precompile_headers(mrtrix-core PRIVATE
+	[["exception.h"]]
+	<fstream>
+    )
```
Lestropie, line 70:
```suggestion
    target_precompile_headers(mrtrix-core PRIVATE
        [["exception.h"]]
        <fstream>
    )
```

src/CMakeLists.txt
```diff
@@ -58,6 +58,14 @@ target_include_directories(mrtrix-headless PUBLIC
     ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
+if(MRTRIX_USE_PCH)
+    target_precompile_headers(mrtrix-headless PRIVATE 
+	[["exception.h"]]
+	<fstream>
+	<complex>
+    )
```
Lestropie, line 66:
```suggestion
    target_precompile_headers(mrtrix-headless PRIVATE 
        [["exception.h"]]
        <fstream>
        <complex>
    )
```

Lestropie:
`MR::Registration::Transform::Base::Base(size_t)` should be easy to fix the `clang-tidy` warnings; those `Eigen` member variables have corresponding initialisation syntax. Not essential for merge, just what I saw when reviewing.

I would have just pushed the change but I wasn't able to verify compilation because of:

```
robertes@9520l-003953-l:~/src/mrtrix3$ cmake --build build_debug/
[29/498] Linking CXX executable cmd/pch_cmd
FAILED: cmd/pch_cmd 
: && /usr/bin/c++ -g -fuse-ld=lld cmd/CMakeFiles/pch_cmd.dir/pch_cmd.cpp.o -o cmd/pch_cmd   && :
ld.lld: error: undefined symbol: MR::File::Config::config[abi:cxx11]
>>> referenced by config.h:34 (/home/unimelb.edu.au/robertes/src/mrtrix3/cmd/../core/file/config.h:34)
>>>               cmd/CMakeFiles/pch_cmd.dir/pch_cmd.cpp.o:(MR::File::Config::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&))
>>> referenced by config.h:35 (/home/unimelb.edu.au/robertes/src/mrtrix3/cmd/../core/file/config.h:35)
>>>               cmd/CMakeFiles/pch_cmd.dir/pch_cmd.cpp.o:(MR::File::Config::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&))
collect2: error: ld returned 1 exit status
[50/498] Building CXX object src/CMakeFiles/mrtrix-headless.dir/dwi/fmls.cpp.o
ninja: build stopped: subcommand failed.
```

src/registration/transform/base.h
```diff
@@ -222,33 +173,15 @@ class Base {
...
+  void compute_halfspace_transformations();
 
   size_t number_of_parameters;
-  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo;
-  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo_half;
-  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo_half_inverse;
-  Eigen::Vector3d centre;
+  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo{};
+  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo_half{};
+  Eigen::Transform<ParameterType, 3, Eigen::AffineCompact> trafo_half_inverse{};
```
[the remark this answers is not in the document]
daljit46: I'm not sure about this check. While I understand that having classes with private *and* public data members can be undesirable, I don't really find this to be a huge concern. Especially in a codebase like ours.

core/app.cpp
```diff
@@ -293,6 +293,17 @@ std::string usage_syntax(int format) {
   return s + "\n\n";
 }
 
+Description &Description::operator+(const char *text) {
+  push_back(text);
+  return *this;
+}
+
+Description &Description::operator+(const char *const text[]) {
+  for (const char *const *p = text; *p != nullptr; ++p)
```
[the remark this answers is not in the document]
daljit46: This should be addressed in some future PR that tries to get rid of all uses of `char*`.

Lestropie:
Windows 11 machine through MSYS2, 16 cores, Ninja 1.12 (already available on MSYS2), average of 3 runs:

|       | `dev`  | new  |
|-------|--------|------|
| GCC   |   202s | 118s |
| Clang |   183s | 106s |

Can't currently test on my work Linux laptop as it's partway through a script doing 72 `bedpostx` runs and I don't particularly want to interrupt it.

I agree regarding abuse of inlining / too many implementations in header files / would benefit from use of `extern template`. Chances are most of the benefits have already been obtained here, and there may be somewhat diminishing returns hunting for more. So I'm not sure it's worth delaying the merge. I will myself try to keep in mind whenever any given area of code is under modification to consider rectifying such things on a case-by-case basis.

Lestropie (approved):
Merge conflicts resolved; should be good to go. Nice work!

PR #2878, opened 2024-04-15
Fix typo in LinkerSetup.cmake

daljit46:
Fixes. a typo introduced in #2875


Lestropie:
*slaps own hand for bad code review*

PR #2879, opened 2024-04-18
Don't use std::move for return values

daljit46:
It prevents NRVO (Named Return Value Optimisation) and in fact can be a pessimisation. See [here](https://devblogs.microsoft.com/oldnewthing/20231124-00/?p=109059).


Lestropie:
Is this just a set of instances that you yourself noticed, or was it a clang tidy modify across the repo?
(Not asking because of the prospect of re-authoring to MRtrixBot---that's only for instances where the changeset is disproportionate to the effort invested---asking because if it's not yet a repo-wide check, maybe we should do that)

daljit46:
I grepped for "return std::move" and these are the only instances in the codebase I could find. There are two more in `core/file/json/json.h` but since that's a third party header file, it wouldn't be wise to change it.
Clang-tidy didn't produce any warning on my local system (although I did expect it too, but there are may be very rare cases where the use of `std::move` is appropriate).

Issue #452, opened 2016-02-26
Alert user when switching between images with lock to grid on.


draffelt:
To further an email discussion with @jdtournier, and an in-house discussion with @thijsdhollander and @Lestropie, it would be nice to alert users when they switch between two images that have grids with different orientations wrt to scanner coordinates.

A lot of users don't appreciate the difference between scanner space and image space, and that the FOV can be differently rotated with respect to scanner space for different images acquired in the same session.  Furthermore I've seen users that do understand this get caught up on why two images of different modalities are not aligned after pre-processing when opening in `mrview`,  because they did not realise lock to axis was on (or what this button even does).

We don't want to have lock-to-axis off by default, since it would require full load of images that are rendered off axis (wrt to the first image). And as @Lestropie pointed out, users may also just want to see the image as it was acquired, and not with respect to another image that is currently loaded. 

So, we have been discussing the option of somehow alerting the user that when switching between two images, they are not in the same space. 

Options so far include:

* Pop-up a box with a text warning, possibly up the top of the view right underneath the lock-to-axis button. This would then fade away fairly quickly. 
* Providing a render of the scanner-space axis (like the ODF preview tool) that then changes when switching between views. I'm not sure where this would go, or if it might clutter the viewer for every day usage. Also, users may not notice subtle changes between grid orientations so it also may require some highlighting upon change.
* wobble the new image (upon changing). This then makes it hard to flick between two images quickly to compare unless lock-to-axis is off. The downside is that most users may not immediately understand what the wobble is for. Note that I've seen fslview do something similar (on a colleagues machine), but I'm not sure what the wobble indicates.

jdtournier:
I was originally in favour of the axes render, just as a little glyph in the corner somewhere (although we're running out of corners...), but it's true that for subtle changes, that wouldn't be noticeable. Besides, the orientation labels at the edges of the display already serve that purpose, and clearly aren't visible enough for people to notice (maybe we could change the colour of those labels to something more obvious?). A full blown pop-up would be too disruptive, but how about a little one liner at the top of the display with a temporary message in a red font, rendered the same way as the header info, etc? Just have it pop up for 3 seconds with the message "locking to image axes" or something...?

Lestropie:
One idea I had in this context that I thought I'd write down:

[...]

PR #2882, opened 2024-04-22
Docs: New page on MRView "lock to image axes"

Lestropie:
This is a long-standing confound with image visualisation that we have regularly attempted to describe but never invested the effort in properly documenting and explaining. Possibly because #452 would be a whole lot more fun. But for a pending article submission with @jdtournier it would be beneficial to be able to provide a hyperlink to a documentation page where the issue is explained, rather than trying to shoehorn in there an inadequate explanation.

Rendered docs link:
https://mrtrix3-dev.readthedocs.io/en/docs_mrview_lock_axes/concepts/mrview_lock_axes/mrview_lock_axes.html


jdtournier (approved):
Looks great, thanks for putting in the effort! It was long overdue... 

I just fixed a couple of typos, but otherwise it's really nice work!

Lestropie:
Requires review by a third party: @jdtournier's review does not satisfy requirement for merge due to being the author of the last commit.

daljit46 (approved):
Small typo on line 48 I think

Issue #1373, opened 2018-06-27
mrview on Windows: fix drag & drop


jdtournier:
As reported [on the forum](http://community.mrtrix.org/t/windows-mrview-dragndrop/1730):

> A small windows-specific issue, it seems like drag and drop does not work in mrview (RC2 and RC3) due to an additional “/” in front of the file specifier (e.g. “/C:/pathtofile/file.nii.gz”) leads to “error opening image…”
> However, I can easily load the file via the Open-dialog.

Ulquoirra-rn:
Hey @jdtournier was this issue fixed?

steso:
I just rebuild my installation (3.0.8-29-ga4077b82) and tested the drag & drop again and I still get the leading "/" which doesn't work on windows in front of the drive. Either it's not in the newest build / current branch or it's not fixed yet?

Lestropie:
#2886 applied the fix to the `dev` branch, not `master`. I noted in that thread that it needs to be back-ported to `master`, so perhaps while the PR was fine to merge this Issue should not have been closed. I'll reopen and try to get to back-porting.

steso:
thank you for fixing this bug, I can confirm after pulling and building drag and drop is now also possible on windows systems!

PR #2886, opened 2024-04-25
Fix drag & drop on Windows for mrview

daljit46:
Should address #1373

### DIFF
--- a/src/gui/mrview/qthelpers.cpp
+++ b/src/gui/mrview/qthelpers.cpp
@@ -1,0 +1,33 @@
+/* Copyright (c) 2008-2024 the MRtrix3 contributors.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Covered Software is provided under this License on an "as is"
+ * basis, without warranty of any kind, either expressed, implied, or
+ * statutory, including, without limitation, warranties that the
+ * Covered Software is free of defects, merchantable, fit for a
+ * particular purpose or non-infringing.
+ * See the Mozilla Public License v. 2.0 for more details.
+ *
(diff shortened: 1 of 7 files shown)


Lestropie (changes requested):
Confirmed resolves issue on my Windows machine. But only for specifically dragging & dropping main images. There are also toolbars where a user may attempt to drag & drop a file, which this changeset currently won't resolve. Would suggest functionalising the decoding of filesystem paths from a drop event, and make use of it across those multiple contexts:

- Overlay tool
- Connectome tool
- Fixel plot
- Tractography tool
- ROI editor

Note that the ODF display tool is intentionally omitted from this list: this is because the *type* of ODF cannot be reliably inferred from just the input image data alone, and so the user is forced to nominate the type of ODF by clocking on the corresponding button.

daljit46:
To address your comment, I've now introduced a new `MR::GUI::MRView::QtHelpers` namespace which contains a `url_to_std_string` helper function that is used when handling drop events. 
One thing I'm not sure is where should `qthelpers.h` should reside. Currently, it's in `src/gui/mrview` but perhaps it makes more sense in `src/gui`?

Lestropie:
-   Tried to test, but for some reason `libmrtrix-gui.dll` was not built. Fresh build directory, QT5. Will continue trying.
-   I think that this would be better targeted at `master` rather than `dev` given that it is a bug fix.

Lestropie (approved):
Fixed non-compilation by bringing up to date with `dev`.

Tested drag & drop on the listed toolbars.

Do maintain this should be back-propagated to `master`. But given the code formatting changes would cause conflicts across the board, could maybe merge this, manually re-implement on `master`, and then on `dev` accept only the `dev`-based change?

daljit46:
> Do maintain this should be back-propagated to `master`. But given the code formatting changes would cause conflicts across the board, could maybe merge this, manually re-implement on `master`, and then on `dev` accept only the `dev`-based change?

Sounds good to me.

PR #2887, opened 2024-04-25
Fix launching mrview from build directory

daljit46:
Changes output location of `mrtrix-gui` to ensure that mrview launches correctly in a MSYS2 environment.


src/CMakeLists.txt
```diff
@@ -87,6 +81,18 @@ if(MRTRIX_BUILD_GUI)
...
 endif()
 
+# On Windows, the libraries need to be in the same directory as the executables
+if(WIN32)
+    set_target_properties(mrtrix-headless PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin
+    )
+    if(MRTRIX_BUILD_GUI)
+	set_target_properties(mrtrix-gui PROPERTIES
+	    RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin
```
Lestropie, line 91:
```suggestion
        set_target_properties(mrtrix-gui PROPERTIES
            RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin
```

Lestropie:
On my own Windows system, running `mrview` without this change does nothing; no window opens, and control returns to the terminal. However I can't confirm that this change results in successful execution, as on my own system `mrview` immediately experiences a segmentation fault. And I'm having other issues trying to compile in debug mode to diagnose. Content to have this merged if you're confident in the change, I just can't confirm the fix myself right now.

daljit46:
> On my own Windows system, running mrview without this change does nothing; no window opens, and control returns to the terminal.

Perhaps #2884 also needs to be applied?

Lestropie (approved):
Confirmed resolution if used in combination with #2884; that's now merged to `dev`, so this will now join it.

PR #2889, opened 2024-04-26
Fix data race in ordered_queue tests

daljit46:
It was caused accessing the global variable *num_items* in the constructor of an `Item` sent to the queue. This is not permissible since items are potentially accessed in parallel by separate threads.

This should be the last item remaining issue before closing #2772 and merging #2763.


Lestropie:
Actually, this one should be made to `master` also; it should be included in `3.0.5`.

PR #2768, opened 2023-12-13
Reslice adapter: allow use with complex numbers

jdtournier:
Simple fix to enable `Adapter::Reslice` to work as expected. Required for complex number support in `mrpeek` (PR: mrtrix3/mrpeek#47).

PR #2891, opened 2024-05-02
Adapter::Reslice: fix handling of complex types

jdtournier:
Not sure whether this is required, but it caused trouble for me at some point. I suggest leaving as a draft for now until the problem comes up again...


Lestropie:
Suspected duplicate of #2768; I posted a suggested alternative as #2785.

jdtournier:
Well spotted! OK, I'll close this and #2768 - I prefer your approach in #2785.

PR #2893, opened 2024-05-07
Master to dev

Lestropie:
Attempting to merge the latest changes on `master` onto the `dev` branch.
Will need to cross-check the commits with the proposed changes, make sure that merge commit 3d4f392a75672040afeb4fb55e12f54da0a8fa17 doesn't accidentally revert anything.

For the record, I accidentally created merge commit 3d4f392a75672040afeb4fb55e12f54da0a8fa17 on my local `dev` branch, pushed it, and GitHub actually accepted it, with the following message:
```
remote: Bypassed rule violations for refs/heads/dev:
remote:
remote: - Changes must be made through a pull request.
remote:
```

I then *couldn't* force-push GitHub's `dev` back one commit because of being locked against force pushes. For now, I've gotten around this by giving myself and @jdtournier the ability to force push to `dev`. However GitHub should not have accepted that commit in the first place according to the branch settings.


Lestropie:
I've done a manual cross-reference of the changeset against the underlying commits, and found one small glitch from manual merge conflict resolution in `python/bin/mrtrix3.py` (which should disappear soon on `dev` anyways as part of #2741). Going to merge this to allow its merge into other PRs to unblock them, since this addresses CI failures on `dev`.

PR #2902, opened 2024-05-13
dwishellmath: Enhancements

Lestropie:
Am currently in the process of generating data & analyses for verifying the handling of DWI metadata, both raw and derivative, both slice / phase encoding and gradient table (especially with the pesky `bvec` esoterics). Am aiming particularly for resolution of #2477 and for verifying correct interpretation of subsequently derived diffusion model fibre orientations in https://github.com/bids-standard/bids-bep016.

A known limitation of the original implementation of `dwigradcheck` is that it considered exclusively axis permutations and flips. It failed to consider the prospect of a historically common misunderstanding where a user believed `dw_scheme` and `bvec`/`bvals` to be merely transposes of one another, and would therefore apply that operation erroneously. Here I've finally gotten around to addressing this. In retrospect it was not the most crucial of enhancements for me to put effort into, given that the dataset I'm working with always has image axes aligned with the scanner axes, but it was nevertheless annoying that the command was doing twice as much work as it needed to in that case. Plus I wanted to have the command emit a  non-zero return code if it considers the input gradient table to be incorrect, as this allows me to use it for automated verification of metadata handling.

### DIFF
--- a/docs/reference/commands/dwigradcheck.rst
+++ b/docs/reference/commands/dwigradcheck.rst
@@ -22,7 +22,9 @@ Description
 
 Note that the corrected gradient table can be output using the -export_grad_{mrtrix,fsl} option.
 
-Note that if the -mask command-line option is not specified, the MRtrix3 command dwi2mask will automatically be called to derive a binary mask image to be used for streamline seeding and to constrain streamline propagation. More information on mask derivation from DWI data can be found at the following link: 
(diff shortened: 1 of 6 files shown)


dchristiaens:
I presume the title should say "dwigradcheck" instead of "dwishellmath"?

Lestropie:
🤦‍♂️

PR #2905, opened 2024-05-15
DICOM command-line selection: abort on error reading from std::cin

jdtournier:
Encountered issues when running DICOM tests: sending EOF through to stdin (e.g. via Ctrl-D) would cause infinite loop.


Lestropie:
Closing; will be addressed once #2908 is included in a merge from `master` back to `dev` (which will likely be manual, but we'll catch it).

PR #2906, opened 2024-05-15
dwidenoise: try using bidiagonal divide and conquer SVD

jdtournier:
This changes the mode of operation from our current fast, but less numerically stable eigenvalue decomposition approach to the fast [bidiagonal divide & conquer SVD implementation](https://eigen.tuxfamily.org/dox/classEigen_1_1BDCSVD.html) now provided with Eigen 3.4. Seems to work on my limited testing, producing results that are indistinguishable (though not identical) from the current implementation on my test data (BATMAN tutorial data). 

However, this comes with a considerable additional computation cost. On my system (Arch Linux, AMD Ryzen 9 3950X 16-Core Processor, 64GB RAM), these are the runtimes:

- current implementation (float): ~26s
- current implementation (double): ~49s
- this implementation (BDC-SVD, float): ~1m03s

Thoughts welcome.

### DIFF
--- a/cmd/dwidenoise.cpp
+++ b/cmd/dwidenoise.cpp
@@ -149,6 +149,7 @@ template <typename F = float> class DenoisingFunctor {
 
 public:
   using MatrixType = Eigen::Matrix<F, Eigen::Dynamic, Eigen::Dynamic>;
+  using VectorType = Eigen::Matrix<F, Eigen::Dynamic, 1>;
   using SValsType = Eigen::VectorXd;
 
   DenoisingFunctor(int ndwi,
@@ -164,8 +165,9 @@ public:
         q(std::max(m, n)),
         exp1(exp1),
         X(m, n),
-        SVD(q, r, Eigen::ComputeThinU | Eigen::ComputeThinV),
+        SVD(q, r, Eigen::ComputeThinU),
         s(r),
+        v(r),
         pos{{0, 0, 0}},
         mask(mask),
         noise(noise),
@@ -211,10 +213,11 @@ public:
     }
 
     if (cutoff_p > 0) {
-      // recombine data using only eigenvectors above threshold:
       const size_t n_sv = r - cutoff_p;
-      X.col(n / 2) = SVD.matrixU().leftCols(n_sv) *
-                     (SVD.singularValues().head(n_sv).asDiagonal() * SVD.matrixV().row(n / 2).head(n_sv).adjoint());
+      // calculate central column of S.V'
+      v = SVD.matrixU().adjoint() * X.col(n / 2);
+      // recombine data using only eigenvectors above threshold:
+      X.col(n / 2) = SVD.matrixU().leftCols(n_sv) * v.head(n_sv);
     }
 
     // Store output
@@ -240,6 +243,7 @@ private:
   MatrixType X;
   Eigen::BDCSVD<MatrixType> SVD;
   SValsType s;
+  VectorType v;
   std::array<ssize_t, 3> pos;
   double sigma2;
   Image<bool> mask;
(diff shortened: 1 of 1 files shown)


dchristiaens:
I've tested this on the Workshop data (single subject Siemens). On these data too, the denoised data and noise level maps are nearly identical. My timings on these data are similar to yours:
- current implementation (float): ~26s
- current implementation (double): ~49s
- JacobiSVD (float): ~4m45s
- JacobiSVD (double): ~10m16s **(baseline)**
- BDC-SVD (float): ~1m11s
- BDC-SVD (double): ~2m14s

The JacobiSVD in double precision should have the best numerical stability of all tested methods. I therefore calculated the error of the current implementation and the BDC-SVD implementation w.r.t. this baseline. For the noise level, the mean absolute and relative errors are:
- current implementation (float): 2.7e-3 +/- 4.5e-2 (absolute) and 3.6e-5 +/- 5.7e-4 (relative)
- BDC-SVD (float): 1.2e-4 +/- 9.8e-3 (absolute) and 1.2e-6 +/- 1.0e-4 (relative)

This means that the result of BDC-SVD is one order of magnitude more accurate than the current EVD-based implementation in single precision. The EVD in double precision leads to an absolute error that is exactly 0, but this isn't a fair comparison because the input data are in single precision. These results confirm the theory I seem to remember from my numerical linear algebra class ages ago :wink: 

Based on this, I'd be in favour of switching to BDC-SVD. I think the longer run time is a fair price to pay for accuracy in single precision, and it's not that much longer than using the current implementation in double precision. I'd then keep float32 as the default, except perhaps when the input data already has datatype float64.

Lestropie:
Given the relatively small differences in code, would be not too difficult to template by decomposition method and provide a command-line option.

dchristiaens:
I've made a change to the code to avoid calculating the full (thin) matrix V. Only the central column of S.V' is needed to recombine the data, so we can take advantage of that to save some computation time. I've tested this for cases M<N and M>N; the output is identical to Donald's BDC-SVD implementation. I've also tried calculating only the non-truncated part of this vector, but this had no significant effect on the run time and would impede adding other shrinkage functions later.

The run time is ~20% faster:
- previous BDC-SVD (float): ~1m11s
- previous BDC-SVD (double): ~2m14s
- new BDC-SVD (float): ~57s
- new BDC-SVD (double): ~1m51s

Lestropie:
RE [previous comment](https://github.com/MRtrix3/mrtrix3/pull/2906#issuecomment-2115232759):

> Given the relatively small differences in code, would be not too difficult to template by decomposition method and provide a command-line option.

I wouldn't go to those lengths here. In `dwidenoise2` I have the ability to switch between decomposition methods. But I'd prefer any investment in capability or flexibility to go there rather than `dwidenoise` given the impact of the aggregation of many enhancements (see [demonstration](https://github.com/Lestropie/dwidenoise2)).

Personally I'm sceptical that the ~ 2x execution time moving from the current implementation to BDC-SVD *in this context* is justified. In the original post it was noted that the results were "indistinguishable" from the current implementation. And it has not been shown that this specific change is successful in rectifying any reported problem. So a greater precision with respect to a Jacobi decomposition may actually convey a false claim of improvement in outcomes. In contrast `dwidenoise2` does yield superior results, but because it changes a lot more than just the decomposition algorithm.
(`dwidenoise2` does use BDC-SVD, but it can afford to for the sake of precision: the runtime is still shorter than current `dwidenoise` despite doing 4 iterations)

I'd be *content* to bring this up to date and merge into `dev`, but my personal choice would be to leave `dwidenoise` alone, and if permitted even add a `CONSOLE()` level message to `dwidenoise` for 3.1.0 indicating that `dwidenoise2` may be the superior choice if the user is willing to go to the effort of installing.

PR #2907, opened 2024-05-15
New option to build non-core code as a static lib

daljit46:
As mentioned in the meeting today, it'd make sense for production builds to have the C++ code in `src` built as a static library. This PR adds a new `MRTRIX_BUILD_STATIC` option to allow for this (@MRtrix3/mrtrix3-devs if anyone has a better name please let me know).
The option is disabled by default and is only intended to be enabled in production and external projects. For every day builds, making shared libraries provides a faster incremental build workflow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ option(MRTRIX_STL_DEBUGGING "Enable STL debug mode" OFF)
 option(MRTRIX_BUILD_TESTS "Build tests executables" OFF)
 option(MRTRIX_STRIP_CONDA "Strip ananconda/mininconda from PATH to avoid conflicts" ON)
 option(MRTRIX_USE_PCH "Use precompiled headers" ON)
-option(MRTRIX_BUILD_STATIC "Build MRtrix's non-core code as a static library" OFF)
+option(MRTRIX_BUILD_NON_CORE_STATIC "Build MRtrix's non-core code as a static library" OFF)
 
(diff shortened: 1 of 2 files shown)


Lestropie (changes requested):
Seems a simple enough modifier to have in place regardless of whether it actually gets used for any end purpose.

Regarding the name, it probably needs to be clear that it's not *the entire build* that's static, only that outside of core, especially given earlier builds have had full static build capability (though from memory it's been a little patchy at times).

daljit46:
> Regarding the name, it probably needs to be clear that it's not the entire build that's static, only that outside of core, especially given earlier builds have had full static build capability (though from memory it's been a little patchy at times).

Unfortunately, I don't have good ideas for the name (neither does ChatGPT). The best I could come up with is `MRTRIX_BUILD_NON_CORE_STATIC`.

Lestropie:
If ChatGPT can't even come up with variable names, is it even fit for purpose? I mean, that's 98% of the work to be done. 🤡  No clear superior alternative to `MRTRIX_BUILD_NON_CORE_STATIC` from me. Will take the slight clumsiness for the sake of accuracy.

daljit46:
@Lestropie can we merge this?

Lestropie (approved):
I'm not very familiar with static builds so can't really feedback on the utility of the change, but as far as the change itself goes I see no issue with merging.

PR #2911, opened 2024-05-21
Remove uses of char* in app CLI

daljit46:
This PR aims to replace uses of C-style `const char*` arguments with `std::string` in our CLI. C-style arrays for list of arguments/options are also replaced in favour of `std::vector<std::string>`. Partially fulfils one of the objectives in #2585 and #2111.


daljit46:
@Lestropie @jdtournier I will need your help with migrating this part of the code in `src/gui/mrview.cpp:721-755`:
```
  if (!MR::App::argument.empty()) {
    if (!MR::App::option.empty()) {
      // check that first non-standard option appears after last argument:
      size_t last_arg_pos = 1;
      for (; MR::App::argv[last_arg_pos] != MR::App::argument.back().c_str(); ++last_arg_pos)
        if (MR::App::argv[last_arg_pos] == nullptr)
          throw Exception("FIXME: error determining position of last argument!");

      // identify first non-standard option:
      size_t first_option = 0;
      for (; first_option < MR::App::option.size(); ++first_option) {
        if (size_t(MR::App::option[first_option].opt - &MR::App::__standard_options[0]) >=
            MR::App::__standard_options.size())
          break;
      }
      if (MR::App::option.size() > first_option) {
        first_option = MR::App::option[first_option].args - MR::App::argv;
        if (first_option < last_arg_pos)
          throw Exception("options must appear after the last argument - see help page for details");
      }
    }

    std::vector<std::unique_ptr<MR::Header>> list;
    for (size_t n = 0; n < MR::App::argument.size(); ++n) {
      try {
        list.push_back(std::make_unique<MR::Header>(MR::Header::open(MR::App::argument[n])));
      } catch (CancelException &e) {
        for (const auto &msg : e.description)
          CONSOLE(msg);
      } catch (Exception &e) {
        e.display();
      }
    }
    add_images(list);
  }
```

Lestropie:
Message "`"options must appear after the last argument - see help page for details"`" isn't quite right; it's specifically *mrview-specific` options, ie. *not* standard options, that must appear after all positional arguments *and* standard options (at least from looking at what the code is doing).

Lestropie:
Decision during yesterday's meeting was that the scope of this PR would be limited to the CLI; while there are uses of `char*` elsewhere in the code base, it would be better to address those separately in the future.

cmd/connectome2tck.cpp
```diff
@@ -39,7 +39,7 @@ using namespace MR::DWI;
 using namespace MR::DWI::Tractography;
 using namespace MR::DWI::Tractography::Connectome;
 
-const char *file_outputs[] = {"per_edge", "per_node", "single", NULL};
+const std::vector<std::string> file_outputs = {"per_edge", "per_node", "single"};
```
Lestropie, line 42: Is `std::vector<std::string>` the best container to use here? These never change dynamically. I wonder if `std::array<std::string>` would be more appropriate? The code that resolves the user input against the available choices is regularly done by integer index, which precludes `std::set<std::string>`.

Lestropie: Also, can they be `constexpr`?

daljit46: I probably should have mentioned this, but I did consider the use of `std::array<std::string,N>` in various places instead of `std::vector<std::string>` and ended up discarding the idea. Firstly because in our code, we often rely on `extern` declarations to allow a cpp file to define a certain variable. This implies that we would need to declare that variable for each `N` instantiated in the code, which would be cumbersome, to say the least. 

Secondly, classes like `MR::App::Argument` need to store lists of strings which means that the use `std::array<std::string>` would introduce the need for manual conversions to `std::vector<std::string>` (or alternatively require that those classes be templated the size of their lists, which is overkill IMO).

The use of `std::array` for this use case has very few benefits (e.g. performance is not even a concern) and would make the code less easy to maintain and read. 

For `constexpr` I believe we cannot do this until C++20 because `std::string` doesn't have a `constexpr` constructor in C++17. 

Additionally, an alternative to `std::vector<std::string>` would be `std::initializer_list<const char*>` to write something like:

```
// type is std::initializer_list<const char*>
constexpr auto file_outputs = {"per_edge", "per_node", "single"}; 
```
In C++20 one could even do:

```
using namespace std::string_literals;
// type is std::initializer_list<std::string>
constexpr auto file_outputs = {"per_edge"s, "per_node"s, "single"s}; 
```

Lestropie:
It seems counter-intuitive to me that you'd use iterator offsets to interrogate one argument position, but then rely on upstream internal storage of argument position of each command-line option to interrogate another argument position. I would have thought that if you need to determine two different command-line argument positions, the same mechanism should be used for both.

1.  Find *last* item that *is* an argument or standard option
    (or an argument that is provided to a standard option)
2.  Find *first* item that *is not* an argument or standard option
    (ie. it's an `mrview`-specific option)
3.  If there's at least one instance of both 1 and 2, make sure that position 1 does not exceed position 2.

Ultimately, whether this does as intended is best verified by adding command tests, with both valid and invalid uses, and make sure that the former succeeds and the latter fails with the intended error message.

daljit46:
> It seems counter-intuitive to me that you'd use iterator offsets to interrogate one argument position, but then rely on upstream internal storage of argument position of each command-line option to interrogate another argument position.

That's a good point and I agree. To improve this, we could add a new `index` variable for `ParsedArgument` (just like I did for `ParsedOption`), which keeps track of the position in the raw command line argument list. Alternatively, we can loop over the raw argument list and compare the `id` of a `ParsedOption` against the raw arguments.
I don't know which one I prefer.

> Ultimately, whether this does as intended is best verified by adding command tests, with both valid and invalid uses, and make sure that the former succeeds and the latter fails with the intended error message.

Yes, but since this change concerns `mrview` it'd be difficult to verify it with an ordinary bash test. The app may fail to launch since GUI commands may not work on the CI machines (e.g. due to missing GUI libraries). This is a case where having the possibility to add *unit tests* that verify the correctness of a *single* function may be useful.

daljit46:
@Lestropie btw can an option be repeated? In other words, is this syntax valid: `command arg1 arg2 -opt1 -opt1`?

jdtournier:
> can an option be repeated? 

Yes, but not by default - only if it's been labelled with `.allow_multiple()`. There are quite a few examples of that in the existing commands (both for arguments or options).

Lestropie:
Also, with respect to the option repeating: you'll sometimes see the value of an argument passed via a command-line option accessed as something like `opt[0][0]`. The first index references one of possibly multiple appearances of that particular command-line option; the second index references one of potentially multiple arguments expected to appear after that command-line option. Good example would be something like `mredit -plane`: [usage](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mredit.cpp#L57-L60); [access](https://github.com/MRtrix3/mrtrix3/blob/master/cmd/mredit.cpp#L129-L132).

daljit46:
This is now ready for review. I fixed all the mistakes that were causing the tests to fail and also did some manual testing to check if things behaved as expected. @Lestropie @jdtournier can you take a look at this?

cmd/mrcalc.cpp
```diff
@@ -480,13 +480,12 @@ class LoadedImage {
 
 class StackEntry {
 public:
-  StackEntry(const char *entry) : arg(entry), rng_gaussian(false), image_is_complex(false) {}
+  StackEntry(std::string entry) : arg(std::move(entry)), rng_gaussian(false), image_is_complex(false) {}
```
Lestropie, line 483: Is the `std::move()` here necessary? If yes it could do with code comment to explain why.

cmd/mrcalc.cpp
```diff
@@ -984,15 +982,14 @@ class OpTernary : public OpBase {
...
 
-  for (int n = 1; n < App::argc; ++n) {
-
-    const Option *opt = match_option(App::argv[n]);
+  for (const auto &argument : App::raw_arguments_list) {
+    const Option *opt = match_option(argument);
     if (opt) {
 
       if (opt->is("datatype"))
-        ++n;
```
Lestropie, line 993: These increments on `n` can't be removed. If a standard option is not just a flag, but receives as input one or more arguments, then it's not just the appearance of the command-line option that needs to be skipped in construction of the operand stack, the arguments attributed to that option need to be skipped also.

cmd/tckgen.cpp
```diff
@@ -40,10 +40,11 @@
 using namespace MR;
 using namespace App;
 
-const char *algorithms[] = {
-    "fact", "ifod1", "ifod2", "nulldist1", "nulldist2", "sd_stream", "seedtest", "tensor_det", "tensor_prob", nullptr};
 #define DEFAULT_ALGORITHM 2 // ifod2
 
+const std::vector<std::string> algorithms = {
+    "fact", "ifod1", "ifod2", "nulldist1", "nulldist2", "sd_stream", "seedtest", "tensor_det", "tensor_prob"};
+
```
Lestropie, line 47: It's quite common to have a list of strings define the set of options, but for both the default selection and the code branching based on user input to be based on integer index. While it's not requisite for this PR to go ahead, it's worth thinking about how this could perhaps be made less fragile.

cmd/mrcalc.cpp
Lestropie: Worth adding a `mrcalc` test that intersperses uses of the standard options between operands.

core/app.cpp
```diff
@@ -831,66 +836,74 @@ std::string restructured_text_usage() {
...
-      const Option *opt = match_option(argv[n]);
-      if (opt) {
-        if (n + int(opt->size()) >= argc)
-          throw Exception(std::string("not enough parameters to option \"-") + opt->id + "\"");
-
-        option.push_back(ParsedOption(opt, argv + n + 1));
-        n += opt->size();
-      } else
-        argument.push_back(ParsedArgument(nullptr, nullptr, argv[n]));
+void sort_arguments(const std::vector<std::string> &arguments) {
```
Lestropie, line 886: What's your level of confidence in the reimplementation of `match_option()` / `sort_arguments()`? It's hard to review from code alone. Could possibly make use of `testing_cpp_cli`: if the testing tool writes out the arguments in sorted order / what options were specified and what their corresponding arguments are, then can run tests with various permutations of argument / option order and make sure that what was parsed was as intended. Could even add tests with non-ASCII "dash" characters.

core/app.cpp
```diff
@@ -1170,9 +1185,9 @@ void init(int cmdline_argc, const char *const *cmdline_argv) {
     }
     return s;
   };
-  command_history_string = argv[0];
-  for (int n = 1; n < argc; ++n)
-    command_history_string += std::string(" ") + argv_quoted(argv[n]);
+  command_history_string = raw_arguments_list.front();
```
Lestropie, line 1188: Is the command history string here erroneously omitting the name of the command being executed? That's popped from the front of the list at [line 1148](https://github.com/MRtrix3/mrtrix3/pull/2911/files#diff-ea9657f1eb7d593a574dd53c8caaf0b7b7ae26233db0e199aad234a0ee849c18R1148).

src/gui/mrview/window.cpp
```diff
@@ -721,22 +721,19 @@ void Window::parse_arguments() {
...
+        return std::none_of(MR::App::__standard_options.begin(),
+                            MR::App::__standard_options.end(),
+                            [&option](const auto &standard_option) { return option.opt == &standard_option; });
+      };
+
+      const auto first_non_standard_option =
+          std::find_if(MR::App::option.begin(), MR::App::option.end(), is_non_standard_option);
+
+      if (first_non_standard_option != MR::App::option.end() && first_non_standard_option->index < last_arg_pos) {
+        throw Exception("options must appear after the last argument - see help page for details");
```
Lestropie, line 736:
```suggestion
        throw Exception("MRView-specific options can only appear on the command-line after all arguments specifying input images; see help page for details");
```

Lestropie (changes requested):
Left a few minor comments. Might be worth adding tests for some of the bespoke CLI handling. Otherwise happy to be merged once you're content.

cmd/mrcalc.cpp
daljit46: Not strictly required, but it's a performance optimisation and conveys the intent better I think. `StackEntry` needs to store `entry` in `arg` which requires a copy if a non-temporary value is passed to the constructor. However, the copy is unnecessary if a rvalue (temporary) is passed, which is what the `std::move` avoids. The alternative would be to pass by const ref, but then copies would occur in both cases (unless an additional && overload is provided). See [here](https://stackoverflow.com/a/51706522).

core/app.cpp
daljit46: Yes, good catch.

core/app.cpp
daljit46: I have tested this manually and using "standard" unit tests with Google Test (as an aside, it's very easy to use Google Test with our codebase since our CMake switch).
It's good idea to tests to the codebase though. I will try to add some tests with `testing_cpp_cli` to ensure correctness.

cmd/tckgen.cpp
daljit46: On this point, could we use enums? They seem to be most natural choice and we could provide functions to/from strings.

daljit46:
Review feedback addressed

cmd/tckgen.cpp
Lestropie: I think I already use enums in some places for this kind of purpose; but it still nevertheless requires two parallel variables defined alongside one another, a strongly-typed enum and a list of strings. If there could be a way to both generate a list of strings from an enum for the CLI, and ideally to have the CLI class directly yield a corresponding enum instance when reading argument contents, that would be appealing, but it likely requires language capabilities I'm not familiar with.

Lestropie (approved):
Should be good once the C++ CLI failure on Windows is identified.

**Edit**: Guessing it's this one:
```
testing_cpp_cli: [ERROR] Expected exactly 0 arguments (2 supplied)
testing_cpp_cli: [ERROR] Usage: testing_cpp_cli
testing_cpp_cli: [ERROR] Yours: testing_cpp_cli �bool false
```
MSYS2 is breaking either the em or en dash?

cmd/tckgen.cpp
daljit46: For enum to string conversion, we could use [magic_enum](https://github.com/Neargye/magic_enum), which is a popular header-only C++17 library.

daljit46:
@Lestropie I've spent a lot of time to try to understand why the parsing of utf-8 characters is failing on MSYS2, but without success. I've found that the MSYS2 port of GCC [only supports](https://stackoverflow.com/questions/75426537/stdlocale-constructor-throws-an-exception-in-a-program-compiled-with-mingw) the "C" and "POSIX" locales. Then I thought that the best thing to do would just to call `std::setlocale(LC_ALL,"")` to try to automatically select the user's preferred locale. This seem to work at first try with a simple test application I wrote, but when trying to parse out dash characters, the program didn't parse them correctly. To make things worse, when running the executables under `CTest` the detection of UTF-8 characters didn't even work. 

I'm not sure what more I could try as the documentation on these matters is rather limited and sparsed all over the internet with no real source of truth.

Lestropie:
If that magnitude of effort invested hasn't yielded a solution, then I suggest that when running under MSYS2 we bypass the tests that verify handling of non-ASCII dashes. It's a nice-to-have to handle cases of people copy&pasting code examples from programs that have auto-bastardised text, but it's not mission critical.

PR #2912, opened 2024-05-21
-export_grad_fsl fix for highly non-RAS acquisitions

Lestropie:
Here we go again...

So I was left scratching my head at this one for quite some time. I am trying to get to #2477 among other things, but was seemingly having issues at the `bvec` write stage, which meant I couldn't yet trust anything I was observing at the `bvec` read stage.

1.  Eigen function `.rotation()` was being used, which is *not* the same as `.linear()`. For a DICOM DWI series acquired on a Siemens magnet with a sagittal plane, these two functions yield very different results. I'm sure we've had at least one issue in the past with conflating these two, though haven't had success in finding anything on GitHub (@MRtrix3/mrtrix3-devs anyone with better recollection?).

2.  Code was doing a post-multiply with the image2scanner transform in order to do a scanner2image transformation., rather than constructing the scanner2image transform and using that.

3.  Code was doing a scanner2image transformation based on the transformation for the internal in-memory representation of the header, and then performing subsequent permutations & flips based on the secondary output of the `File::NIfTI::adjust_transform()` function and the internal strides, knowing that at NIfTI export time there would be similar modifications applied. However the `File::NIfTI::adjust_transform()` function returns what will become the header transformation of the output image. So it seems to me that if one wants to perform an image2scanner transformation of gradient directions that will correspond to the output image, and one has what will be the transformation of the output image, might as well do it in one step?

I'm reasonably confident given the test data & tool I have:

- Direct verification of bvec contents based on non-realigned header transformation and known gradient table
- Manual checking of `dwi2tensor` outputs:
    DICOM `mrconvert`ed to NIfTI / `bvec` / `bval`, both with & without internal header transform realignment

In the absence of this change, the specific combination of sagittally-acquired DICOM DWI data, running `mrconvert` to NIfTI / `bvec` / `bval` with internal header transform realignment disabled, produces an incorrect result.

@jdtournier I'll fill in more details in a separate email chain; content to leave as a draft until I can explain the broader project there.

### DIFF
--- a/core/dwi/gradient.cpp
+++ b/core/dwi/gradient.cpp
@@ -178,30 +178,20 @@ namespace MR
     {
       const auto grad = parse_DW_scheme (header);
 
-      // rotate vectors from scanner space to image space
-      Eigen::MatrixXd G = grad.leftCols<3>() * header.transform().rotation();
-
-      // deal with FSL requiring gradient directions to coincide with data strides
-      // also transpose matrices in preparation for file output
+      // rotate vectors from scanner space to image space for expected NIfTI export;
+      //   requisite transposition is a natural consequence
       vector<size_t> order;
-      auto adjusted_transform = File::NIfTI::adjust_transform (header, order);
-      Eigen::MatrixXd bvecs (3, grad.rows());
-      Eigen::MatrixXd bvals (1, grad.rows());
-      for (ssize_t n = 0; n < G.rows(); ++n) {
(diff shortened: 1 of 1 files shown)


Lestropie:
Realised that the testing I'm doing currently has commented out the evaluation of manually requested output strides. Will leave as draft until my tests have been expanded in scope.

Lestropie:
Superseded by #3011.

PR #2922, opened 2024-06-11
SIFT2: Changes to output streamline weights units

### LINKED ISSUE #2138: tcksift(2): Change handling of mu for outputs

Lestropie:
Closes #2138.

I've for years questioned myself about exactly how this confound should be treated. But in retrospect it's remained unaddressed for too long, and I regularly find myself having to re-explain what factors to apply when, what is automatically applied vs. what isn't, what the default should be, and what the interface should look like.

(I like having the capability to account for differences in response function magnitudes after the fact rather than require common response functions, but in retrospect it just confuses people).

I think I've landed on the right long-term solution. This *does* result in a difference in command output compared to `3.0.x`, but the prior behaviour can also be achieved if desired. Rather than having the interface relate to "has been scaled by X / has not been scaled by X", I think the right approach is to focus on the *units* in which those streamline weights are provided. That is then easily reportable as metadata in the file header. The new tests hopefully prove the corresponding interpretations.

### DIFF
--- a/cmd/tcksift2.cpp
+++ b/cmd/tcksift2.cpp
@@ -19,6 +19,7 @@
 #include "header.h"
 #include "image.h"
 
+#include "file/config.h"
 #include "file/path.h"
 
 #include "dwi/directions/set.h"
@@ -29,6 +30,7 @@
 #include "dwi/tractography/SIFT/sift.h"
 
 #include "dwi/tractography/SIFT2/tckfactor.h"
+#include "dwi/tractography/SIFT2/units.h"
 
 using namespace MR;
 using namespace App;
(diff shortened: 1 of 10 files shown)


Lestropie:
From #2138:

- [ ] Update online documentation page dedicated to SIFT, adding information about SIFT2, link to the Aperture manuscript, and explaining the circumstances in which one might do it one way vs another

Lestropie:
- [ ] Add to new documentation page explanation about obtaining FBC from `tcksift`.

PR #2923, opened 2024-06-14
mrdegibbs: update doc to soften warning about partial Fourier

jdtournier:
As discussed on forum some time ago:
https://community.mrtrix.org/t/mrdegibbs-in-partial-fourier/2457/2?u=jdtournier

### DIFF
--- a/cmd/mrdegibbs.cpp
+++ b/cmd/mrdegibbs.cpp
@@ -42,11 +42,16 @@ void usage ()
     "you should run denoising before this command to not alter the noise structure, "
     "which would impact on dwidenoise's performance."
 
+    + "For best results, any form of filtering performed by the scanner should be disabled, whether "
+    "performed in the image domain or k-space. This includes elliptic filtering and other filters that "
+    "are often applied to reduce Gibbs ringing artifacts. While this method can still safely be applied "
+    "to such data, some residual ringing artefacts may still be present in the output."
+
     + "Note that this method is designed to work on images acquired with full k-space coverage. "
-    "Running this method on partial Fourier ('half-scan') data may lead to suboptimal and/or biased "
-    "results, as noted in the original reference below. There is currently no means of dealing with this; "
-    "users should exercise caution when using this method on partial Fourier data, and inspect its output "
-    "for any obvious artefacts. ";
+    "Running this method on partial Fourier (‘half-scan’) may not fully remove all ringing artifacts, "
+    "and you may observe residuals of the original artifact in the partial-fourier direction. "
+    "Nonetheless, application of the method is still safe and worthwhile. Users are however encouraged "
+    "to acquired full-Fourier data where possible.";
 
 
   ARGUMENTS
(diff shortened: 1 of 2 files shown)


bjeurissen (dismissed):

Lestropie (changes requested):
- Suggest having `master` as base branch

- Currently a note about filtering is slid in at the last minute on a paragraph about partial Fourier. I would suggest a separate `DESCRIPTION` paragraph on disabling on-scanner filtering (both k-space-based and image-based).

jdtournier:
OK, pull request redirected to `master`, and more explicit discussion of filtering, as requested. Here's the full text of the command description, for convenience - changes in bold at the end. Comments & suggestions welcome as always.

> This application attempts to remove Gibbs ringing artefacts from MRI images using the method of local subvoxel-shifts proposed by Kellner et al. (see reference below for details).
>
> This command is designed to run on data directly after it has been reconstructed by the scanner, before any interpolation of any kind has taken place. You should not run this command after any form of motion correction (e.g. not after dwifslpreproc).    Similarly, if you intend running dwidenoise, you should run denoising before this command to not alter the noise structure, which would impact on dwidenoise's performance.
>
> **For best results, any form of filtering performed by the scanner should be disabled, whether performed in the image domain or k-space. This includes elliptic filtering and other filters that are often applied to reduce Gibbs ringing artifacts. While     this method can still safely be applied to such data, some residual ringing artefacts may still be present in the output.**
> 
> **Note that this method is designed to work on images acquired with full k-space coverage.  Running this method on partial Fourier (‘half-scan’) may not fully remove all ringing artifacts, and you may observe residuals of the original artifact in the      partial-fourier direction.  Nonetheless, application of the method is still safe and worthwhile. Users are however encouraged to acquired full-Fourier data where possible.**

cmd/mrdegibbs.cpp
```diff
@@ -42,11 +42,16 @@ void usage ()
...
+
     + "Note that this method is designed to work on images acquired with full k-space coverage. "
-    "Running this method on partial Fourier ('half-scan') data may lead to suboptimal and/or biased "
-    "results, as noted in the original reference below. There is currently no means of dealing with this; "
-    "users should exercise caution when using this method on partial Fourier data, and inspect its output "
-    "for any obvious artefacts. ";
+    "Running this method on partial Fourier (‘half-scan’) may not fully remove all ringing artifacts, "
+    "and you may observe residuals of the original artifact in the partial-fourier direction. "
+    "Nonetheless, application of the method is still safe and worthwhile. Users are however encouraged "
+    "to acquired full-Fourier data where possible.";
```
Lestropie, line 54:
```suggestion
    "If this method is executed on data acquired with partial Fourier (eg. \"half-scan\") acceleration, "
    "it may not fully remove all ringing artifacts, "
    "and you may observe residuals of the original artifact in the partial Fourier direction. "
    "Nonetheless, application of the method is still considered safe and worthwhile. "
    "Users are however encouraged to acquired full-Fourier data where possible.";
```

Lestropie:
Will require executing `docs/generate_user_docs.sh` also.

jdtournier:
@Lestropie, feel free to merge if you're OK with these last few changes.

Lestropie (approved):

PR #2933, opened 2024-06-25
New workflow for releases

daljit46:
This PR succeeds the work done in #2835. As mentioned there, instead of having a separate workflow for weekly releases, it's better to have a single CI workflow that manages both tagged and periodic releases. 
As a first step towards that, this PR introduces a new CI workflow which creates artefacts of release builds using GitHub Actions. The artefacts are named as `mrtrix3-{os}-{branch}.{extension}` (@MRtrix3/mrtrix3-devs let me know if we want to add further information like build dates).
The workflow can be triggered manually, and the developer can specify the branch of MRtrix3 they want to build (the branch will need to be based on `cmake`).
This will be useful firstly because it allows a developer to create a release for a particular feature branch of their choice. Secondly, it will allow us to test our build scripts for our upcoming releases without creating a tagged release.

Since GitHub only allows [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) triggers if the workflow file is on the default branch, the merge needs to be against `master`. This is slightly unfortunate since `master` is still not on `cmake`.
To make sure things behave as expected, I have already tested that the workflow succeeds on a separate fork of MRtrix3.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -1,0 +1,137 @@
+name: Releases
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to release from'
+        required: true
+        default: 'dev'
+
+jobs:
+  linux-release:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.inputs.branch }}
+      
+      - name: Set env vars
+        id: envs
+        run: |
+          commit_sha=$(git rev-parse HEAD)
+          date=$(date +%Y_%m_%d)
+          echo "commit_sha=$commit_sha" >> $GITHUB_OUTPUT
+          echo "date=$date" >> $GITHUB_OUTPUT
+          echo "output_name=mrtrix3-linux-$commit_sha-$date.tar.gz" >> $GITHUB_OUTPUT
+
+      - name: Install Eigen3
+        run: |
+          git clone https://gitlab.com/libeigen/eigen.git && cd eigen && git checkout 3.4.0
+          cmake -B build && cmake --build build
+          sudo cmake --install build
+
+      - name: Install Qt 6
+        uses: jurplel/install-qt-action@v3
+        with:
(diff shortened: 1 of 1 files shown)


Lestropie:
I would have thought that for anything that is not based on an immutable tag, you would want the build date (yyyymmdd) embedded in the file name.

daljit46:
> I would have thought that for anything that is not based on an immutable tag, you would want the build date (yyyymmdd) embedded in the file name.

Ok, will make changes to address this. Would a commit hash be even more preferable?

Lestropie:
> Would a commit hash be even more preferable?

I think that if a user requires a version with greater specificity than a branch name and date to the nearest week, they're probably capable of compilation from source. I'd be content with naming by date rather than commit since it's intended for interpretation by users, though also wouldn't be against having both.

Maybe another case worth considering here is where a developer has made some change to a branch, and then wants to make that code version immediately available to a user to utilise / test. There, telling the user that they have to wait up to a week for GitHub to build and package it is a bit clumsy. Will developers be able to trigger regeneration of the package for a specific branch on demand?

jdtournier:
I'd favour both commit & date in the filename, personally. 

>  Will developers be able to trigger regeneration of the package for a specific branch on demand?

Yes, I think that's the idea - @daljit46 can correct me if I'm wrong here...

daljit46:
Yes, that's correct. In fact, this PR *only* implements manually triggered releases which don't create a tagged release, but only GA artefacts that can be downloaded by anyone who has a GitHub account.

daljit46:
I've changed the workflow to produce artefacts as `mrtrix3-${commit_sha}-${date}.${extension}. An example run can be seen [here](https://github.com/daljit46/mrtrix3-ci-test/actions/runs/9697650687). 

This is ready for review.

daljit46:
Actually, just realised that there is a mistake in the Windows artefacts. Let me check that.

daljit46:
Another little thing I noticed is that Github by default packages artefact using `.zip` as an extension. Should I remove the extensions from the artefacts name?

PR #2940, opened 2024-07-03
Fix compilation errors with C++20

daljit46:
`PerThread` and `Shared` classes in `threaded_loop.h` now longer use references as data members. This fixes some errors when compiling with C++20 as the minimum standard (due to errors in initialising the variables in the constructors) and also allows us to avoid following the rule of 5 (e.g. no need to define custom copy/move constructors and assignment operators).
More generally, I think we should try to avoid using references as data members of classes because they limit the implementation of a class and also slightly hurt readability (for example, it's not obvious when accessing the member that it is a reference).


Lestropie:
If it is desirable to support multiple language standards in this way, it might be worth investing in expanding the CI actions to test such compilation. We've previously avoided doing this across major versions of compilers to save trees, but maybe with compiler caching it's now more reasonable, and certainly the list of language standards would be shorter.

(We've also not yet done it for Python minor versions---might have caught the 3.12 deprecations earlier---but that would be better addressed in a separate PR as a follow-up to #2865)

PR #2943, opened 2024-07-17
New command: dirrotate

### LINKED ISSUE #2749: New command: dirrotate
Contemplating having to write a command to deal with premature DWI sequence termination on older hardware.

I'm using a gradient table that's been generated with `dirgen`. It turns out that the scanner is regularly aborting execution of the sequences at those volumes where the *b*-value is maximal and the direction of diffusion sensitisation happens to align quite closely with one of the scanner's physical axes. So it doesn't like the fact that it's being asked to drive one specific gradient at near maximal strength for a long period of time, and aborts. But for whatever reason the sequence is failing to detect the fact that this is going to be a problem and refuse that particular set of sequence parameters.

What I might attempt is an algorithm that finds the spherical rotation that results in the minimum peak gradient amplitude for any volume along any scanner axis. So kind of a soft version of what from memory was a GE "DWI+" mode, where it would only acquire diffusion sensitisations where it could combine multiple gradients, leading to poor distribution of the direction set because it would omit the XYZ axes. Here I'm looking to do the same thing, but with a well-distributed direction set, just trying as best as possible to avoid having directions *exactly* along the scanner axes.

In its simplest form this could just be brute force random number based. Maybe there's a better algorithmic solution where you align each face of the convex hull with a physical axis and then find the best orthogonal rotation. But lots of `dir*` commands are based on brute force so I'd probably just try that.

Lestropie:
Closes #2749.

Have had a lot of issues locally with getting DWI protocols to execute reliably without yielding scanner hardware complaints or outright termination. One of many contributing factors was the presence of volumes for which the direction of diffusion sensitisation aligns very closely with one of the scanner physical axes. This maximises power amplifier & cooling load, and risks exceeding peak amplitude capability once gradient pre-emphasis is added.

This command finds (via brute force) an axis & angle of rotation for which its application results in the *minimum of maximum cosines* across the direction set. By integrating this into my scheme generation script, I was able to reduce the maximal requested gradient amplitude from 0.995 to 0.980. There is zero consequence to other aspects of the scheme such as nearest neighbour angles and SH transform condition number.

-----

- [x] Add a test to prove successful execution / reduction of peak amplitude / preservation of direction scheme properties


cmd/dirrotate.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+protected:
+  const Eigen::MatrixXd &directions;
+  const size_t target_num_permutations;
+  size_t num_permutations;
+  ProgressBar progress;
+  rotation_type best_rotation;
+  default_type min_peak;
+  default_type original_peak;
+  std::mutex mutex;
```
daljit46, line 106: We can probably use `MutexProtected` (in `core/mutexprotected.h`) here?

Lestropie: I created this command by duplicating `dirflip`. On `dev` that's currently not using that class. So should modify both.

Lestropie: In this scenario, there's *two* member variables that should be checked and updated within a lock: `best_rotation` and `min_peak`. I *think* that given it's a full lock rather than an atomic with limited memory ordering, just locking one and then modifying the other while the lock is held shouldn't cause any memory incongruency issues. However given the design of `MutexProtected` I think the better solution is to define a `struct` that encompasses both the rotation and the resulting minimal peak value, and then that becomes the templated object within the `MutexProtected` class.

(This logic potentially extends to other places where `MutexProtected` could be used; essentially a form of "shared" class that encompasses everything controlled under a singular lock)

daljit46: Yes, I agree that using a `struct ProtectedFields{...}` is the most natural way of using `MutexProtected` in this case.

Lestropie: Resolved by 07693ac799ae75e8ba97bf95631e82aafa16c2cd.

PR #1828, opened 2019-12-04
run.command(): Use same Python interpreter

Lestropie:
Works on Linux; requires test on Windows.

Closes #1825.

PR #2958, opened 2024-08-12
mrtrix3.run: Improved shebang parsing

Lestropie:
In doing #2957 I noticed an issue whereby whenever I manually invoked my compiled-from-source Python 3.12, any Python scripts that themselves invoke Python scripts would revert to Python 3.10 that was in my `PATH`. This is contrary to previously constructed code whose intent was to ensure that if a Python script invoked another Python script, the same interpreter would be used (#1828). This occurred because on `dev`, in the course of dropping Python2 support on `dev` (#2215), the script shebangs were changed to `/usr/bin/python3` rather than `/usr/bin/env python3`---note that this persists through #2741---but the code for explicit shebang parsing was not tested on such.

I note that a reversion back to using `/usr/bin/env` seems to be the most appropriate course of action from discussion with @daljit46. But this shebang parsing code is not just for *MRtrix3* executables, it tries to choose the most suitable interpreter for any shebang.

This change makes the relevant function more robust to both shebang styles, and enhances the version matching logic. If a script specifies a specific *minor* Python version (or indeed even *patch* version), the current interpreter will only be used if it is directly compatible with such; otherwise the shell will get to choose (including if the shell itself invokes `env`).

Tough to integrate the following as unit tests given that it depends on the contents of `PATH`. So I'm just doing a dump here. For context, the system Python version is 3.10, but I'm explicitly running a 3.12.4 from outside of `PATH`; this best shows the selection of when the current interpreter is run vs. other behaviour. Bold indicates cases where behaviour is different on Windows.

- [x] Test behaviour on Windows.
    In particular, the API supersedes `env`, because this was requisite at some point in the past to allow Python scripts to execute other Python scripts on Windows. But this may have been an MSYS2 problem that has since been rectified.

- [ ] Discuss any contentious scenarios.
    Possible that @daljit46 may disagree with some of these, preferring to allow the system to decide what to use in all scenarios. However the case that made me pursue this (#2957) was specifically about my using a non-system version of Python yet having subsequent scripts then executing on the system Python. So my current intuition is to preserve this behaviour. The other one that might be contentious is what to do if just "`python`" is specified as the destination interpreter. Here I'm invoking the current interpreter in that case.

| Shebang contents                         | POSIX parsed / error                                                                   | Windows Parsed / error                                                                 |
| ------------------                       | -------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------- |
| `b'Not a shebang'`                       | No shebang found                                                                       | No shebang found                                                                       |
| `b'\n#!/newline/before shebang'`         | ['/newline/before', 'shebang']                                                         | ['/newline/before', 'shebang']                                                         |
| `b'#!/usr/bin/env'`                      | Invalid shebang: missing interpreter after "env"                                       | Invalid shebang: missing interpreter after "env"                                       |
| `b'#!/usr/bin/env python'`               | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| **`b'#!/usr/bin/env python2'`**          | ['/usr/bin/env', 'python2']                                                            | ['/usr/bin/python2']                                                                   |
| `b'#!/usr/bin/env python3'`              | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| **`b'#!/usr/bin/env python3.10'`**       | ['/usr/bin/env', 'python3.10']                                                         | ['/usr/bin/python3.10']                                                                |
| **`b'#!/usr/bin/env python3.10 extra'`** | ['/usr/bin/env', 'python3.10', 'extra']                                                | ['/usr/bin/python3.10', 'extra']                                                       |
| **`b'#!/usr/bin/env python3.11'`**       | ['/usr/bin/env', 'python3.11']                                                         | Invalid shebang: "env" executing on Windows, but unable to find command "python3.11"   |
| `b'#!/usr/bin/env python3.12'`           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| `b'#!/usr/bin/env python3.12 extra'`     | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3', 'extra']                  | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3', 'extra']                  |
| **`b'#!/usr/bin/env python3.10.1'`**     | ['/usr/bin/env', 'python3.10.1']                                                       | Invalid shebang: "env" executing on Windows, but unable to find command "python3.10.1" |
| **`b'#!/usr/bin/env python3.12.3'`**     | ['/usr/bin/env', 'python3.12.3']                                                       | Invalid shebang: "env" executing on Windows, but unable to find command "python3.12.3" |
| `b'#!/usr/bin/env python3.12.4'`         | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| **`b'#!/usr/bin/env python3.12.5'`**     | ['/usr/bin/env', 'python3.12.5']                                                       | Invalid shebang: "env" executing on Windows, but unable to find command "python3.12.5" |
| `b'#!/usr/bin/env python3.x`'            | Invalid shebang: unable to extract Python version from text "#!/usr/bin/env python3.x" | Invalid shebang: unable to extract Python version from text "#!/usr/bin/env python3.x" |
| `b'#!/usr/bin/python'`                   | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| `b'#!/usr/bin/python2'`                  | ['/usr/bin/python2']                                                                   | ['/usr/bin/python2']                                                                   |
| `b'#!/usr/bin/python3'`                  | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| `b'#!/usr/bin/python3.10'`               | ['/usr/bin/python3.10']                                                                | ['/usr/bin/python3.10']                                                                |
| `b'#!/usr/bin/python3.10 extra'`         | ['/usr/bin/python3.10', 'extra']                                                       | ['/usr/bin/python3.10', 'extra']                                                       |
| `b'#!/usr/bin/python3.11'`               | []                                                                                     | []                                                                                     |
| `b'#!/usr/bin/python3.12'`               | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| `b'#!/usr/bin/python3.12 extra'`         | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3', 'extra']                  | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3', 'extra']                  |
| `b'#!/usr/bin/python3.10.1'`             | []                                                                                     | []                                                                                     |
| `b'#!/usr/bin/python3.12.3'`             | []                                                                                     | []                                                                                     |
| `b'#!/usr/bin/python3.12.4'`             | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           | ['/home/unimelb.edu.au/robertes/src/Python3.12/bin/python3']                           |
| `b'#!/usr/bin/python3.12.5'\             | []                                                                                     | []                                                                                     |
| `b'#!/usr/bin/blah'`                     | ['/usr/bin/blah']                                                                      | ['/usr/bin/blah']                                                                      |
| `b'#!/usr/bin/foo bar'`                  | ['/usr/bin/foo', 'bar']                                                                | ['/usr/bin/foo', 'bar']                                                                |
| `b'#!'`                                  | No shebang found                                                                       | No shebang found                                                                       |
| `b'#! /gap/in/shebang'`                  | ['/gap/in/shebang']                                                                    | ['/gap/in/shebang']                                                                    |

### DIFF
--- a/python/mrtrix3/run.py
+++ b/python/mrtrix3/run.py
@@ -613,7 +613,7 @@ def _shebang(item):
 
   class ShebangParseError(Exception):
     pass
-  def parse_shebang(line):
+  def parse_shebang(line, resolve_env):
     # Need to strip first in case there's a gap between the shebang symbol and the interpreter path
     shebang = line[2:].strip().split(' ')
     # On Windows, /usr/bin/env can't be easily found
@@ -624,23 +624,29 @@ def _shebang(item):
     # This selection should apply to shebangs both of the form "/usr/bin/env python3" and "/usr/bin/python3"
     shebang_firstitem_basename = os.path.basename(shebang[0])
     shebang_python_version = None
+    shebang_extras = None
     if shebang_firstitem_basename == 'env':
+      shebang_extras = shebang[2:]
       if len(shebang) < 2:
         raise ShebangParseError('missing interpreter after "env"')
-      if shebang[1].startswith('python'):
+      if shebang[1] == 'python':
+        shebang_python_version = tuple()
+      elif shebang[1].startswith('python'):
         try:
           shebang_python_version = tuple(map(int, shebang[1][len('python'):].split('.')))
         except ValueError as exc:
           raise ShebangParseError(f'unable to extract Python version from text "{line}"') from exc
-      elif utils.is_windows():
-        return [ os.path.abspath(shutil.which(exe_name(shebang[1]))) ] + shebang[2:]
-    elif shebang_firstitem_basename == 'python':
-      shebang_python_version = tuple()
-    elif shebang_firstitem_basename.startswith('python'):
-      try:
-        shebang_python_version = tuple(map(int, shebang_firstitem_basename[len('python'):].split('.')))
-      except ValueError as exc:
-        raise ShebangParseError(f'unable to extract Python version from text "{line}"') from exc
+      elif resolve_env:
+        return [ shutil.which(shebang[1]) ] + shebang[2:]
+    else:
+      shebang_extras = shebang[1:]
+      if shebang_firstitem_basename == 'python':
+        shebang_python_version = tuple()
+      elif shebang_firstitem_basename.startswith('python'):
+        try:
+          shebang_python_version = tuple(map(int, shebang_firstitem_basename[len('python'):].split('.')))
+        except ValueError as exc:
+          raise ShebangParseError(f'unable to extract Python version from text "{line}"') from exc
     if shebang_python_version is None:
       return shebang
     if len(shebang_python_version) > 3:
@@ -650,7 +656,16 @@ def _shebang(item):
     #   or the shebang just requests "python"
     if this_version[:len(shebang_python_version)] == shebang_python_version \
         or not shebang_python_version:
-      return [ sys.executable ] + shebang[2:] if sys.executable else []
+      return [ sys.executable ] + shebang_extras if sys.executable else []
+    if shebang_firstitem_basename == 'env' and resolve_env:
+      exe_path = shutil.which(shebang[1])
+      if not exe_path:
+        raise ShebangParseError(f'on Windows with "env" shebang, but unable to find command "{shebang[1]}"')
+      return [exe_path] + shebang_extras
+    if os.path.exists(shebang[0]):
+      if not os.access(shebang[0], os.X_OK):
+        raise ShebangParseError(f'no execution access to "{shebang[1]}"')
+      return shebang
     return []
 
   # Try to find the shebang line
@@ -665,7 +680,7 @@ def _shebang(item):
     if not (len(line) > 2 and line[0:2] == '#!'):
       continue
     try:
-      shebang = parse_shebang(line)
+      shebang = parse_shebang(line, utils.is_windows())
       app.debug(f'File "{item}": shebang line "{line}"; utilised shebang {shebang}')
     except ShebangParseError as exc:
       app.warn(f'Invalid shebang in script file "{item}": {exc}')
--- a/python/mrtrix3/run.py
+++ b/python/mrtrix3/run.py
@@ -369,14 +369,15 @@ def command(cmd, **kwargs): #pylint: disable=unused-variable
           line.append('-force')
       else:
         line[0] = exe_name(line[0])
-      shebang = _shebang(line[0])
-      if shebang:
+      interpreter = get_interpreter(line[0])
+      if interpreter:
         if not is_mrtrix_exe:
-          # If a shebang is found, and this call is therefore invoking an
-          #   interpreter, can't rely on the interpreter finding the script
-          #   from PATH; need to find the full path ourselves.
+          # If a shebang is found,
+          #   and this call is therefore explicitly invoking an interpreter,
+          #   can't rely on the interpreter finding the script from PATH;
+          #   need to find the full path ourselves.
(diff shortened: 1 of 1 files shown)


Lestropie:
Took a quick look on MSYS2. The `/usr/bin/env` workaround (running a `shutil.which()` on the specified interpreter) is still necessary on Windows; removing it results in "`The system cannot find the file specified`".

@daljit46 Do you want to make a case against the pre-existing behaviour where we ensure that executed Python scripts use the same interpreter as that currently executing, potentially overruling the system config? Otherwise I'll merge this.

daljit46:
I don't have strong objections against this PR. 

> I note that a reversion back to using /usr/bin/env seems to be the most appropriate course of action from discussion with @daljit46. 

Yes, this is the expectation I have when running a script. I think it's the simplest solution and the most in line with Python's idea of venvs.
However, the author of a script is always in a better position to determine the Python versions (just like in C++ code the project author sets up the C++ standard). So if we want to have a custom logic to choose/impose the Python interpreter, I'm not opposed to it (as long as this is adequately documented).

Lestropie:
In responding on #2261, I realised there may be an outstanding problem with this kind of solution.

One of the approaches that had been proposed earlier for controlling the Python module search environment was to have executable files where the shebang points to a softlink to Python. That softlink would reside within the *MRtrix3* build / installation directory, and just point to a suitable Python version. However by invoking via that softlink, the *MRtrix3* API would intrinsically be included within Python's `sys.path`.

Say hypothetically that a third party software uses a similar approach, but instead of the shebang pointing to a softlink, it instead points to a Python in eg. a custom Conda environment, which contains custom libraries / library versions requisite for that command. Say for instance:
```
#!/usr/local/fsl/bin/python
```
With the code in this PR, that would result in use of the current interpreter, which may cause problems for that command.

For the FSL I have installed locally, this isn't the case, they instead use eg.:
```
#!/usr/bin/env bash
/usr/local/fsl/bin/python -I /usr/local/fsl/bin/eddy_quad "$@"
```
, which would be fine. But I don't know if this is what FSL has always done, or whether any other third party piece of software could hard-code a custom Python into their shebang.

-----

There's a possible alternative strategy that comes to mind:

-   If the command is an *MRtrix3* command, and it is a Python command (which is now known without even having to parse the file, CMake generates a list), then use the same interpreter as that currently running;
-   If the command is a non-*MRtrix3* command, make no attempt to impose the currently executing interpreter; just take the shebang and use it.

I don't recall whether there have been any historical edge cases where this would cause problems. But some of them may have had to do with malformed shebangs, which may have been resolved upstream in third-party softwares.

I'll let this simmer for a bit, but my current instinct is that this would be a better solution.

PR #2966, opened 2024-08-16
Python executables shebang: use env

### LINKED ISSUE #2261: Python3 shebang for 3.1.0

Lestropie:
I keep alluding to it but not actually committing.

Closes #2261.

Based on #2261 and many other conversations (eg. #2215), I think that reverting `dev` to using `env` but specifying `python3` as the interpreter is the right way to go.

### DIFF
--- a/cmake/MakePythonExecutable.cmake
+++ b/cmake/MakePythonExecutable.cmake
@@ -4,7 +4,7 @@
 #   - OUTPUT_DIR: Directory in which to create the executable
 
 set(BINPATH_CONTENTS
-    "#!/usr/bin/python3\n"
+    "#!/usr/bin/env python3\n"
     "# -*- coding: utf-8 -*-\n"
     "\n"
     "import importlib\n"


Lestropie:
Cherry-picked ec09027 to get around the reverted history of #2859.

PR #2974, opened 2024-08-21
Fix invalid memory access when reading colour maps

daljit46:
In 62861585ac9dedfdc64d2879124ab392ac4736a4, as part of #2911, `ColourMap::maps` was converted from a null-terminated C array to a `std::vector`. However, the loops iterating over this list weren't updated and were still relying on null-termination causing an invalid memory access. This also resulted `mrview` crashing on MacOS.
@bjeurissen Could you check if this fixes the crash on your system?

### DIFF
--- a/src/colourmap.h
+++ b/src/colourmap.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include "exception.h"
 #include "types.h"
 #include <functional>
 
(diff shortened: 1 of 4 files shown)


bjeurissen (approved):
Works!

daljit46:
Mentioning this for bookkeeping: the changes in this PR have been rebased in `dev` following discussions in #2859.

PR #2976, opened 2024-08-23
Update .git-blame-ignore-devs

daljit46:
This PR updates our `.git-blame-ignore-devs` file with the addition of 10c49930a63e6a2b0ce7d25d3f6d8dde290b2719, authored by @MRtrixBot , to remove third-party headers files from the codebase.


Lestropie:
Not technically required, since `git-blame-ignore-revs` only affects `git blame` reporting and not contribution statistics, but content for it to be in there.

Issue #2685, opened 2023-07-19
Documentation: Revisit FBA instructions


Lestropie:
This is something that has come up repeatedly in discussion, but I've just looked and realised that it's not explicitly listed.

Going to create a checklist as there's a decent amount of separate changes to make. Individual points are however open to debate.

- [ ] Merge two separate single-shell and multi-shell FBA instruction documents into one.
    While having two separate documents is more copy-paste-able, people are often curious about the differences between them, and it's not obvious, given that the content is almost identical. Emerging consensus is that it would be better to have a single document, with bright lights indicating where decisions may need to be made depending on the number of unique *b*-values present in the input data.

- [ ] More verbose discussion of "AFD" vs. "FD".
    Relates to #2299. That nuance may not require a whole page in its own right, but perhaps at the point at which the "`-afd`" command-line option is used to generate a file called "`fd.mif`", it could do with a paragraph explaining the distinction.

- [ ] Modifications of tracking / SIFT usage
    Combination of #1704 and #2273. Opportunity to combine dynamic rather than homogeneous WM seeding, and SIFT2 rather than (/ in addition to?) SIFT, for calculation of the fixel-fixel connectivity matrix. Lazy solution is to just generate 2M with dynamic seeding and then run SIFT2 and `fixelconnectivity -tck_weights_in`.

- [ ] Multi-contrast template
    I've mucked around with this a little myself with a current experiment. Not sure if there is any kind of consensus around what's best to do. @maxpietsch Care to chime in?
    Another little trick that can be used here in concert with the previous point is to use a three-tissue template to derive appropriate model weights for the SIFT model, down-regulating the influence of template voxels with partial volume. Not sure how much effect it has, but it can be done.

- [ ] `fixelcorrespondence`
    When the new `fixelcorrespondence` comes (discussion on data format in #2200; abstract [here](https://www.researchgate.net/publication/326146373_Mitigating_the_effects_of_imperfect_fixel_correspondence_in_Fixel-Based_Analysis) but no PR yet), this will require multiple changes:
    - [ ] Different utilisation even to achieve the same outcomes as the current code (two commands instead of one);
    - [ ] Decide whether to advocate for using one of the more advanced correspondence techniques over the legacy technique;
    - [ ] Include optional details for propagating additional correspondence information for later use as regressors.

- [ ] NaN masking
[...]

PR #2985, opened 2024-09-03
Rename sh2power to sh2metric, new "entropy" operation

Lestropie:
Had previously thought about this from the perspective of registration, but have also been looking at a collaborator dataset where doing satisfactory masking for FBA has been difficult, and using that to revised FBA documentation (#2685).

Here is the WM FOD *l*=0 image, which you can see doesn't do a very good job of separating what one would like to analyse vs. not:

Screenshot from 2024-09-03 16-13-43

Then this is the entropy of the WM FOD template (normalised & inverted):

Screenshot from 2024-09-03 16-14-08

-----

- [ ] Implement the Spherical Harmonic Entropy (SHE) metric as per:
    https://arxiv.org/pdf/1805.08084
    This may yield the same or similar result much faster
    
- [x] Add command-line option to modify the direction set that's used for amplitude sampling

- [x] Implement entropy operations for:
    - [x] mrmath
    - [x] fixel2voxel

- [x] Revisit templated entropy calculation code
    For some reason had trouble specifying the logarithm function as a template. The `switch` statement should get optimised out by the compiler, but maybe I'll have another go before merging.


Lestropie:
Was potentially a little too idealistic about this metric being useful as a scalar surrogate for multi-contrast registration (as multiple SH images can be provided as input to the entropy calculation). Kind of interesting, and maybe useful in specific use cases such as the unusual template above, but not a game-changer.

Single-subject HCP data:
MSMT_tissues
MSMT_entropy

PR #2994, opened 2024-09-12
Merge SHARDRecon into main MRtrix3 codebase

dchristiaens:
This PR aims to merge the [SHARDRecon](https://github.com/dchristiaens/shard-recon) module into the main MRtrix3 repository. The recent changes to the build system and python interface make it challenging to maintain external projects that are not linked to one particular version of MRtrix3. Merging SHARDRecon into `dev` will ensure that the code remains compatible.

Key tasks:
- [x] merge code, preserving git history
- [x] fix compiler errors
- [x] update python scripts for new command interface
- [ ] testing if output is correct
- [ ] add documentation
- [ ] add testing data


cmd/msshsvd.cpp
```diff
@@ -0,0 +1,192 @@
...
+  auto key = in.keyval().find("shellcounts");
+  opt = get_options("weights");
+  if (opt.size()) {
+    W = File::Matrix::load_vector<float>(opt[0][0]).cwiseSqrt();
+    if (W.size() != nshells)
+      throw Exception("provided weights do not match the no. shells.");
+  } else if (key != in.keyval().end()) {
+    size_t i = 0;
+    for (auto w : parse_floats(key->second))
+      W[i++] = std::sqrt(w);
```
daljit46, line 129: Is this intentional? Are we not skipping the first element of `W` here?

daljit46: Never mind I forgot my post-increment operator rules.

cmd/dwirecon.cpp
```diff
@@ -0,0 +1,413 @@
...
+  auto fieldmap = Image<value_type>();
+  size_t fieldidx = 0;
+  Eigen::MatrixXf PE;
+  if (hasfield) {
+    auto petable = PhaseEncoding::get_scheme(dwi);
+    PE = petable.leftCols<3>().cast<float>();
+    PE.array().colwise() *= petable.col(3).array().cast<float>();
+    // -----------------------  // TODO: Eddy uses a reverse LR axis for storing
+    PE.col(0) *= -1;            // the PE table, akin to the gradient table.
+    // -----------------------  // Fix in the eddy import/export functions in core.
```
Lestropie, line 181: Need to determine whether this is an artifact of bad metadata handling in *MRtrix3* (#2917), or whether it is in fact the case that the FSL text file formats for indicating the phase encoding table are expected to be based on a negative determinant image. My expectation was that it would be the former. If the comment is true and the issue is the latter, then there will need to be an explicit distinction between storage of the phase encoding table w.r.t. image axes and the import/export of that information from/to FSL.

src/dwi/svr/mapping.h
```diff
@@ -0,0 +1,300 @@
...
+#include "interp/linear.h"
+#include "interp/cubic.h"
+#include "algo/threaded_loop.h"
+
+#include "dwi/svr/param.h"
+#include "dwi/svr/psf.h"
+#include "dwi/svr/qspacebasis.h"
+
+
+namespace MR
```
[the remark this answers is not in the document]
dchristiaens: Yes, this is now done.

cmd/dwirecon.cpp
jdtournier: I've removed all of these on the version we're using at KCL now. You can see these changes on my forlk: https://github.com/dchristiaens/shard-recon/compare/master...jdtournier:shard-recon:master

python/mrtrix3/commands/dwimotioncorrect.py
```diff
@@ -0,0 +1,333 @@
...
+    except ImportError:
+        raise MRtrixError('SciPy required: https://www.scipy.org/install.html')
+    # import metadata
+    grad_import_option = app.dwgrad_import_options()
+    pe_import_option = []
+    if app.ARGS.pe_table:
+        pe_import_option = ['-import_pe_table', app.ARGS.pe_table]
+    elif app.ARGS.pe_eddy:
+        pe_import_option = ['-import_pe_eddy', app.ARGS.pe_eddy[0], app.ARGS.pe_eddy[1]]
+    # check output path - code no longer works; now done internally??
```
Lestropie, line 101: Correct: arguments of type `(Directory|File|Image)Out` are now automatically checked, so devs are no longer responsible for adding these function calls.
https://github.com/MRtrix3/mrtrix3/blob/1f9d3b485e0e3505a89848fc5e7747b7f50d350c/python/mrtrix3/app.py#L181-L192

python/mrtrix3/commands/dwimotioncorrect.py
```diff
@@ -0,0 +1,333 @@
...
+        PE_scheme = image.mrinfo('in.mif', 'petable').split('\n')
+        if len(PE_scheme) != int(dwi_sizes[3]):
+            raise MRtrixError('Input image does not contain valid phase encoding scheme')
+
+    # Generate a brain mask if required, or check the mask if provided
+    if app.ARGS.mask:
+        if not image.match(header, 'mask.mif', up_to_dim=3):
+            raise MRtrixError('Provided mask image does not match input DWI')
+    else:
+        run.command('dwi2mask in.mif mask.mif')
```
Lestropie, line 136: Needs to be updated to new `dwi2mask` interface

python/mrtrix3/commands/dwimotioncorrect.py
```diff
@@ -0,0 +1,333 @@
...
+        rankreduxstep(k, liveconfig)
+        # slice weights
+        sliceweightstep(k)
+        # register template to volumes
+        registrationstep(k, liveconfig)
+
+
+    #   __________ Copy outputs __________
+
+    run.command(['mrconvert', f'recon-{nepochs}.mif', app.ARGS.output], mrconvert_keyval=f'recon-{nepochs}.mif', force=app.FORCE_OVERWRITE)
```
Lestropie, line 326:
```suggestion
    run.command(['mrconvert', f'recon-{nepochs}.mif', app.ARGS.output],
                mrconvert_keyval=app.ARGS.input,
                force=app.FORCE_OVERWRITE)
```

Purpose here is to have `command_history` in the output image reflect the invocation of `dwimotioncorrect`, rather than listing all of its internal steps.

python/mrtrix3/commands/dwimotioncorrect.py
```diff
@@ -0,0 +1,333 @@
...
+    if app.ARGS.export_motion:
+        run.command(f'cp motion.txt {app.ARGS.export_motion}')
+    if app.ARGS.export_weights:
+        run.command(f'cp sliceweights.txt {app.ARGS.export_weights}')
```
Lestropie, line 330: In addition to using Python functions rather than Unix `cp`, this could potentially be improved by adding `command_history` to the header comments of these files.

Lestropie:
I'm going to generate a list here of things that come to mind; nothing major, just thoughts that I think are worth contemplating. Don't want to block progress by demanding massive pieces of work, just starting dialogues. Some I can help out with.

- [ ] `dwirecon` command (#2915 relevant):
    At the moment this is hard-coded to yield multi-shell spherical harmonics data, however from the command name one would intuitively expect DWI data. There's a few different options here, and it might need a combination of things to get fully resolved with #2915:
    -   Change name of this command to `msshrecon`. Completely unblocks #2915.
    -   Make `dwirecon` an operation-based command. Essentially take what's in #2915 and make what is contained within `cmd/dwirecon.cpp` in this PR another operation within that command, with a set of command-line options that are applicable only to that operation.
    -   Have a toggle that changes the output of this code between multi-shell spherical harmonics and reconstructed DWI data.
        (I would possibly lean towards a toggle here, rather than having a default output and then a command-line option that outputs the data in the other form, since it's large data and users are likely to want one or the other)

- [ ] Data type for command naming:
    Currently here "`mssh`" is used for multi-shell spherical harmonics. I am however reminded that `dwi2fod msmt_csd` started its life as `msdwi2fod`. Here there is a stronger argument for a different datatype indicator: `mssh` and `sh` are less interchangeable from a data description perspective that single-shell vs. multi-shell DWI data. It may however be the case that many operations intended for SH data can be extended to multi-shell SH data.
    -   Possibility of just using `sh` from a command name perspective, and anything that expects multi-shell SH but receives a single SH, or vice versa, throws an Exception? Whether this might be reasonable would require looking across the package for SH interfaces.
    -   Is `mssh` the best indicator? Would "multi-spherical-harmonics (`msh`)" work? Are there any contexts in which one may have multiple spherical harmonics functions, where the groups of coefficients do *not* correspond to shells? Multiple ODFs within a single 5D image comes to mind. I suppose `shard` isn't correct from a data description perspective?

- [ ] `dwirecon` needs a more comprehensive Description.

- [ ] I'd like for there to be more explanation of how the slice profile is derived / interfaced so that I can try to make proper use of it for my own data.

- [ ] I know that for my own data, having the ability to correct for eddy current distortions is going to be important. I tried to read up on the Lie algebra used for the motion modeling, but it's well outside my current expertise to know how it might be extended from rigid-body to an affine model, or what the consequences of doing so on other code might be. @dchristiaens Do you have an intuition of how involved such a change might be? I'm happy to help on this one where I can from a code perspective , I'm moreso doubtful on my capacity to derive / follow the math.

- [ ] Some commands involve specification of multi-band factors to influence slice-to-volume registration. This could be streamlined by instead making use of `SliceTiming`.

- [ ] With `mrfieldunwarp`, I'm wondering if this could be handled differently. If one has a susceptibility field in Hz, then that can be converted to a deformation field based on phase encoding direction & readout time. That's just a non-linear warp field that can be composed with any other transformation, whether it be rigid-body motion (as `mrfieldunwarp currently handles on a per-slice-stack basis) or anything else (eg. I'm imagining composing this with the gradient non-linearity geometric distortion field). Therefore, is this something that could be adequately handled by:
    -   Giving `warpconvert` the ability to convert inhomogeneity field to deformation field based on phase encoding information
    -   Having some other way to compose this deformation with per-slice-stack motion parameters
    ? It's not immediately clicking what the best form of the second point would be. Explicitly composing transforms and then subsequently applying those transforms would be the most modular, but highly inefficient. Maybe if `mrtransform` could accept both the deformation field and the rigid-body motion parameters and would then be responsible for the composition?

- [ ] @dchristiaens Are you able to share some exemplar data for which you know the reconstruction pipeline behaves properly and yields an acceptable result, whether privately or publicly? Just if I'm going to muck around with the code I'd like to know that I'm not breaking anything.
    (If I do come up with any changes, I won't commit directly to this branch, I'll make a PR)

- [ ] Some Python commands need to have their typed arguments updated.

- [ ] Python commands using 4 spaces for indentation, whereas the internal convention is 2.

- [ ] Some Python code could make better use of F-strings, which are now used throughout the Python code base.

- [ ] File copyright headers currently differ from the main code base.

cmd/dwirecon.cpp
dchristiaens: @jdtournier Good point. I forgot to merge [your PR](https://github.com/dchristiaens/shard-recon/pull/20) before creating this one. Your last update is that you were going to confirm if the PE direction had been fixed in the main repo. Is this the case? If so, I don't understand why the test I did last week didn't flag a change in PE.

jdtournier: Well, all I can say is that the code we're currently running has these lines removed - but I did also have to modify the corresponding `acqp.txt` for it to work correctly... So that would suggest that it isn't fixed after all... 

It terms of testing, I guess what's required is to first run `eddy` to verify the correctness of the `acqp.txt` file, and then run `shardrecon` with same `acqp.txt`, right? Must admit I haven't done that, and given the above, I suspect that will show that the problem persists...

Lestropie: 1.  Is the derivation of your `acqp.txt` file going through *MRtrix3* processing at all? I'm presuming it's not being loaded from DICOM headers, but are you loading it into an image header and then doing an export to text file? If so then the issues rectified in #2917 could be playing a role.

2. A useful test to do, with this line removed, would be to find the `acqp.txt` that results in `topup` doing the right thing, then convert the NIfTI between LAS and RAS and re-run with the same `acqp.txt`. If it then produces the wrong result, then the FSL interpretation of the phase encoding table is with respect to the image axes, as has always been my understanding. If it produces the correct result, then FSL expects the phase encoding table to be defined with respect to its internal left-handed coordinate system, and our command-line interface for phase encoding data handling has to be changed.

cmd/dwirecon.cpp
```diff
@@ -0,0 +1,372 @@
...
+  R.setWeights(Wsub);
+
+  R.setVoxelWeights(Wvox);
+
+  // Read input data to vector (this enforces positive strides!)
+  Eigen::VectorXf y(R.rows());
+  y.setZero();
+  ssize_t j = 0;
+  for (auto lv = Loop("loading image data", {0, 1, 2, 3})(dwisub); lv; lv++, j++) {
+    float const w = Wsub(size_t(dwisub.index(2)), size_t(dwisub.index(3))) * Wvox[j];
```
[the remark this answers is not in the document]
dchristiaens:
```suggestion
    float const w = Wsub(dwisub.index(2), dwisub.index(3)) * Wvox[j];
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+        mask(mask, false),
+        motion(mot),
+        E(new Eigen::MatrixXf(nz, nv)),
+        N(new Eigen::MatrixXi(nz, nv)) {
+    E->setZero();
+    N->setZero();
+  }
+
+  void operator()(Image<value_type> &data, Image<value_type> &pred) {
+    size_t v = data.get_index(3);
```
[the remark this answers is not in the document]
dchristiaens:
```suggestion
    ssize_t const v = data.get_index(3);
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+    size_t z = data.get_index(2);
```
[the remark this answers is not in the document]
dchristiaens:
```suggestion
    ssize_t const z = data.get_index(2);
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+    (*E)(z, v) = e;
+    (*N)(z, v) = n;
+  }
+
+  Eigen::MatrixXf result() const {
+    Eigen::MatrixXf Emb(ne, nv);
+    Emb.setZero();
+    Eigen::MatrixXi Nmb(ne, nv);
+    Nmb.setZero();
+    for (size_t b = 0; b < nz / ne; b++) {
```
dchristiaens, line 108:
```suggestion
    for (ssize_t b = 0; b < nz / ne; b++) {
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+      Emb += E->block(b * ne, 0, ne, nv);
```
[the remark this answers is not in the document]
dchristiaens:
```suggestion
      Emb += E->block(b * ssize_t(ne), 0, ne, nv);
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+      Nmb += N->block(b * ne, 0, ne, nv);
+    }
+    Eigen::MatrixXf R = (Nmb.array() > 0).select(Emb.cwiseQuotient(Nmb.cast<float>()), Eigen::MatrixXf::Zero(ne, nv));
+    return R.cwiseSqrt();
+  }
+
+private:
+  const size_t nv, nz, ne;
```
dchristiaens, line 117:
```suggestion
  const ssize_t nv, nz, ne;
```

cmd/dwisliceoutliergmm.cpp
```diff
@@ -0,0 +1,294 @@
...
+        N(new Eigen::MatrixXi(nz, nv)) {
+    E->setZero();
+    N->setZero();
+  }
+
+  void operator()(Image<value_type> &data, Image<value_type> &pred) {
+    size_t v = data.get_index(3);
+    size_t z = data.get_index(2);
+    // Get transformation for masking. Note that the MB-factor of the motion table and the OR settings can be different.
+    size_t ne_mot = motion.rows() / nv;
```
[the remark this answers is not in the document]
dchristiaens:
```suggestion
    ssize_t const ne_mot = motion.rows() / nv;
```

PR #3006, opened 2024-09-18
Fix git version mismatch logic with shallow clones

### LINKED ISSUE #3000: cmake: Fix for shallow clones
If performing a shallow git clone (eg. for a Docker image), `cmake` function `FindVersion()` [here](https://github.com/MRtrix3/mrtrix3/blob/39d61f88a91e7ea3f783f1472bc777d3b5049c62/cmake/FindVersion.cmake#L20-L22) reports mismatch between git tag and base version. However `git describe` isn't yielding a mismatched tag: it's yielding an error message, the function is then comparing that error message to `MRTRIX_BASE_VERSION`, and compilation cannot proceed.

daljit46:
The logic in `FindVersion.cmake` uses `git describe` to find the git commit and tags of the project. Previously, we were mistakenly aborting the configuration of the project when `git describe` returned an error. This PR fixes that by checking the version mismatch only if the parsing of the git tag and commit is successful. A warning is emitted otherwise.
Closes #3000


daljit46:
@Lestropie Sometimes the binary tests for `dirrotate` fail on our CI. The failure doesn't seem to be deterministic because it doesn't always happen, but I have seen it also in other PRs.

PR #3013, opened 2024-09-26
dir* command modifications for fixed direction handling

Lestropie:
Modifications to `dirgen`, `dirflip` and `dirorder`.

These are augmentations I made for the purpose of generating [this dataset](https://osf.io/p96tj/) to produce [this tool](https://github.com/Lestropie/DWI_metadata) to validate #2917 / #3011. I need to be able to re-run this, so it would be convenient if it would make it to `dev` and resolve it against other changes.


Lestropie:
- [ ] Upon merge, update `dev` branch and delete feature branch on `test_data` repository.

PR #3014, opened 2024-10-02
dirrotate: Test using awk rather than bc

Lestropie:
Should fix issue [reported](https://github.com/MRtrix3/mrtrix3/pull/3012#issuecomment-2376982389) in #3012. Seems to be some kind of race condition in the BASH `[[ ]]` / `[ ]` implementation, where sometimes the "`1`" yielded by `bc` would be invoked as a command name. But it was inconsistent. And running the script manually rather than via `ctest` would never fault. So kinda weird. I've verified that this is indeed testing for expected behaviour by flipping the inequality test.

### DIFF
--- a/testing/binaries/tests/dirrotate/default
+++ b/testing/binaries/tests/dirrotate/default
@@ -23,7 +23,6 @@ testing_diff_matrix tmp1_stats.txt tmp2_stats.txt
 cat tmp1.txt | \
 tail -n +2 | \
 tr -d "-" | \
-tr -d "n" | \
 tr " " "\n" | \
 sort -g -r | \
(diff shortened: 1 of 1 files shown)


Lestropie:
Ie. This line of the test:
```
[[ $(echo "${b} < ${a}" | bc) -eq 1 ]]
```

Would yield:
```
(standard_in) 1: syntax error
CMake Error at /home/runner/work/mrtrix3/mrtrix3/cmake/RunTest.cmake:17 (message):
  Test
  /home/runner/work/mrtrix3/mrtrix3/testing/binaries/tests/dirrotate/default
  failed!
```

PR #3025, opened 2024-10-10
5TTgen MSMT

arkiev:
This script is intended for generating a 5-Tissue Type (5TT) image from Orientation Distribution Function (ODF) images in the MRtrix3 neuroimaging framework. 

By default, white-matter, grey-matter,  cerebrospinal fluid ODFs  along with a binary brain mask is used to generate a 5TT image with empty volumes for subcortical grey-matter and pathological tissue. If SGM and/or PATH segmentations are available, there is an option to include them in the 5TT image.


python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
+# HSVS SCRIPT
```
Lestropie, line 1:
```suggestion
```

(Hold off on merging: I'll make a bunch of suggestions on GitHub and they can all be merged in a single batch commit)

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def usage(base_parser, subparsers):  # pylint: disable=unused-variable
+    parser = subparsers.add_parser('odf', parents=[base_parser])
+    parser.set_author("Arkiev D'Souza (arkiev.dsouza@sydney.edu.au) & Robert E. Smith (robert.smith@florey.edu.au)")
+    parser.set_synopsis('Generate a 5TT image from ODF images.')
+    
+    parser.add_argument('template_odf_wm', type=app.Parser.ImageIn(), help='The input white-matter ODF')
+    parser.add_argument('template_odf_gm', type=app.Parser.ImageIn(), help='The input grey-matter ODF')
+    parser.add_argument('template_odf_csf', type=app.Parser.ImageIn(), help='The input cerebrospinal fluid ODF')
```
Lestropie, line 30:
```suggestion
    parser.add_argument('odf_wm', type=app.Parser.ImageIn(), help='The input white-matter ODF')
    parser.add_argument('odf_gm', type=app.Parser.ImageIn(), help='The input grey-matter ODF')
    parser.add_argument('odf_csf', type=app.Parser.ImageIn(), help='The input cerebrospinal fluid ODF')
```

Algorithm is not exclusively applicable to template ODFs; should be at least possible to run on subgle-subject data.

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+    if not os.path.isfile(filepath):
+        raise MRtrixError(f'Required input file missing (expected location: {filepath})')
+
+
+def execute():  # pylint: disable=unused-variable
+    # Check for input files
+    check_file(app.ARGS.template_odf_wm)
+    check_file(app.ARGS.template_odf_gm)
+    check_file(app.ARGS.template_odf_csf)
+    check_file(app.ARGS.mask_image)
```
Lestropie, line 49:
```suggestion
def execute(): # pylint: disable=unused-variable
```

On `dev`, this type of input / output file checking is no longer done explicitly by the developer, it's instead handled by the typed CLI. 

In this case it's actually worse; doing this check would prevent valid use cases. One of these inputs could be a using the multi-file numbered image format. In this case, there will not exist a filesystem entry corresponding to the string specified at the command-line, yet a valid input image can nevertheless be read once that string is parsed by the C++ API.

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Extract l=0 term from WM
+    run.command(f'mrconvert', app.ARGS.template_odf_wm, '-coord 3 0 -axes 0,1,2 wm_vol.mif')
```
Lestropie, line 52:
```suggestion
    run.command(f'mrconvert {app.ARGS.odf_wm} -coord 3 0 -axes 0,1,2 wm_vol.mif')
```

At least this is the syntax style that I've used elsewhere with the changes on `dev`.

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Set negative values to zero
+    run.command(f'mrcalc wm_vol.mif 0.0 -max wm_vol_pos.mif')
```
Lestropie, line 55:
```suggestion
    run.command('mrcalc wm_vol.mif 0.0 -max wm_vol_pos.mif')
```

`pylint` will complain if you use an F-string with zero substitutions.

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+    run.command(f'mrcalc', app.ARGS.template_odf_gm, '0.0 -max gm_vol_pos.mif')
+    run.command(f'mrcalc', app.ARGS.template_odf_csf, '0.0 -max csf_vol_pos.mif')
```
Lestropie, line 57:
```suggestion
    run.command(f'mrcalc {app.ARGS.odf_gm} 0.0 -max gm_vol_pos.mif')
    run.command(f'mrcalc {app.ARGS.odf_csf} 0.0 -max csf_vol_pos.mif')
```

Variable rename + different F-string syntax

python/mrtrix3/commands/5ttgen/odf.py
Lestropie: Change to be applied elsewhere also

python/mrtrix3/commands/5ttgen/odf.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Normalize each volume
+    run.command(f'mrmath wm_vol_pos.mif gm_vol_pos.mif csf_vol_pos.mif sum totalvol.mif')
+    run.command(f'mrcalc wm_vol_pos.mif totalvol.mif -divide wm_vol_pos_norm.mif')
+    run.command(f'mrcalc gm_vol_pos.mif totalvol.mif -divide gm_vol_pos_norm.mif')
+    run.command(f'mrcalc csf_vol_pos.mif totalvol.mif -divide csf_vol_pos_norm.mif')
+
+    # Create empty volume for SGM and pathology
+    run.command(f'mrcalc wm_vol.mif 0 -neq wm_vol_1.mif')
+    run.command(f'mrcalc wm_vol_1.mif 1 0 -replace empty_vol.mif')
```
Lestropie, line 67: This could instead be done as:

```suggestion
    run.command(f'mrcalc wm_vol.mif 0 -neq - | '
                f'mrcalc - 1 0 -replace empty_vol.mif')
````

(Since the intermediate image isn't used anywhere else)

Or better, something like:

```suggestion
    run.command(f'mrcalc wm_vol.mif inf -gt empty_vol.mif')
````

**Comment by Lestropie** [comment]
- [ ] I would consider renaming to "`5ttgen msmt`", given it's specifically information from a multi-tissue decomposition that is used to generate a 5TT image; a single ODF couldn't be used.

- [ ] Input mask could technically be made a command-line option rather than a compulsory positional argument. One way to do it would be similar to how `dwibiasnormmask` operates. Generate a mask that contains only those voxels where the sum of ODF *l*=0 terms exceeds 0.5/sqrt(4pi); then select the largest component and fill any holes. A user-specified mask acts as a substitute for such.

- [ ] Standard indentation for *MRtrix3* is two spaces, not four.

- [ ] Need to run `docs/generate_user_docs.sh` and contribute changes.

- [ ] Currently `mrgrid regrid` is always run on the input mask image. Better would be to detect whether or not the input mask image already uses the same voxel grid as the input ODFs. If it does, then a straight multiplication can be done. If not, then a warning should be issued; then, marginally preferable to a nearest-neighbour regridding is to instead do a resampling of the input binary image and then threshold the result at 0.5.

- [ ] Should be an explicit check regarding whether the input ODF images are stored on the same voxel grid, prior to attempting to concatenate them.

- [ ] For SGM and "pathology" tissue types, technically the output of `5ttgen` when run with this algorithm could be piped directly to `5ttedit`. So I'm wondering if it would actually be better to remove that functionality from here.

**Comment by Lestropie** [comment]
Rather than accepting suggestions individually, can you please add them to a batch commit; otherwise the commit history becomes non-informative.

**Comment by arkiev** [comment]
- [ ] name has been changed to msmt.py
- [ ] mask changed to optional input. By default, brain mask created using the `dwibiasnormmask ` approach: `mrcalc totalvol.mif 0.1410473959 -gt - | maskfilter - clean - | maskfilter - fill - | mrcalc - fTT_dirty.mif -mult result.mif`
- [ ] indentation updated to two spaces
- [ ] Still need to run docs/generate_user_docs.sh and contribute changes
- [ ] `mrgrid regrid` is only conducted on input mask if input mask has different spacing to the 5TT image, and nearest neighbour interpolation has been replaced by resampling and thresholding: `mrgrid {app.ARGS.mask} -template wm_vol.mif regrid - | mrcalc - 0.5 -gt - | mrcalc - fTT_dirty.mif -mult result.mif`
- [ ] still to do explicit check regarding whether the input ODF images are stored on the same voxel grid, prior to attempting to concatenate them. Can a simple comparison of `mrinfo -size` be used here, or is there more to it?
- [ ] SGM and PATH have been removed

File: python/mrtrix3/commands/5ttgen/msmt.py [hunk 8]
```diff
@@ -0,0 +1,112 @@
...
+
+import os
+import subprocess
+from mrtrix3 import MRtrixError
+from mrtrix3 import app, path, run
+
+
+def usage(base_parser, subparsers):  # pylint: disable=unused-variable
+  parser = subparsers.add_parser('msmt', parents=[base_parser])
+  parser.set_author("Arkiev D'Souza (arkiev.dsouza@sydney.edu.au) & Robert E. Smith (robert.smith@florey.edu.au)")
```
Lestropie, line 24:
```suggestion
  parser.set_author('Arkiev D'Souza (arkiev.dsouza@sydney.edu.au) & Robert E. Smith (robert.smith@florey.edu.au)')
```

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -0,0 +1,112 @@
...
+  parser.set_synopsis('Generate a 5TT image from ODF images.')
+
+  parser.add_argument('odf_wm', type=app.Parser.ImageIn(), help='The input white-matter ODF')
+  parser.add_argument('odf_gm', type=app.Parser.ImageIn(), help='The input grey-matter ODF')
+  parser.add_argument('odf_csf', type=app.Parser.ImageIn(), help='The input cerebrospinal fluid ODF')
+  parser.add_argument('fTT_image', type=app.Parser.ImageOut(), help='The output 5TT image')
+
+  options = parser.add_argument_group('Options specific to the "odf" algorithm')
```
Lestropie, line 32:
```suggestion
  options = parser.add_argument_group('Options specific to the "msmt" algorithm')
```

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -0,0 +1,112 @@
...
+    # case 1: using provided brainmask
+    # check if dimensions of brainmask match 5TT_dirty. If not, regrid
+    command_fTTdirty = 'mrinfo -spacing fTT_dirty.mif'
+    spacing_fTTdirty = subprocess.check_output(command_fTTdirty, shell=True).decode('utf-8')
+
+    command_mask = 'mrinfo -spacing {app.ARGS.mask}'
+    spacing_mask = subprocess.check_output(command_mask, shell=True).decode('utf-8')
+
+      if spacing_fTTdirty == spacing_mask:
+        print("mask has equal dimensions to 5TT image. No regridding of mask required")
```
Lestropie, line 94:
```suggestion
        app.console("mask has equal dimensions to 5TT image. No regridding of mask required")
```

Note second instance below.

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -0,0 +1,112 @@
...
+        run.command('mrcalc fTT_dirty.mif {app.ARGS.mask} -mult result.mif')
+
+      else:
+        print("WARNING: mask has different dimeonsions to 5TT image. Regridding mask")
+        run.command('mrgrid {app.ARGS.mask} -template wm_vol.mif regrid - | mrcalc - 0.5 -gt - | mrcalc - fTT_dirty.mif -mult result.mif')
```
Lestropie, line 99:
```suggestion
        run.command(f'mrgrid {app.ARGS.mask} -template wm_vol.mif regrid - | mrcalc - 0.5 -gt - | mrcalc - fTT_dirty.mif -mult result.mif')
```

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -0,0 +1,112 @@
...
+
+    command_mask = 'mrinfo -spacing {app.ARGS.mask}'
+    spacing_mask = subprocess.check_output(command_mask, shell=True).decode('utf-8')
+
+      if spacing_fTTdirty == spacing_mask:
+        print("mask has equal dimensions to 5TT image. No regridding of mask required")
+        run.command('mrcalc fTT_dirty.mif {app.ARGS.mask} -mult result.mif')
+
+      else:
+        print("WARNING: mask has different dimeonsions to 5TT image. Regridding mask")
```
Lestropie, line 98:
```suggestion
        app.warn('Mask has different dimensions to 5TT image; regridding')
```

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -0,0 +1,112 @@
...
+
+def usage(base_parser, subparsers):  # pylint: disable=unused-variable
+  parser = subparsers.add_parser('msmt', parents=[base_parser])
+  parser.set_author("Arkiev D'Souza (arkiev.dsouza@sydney.edu.au) & Robert E. Smith (robert.smith@florey.edu.au)")
+  parser.set_synopsis('Generate a 5TT image from ODF images.')
+
+  parser.add_argument('odf_wm', type=app.Parser.ImageIn(), help='The input white-matter ODF')
+  parser.add_argument('odf_gm', type=app.Parser.ImageIn(), help='The input grey-matter ODF')
+  parser.add_argument('odf_csf', type=app.Parser.ImageIn(), help='The input cerebrospinal fluid ODF')
+  parser.add_argument('fTT_image', type=app.Parser.ImageOut(), help='The output 5TT image')
```
Lestropie, line 30: Rather than "`fTT`", which is inconsistent with the adopted 5TT (but presumably that is not a permissible variable name), you could instead do something like "`out_5tt`".

Lestropie:
I'll have to brush up on my `git rebase` skills to combine those suggestion acceptances.

For checking images being on the same voxel grid see [`image.match()`](https://github.com/MRtrix3/mrtrix3/blob/4d6f20f17e83c8f3d6031b099e706bf76f137223/python/mrtrix3/image.py#L170).

Did you duplicate checkboxes because you weren't able to activate the ones I posted? I could imagine GitHub might not give you permission to do that.

arkiev:
thanks for sharing `image.match()` which has been incorporated 

correct - wasn't able to tick off your checklist

Lestropie:
Rebased to clean up commit history. For GitHub code reviews with a lot of minor changes in different locations of the code you can add the agreed-upon changes to a batch, and GitHub will then create a single commit containing all such changes with appropriate attribution: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

- [ ] Check dimensionalities of ODFs
    - [ ] Issue warning if WM ODF is *lmax*=0, but can technically still proceed
    - [ ] Issue warning if GM ODF is *lmax*>0, but can technically still proceed
    - [ ] Issue warning if CSF ODF is *lmax*>0, but can technically still proceed

- [ ] Combine some operations to reduce scratch storage

- [ ] Add description to explain how brain masking will operate in the absence of an input brain mask

- [ ] Derive some suitable test data & add tests

python/mrtrix3/commands/5ttgen/__init__.py
```diff
@@ -14,4 +14,4 @@
 # For more details, see http://www.mrtrix.org/.
 
 # pylint: disable=unused-variable
-ALGORITHMS = ['freesurfer', 'fsl', 'gif', 'hsvs']
+ALGORITHMS = ['freesurfer', 'fsl', 'gif', 'hsvs', 'msmt']
```
Lestropie, line 17: @arkiev Code would not have been possible to execute via `5ttgen` until it was added here.

python/mrtrix3/commands/5ttgen/msmt.py
```diff
@@ -13,92 +13,116 @@
...
+                         ' of what is expected from a voxel that contains a DWI signal'
+                         ' that exactly matches one of the response functions.')
+
+  parser.add_argument('odf_wm', type=app.Parser.ImageIn(), help='The input white matter ODF')
+  parser.add_argument('odf_gm', type=app.Parser.ImageIn(), help='The input grey matter ODF')
   parser.add_argument('odf_csf', type=app.Parser.ImageIn(), help='The input cerebrospinal fluid ODF')
   parser.add_argument('output', type=app.Parser.ImageOut(), help='The output 5TT image')
 
-  options = parser.add_argument_group('Options specific to the 'msmt' algorithm')
-  parser.add_argument('-mask', nargs='?', default=None, type=app.Parser.ImageIn(), help='The input binary brain mask image')
```
Lestropie: @arkiev "`nargs='?'`" would make it inconsistent with the *MRtrix3* interface. The expectation is that the user either provides nothing, or the specify "`-mask`" followed by an image path. However "`nargs='?'`" specifically permits the absence of a trailing argument, which wouldn't make sense here.

Lestropie:
@arkiev I had a bit of a go at the code here. Sorry if I ended up hacking away much of it. But hopefully there's a few teachable things in there:

-   Avoid code duplication wherever possible. If there's three lines of code that are almost identical except for some small change to a string, do it in a loop or list comprehension or the like.

-   I nowadays try to avoid having filesystem paths as strings that get duplicated across multiple command invocations. It's too prone to error. So I try to store them in named variables.

-   While it's possible to run a series of commands and store all of the intermediate images in the scratch directory, I'm conscious of MRtrix scratch directories blowing out in size, and unnecessarily using hard drives. So I try to make use of the fact that `run.command()` can deal with image piping, and the underlying MRtrix commands might be able to instead use a RAM filesystem for the intermediate data. The actual command or set of commands to be executed can be constructed according to more complex underlying logic.

arkiev:
Thanks for the feedback @Lestropie. It is looking good!!

I address the errors from pylint - there was another error that I wasn't familiar with (I think it was a build error, but since my last commit a new set of checks needs to be done to reproduce the error). How do I run those checks (or can I leave this with you)?

Lestropie:
Looks like in 2bc20890729d43d3216ffa4e2932e2772700c8ee you did a find & replace from `"` to `'` and inadvertently edited the copyright header in the process, which now fails tests :rofl:

The clang-tidy failure is irrelevant to this PR. Might be spontaneous and just need a re-run at a later date. `dev` isn't locked to CI passing so I can merge it anyway if I want to. But I want to add test data and tests first.

arkiev:
ah yes I did do that 😅

Okay, sounds good - let me know what is needed from my end for testing. 

At some point we should think about comparing the SGM segmentation generated using the vis image (using `run_first_all <vis>` followed by `5ttedit`) against the ideal way to create the population template SGM (create subject-specific SGM segmentation, apply the warp to get the segmentation into template space, compute the average across subjects). If they are comparable, perhaps include an option to do the SGM segmentation generated using the vis image.

Lestropie:
dddce99904b75037d76ea1851067cce912d75ed0 shows how you can add tests. These aren't evaluated in CI, but you can run them if you configure CMake accordingly. Also see how I had to commit data to a different repo and then update the reference; this is now a bit different on `dev`---see updated instructions [here](https://github.com/MRtrix3/mrtrix3/tree/ad951f54cfe645667f48e3545f9363ef5aaf88dc/testing#adding-test-data)---so I did this one myself.

If you want to add the ability to invoke `run_first_all` on the pseudo-T1w image and integrate its output into the 5TT image, that can probably be proposed as a separate augmentation. It's not a trivial addition, it would be preferred to make use of #2558, and it would need its own additional test (which you can now see how to do). So this PR could be merged in the absence of such.
I'm a little hesitant about adding to *MRtrix3* itself even more scripts that invoke non-*MRtrix3* functionalities; I'd kinda prefer that any such things go into a separate repository. Here that functionality doesn't strictly need to be in the same script, even it would just be more convenient for it to be there. You can take a 5TT image generated by this command and go `5tt2vis` -> `run_first_all` -> `5ttedit` as a little standalone modifier script.

RE segmenting SGM in each subject and transforming those results to template space and aggregating, I would conceptualise that as a separate discussion to what does or does not get included as an *MRtrix3* software feature. Comparability of the two outcomes isn't the only thing determining whether that feature would be useful. Recall `5tt msmt` can be run on single-subject data.

PR #2451, opened 2022-03-12
tcksample: Permit 4D input image

Lestropie:
Implemented for a specific project; listing here as draft for attribution to `3.1.0` milestone so that it doesn't get forgotten.

PR #3028, opened 2024-10-24
tcksample: Ability to sample SH amplitudes

### LINKED ISSUE #2942: tcksample: Sample FOD amplitudes
Propagating from [code comment](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/cmd/tcksample.cpp#L83) given external feature request.

- [x] #2451 allows the input image to be 4D, in which case each volume is sampled independently. Therefore, similarly to #1485, I believe that if the input number of volumes match that of an SH series the user would need to be forced to specify a command-line option indicating that the input image is to be interpreted as an SH series.

- [x] Consider not permitting use of nearest-neighbour interpolation if sampling FOD amplitudes.

- [x] For `-precise`, will need to generate `Mapping::SetVoxelDir` to have access to streamline-voxel intersection tangents.

Lestropie:
Closes #2942.

- [x] Restructure code to make better use of templating.
    The input being a 4D SH should perhaps be a derived class, which instantiates the precomputer and has a local buffer for loading SH coefficients from the image.

- [ ] Add test data and tests.

### DIFF
--- a/cmd/tcksample.cpp
+++ b/cmd/tcksample.cpp
@@ -29,6 +29,7 @@
 #include "interp/linear.h"
 #include "interp/nearest.h"
 #include "math/median.h"
+#include "math/SH.h"
 
 
 
@@ -59,7 +60,13 @@ void usage ()
   + "In the circumstance where a per-streamline statistic is requested, the input "
     "image can be 4D rather than 3D; in that circumstance, each volume will be sampled "
     "independently, and the output (whether in .npy or text format) will be a matrix, "
-    "with one row per streamline and one column per metric.";
+    "with one row per streamline and one column per metric."
+
+  + "If the input image is 4D, "
+    "and the number of volumes corresponds to an antipodally symmetric spherical harmonics function, "
+    "then the -sh option must be specified, "
+    "indicating whether the input image should be interpreted as such a function "
+    "or whether the input volumes should be sampled individually.";
 
   ARGUMENTS
   + Argument ("tracks", "the input track file").type_tracks_in()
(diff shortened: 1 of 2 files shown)


cmd/tcksample.cpp
```diff
@@ -186,11 +202,24 @@ class SamplerNonPrecise
...
-        if (interp.scanner (tck[i]))
-          out[i] = interp.value();
-        else
-          out[i] = NAN;
+      if (sh_precomputer) {
+        for (size_t i = 0; i != tck.size(); ++i) {
+          if (interp.scanner (tck[i])) {
+            for (interp.index(3) = 0; interp.index(3) != interp.size(3); ++interp.index(3))
+              sh_coeffs[interp.index(3)] = interp.value();
+            const Eigen::Vector3f dir = (tck[(i == ssize_t(tck.size()-1)) ? i : (i+1)] - tck[i ? (i-1) : 0]).normalized();
```
daljit46, line 210: Perhaps `std::clamp` conveys the intention slightly better here? Something like:
`const Eigen::Vector3f dir = (tck[std::clamp(i + 1, size_t(0), tck.size() - 1)] - tck[std::clamp(i - 1, size_t(0), tck.size() - 1)]).normalized();`

Lestropie: I will need to reprogram my brain to be accustomed to having that one available. However the base for this branch is pre-`cmake` and so pre-C++17. 
(This particular calculation would also be better handled via #2189)

Lestropie: Also `clamp()` to me only makes sense if it's possible for one limit or the other to kick in; whereas in both instances here it's only one of the two that should really need to be checked.
(While not present at time of posting, if there's only 1 vertex in the streamline, it's not possible to derive a tangent)

Lestropie:
This PR targets the branch in #2451; will resolve against `dev` and add test data and tests there.

Issue #3023, opened 2024-10-09
dwidenoise enhancements list


Lestropie:
There have been multiple potential enhancements to `dwidenoise` discussed. While there may be additional ideas for improving denoising that may be kept private as research developments, this list is items that are hopefully acceptable for public listing for the purpose of participant recruitment.

- [ ] Evaluate demeaning: #2363
    Use of single-precision floating-point with an imprecise decomposition algorithm can lead to artifacts in noise estimation & rank selection. One potential solution proposed some time ago was to explicitly demean the signal, storing that mean as a signal component to be included in the output series. Doing so however necessitates ensuring that the rank estimation does not become biased.
 
- [ ] Alternative decomposition: #2906
    Simply changing the decomposition used from the Eigen library may improve precision, though at the cost of computation time. This should ideally be evaluated across multiple use cases.

- [ ] Modify rank estimation: #2591
    May be possible to directly translate the modified estimators from externally published work to the *MRtrix3* implementation. Would need to contrast against the existing modification of @jelleveraart's estimator as "`Exp2`".

- [ ] Facilitate two-pass approach: #2274
    This could be used to stabilise against outliers in noise level estimation. It could also be used in pipelines that explicitly separate between the estimation of corrections to be applied vs. the application of those corrections.

- [ ] Spherical kernel: #2742
    Already published; having implemented would reduce issues around kernel size selection, and could also be used to mitigate against issues of kernel size / independence at FoV extremities.

- [ ] Handle >4D data: #3021
    Would make the method more amenable in particular to multi-echo fMRI data.

- [ ] Soft component thresholding: #3022
    Would mitigate quantisation effects due to a component being just above vs. just below the MP distribution cutoff. May improve denoising of datasets with a very small number of volumes.

Also:

- [ ] Build dataset of exemplar series for denoising efficacy evaluation
    Sometimes we hear about a particular dataset where the denoising performs poorly. We're also now using this method for multi-echo fMRI data in addition to DWI data. So it would be good to start building a dataset of different acquisitions of different quality, so that methods and heuristics can be refined to try to make the implementation robust against such variation, or to provide instructions on recommended settings in different scenarios.

Issue #2742, opened 2023-10-29
ENH dwidenoise: Spherical kernel


Lestropie:
Been on my list for a long time but seemingly never listed.

As performed in https://www.sciencedirect.com/science/article/pii/S1053811919305348.

A spherical kernel should:
-   Offer better rotation invariance
-   Allow precisely tuning kernel size to number of volumes for optimal estimator behaviour
-   Allow increase of kernel size at FoV edges such that kernel is always centred at voxel of interest and always of (almost?) equal size

Lestropie:
- [ ] Ideally, should be possible to denoise using *either* a square *or* spherical kernel; generic code would fill the 2D data matrix with image data from the local neighbourhood, and know how to reconstruct image data for the voxel at the centre of the kernel from the PCA components.

- [ ] Kernel should be spherical even in the presence of voxel anisotropy; the voxels chosen for inclusion in the kernel should be based on ordering of Euclidean distances from the kernel centre.

- [ ] Command-line option to modulate the size of the spherical kernel. What makes most sense to me is to have the number of voxels as a minimal multiple of the number of volumes, but it should also be possible to specify an absolute size.
    I say "minimal" deliberately: for a given Euclidean distance from the kernel centre there may be multiple voxels with an equivalent distance from the kernel centre. Where this happens I think that all such voxels should be included in the kernel, rather than selecting a subset of them arbitrarily in order to satisfy a kernel size target.

PR #3029, opened 2024-11-04
New commands: dwidenoise2, dwi2noise

Lestropie:
Doing some work on #3023 (specifically #2742) for a Hackathon.

Draft PR; scope will change over the next couple of days.

### DIFF
--- a/cmd/mrcalc.cpp
+++ b/cmd/mrcalc.cpp
@@ -529,6 +529,7 @@ public:
         }
       }
     }
+    arg.erase();
   }
 
   std::string arg;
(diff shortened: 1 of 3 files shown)


Lestropie:
@arshiyasan: The code base has changed a whole lot since you last looked at it. If you want to contribute code for `dwidenoise` taking an input noise map, it'll need to be refactored, achieving it with a new class derived from `Estimator::Base`. If you still want to try it, I'll hold out, otherwise LMK and I'll get it done, I've kind of been scratching an itch here.

Lestropie:
I have revised the proposal here to retain the version of `dwidenoise` currently on `dev` without consequential modification (and without corrupting Git history or `git blame`), and instead propose my new implementation as `dwidenoise2`. That way the code produced under the NYU copyright and a different author list can reside effectively untouched, and new decisions can be made on what needs to appear for `dwidenoise2`. See discussion in #3059.

Still a couple of items in #3023 to implement before this can be proposed for merge.

PR #3033, opened 2024-11-06
New command: fixeltransform

Lestropie:
Functionality required to establish spatial correspondence between two independent fixel datasets.


Lestropie:
Got away without #3188 by performing tests with `-abs` rather than `-frac`.

PR #2929, opened 2024-06-19
Fixes to containers FSL dependency

Lestropie:
Closes #2921.

See also https://github.com/MRtrix3/containers/pull/26, which includes changes necessary to produce the minified FSL dependency that this change then downloads from OSF.

While the container minification process *intended* to capture EddyQC's `eddy_quad`, the fact that it was not being properly executed was missed because `dwifslpreproc` does not return non-zero if it fails to execute.

Unfortunately I couldn't easily resolve just this one thing. I've ended up having to update the FSL version as well. Which is not fantastic for a patch MRtrix update, since there's some chance of a change in behaviour out of our control.

Regarding #2921, if this change is deemed acceptable I can push it to DockerHub as `latest`, which would make it easy for people to access immediately but without necessitating an MRtrix patch update.

PR #3039, opened 2024-11-20
[ENH] Update Dockerfile

arnaudbore:
Dockerfile has not been update for quite some time, I based my modifications on your branch container_eddyqc_fix and based the whole container on the non slim version of python3.8 to get more core system commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 ARG MAKE_JOBS="1"
 ARG DEBIAN_FRONTEND="noninteractive"
 
-FROM python:3.8-slim AS base
-FROM buildpack-deps:buster AS base-builder
+FROM python:3.8 AS base
+FROM buildpack-deps:bookworm AS base-builder
 
 FROM base-builder AS mrtrix3-builder
 
@@ -22,7 +22,8 @@ RUN apt-get -qq update \
         libqt5opengl5-dev \
         libqt5svg5-dev \
         libtiff5-dev \
-        qt5-default \
+        python3 \
+        qtbase5-dev \
         zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


Lestropie:
Is there a way to add just the requisite runtime system dependencies to the image recipe rather than resorting to the non-slim base? If it’s just a couple of things missing here and there, it’d be preferable to add those explicitly, just to keep the size as small as possible. If it’s a very long list, or the reduction in size that would be achieved is minimal, then just adopting the non-slim base is justified.

arnaudbore:
Thanks for the quick review, I'm going to check. I also sent you an email and put @jdtournier in cc for more context.

Dockerfile
```diff
@@ -51,10 +52,10 @@ FROM base-builder as freesurfer-installer
 WORKDIR /opt/freesurfer
 RUN curl -fsSLO https://raw.githubusercontent.com/freesurfer/freesurfer/v7.1.1/distribution/FreeSurferColorLUT.txt
 
-# Download minified FSL (6.0.4-2)
+# Download minified FSL (6.0.7.7)
```
Lestropie, line 55: Is there a strong motivation to update the FSL version here? We try to avoid introducing anything that would modify results (over and above "bug fixes") within a patch release.

arnaudbore: I created this PR based on your modifications: https://github.com/MRtrix3/mrtrix3/blob/container_eddyqc_fix/Dockerfile.

Lestropie:
Ah OK; this is actually a superset of #2929. Clearly there I made the decision that along with fixing how the dependencies are captured I also wanted to increment the dependency version. I'd prefer to merge #2929 first, after which this PR should show only the proposed change that is specific to your need.

arnaudbore:
@Lestropie  @MRtrix3/mrtrix3-devs IMO Ready to be merged :+1:

jdtournier (approved):
:+1:

PR #3044, opened 2024-11-24
dwi2adc: Change IVIM interface

Lestropie:
As discussed in #2578.

After trialling the IVIM fit in `dwi2adc` for a facility user, I had to change the interface out of shear frustration.


Lestropie:
- [ ] Currently the S0 and ADC images are concatenated together in order to compare the outcome against the output from a prior version of `dwi2adc` where these two images were concatenated in a 4D image. It would likely be preferable to instead split these within the test data.
- [ ] Add test data for IVIM

PR #2330, opened 2021-06-01
edited documentation 5ttgen hsvs -white_stem

nicdc:
Please enter a concise description of the changes proposed in the pull request,
along with any relevant links to external discussions.

-  If your pull request involves fixing a bug, then the relevant code
   should be in a branch that is a derivative of `master`, and you should
   select `master` as the target branch in the pull request; for all other
   pull requests, the relevant code should be in a branch that is a
   derivative of `dev`, and you should select `dev` as the target branch
   of the pull request (assuming no other more appropriate branch exists).

-  If your proposed changes address an [existing Issue](https://github.com/MRtrix3/mrtrix3/issues)
   listed on GitHub, please add a reference to that Issue (a "`#`"
   character followed by the issue number).

-  If your proposed changes are not yet ready to be merged, but you are
   creating a pull request in order to initiate communications on such,
   please flag it as a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

-  You may optionally add `@mentions` of a person or team appropriate
   for reviewing proposed changes.
@Lestropie

PR #3045, opened 2024-11-24
Fsl first precalc

Lestropie:
Work In Progress.

Final relevant changeset being extracted from #2330 to allow its closure. I have extracted the code changes relating to addition of the `-first_dir` option to `labelsgmfirst` / `5ttgen fsl` / `5ttgen hsvs` and applied them to `dev` with authorship to @nicdc. From here I will need to:

- [ ] Resolve against various changes made to `dev` since #2330 was listed. This will include use of f-strings, typed command-line argument variables, ...
- [ ] Re-arrange some code; eg. checking validity of user input before any processing commences
- [ ] Make code more robust to user input; eg. ideally not requiring specific filename prefixes to be used
- [ ] Add tests that manually run FIRST, then run these scripts providing those data as input

### DIFF
--- a/docs/reference/commands/5ttgen.rst
+++ b/docs/reference/commands/5ttgen.rst
@@ -213,6 +213,10 @@ Options specific to the "fsl" algorithm
 
 - **-premasked** Indicate that brain masking has already been applied to the input image
 
+- **-first_dir /path/to/first/dir** use pre-calculated output of FSL FIRST previously run on input T1-weighted image; data must be defined in the same space as input T1w
+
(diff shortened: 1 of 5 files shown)


Lestropie:
Noting relevance to #3073. For some (not all) *MRtrix3* Python commands, it should be possible to take the outputs of third-party software commands as input rather than executing them internally. This would make it possible to utilise an *MRtrix3* container that does not encapsulate those third-party softwares; the user would be responsible for generating their derivatives externally.

PR #3049, opened 2024-12-02
voxel2mesh fixes

Lestropie:
Two minor glitches I found in `voxel2mesh` when diagnosing issues reported by a colleague.

### DIFF
--- a/cmd/voxel2mesh.cpp
+++ b/cmd/voxel2mesh.cpp
@@ -70,8 +70,10 @@ void run ()
     Surface::Algo::image2mesh_blocky (input, mesh);
 
   } else {
+
     auto input = Image<float>::open (argument[0]);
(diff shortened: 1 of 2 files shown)


Lestropie:
For discoverability: what is rectified in f666c767214fa2ed096011356170c36b962295e1 is that feeding the result to `meshfilter` would result in "Cannot perform smoothing on this mesh: no triangulation information" from [here](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/src/surface/filter/smooth.cpp#L45). The original `voxel2mesh -blocky` implementation just generated vertices at voxel corners as required, resulting in duplicate vertices; a later change was supposed to re-use existing vertices so that the resulting surface would be closed, but was not functional.

daljit46 (approved):

PR #3052, opened 2024-12-16
Print build information at the end of configuration

daljit46:
A small addition that prints build instructions when CMake has finished configuring the project. Also add a suggestion to try to remove `CMakeCache.txt` (or use the `--fresh` flag) if reconfiguration fails (this is a less drastic measure than `rm -rf build`).

### DIFF
--- a/cmake/BuildInfo.cmake
+++ b/cmake/BuildInfo.cmake
@@ -1,7 +1,8 @@
 function(mrtrix_print_build_instructions)
     set(message_list "\n")
+    file(RELATIVE_PATH relative_build_path "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}")
     list(APPEND message_list "MRtrix3 has been successfully configured for compilation.")
-    list(APPEND message_list "To build MRtrix3, run 'cmake --build .' from the build directory.")
+    list(APPEND message_list "To build it, run 'cmake --build ${relative_build_path}'")
 
     list(APPEND message_list "To reconfigure the project, run CMake again with the same arguments.")
     list(APPEND message_list "If reconfiguration fails, remove the CMakeCache.txt file in the build "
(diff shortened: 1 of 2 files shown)


cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,14 @@
+function(mrtrix_print_build_instructions)
+    set(message_list "\n")
+    list(APPEND message_list "MRtrix3 has been successfully configured for compilation.")
+    list(APPEND message_list "To build MRtrix3, run 'cmake --build .' from the build directory.")
```
Lestropie, line 4: Personally I tend to build from the source root, and it would be expected that the source root would be the current working directory when running the configuration step. So would it be better to have the example build invocation refer to the relative path to the build directory by name (eg. `"To build MRtrix3, run: 'cmake --build build/'."`)?

daljit46: That makes sense to me. I've now changed the instructions to call 'cmake --build build_directory' instead

Lestropie:
The CI failures from `pylint` reflect code present on `master`, so I am addressing them there in #3054; ideally merge that first and then have a mass merge of `master` to `dev` after tagging `3.0.5`.

cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,15 @@
+function(mrtrix_print_build_instructions)
+    set(message_list "\n")
+    file(RELATIVE_PATH relative_build_path "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}")
+    list(APPEND message_list "MRtrix3 has been successfully configured for compilation.")
+    list(APPEND message_list "To build it, run 'cmake --build ${relative_build_path}'")
+
+    list(APPEND message_list "To reconfigure the project, run CMake again with the same arguments.")
+    list(APPEND message_list "If reconfiguration fails, remove the CMakeCache.txt file in the build "
+        "directory or run CMake adding the --fresh flag.")
```
Lestropie, line 9:
```suggestion
        "directory, or include the --fresh flag when running cmake --build")
```

"Run CMake" is ambiguous as to whether running a configuration or a build.

PR #3054, opened 2024-12-26
Pylint fixes for master branch

Lestropie:
#3052 is failing CI for multiple reasons. One of them is errors during `pylint` testing, presumably due to the `pylint` version on the Action being incremented. Here I have addressed those issues on `master`, as well as some other Errors that `pylint` raised on my own system when checking `master`.

Some changes are due to warnings about reading from uninitialised variables, which arise due to using a "`elif variable == 'string'`" substitute-for-`switch` pattern that is not guaranteed to be exhaustive from the perspective of the linter. I tend to prefer to keep that pattern rather than using an "`else:`" for the last option as it's more robust to future augmentation.

I have *not* addressed the warning about `distutils` being deprecated.

On my local system I also get a lot of these, will wait to see if GitHub complains about it also:
```
pylint: Command line or configuration file:1: UserWarning: 'Exception' is not a proper value for the 'overgeneral-exceptions' option. Use fully qualified name (maybe 'builtins.Exception' ?) instead. This will cease to be checked at runtime when the configuration upgrader is released.
************* Module testing/pylint.rc
testing/pylint.rc:1:0: E0015: Unrecognized option found: argument-name-hint, attr-name-hint, class-attribute-name-hint, class-name-hint, const-name-hint, function-name-hint, inlinevar-name-hint, method-name-hint, module-name-hint, variable-name-hint, no-space-check (unrecognized-option)

```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -27,7 +27,7 @@ jobs:
     - name: install dependencies
       run: |
          sudo apt-get update
-         sudo apt-get install clang libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev python3-distutils
+         sudo apt-get install clang libqt5opengl5-dev libqt5svg5-dev libglvnd-dev libeigen3-dev zlib1g-dev libfftw3-dev
 
     - name: configure
       run: ./configure -assert || { cat configure.log; false; }
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -182,13 +182,7 @@ jobs:
       run: python3 -m sphinx -n -N -W -w sphinx.log docs/ tmp/
 
     - name: install build dependencies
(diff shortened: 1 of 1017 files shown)


Lestropie:
While c5c41a81cf49e9953fa0a79ae4d6b526e30d8644 is unrelated to `pylint`, it is requisite to pass the Linux Clang CI test. If proposed separately it would not be possible to merge either set of changes onto `master` independently as neither would pass all tests. While it would be better to create two separate PRs for the two separate changes and then construct a PR containing both sets of changes that could actually be merged, I don't think it's worth the effort in this particular instance, especially given it's holiday season.

daljit46 (dismissed):

Lestropie:
Requires a renewed review; merge failed due to needing to re-run `update_copyright`.

daljit46 (dismissed):

Lestropie:
Sorry, need another one; forgot to run `generate_user_docs.sh` after `update_copyright` :-/

daljit46 (approved):

PR #3056, opened 2024-12-30
dwi2tensor: fix operation with -iter 1

Lestropie:
Problem has been present since initial addition of iterative reweighted least squares for tensor fitting in #1542.

Does *not* affect default operation; only affects the specific use case of `-iter 1`.


bjeurissen (dismissed):

Lestropie:
Test data contained sample `dwi2tensor` outputs with different numbers of iterations, but images with 0 and 1 iterations were identical. So friendly reminder that regression testing doesn't ensure that the generated outputs are correct :melting_face:

daljit46 (approved):

PR #3067, opened 2025-02-16
Fix C++ CLI for .type_various()

Lestropie:
Downstream of #2911.

Found that when transitioning a command-line option from a fixed type to a `.type_various()`, I got an error relating to attempting to resolve a bad variant. Seems my CLI tests were not comprehensive enough to catch; I might look at expanding that before merging.

#2580 would fix this problem, as each individual possible type would have its own limits data instantiated.


Lestropie:
Superseded by #3074.

Issue #3068, opened 2025-02-18
ENH cmake: Target per Python command


Lestropie:
On `dev` it's possible to build only one specific C++ binary (and all requisite dependencies to do so) by specifying `--target` at the command-line. This however can't be done for Python commands: everything is all done within a small number of targets: copying / linking API files, copying / linking source files, generating version file, building executable Python files.

It would be preferable to have one target per command, so that `--target` works as expected. Doing so would still need to compile all C++ binaries, copy / link all API files, and generate the version file, but would only copy / link source files and generate the executable Python file for that one command.

@daljit46 Is this something you could take on?

Lestropie:
Mostly on principle. I have a couple of projects where I'm building multiple different versions of the software within a single container, and want full control over what I build within each. It makes more sense to me to have each individual commands show up in the build progress / logs. Also I'm getting into the routine of using `cmake --target` when doing development work on specific commands, and it makes sense if testing changes to a Python command to be able to do the same.

It'd be more effective if we had the ability to explicitly specify for each Python command which MRtrix C++ binaries are required, such that if a build nominates a specific Python command target, only those MRtrix C++ commands potentially invoked by that command need be built. Slightly trickier so deferring on that for now.

Lestropie:
Closed by #3069.

PR #3069, opened 2025-02-20
Add individual targets for Python commands

daljit46:
This PR aims to address the issue raised in #3068 . The developer can now specify an individual Python target to build (e.g. `cmake --build build --target dwicat`), just like it's already possible for C++ commands. It's also possible to build all Python commands by building the `PythonCommands` target.


Lestropie (approved):
Can request to build only one specific Python command, and this will copy the source code & generate the executable file for just that one command, and build all C++ binaries, but omit all elements relating to testing; subsequently doing a complete build adds all the missing pieces. :+1:

PR #3071, opened 2025-03-10
Fix tck writer when a single streamline exceed the buffer size

jdtournier:
This addresses an issue raised on the forum, and is due to our use of a write-back buffer (to prevent file segmentation) with a default size of 16MB. This is ample for regular streamlines, but clearly not enough in all cases...

Original issue:
https://community.mrtrix.org/t/tckgen-seed-point-visualization/8271/2


Lestropie (approved):
LGTM

Issue #2684, opened 2023-07-18
Alternative minification encoding


Lestropie:
In discussing the utility of the NeuroDocker minification process for another project, a prospective alternative utilisation came to mind.

Currently, what happens for a given external dependency (eg. FSL) is:

1.   A container is built that contains the full installation of that software.
2.   A set of exemplar *MRtrix3* scripts is executed to determine which features of that external software may be utilised.
3.   The features of that software not utilised are deleted from the container.
4.   The remaining features are tarballed and uploaded online.
5.   When building the *MRtrix3* image, that tarball is downloaded.

What I dislike about this strategy is that we are very explicitly providing parallel downloads of other peoples' softwares.

What we could do instead is:

4.   Generate and store in the main repository a text file that encodes which files need to be retained / removed from the installation of that software.
5.   When building the *MRtrix3* image, retain / remove the relevant files.

This means that if anyone is logging download counts of their software at their end, image build triggers will still count toward that total.

(Note this isn't the only change I want to make w.r.t. containers, but it might be a contribution in the right direction)

Lestropie:
In looking into #2921, I'd like to promote this to the `3.1.0` milestone. In retrospect I really don't like bundling substantial portions of other people's software, and then having downloads of such that the creators of those softwares can't track, even if it is technically more reliable if it's fully under our control in that way.

@daljit46 is this something you could take on?

We'd need:
- [ ] For each external dependency:
    - [ ] After generating the minified container, run something like `find -print` to get a list of all preserved files and directories
    - [ ] Add text files containing those lists to the main repository
    - [ ] Modify container recipes to install / build the full external dependency, then delete all files / directories that don't appear in the corresponding preservation list
- [ ] Soft deprecation of https://github.com/MRtrix3/containers repository:
    - [ ] Move content to an eg. `containers/` sub-directory, to live alongside the file lists
    - [ ] Update README instructions to instead instruct on committing updates to external software versions / file lists rather than uploading tarballs to OSF

daljit46:
> @daljit46 is this something you could take on?

Sure :)

Lestropie:
See implementation of alternative encoding at: https://github.com/MRtrix3/mrtrix3-with3p/commits/dev/

PR #3073, opened 2025-03-11
Docker: No third-party software embedding

### LINKED ISSUE #2601: Containers: Split w.r.t. dependencies
The topic of *MRtrix3* Python scripts making use of commands from other neuroimaging software packages has always been thorny. With `dwi2mask` in #2197 expanding that list, requiring a greater number of dependencies to be encapsulated into the containers in #2529, it becomes even thornier, and it's getting harder to justify dragging all of those external tools into "**the** *MRtrix3* official containers".

What I wish to propose here is that we provide *two* image recipes:

1.  One that contains **only** *MRtrix3* commands.
    Ideally, attempting to execute any *MRtrix3* command that has an external dependency would yield a very specific error, saying that it is not possible to execute that command because of the absence of the requisite external dependency within the container.

3.  One that contains **both** *MRtrix3* **and** any tools from other neuroimaging software packages that the scripts within such may make use of.
    This would essentially be the same as the containers that are already provided as at `3.0.3` & `3.0.4`. However the proposed change would provide an important distinction: the existence of the first container above would implicitly highlight that this second container would be *MRtrix3* **and** a bunch of other things that certain scripts may use. So it would be a container of broader applicability, including components of multiple neuroimaging packages, but it happens to be maintained by the *MRtrix3* developers since that's its primary purpose.

Thoughts?

### LINKED ISSUE #2736: Containers: cmake support
#2689 did not make the requisite changes to the container recipe files to build & install MRtrix3.

Lestropie:
Closes #2601.
Closes #2736.

-----

Working toward desired long-term solution for software containerisation.

The goal here is that "the *MRtrix3* container" will *not* contain any third-party neuroimaging software tools. What will therefore happen following this PR, if one executes within this container an *MRtrix3* Python command that attempts to execute such a third-party command, that *MRtrix3* command will error out, with a suitable error message explaining the situation. 

In conjunction with this, I aim to create a new repository within the *MRtrix3* organisation, which provides the requisite data to construct a container that encapsulates *both* *MRtrix3* *and* the set of third-party neuroimaging software commands that may be invoked by MRtrix3 Python commands. However that container will be explicitly labelled as having additional contents over and above just *MRtrix3* (ideas for suitable naming welcome; "*MRtrix3* and friends"? :upside_down_face:).

-----

- [ ] Make corresponding changes to `Singularity`
    I've been having trouble doing Apptainer builds for some reason, so haven't finished the requisite work there.

- [x] Create new page in documentation explaining the dependence of some *MRtrix3* commands on third-party tools, why those are unavailable in the official "core" "exclusively *MRtrix3*" container, and what options are available to users
    - [x] Update bash file that generates the error message to link to this new documentation page

- [x] Create new repository for container that includes third-party commands
    - [x] Encode minified dependencies based on text files listing items to be preserved, rather than tarballing and distributing other people's software (#2684)
        (I already have some progress on this)


Lestropie:
I have created the new repository and the first working version of the standalone container with alternative minification encoding: https://github.com/MRtrix3/mrtrix3-with3p/commits/dev/

PR #3074, opened 2025-03-11
API: Replace .type_various()

### LINKED ISSUE #2580: Reduce generalization of `.type_various()`

Lestropie:
Closes #2580.
Supersedes #3067.
More info via 3ce5b54323136c6720d07c0524d652d77f145c78.

Knew that fixing what I think is the wrong implementation via #3067 was going to irritate me; so went ahead and re-implemented in the way I think is preferable. The changes to the CLI tests should show what's different.

For `__print_full_usage__` (which remains unused elsewhere to the best of my knowledge), I chose to keep all possible types for a given command-line argument on the same line, with `INT` / `FLOAT` always followed by two lower and upper bounds, and `CHOICE` is always the last one in the list given the risk of one of the possible input strings clashing with one of the argument type identifiers.

### DIFF
--- a/cpp/core/app.cpp
+++ b/cpp/core/app.cpp
@@ -185,45 +185,6 @@ std::string underline(const std::string &text, bool ignore_whitespace = false) {
 
 } // namespace
 
-const char *argtype_description(ArgType type) {
-  switch (type) {
-  case Boolean:
-    return ("boolean");
-  case Integer:
-    return ("integer");
-  case Float:
-    return ("float");
-  case Text:
-    return ("string");
-  case ArgFileIn:
-    return ("file in");
-  case ArgFileOut:
-    return ("file out");
-  case ArgDirectoryIn:
-    return ("directory in");
-  case ArgDirectoryOut:
-    return ("directory out");
-  case ImageIn:
-    return ("image in");
-  case ImageOut:
-    return ("image out");
-  case Choice:
-    return ("choice");
-  case IntSeq:
-    return ("int seq");
-  case FloatSeq:
-    return ("float seq");
-  case TracksIn:
-    return ("tracks in");
-  case TracksOut:
-    return ("tracks out");
-  case Various:
-    return ("various");
-  default:
-    return ("undefined");
-  }
-}
-
 std::string help_head(int format) {
   if (!format) {
     return std::string(NAME) + ": " +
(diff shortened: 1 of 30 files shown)


Lestropie:
- [ ] Given strong CLI typing now present for Python on `dev`, need to remove `app.Parser.Various` and replace any instances of such with new classes that will infer the type based on the set of possible data types and the user input string and return a value of appropriate type accordingly.

cpp/core/cmdline_option.h
```diff
@@ -154,50 +172,44 @@ class Argument {
...
-    type = ImageIn;
+    types |= ImageIn;
     return *this;
   }
 
   //! specifies that the argument should be an output image
   Argument &type_image_out() {
-    assert(type == Undefined);
-    type = ImageOut;
+    types = ImageOut;
```
daljit46: Typo, I think it should be `types |= ImageOut`.

cpp/core/app.h
```diff
@@ -141,10 +139,16 @@ class ParsedArgument {
 public:
   operator std::string() const { return p; }
 
-  const std::string &as_text() const { return p; }
-  bool as_bool() const { return to<bool>(p); }
+  const std::string &as_text() const {
+    assert(arg->types | Text);
```
daljit46: We should & here, otherwise the assertion is always true.

cpp/core/app.h
```diff
@@ -141,10 +139,16 @@ class ParsedArgument {
...
+    return p;
+  }
+  bool as_bool() const {
+    assert(arg->types | Boolean);
```
daljit46: We should & here, otherwise the assertion is always true.

cpp/core/app.cpp
```diff
@@ -1076,61 +1008,62 @@ void parse() {
...
       }
-      if (arg.type == ArgDirectoryOut)
+      if (arg.types == ArgDirectoryOut)
         check_overwrite(text);
-      if (arg.type == TracksIn && !Path::has_suffix(text, ".tck"))
+      if (arg.types == TracksIn && !Path::has_suffix(text, ".tck"))
         throw Exception("input file \"" + text + "\" for option \"-" + std::string(i.opt->id) +
                         "\" is not a valid track file");
-      if (arg.type == TracksOut && !Path::has_suffix(text, ".tck"))
+      if (arg.types == TracksOut && !Path::has_suffix(text, ".tck"))
```
daljit46: I think here we should use bitwise & since `Argument::types` is no longer an enum, so equality wouldn't work for multityped arguments. So something like:
```
if ((arg->types & (ArgFileIn | TracksIn)))...
````

File: cpp/core/cmdline_option.h [hunk 5]
```diff
@@ -39,33 +39,39 @@ namespace MR::App {
...
+constexpr ArgModifierFlags AllowMultiple = 0x2;
 //! \endcond
 
+namespace {
+template <typename T> typename std::enable_if<std::is_integral<T>::type, T>::type void_rangemax() {
+  return std::numeric_limits<T>::max();
+}
+template <typename T> typename std::enable_if<std::is_floating_point<T>::type, T>::type void_rangemax() {
+  return std::numeric_limits<T>::infinity();
+}
```
daljit46, line 88: Here, I think it should be `std::is_integral<T>::value` (as we want `std::enable_if<true/false,T>`), in C++17 there's a [shortcut](https://en.cppreference.com/w/cpp/types/is_integral.html): `std::is_integral_v<T>`. Even better, we could make this clearer and explicit:

```
template <typename T>
inline constexpr T range_max() {
  static_assert(std::is_integral_v<T> || std::is_floating_point_v<T>,
   "range_max<T>: T must be integral or floating point");

  if constexpr (std::is_floating_point_v<T>) {
    return std::numeric_limits<T>::infinity();
  } else {
    return std::numeric_limits<T>::max();
  }
}
```

cpp/core/app.cpp
Lestropie: These are trickier again. Some up-front checks should only be performed if the argument cannot be any type other than that for which the check is being performed.

So for instance, in the line you picked above, your suggestion:
```cpp
if ((arg->types & (ArgFileIn | TracksIn)))
```

will perform that check if the argument is flagged either as potentially being an input file, or potentially being an input tractogram. However if the argument could *also* alternatively be eg. an input directory, then performing the subsequent `Path::is_file()` check would be inappropriate. So what's actually required in a scenario like this one I think is to make sure that it is *not* possible for the argument to be interpreted as a type other than one of those in the list:
```cpp
if (!(i.arg->types & ~(ArgFileIn | TracksIn))) {
```

cpp/core/cmdline_option.h
Lestropie: Fixed by 58728a9581b9b914f33bb1c3c2a1a49aa7c928a8.

cpp/core/app.h
Lestropie: Fixed by 58728a9581b9b914f33bb1c3c2a1a49aa7c928a8.

cpp/core/app.cpp
Lestropie: Hopefully all fixed by 58728a9581b9b914f33bb1c3c2a1a49aa7c928a8.

cpp/core/app.h
Lestropie: Fixed by 58728a9581b9b914f33bb1c3c2a1a49aa7c928a8.

cpp/core/cmdline_option.h
Lestropie: Fixed by 58728a9581b9b914f33bb1c3c2a1a49aa7c928a8.

Did just the simple fix; have not yet fully embraced C++17, might try to change developer behaviour if we get further through #2585.

cpp/core/cmdline_option.h
```diff
@@ -39,33 +36,58 @@ namespace MR::App {
...
-  ImageOut,
-  IntSeq,
-  FloatSeq,
-  TracksIn,
-  TracksOut,
-  Various
+class ArgTypeFlags : public std::bitset<15> {
+public:
+  ArgTypeFlags() = default;
+  static const ssize_t Text = 0;
```
daljit46: Do we need `static const` here or can we use something like `inline static constexpr`? Although probably there is not much practical difference.

daljit46:
This LGTM. I also agree with renaming tests that aren't independent (unit tests should always be IMO).

cpp/core/cmdline_option.h
Lestropie: `inline static constexpr` is C++17, so I'd never seen it previously. I avoided `enum` just because of the amount of `static_cast<>`'ing necessary. But `inline static constexpr` might prevent all of the `static`'s from being printed out in `gdb` backtraces, which is always annoying.

Lestropie: Resolved by e594b6dd1619883ef9e4144d5eed49003eabd6c9.

PR #3080, opened 2025-03-26
Fixes for new GA releases workflow

daljit46:
This PR fixes a mistake in the logic of uploading artefacts on our new CI workflow. Relevant only to dev, but needs to be on master (see https://github.com/MRtrix3/mrtrix3/pull/2933 for more details). Additionally, since for MacOS packaging we build eigen and fftw from source, their installation through brew is omitted.


daljit46:
@MRtrix3/mrtrix3-devs I think this can be merged.

jdtournier (approved):

PR #3084, opened 2025-04-09
Use premultiplied image format for text rendering

daljit46:
Using non-premultiplied format was causing antialiasing issues on MacOS with Qt 6.

### DIFF
--- a/cpp/gui/opengl/font.cpp
+++ b/cpp/gui/opengl/font.cpp
@@ -62,7 +62,7 @@ void Font::initGL(bool with_shadow) {
     tex_width += metric.horizontalAdvance(QChar(c)) + 2;
 #endif
 
-  QImage pixmap(max_font_width, font_height, QImage::Format_ARGB32);
+  QImage pixmap(max_font_width, font_height, QImage::Format_ARGB32_Premultiplied);
   const GLubyte *pix_data = pixmap.bits();
 
   VLA(tex_data, float, 2 * tex_width * font_height);


bjeurissen (approved):
Works here! How did you manage to come up with that solution? Feels pretty random...

daljit46:
Just guessing really. I suspected that the problem would have to be somewhere in `opengl/font.h` or `opengl/font.cpp`. Then I tried to change the format of the pixmap for the fonts to premultiplied (mainly because I vaguely remembered that at least for raster backends that's what [Qt recommends](https://doc.qt.io/qt-6/qpainter.html#performance)). 
Now I don't exactly know why things get weird with the non-premultiplied format. Anyway, I think we can merge this if it doesn't break the rendering on other platforms. One thing that might need some double-checking is whether we need to change the OpenGL blend functions in `font.h`:

```
            gl::Enable (gl::BLEND);
            gl::BlendEquation (gl::FUNC_ADD);
            gl::BlendFunc (gl::SRC_ALPHA, gl::ONE_MINUS_SRC_ALPHA);
```
If we're using premultiplied pixel values, I guess we can use `gl::BlendFunc (gl::ONE, gl::ONE_MINUS_SRC_ALPHA);`? @jdtournier any thoughts?

bjeurissen:
Maybe this is a naive suggestion, but I was wondering if we could just use Qt font rendering capabilities more directly to display text in the OpenGL context, as in:
```
#include <QOpenGLWidget>
#include <QPainter>
#include <QFont>
#include <QFontMetrics>

class GLWidget : public QOpenGLWidget {
    Q_OBJECT
public:
    using QOpenGLWidget::QOpenGLWidget;

protected:
    void paintEvent(QPaintEvent *) override {
        QPainter painter(this);
        painter.setRenderHint(QPainter::Antialiasing);
        painter.setRenderHint(QPainter::TextAntialiasing);

        QFont font("Helvetica", 20, QFont::Bold);
        painter.setFont(font);
        painter.setPen(Qt::white);

        painter.drawText(rect(), Qt::AlignCenter, "Hello OpenGL + Qt6!");
    }
};
```

daljit46:
Yes, I think that's feasible. The only caveat is that QPainter's OpenGL backend does not provide antialiasing unless multisampling is enabled. Rendering on the CPU and uploading an atlas to the GPU, as we're currently doing, doesn't have that problem.

PR #3085, opened 2025-04-09
Correct C++ file paths for doxygen

papadop:
Some paths have changed in the dev version vs the master. Consequently several paths in Doxyfile are wrong.
This simple PR just corrects that.
And while I was there I also update Doxyfile using doxygen -u (doxygen version 1.12.0).

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -161,9 +161,8 @@ STRIP_FROM_PATH        =
 # specify the list of include paths that are normally passed to the compiler
 # using the -I flag.
 
-STRIP_FROM_INC_PATH    = core \
-                         src \
-                         cmd
+STRIP_FROM_INC_PATH    = cpp/core \
+                         cpp/cmd
 
 # If the SHORT_NAMES tag is set to YES, doxygen will generate much shorter (but
 # less readable) file names. This can be useful is your file systems doesn't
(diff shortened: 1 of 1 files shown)


Lestropie:
Thanks for the PR @papadop. I never touch doxygen myself so would never have caught it.

I did make one change on your branch. For commits that involve large amounts of automatically-generated content, we re-author using the @MRtrixBot account, to try to maintain faithfulness of the number-of-line-changes statistic as "volume of contributed effort toward the software" per author. Only commenced using this approach after we realised the deleterious effects of its absence, but we're trying to stick to it from now on. The first two commits remain unchanged, and you'll be attributed in the curated changelog for 3.1.0.

Thanks again for the effort!

daljit46:
The CI failures have been fixed by #3089. The easiest path to merge this PR would be to rebase the changes onto `dev`.

Lestropie:
@daljit46 Looks like more CI failures unrelated to the content of the PR; are you able to take a look?

Lestropie:
> The CI failure on Windows is due to dirrotate. It should have been fixed in https://github.com/MRtrix3/mrtrix3/pull/3008 but occasionally the default test still fails.

Whoops, didn't spot that one of them was my doing. Yep, that one's entirely stochastic, and isn't given many iterations for speed, so will fail from time to time. Using a fixed input rather than running `dirgen` would make it more consistent, but there's always the risk of missing some fault by only testing on a prescribed input...

Lestropie:
Looks like `clang-tidy-review` is still failing (though I'm happy to merge this PR without it), seemingly a permissions issue in its trying to post its comment to GitHub.

PR #3086, opened 2025-04-11
Fix packaging workflows on master

### LINKED ISSUE #3007: Fix CI release workflows for master branch
In light of our upcoming 3.0.5 release, we should test and fix the CI workflows on `master`. Currently, according to my testing on a separate fork, all of them seem to be broken and result in immediate failure.

daljit46:
This PR fixes all of our packaging workflows in preparation for the 3.0.5 release. Main changes:
- Windows workflow now uses the same Github Action used on `dev` to set up an MSYS2 environment (MINGW64 instead of UCRT64 for compatibility reasons). Additionally, an explicit addition of `-lopengl32` linker flag has been made (otherwise the build was failing).
- The MacOS conda workflow now uses the [conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda) to set up a `miniconda` environment.
- Qt 5.15.2 is used for the MacOS native build (we may want to discuss this further).
- Extensions of release artefacts have been changed to match the new standards used by package managers (`conda` and `pacman`).

Since GitHub Actions are painful to debug, I tested these on a separate fork of MRtrix3. The workflows seem to produce the expected results with the exception of the MacOS native build script, which is configured to directly clone and build the MRtrix3 repo (the release tag that triggers the workflow must be present in the MRtrix3 official repo). We should probably do a test release to check that all of them work as intended.

Closes #3007.

### DIFF
--- a/.github/workflows/package-linux-anaconda.yml
+++ b/.github/workflows/package-linux-anaconda.yml
@@ -28,6 +28,6 @@ jobs:
     - name: Upload package to GitHub Release
       uses: AButler/upload-release-assets@v2.0
       with:
-        files: '*.tar.bz2'
+        files: '*.conda'
         repo-token: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/package-macos-anaconda.yml
+++ b/.github/workflows/package-macos-anaconda.yml
@@ -6,33 +6,34 @@ on:
 
 jobs:
   package:
-
     runs-on: macos-latest
 
     steps:
+      - name: Set up Miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniconda-version: "latest"
 
-    - name: fetch MacOSX 10.11 SDK
-      run: curl -L https://github.com/phracker/MacOSX-SDKs/releases/download/MacOSX10.11.sdk/MacOSX10.11.sdk.tar.xz | sudo tar xf - -C /opt/
+      - name: Install conda-build
+        run: conda install -y conda-build
 
-    - name: install conda-build # and anaconda-client
-      run: sudo $CONDA/bin/conda install -y conda-build # anaconda-client
+      - name: fetch MacOSX 10.11 SDK
+        run: curl -L https://github.com/phracker/MacOSX-SDKs/releases/download/MacOSX10.11.sdk/MacOSX10.11.sdk.tar.xz | sudo tar xf - -C /opt/
 
-    - name: fetch recipe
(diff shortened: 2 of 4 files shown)


Lestropie (dismissed):
Very difficult one to review. Technically could look at the results you obtained on your fork, but I'm trusting that you've verified everything as well as can be. It's probably a matter of re-verifying as we generate the tag and correcting anything that comes up before announcing that the version is available. So if you're content yourself, I'm content to merge whenever you see fit.

daljit46:
Further testing of the MacOS native workflow (the only one I couldn't verify because of the reason mentioned above) seems to show some problems:
- In order to, minimise build conflicts, the [build script](https://github.com/MRtrix3/macos-installer) strips PATH to: `/bin:/sbin:/usr/bin:/usr/sbin`.  This causes a failure because the build scripts can't find Python. A (hacky but workable) solution is to symlink Python in a local directory and that to PATH when running the build script.
- There seems to be an error in linking Qt.

So will need to solve this before merging.

bjeurissen:
> * In order to minimise build conflicts, the [build script](https://github.com/MRtrix3/macos-installer) strips PATH to: `/bin:/sbin:/usr/bin:/usr/sbin`.  This causes a failure because the configure/build scripts can't find Python. A (hacky but workable) solution is to symlink Python in a local directory and add that to PATH when running the build script.

The macOS bundled python version should be in the default PATH, no? Is this not a python2 vs python3 issue? As far as I know python3 should be in the default PATH.

The main reason I reset the PATH to the bare macOS path is to make sure we only rely on OS-bundled libs and do not accidentally use third party libs (from homebrew, cmake installations, custom python installs, etc), which can’t be relied on for a binary installer.

daljit46:
> Is this not a python2 vs python3 issue? 

Sorry, yes that's correct. However, we can't create a symlink in `/usr/bin` so either we have to call `python3 ./configure` (similarly for `build`) or we can use the solution I proposed above.

daljit46:
I have now updated the MacOS build script to support the building of universal binaries. To do this, I have upgraded the Qt version to 5.15.16 (the support for building universal binaries was introduced in version 5.15.9). I have tested the script locally and also using a test workflow on GitHub Actions. It seems to produce the expected results.
Additionally, to obviate the Python issue as outlined above, the workflow now creates a local symlink to `/usr/bin/python3` and passes this to the build script. 

I think this can now be merged.

jdtournier (approved):

PR #3096, opened 2025-05-14
Experimental GPU-compute abstraction using WebGPU

daljit46:
This PR is a preview of a highly experimental, cross-platform GPU abstraction layer built on top of [WebGPU](https://www.w3.org/TR/webgpu/). The aim is to enable high-performance GPU compute in MRtrix3. The intention is *not* to merge this into `dev`, but to gather feedback on the API and potentially start a discussion about future integration into the codebase.

WebGPU is a new graphics and compute API developed by the W3C, the World Wide Web Consortium, providing a common interface over Vulkan, Metal, and DirectX. While originally designed for the web, there are native implementations written in C++ ([dawn](https://dawn.googlesource.com/)) and Rust ([wgpu](https://github.com/gfx-rs/wgpu)). This work builds on top of Dawn by wrapping common tasks to avoid boilerplate—such as buffer uploads, shader loading, and dispatch. The goal is to keep the API simple and ergonomic, extending functionality later as needed.

Modern GPU programming requires a significantly different mental model than CPU programming. For the uninitiated (with much oversimplification due to brevity), here are some key concepts:

* The GPU executes *compute shaders* (or “kernels” in CUDA parlance) by launching many lightweight “threads” in parallel.
* Threads are grouped into *workgroups*. You define a workgroup by its 3D dimensions (`x, y, z`), so that each workgroup contains `x × y × z` threads.
* To actually run your kernel, you issue a launch with a “dispatch grid,” specifying how many workgroups to spawn in each dimension (`wgCountX × wgCountY × wgCountZ`). The GPU scheduler then maps those workgroups onto its compute units.
* Each thread in a workgroup has built-in IDs like:

  * `global_invocation_id`: which workgroup it belongs to in the grid,
  * `local_invocation_id`: its index within the workgroup,
  * `workgroup_invocation_id`: the index of the workgroup that the thread is part of.
* The total number of threads per launch is `(wgCountX × wgCountY × wgCountZ) × (x × y × z)`.
* Kernels are written from the point of view of a single thread: you author your shader as if it were running on exactly one thread, using those IDs to process data. Within a workgroup, threads can share fast local memory and synchronize with barriers, but there is no cross-workgroup synchronization at the end of each dispatch.

For example, to cover a 2D image of size (W, H) with a 16 × 16 workgroup, you’d dispatch `(ceil(W/16), ceil(H/16), 1)` workgroups, yielding one thread per pixel.

For a complete overview of the API, see the `gpu.h` header.

---

Here is a usage example that uploads a buffer to the GPU, performs an element-wise `sqrt` operation, and writes the results into a second buffer of the same size.

In C++, we create the buffers, schedule the GPU kernel, and then download the output:

```cpp
// CPU-side input
std::vector<float> inputVector(1'000'000);
std::iota(inputVector.begin(), inputVector.end(), 0.0F);

// Create GPU buffers
Buffer<float> inputBuffer  = context.newBufferFromHostMemory<float>(inputVector);
Buffer<float> outputBuffer = context.newEmptyBuffer<float>(1'000'000);

// Define our kernel
const KernelSpec kernelSpec {
  .computeShader    = { .shaderSource = ShaderFile { "sqrt_kernel.wgsl" } },
  .readOnlyBuffers  = { inputBuffer },
  .readWriteBuffers = { outputBuffer }
};

// Create the kernel and dispatch it
// We assume a 1D workgroup size of (64, 1, 1) = 64 threads per workgroup.
// Since we have 1 000 000 elements, we need 1'000'000/64 = 15'625 workgroups so each thread handles one element.
const Kernel       kernel       = context.newKernel(kernelSpec);
const DispatchGrid dispatchGrid = { 1'000'000/64, 1, 1 };
context.dispatchKernel(kernel, dispatchGrid);

// Download the data and wait for all work to finish
std::vector<float> outputData(inputVector.size());
context.downloadBuffer<float>(outputBuffer, outputData);
```

The kernel itself is written in [WGSL](https://www.w3.org/TR/WGSL/), a new shader language for WebGPU that resembles Rust:

```wgsl
// Define the “slots” of our kernel: two arrays of floats.=
@group(0) @binding(0) var<storage, read>      input:  array<f32>;
@group(0) @binding(1) var<storage, read_write> output: array<f32>;

// Define the size of the workgroup: a 1D workgroup with 64 threads along x
@compute(64, 1, 1)
fn main(@builtin(global_invocation_id) globalId: vec3<u32>) {
  output[globalId.x] = sqrt(input[globalId.x]);
}
```

I have also created a test command `test_gpu` that performs the slightly more complicated task of applying an affine transformation to an image on the GPU.

As of now, the API is still in its early stages and has been shaped primarily by my work on performing affine registration using WebGPU, which builds on top of this work. It may change significantly based on feedback and refinements needed for the registration work, but it’s in a usable state, so feel free to experiment with it. You can create a shader in `cpp/gpu/shaders` (automatically symlinked to `bin/shaders`) and then inspect how the `test_gpu` command works.

There are many questions for discussion, notably how to integrate these dependencies into our project structure. Currently, everything lives in `cpp/gpu` without much organization. The unit tests, inappropriately residing in `cpp/gpu/testing`, use the [Google Test](https://github.com/google/googletest) framework and are invaluable both for debugging GPU code and can also serve as examples of API usage. This is slightly tangential, but I think we should integrate a unit testing framework like Google Test into the codebase.

Additionally, the C++ library used to interface with the GPU (dawn) is built from source at each build (fetched over the network), which takes considerable time and clutters the CMake's output at configure time. We should consider providing prebuilt binaries to avoid long build times.

@MRtrix3/mrtrix3-devs  any feedback is much appreciated!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(MRTRIX_USE_PCH "Use precompiled headers" ON)
 option(MRTRIX_PYTHON_SOFTLINK "Build directory softlink to Python source code rather than copying" ON)
 option(MRTRIX_BUILD_STATIC "Build MRtrix's library statically" OFF)
 option(MRTRIX_USE_LLD "Use lld as the linker" OFF)
+option(MRTRIX_BUILD_GPU "Build GPU support" ON)
 
 set(MRTRIX_DEPENDENCIES_DIR "" CACHE PATH
     "An optional local directory containing all thirdparty dependencies:\n \
--- a/cmake/FetchDawn.cmake
+++ b/cmake/FetchDawn.cmake
@@ -1,0 +1,45 @@
+# Adapted from https://github.com/eliemichel/WebGPU-distribution/blob/dawn/cmake/FetchDawn.cmake
+
+include(FetchContent)
+set(FETCHCONTENT_QUIET OFF)
+FetchContent_Declare(
+    dawn
+    DOWNLOAD_COMMAND
+            cd ${FETCHCONTENT_BASE_DIR}/dawn-src &&
+            git init &&
+            git fetch --depth=1 https://dawn.googlesource.com/dawn chromium/7060 &&
+            git reset --hard FETCH_HEAD
+)
+
+FetchContent_GetProperties(dawn)
+if (NOT dawn_POPULATED)
+    FetchContent_Populate(dawn)
+    if (APPLE)
+        set(USE_VULKAN OFF)
+        set(USE_METAL ON)
+    else()
+        set(USE_VULKAN ON)
+        set(USE_METAL OFF)
+    endif()
+
+    set(DAWN_FETCH_DEPENDENCIES ON)
+    set(DAWN_USE_GLFW OFF)
+    set(DAWN_USE_X11 OFF)
+    set(DAWN_USE_WAYLAND OFF)
+    set(DAWN_ENABLE_D3D11 OFF)
+    set(DAWN_ENABLE_D3D12 OFF)
+    set(DAWN_ENABLE_METAL ${USE_METAL})
+    set(DAWN_ENABLE_NULL OFF)
+    set(DAWN_ENABLE_DESKTOP_GL OFF)
+    set(DAWN_ENABLE_OPENGLES OFF)
+    set(DAWN_ENABLE_VULKAN ${USE_VULKAN})
+    set(TINT_BUILD_SPV_READER OFF)
+    set(DAWN_BUILD_SAMPLES OFF)
+    set(TINT_BUILD_CMD_TOOLS OFF)
+    set(TINT_BUILD_TESTS OFF)
+    set(TINT_BUILD_IR_BINARY OFF)
+
+    add_subdirectory(${dawn_SOURCE_DIR} ${dawn_BINARY_DIR} EXCLUDE_FROM_ALL)
+endif()
+
+set(FETCHCONTENT_QUIET ON)
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -35,3 +35,7 @@ add_subdirectory(core)
 if(MRTRIX_BUILD_GUI)
     add_subdirectory(gui)
 endif()
+
+if(MRTRIX_BUILD_GPU)
+    add_subdirectory(gpu)
+endif()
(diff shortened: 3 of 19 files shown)


Cewein:
This is pretty nice, have you consider looking into https://shader-slang.org/ instead of WGSL for the shading language ?
It is pretty new but could be use to with other application very simply. and would also help by not rewritting all the previous glsl shaders.

daljit46:
Hi. Yes, I’m aware of Slang. In fact, I discussed with @jdtournier the possibility of using it instead of WGSL.

In terms of language capabilities, Slang is clearly a step up from WGSL. Its support for generics and modules would eliminate the need for the custom preprocessor mechanism used in the current implementation and could potentially reduce duplication in some tasks (for example, handling images with different dimensions). Automatic differentiation could also be quite handy.

However, there are some downsides to consider:
- Adding Slang would introduce an extra dependency to the project. The Slang team does provide prebuilt binaries, so this may not be a serious issue, but it’s nonetheless worth keeping in mind.
- WGSL acts as the lowest common denominator across all graphics APIs, meaning it supports only the features common to its targets. As a result, if you find WGSL code online, you can copy and paste it without worrying about whether it will be supported on your target platform (although, due to extensions, this isn’t strictly guaranteed).

That said, I haven’t given Slang a thorough trial yet, but I’m not ruling it out. If the benefits outweigh the issues I’ve raised, we should consider adopting it for MRtrix3.

Cewein:
I looked at `gpu.h` and `gpu.cpp`. It is a great but I think if MRtrix move deep into GPU driven state we should make specific file for each part of the gpu pipeline such as `texture`, `kernel`, `buffer` and more. Nevertheless, I understand this is just the start.

I built the branch under ubuntu 24, build just fine. I Also agree that dawn clutter a lot the build time, i think having a fork of dawn that we update only for major wgpu realese could do it. Also since the spec of webgpu are not fully define yet it could be a great way to go around this caviate.

Also are you also thinking to move away from Qt6 and for example using Dear ImGui or something like this ? Qt6 is nice but ImGui provide better and faster UI in my personal opinion since ImGui is designed for GPU driven application. There is binding for WGPU. it allow interactive data interaction for GPU-CPU interaction.

PR #3108, opened 2025-06-11
Add method for canonical direct I/O layout

daljit46:
This PR introduces a new `Image::with_canonical_axes_order_layout()`, a utility method to obtain an `Image` with its data preloaded into RAM in a standardized, contiguous layout. This layout ensures that axis 0 is the most contiguous (fastest varying), followed by axis 1, then axis 2, and so on, up to the highest dimension. All axes involved in this ordering will have positive effective strides. This is a common packed layout suitable for interoperability with GPU APIs (where axis 0 often maps to width, axis 1 to height, axis 2 to depth) and possibly other external libraries. 

Checking the implementation of `Image::with_direct_io`, I believe that the implementation of the function should also ensure that the resulting stride should have all positive directions.

I'm slightly unsure about the naming though. I called the function `canonical` but I admit that may be confusing (the other one I was debating was `standard`).

### DIFF
--- a/cpp/core/image.h
+++ b/cpp/core/image.h
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <numeric>
+#include <utility>
 #define IMAGE_H
 
 #include <functional>
@@ -181,6 +183,17 @@ public:
     return with_direct_io(axis < 0 ? Stride::contiguous_along_spatial_axes(*buffer)
                                    : Stride::contiguous_along_axis(axis, *buffer));
   }
+
+  //! return new Image with data preloaded into RAM using a canonical axes order layout.
+  /*!
+   * This layout ensures that axis 0 is the most contiguous (fastest varying),
+   * followed by axis 1, then axis 2, and so on, up to the highest dimension.
+   * All axes involved in this ordering will have positive effective strides.
+   * This is a common packed layout suitable for interoperability with GPU APIs
+   * (where axis 0 often maps to width, axis 1 to height, axis
+   * 2 to depth).
+   */
+  Image with_canonical_axes_order_layout() const;
 
   //! return RAM address of current voxel
   /*! \note this will only work if image access is direct (i.e. for a
@@ -394,6 +407,12 @@ template <typename ValueType> Image<ValueType> Image<ValueType>::with_direct_io(
   }
 
   return Image(buffer, with_strides);
+}
+
+template <typename ValueType> Image<ValueType> Image<ValueType>::with_canonical_axes_order_layout() const {
+  std::vector<ssize_t> strides(ndim());
+  std::iota(strides.begin(), strides.end(), static_cast<ssize_t>(1));
+  return with_direct_io(strides);
 }
 
 template <typename ValueType> std::string Image<ValueType>::dump_to_mrtrix_file(std::string filename, bool) const {


Lestropie:
There's already the function `.with_direct_io()`, which both ensures explicit load into RAM and allows any custom axis strides. What you're doing here is one use case of that, where you want the strides of `[+1,+2,+3,...,+ndim()]`. So I think the better solution is to have a function in `cpp/core/stride.h` that yields these canonical strides, which you can then pass to `.with_direct_io()`.

One of the reasons for this more general solution is that it's quite common when working with 4D data to want data along the fourth direction to be contiguous. We don't care all that much about the spatial axes, its good to be able to manipulate them arbitrarily and that's exploited at image load but it's primarily making all data for a voxel contiguous that yields performance benefits. You might find that for what you're currently doing with 3D data you're seeking trivial spatial strides, whereas if you were to be trying to do something with 4D data on the GPU you'd instead be seeking the contiguous-along-nonspatial-axes solution.

daljit46:
Ok, so we could probably do something like this:

```
enum class StrideInterpretation {
  BestMatch,
  Exact
};
```

Then change `with_direct_io` to:

```
  // Currently we have Image with_direct_io(Stride::List with_strides = Stride::List())
  Image with_direct_io(Stride::List with_strides = Stride::List(), StrideInterpretation strideInterpretation = StrideInterpretation::BestMatch);
```

Then in `stride.h` we could also add a helper function to ensure positive strides:

```
void ensure_positive_symbolic_signs(List& symbolic_strides) {
  for (ssize_t& stride_val : symbolic_strides) {
    if (stride_val != 0) {
      stride_val = std::abs(stride_val);
    }
  }
}
```

Then, passing in a canonical symbolic list should do the job. If we want an even more general solution, I guess modifying `Stride::get_nearest_match` to accept something like `enum class SignPreference { PreserveOriginal, PreferPositive, PreferNegative, MatchDesired };` could be an option.

daljit46:
@jdtournier @Lestropie I've modified the PR based on your feedback and suggestions.

jdtournier:
The use case @daljit46 is specifically interested in here is getting images in & out of the GPU in an efficient manner, which means a single memory transfer. For that to work, we need to make sure the data are stored in the correct data type with _exactly_ the strides specified (what @daljit46 referred to as 'canonical') - otherwise we'd need to handle arbitrary strides in the GPU shader, which seems likely to complicate things too much. 

The current infrastructure _almost_ covers this, up to the guaranteeing the sign of the strides. What we would need is a call like:
```
auto in = Image<float>::open ("filename").with_direct_io (Strides::InOrder, Strides::Strict);
```
or something like that, where `Strides::InOrder` is understood to mean 1 -> n_dim, and `Strides::Strict` is understood to mean "do not treat negative strides as equivalent to their positive counterpart". 

We can easily do the `Strides::InOrder` bit with the current functionality (though it's probably a little clunky), but we don't currently cater for the strict interpretation of the strides. As I understand it, that's the issue that requires changes deeper in the API.

Lestropie:
> The current infrastructure almost covers this, up to the guaranteeing the sign of the strides.

OK, I wasn't aware of the failure to enforce stride *sign* despite a developer very explicitly providing manual strides. I think that's the origin of the divergence.

-----

If that's the case though, I'm still not sure that the proposed interface is the right approach. Having a singular data structure called "`Strides`", but then having a secondary flag indicating how the data in that structure should be interpreted only within some specific function, feels wrong.
If there are multiple types of image strides:
1.  Actualised, ie. the actual file pointer offsets to apply to step along a given axis;
2.  Symbolic, where both order and sign of non-zero entries are important;
3.  Ordinal, where the relative order of different axes matters but there is no notion of sign

, and your issue is that you need to achieve 2. but the API is only facilitating 3., then arguably these warrant being stored as different structures. The structure storing your strides should intrinsically reflect what they encode / how they are to be imposed, and not be dependent on conditional interpretation. 3 is just a permutation and should enforce non-negativity. Further, there may be elsewhere in the code base where there is a discrepancy between a developer's expectations of whether the signs of strides matters and how those data are utilised. Certainly that wasn't obvious to me, so this could well be a source of problems elsewhere...

Lestropie:
> The [implementation](https://github.com/MRtrix3/mrtrix3/blob/85290c24913598e9d6034855d0e62a6077c52d22/cpp/core/image.h#L368) changes the symbolic strides by calling `Stride::get_nearest_match()` before actualising them.

If that's the case, then perhaps rather than adding a flag that modulates whether `Image::with_direct_io()` internally calls `Stride::get_nearest_match()`, anything that is feeding `Image::with_direct_io()` with data that necessitates a call to `Stride::get_nearest_match()` should be instead giving it data that doesn't need to be run through that function (eg. by itself calling `Stride::get_nearest_match()` beforehand). 

Or instead, check the contents of Stride::List for zeros:
-   If there are any zeros present, call `Stride::get_nearest_match()`, on the basis that the developer has provided as input to the function *incomplete* symbolic strides, where the missing contents should be filled in based on the existing image layout since that's the most likely to be able to persist with a memory mapping.
-   If the input is *complete* symbolic strides, don't modify it in any way.

This is probably the more direct redress of the issue you're encountering, and doesn't require a new enum. You just provide as input a `Stride::List` with [1,2,3(,4,...)], and it gets actualised as specified.

Lestropie:
Given comments in #3238, flagging here that I've been working on a changeset intended to supersede this one. Goal is to convert stride handling from function calls to OOP, with the side-effect that symbolic (signed) strides and axis ordering (potentially with ties) will be different classes, such that "I want symbolic strides with +1,+2,+3,+4" and "I want my axes to be spatial first, then volumes, and I don't care about sign" will be different by class name rather than function argument. Couple of weeks later than intended in finishing so just flagging ahead of time.

Issue #3122, opened 2025-07-03
dwifslpreproc: incorrect pe_table fed to applytopup


bjeurissen:
When using ``dwifslpreproc`` with a simple pair + dwi combo, I noticed that ``applytopup`` after ``topup`` is not correctly applying the field. The distortions are made worse instead of less and the resulting brain mask suffers as a result from that.

~Looking into this, I noticed that ``mrinfo -export_pe_table`` produces a txt file that has the phase encoding direction flipped as opposed to what ``mrinfo -export_pe_eddy`` produces. Since ``applytopup`` is using the txt file from ``mrinfo -export_pe_table``, it appears to apply the opposite correction. Subsequent ``eddy`` on the other hand is using the txt file from ``mrinfo -export_pe_eddy``, so it seems to apply the field correctly.~ (NOTE: the issue turned out not to be pe_table vs pe_eddy, but mrinfo vs mrconvert with strides). 

With ``dwi.mif`` containing a single PE direction:
```
mrinfo dwi.mif -export_pe_table pe_table.txt -export_pe_eddy pe_eddy.txt pe_eddy_idx.txt -force
cat pe_eddy.txt
```
produces
```
0 1 0 0.0321
```
whereas

```
head -n 1 pe_table.txt
```
produces
```
 0 -1  0 0.0321
```

The cause seems to be that ``-export_pe_eddy`` boils down to ``save(transform_for_nifti_write(PE, header), path)``, which includes https://github.com/MRtrix3/mrtrix3/blob/2e4cb20f05247a8baf4f18ea8839fb3637fa7cd2/core/metadata/phase_encoding.cpp#L262-L283, whereas  ``-export_pe_table`` boils down to ``save(PE, path)``, missing the nifti-specific logic.

Lestropie:
Closed by #3123.

PR #3123, opened 2025-07-03
Generate correct applytopup_config.txt

bjeurissen:
This provides a workaround for #3122

### DIFF
--- a/bin/dwifslpreproc
+++ b/bin/dwifslpreproc
@@ -823,9 +823,9 @@ def execute(): #pylint: disable=unused-variable
     # applytopup can't receive the complete DWI input and correct it as a whole, because the phase-encoding
     #   details may vary between volumes
     if dwi_manual_pe_scheme:
-      run.command('mrconvert ' + dwi_path + import_dwi_pe_table_option + ' - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
+      run.command('mrconvert ' + dwi_path + import_dwi_pe_table_option + ' - | mrconvert - dummy.nii -strides -1,+2,+3,+4 -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
     else:
-      run.command('mrinfo ' + dwi_path + ' -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
+      run.command('mrconvert ' + dwi_path + ' dummy.nii -strides -1,+2,+3,+4 -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
 
 
     # Update: Call applytopup separately for each unique phase-encoding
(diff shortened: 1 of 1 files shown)


bjeurissen:
> Given the magnitude of the effort over at https://github.com/Lestropie/DWI_metadata I don't want to do a hack fix here. That test suite needs to evaluate `-export_pe_table` and `-export_pe_eddy` for both commands for all possible data.

Agreed!

To be clear, I think the problem is not an `-export_pe_table` vs `-export_pe_eddy` thing, but an `mrinfo input.mif -export_pe_*` vs `mrconvert input.mif -strides <strides> -export_pe_*` thing. Since `mrinfo` does not have an option to specify output strides, it can't take into account output strides in the same way that `mrconvert -strides <strides>` can. Hence, `dwifslpreproc` can be fixed by replacing `mrinfo` with `mrconvert -strides <strides>`. A cleaner solution would be for `mrinfo` to also support output strides.

bin/dwifslpreproc
```diff
@@ -823,9 +823,9 @@ def execute(): #pylint: disable=unused-variable
     # applytopup can't receive the complete DWI input and correct it as a whole, because the phase-encoding
     #   details may vary between volumes
     if dwi_manual_pe_scheme:
-      run.command('mrconvert ' + dwi_path + import_dwi_pe_table_option + ' - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
+      run.command('mrconvert ' + dwi_path + import_dwi_pe_table_option + ' - | mrconvert - dummy.nii -strides -1,+2,+3,+4 -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
     else:
-      run.command('mrinfo ' + dwi_path + ' -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
+      run.command('mrconvert ' + dwi_path + ' dummy.nii -strides -1,+2,+3,+4 -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
```
Lestropie:
```suggestion
      run.command('mrconvert ' + dwi_path + import_dwi_pe_table_option + ' -strides -1,+2,+3,+4 - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
    else:
      run.command('mrconvert ' + dwi_path + ' -strides -1,+2,+3,+4 - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt')
```

This is at least how I would be doing it from a naive first look. Should not be any need to create a dummy image; the only issue is that what's going into `applytopup` is having explicit strides set.

I still however need to diagnose properly. `Metadata::PhaseEncoding::save_eddy()` is calling `Metadata::PhaseEncoding::ransform_for_nifti_write()`, which is calling `File::NIfTI::axes_on_write()`; so it's not that the issue has been completely overlooked, there's just some assumption that's being overlooked.

bjeurissen: I have modified it accordingly, below. Github would not allow me to directly commit your suggestion because it contained deleted lines?!

bjeurissen:
For reference, here is what the eddy masks before (red) and after the patch (green) look like for this particular dataset:

screenshot0000

bjeurissen:
Indeed

Issue #3127, opened 2025-07-12
Handling of external phase encoding tables


Lestropie:
In response to #3122 I have started extending my tests over at https://github.com/Lestropie/DWI_metadata to look into the handling of *external* vs. *internal* phase encoding metadata, and the corresponding compatibility with FSL. This admittedly should have been included as a primary outcome of the original testing. But frankly I'm tired of stuff like this getting in the way of actual productivity...

There appears to be an outstanding issue over and above #3122 that should have registered in my mind a long time ago.

Just as the `bvec` format is defined with respect to flip-first-axis-if-necessary-to-make-negative-determinant image axes, this seems to apply to phase encoding information also. Importantly however, this *doesn't* only apply to the `eddy` "config" / "index" text file pair; it *also* applies to the phase encoding table with one row per image volume.

I've gotten away with this in the context of `dwifslpreproc` by using `-strides -1,+2,+3(,+4)`.

This is not something for which I can simply change the implementation. The interface `-(im|ex)port_pe_table` implies that the table is just a table, and has no implicit relationship with FSL. What is going to be necessary is for there to be **two distinct command-line options** for importing / exporting four-column phase encoding tables: one where everything is defined with respect to the image axes, and one that conforms to the FSL internal convention.

Lestropie:
`import_pe_table` and `-export_pe_table` removed in https://github.com/MRtrix3/mrtrix3/commit/38910df5149b77c8ed11d277635d35564ff8909f. Possible changes to the internal phase encoding table representation used by *MRtrix3* to be deferred for now.

One possibility that came to mind in that respect would be to encode a signed k-space traversal time per image axis. So it would be a three-column table rather than four-column (which technically the current `pe_scheme` can be three columns if the phase encoding direction varies but the readout time is unknown, but we could use a different metadata key to disambiguate). This could have the added advantage of not explicitly specifying "phase encoding direction" in the name. Eg. high-resolution anatomicals can have non-zero susceptibility distortions along the readout axis. Technically you could even have more than one axis with a non-zero value, though this would be a highly unusual acquisition given for EPI the readout direction changes per line resulting in blurring rather than distortion.

Lestropie:
Re-escalating due to #3142.

The problem with removing these two options outright is that for certain code paths, `dwifslpreproc` manually constructs the content of these tables in Python code, and uses `-import_pe_table` to get the data into the header, eg. so that `-export_pe_eddy` can then be called based on the C++ implementation.

[...]

PR #3128, opened 2025-07-13
Fixes to external phase encoding table handling

Lestropie:
As suspected following posting of #3122 and reported in #3127, there are residual issues with the handling of phase encoding data. I am creating this PR so that I can more easily pull the data into https://github.com/Lestropie/DWI_metadata and perform more exhaustive tests.

Certainly as a first pass, with current `master` code, when I combine data with more complex phase encoding contrast than just reversal, results are OK if the NIfTI is left-handed but erroneous if right-handed.

- [x] Confirm that results from `topup` and `applytopup` with > 2 phase encoding directions are correct with the code posted here.

- [x] Confirm that results from `eddy` with > 2 phase encoding directions are correct with the code posted here.
    (currently trying to run `eddy` utilising all of the data for each candidate stride list is prohibitively slow, so I will need to derive some trickery to utilise only a subset of the available data in such a way that `eddy` still works OK and can confirm that interpretation of phase encoding is correct)

- [x] Implement tests to confirm that exporting to any phase encoding text file format, then importing back in, yields the same internal representation
    (If data exported from *MRtrix3* are interpreted correctly by FSL, then making sure that an *MRtrix3* export / import pair yields no change should be adequate validation of the *MRtrix3* imports)

### DIFF
--- a/bin/dwifslpreproc
+++ b/bin/dwifslpreproc
@@ -805,7 +805,7 @@ def execute(): #pylint: disable=unused-variable
 
 
     # Do the conversion in preparation for topup
-    run.command('mrconvert ' + se_epi_path + ' topup_in.nii' + se_epi_manual_pe_table_option + ' -strides -1,+2,+3,+4 -export_pe_table topup_datain.txt')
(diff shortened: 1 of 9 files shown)


Lestropie:
I have confirmed compatibility with FSL to the best of my ability using the contents of https://github.com/Lestropie/DWI_metadata/pull/12; this is now ready for review.

bjeurissen (dismissed):

PR #3132, opened 2025-07-19
mrdegibbs: Describe complex-valued capability, update types

Lestropie:
In confirming whether I needed to put `mrcalc -abs` upstream or downstream of `mrdegibbs`, it looks like `3.0.x` reads the input image as real-valued, whereas `dev` permits complex-valued data for both 2D and 3D variants. This capability was however not mentioned at all in the description section of the command help page.

While resolving that I decided that the type definitions in this context were ambiguous (is "`value_type`" real or complex?).

### DIFF
--- a/cpp/core/degibbs/degibbs.h
+++ b/cpp/core/degibbs/degibbs.h
@@ -1,0 +1,26 @@
+/* Copyright (c) 2008-2025 the MRtrix3 contributors.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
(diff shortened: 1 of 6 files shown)


cpp/core/degibbs/degibbs.h
```diff
@@ -0,0 +1,26 @@
...
+ * For more details, see http://www.mrtrix.org/.
+ */
+
+#pragma once
+
+#include <complex>
+
+namespace MR::Degibbs {
+
+typedef double real_type;
```
jdtournier: Suggest using the more modern version instead of `typedef`:
```
using real_type = double;
using complex_type = std::complex<real_type>;
```

Lestropie: Closed by 7dfe81b5289fc5b8676909a6f1e80912cf31a293.

PR #3137, opened 2025-07-24
Terminal fixes on dev

Lestropie:
Just couple of little snags discovered while building up the next-gen FBA pipeline.

I was about to start doing a `git bisect` to try to figure out why on `dev` for C++ commands, there is often, preceding the appearance of a progress bar, an extra erroneous progress bar with no message that immediately shows 100% completion. Eg.:
````ShellSession
~/src/mrtrix3/build_debug/bin/mrconvert mask.nii deleteme.mif -force
mrconvert: [WARNING] existing output files will be overwritten
mrconvert: [100%] 
mrconvert: [100%] copying from "mask.nii" to "deleteme.mif"
```

Line 3 shouldn't be there. It does not occur if `-nthreads 0` is used.

But from my first backtrace, it's almost certainly due to #2755 copying rather than moving something. @daljit46 Could you take that one on?


Lestropie:
@daljit46 Makes sense to me, I think we just go with that.

jdtournier:
> I think what we should do here is provide the ProgressBar class with a custom move constructor that ensures the original moved-from instance has show set to false. This requires making the show member non-const (generally speaking, I think we should [avoid const member variables in a class](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type)). @jdtournier any thoughts?

Sounds spot on to me too :+1:

daljit46:
> This I believe to be the "correct" current tip of dev. I would however appreciate external verification on that one from anyone with a clone that hasn't been synced within the last 24 hours.

I only verified until a couple of commits before HEAD (as my local clone of dev is outdated by a couple of weeks), but this looks correct to me. I think you could also double-check with something like `git rev-parse "origin/dev@{2025-08-26 00:00}"` on your local clone.
If you don't want to go through the hassle of updating every PR that is out of sync with `dev`, it should be fairly straightforward to use the GitHub CLI to update all open PRs (rather than doing a local merge + push).

Lestropie:
There's a good chance that any work that anyone else had pushed to a branch on GitHub after the most recent `git fetch` that I had performed on my own system will have been lost.

It is however likely that this is flagging that there's still more fundamental issues. 2b42c34614a36f37ad52d655329bea1bcc5273b0 shows as its parent dda9fd2c2c9c765cdd888b98bd0b0f0610d24d45, which is a copy of what's currently appearing in this PR as d21d986c191342356dbcee3d0712740e803d704f.

I expect more likely what's happening is that a lot of PRs have been moved to point to rebased commits. When I force-pushed `dev` back to 70031c348fcefb16b534e342a9f87a5d1a343c94, which is where it should be, that resulted in all PRs showing massive changes, because those feature branches are now pointing to unexpectedly rebased versions. An explicit merge resolves because there's no actual differences in code content, but it's erroneous duplication nevertheless.

I'm hoping that on my local clone, I can use the `rev-parse` usage shown above to query what the tip of each branch individually used to be, and force push each individually back to where it should be.

Issue #3142, opened 2025-08-04
dwifslpreproc: mrinfo -export_pe_eddy with mif files


kpannek:
MRtrix version 3.0.7

`mrinfo -export_pe_eddy` was changed to work only in conjunction with nii files. dwifslpreproc calls this on a mif file, resulting in an error.

```
dwifslpreproc denoised_degibbsed_ap.mif -se_epi b0_appa.mif denoised_degibbsed_preproc.mif -rpe_pair -pe_dir ap
[...]
Command: mrconvert dwi.mif -strides -1,+2,+3,+4 - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt

dwifslpreproc: [ERROR] mrconvert dwi.mif -strides -1,+2,+3,+4 - | mrinfo - -export_pe_eddy applytopup_config.txt applytopup_indices.txt (dwifslpreproc:828)
dwifslpreproc: [ERROR] Information from failed command:
dwifslpreproc:
               mrconvert: copying from "dwi.mif" to "/Users/pan...18J0L/mrtrix-tmp-AMTUtP.mif"... [==================================================]
               mrinfo: [ERROR] Only export phase encoding table to FSL eddy format in conjunction with MGH / NIfTI format images
```

Lestropie:
🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤕

PR #3146, opened 2025-08-05
Fix phase encoding handling on 3.0.7

Lestropie:
Initial work on resolving #3142. Committing to GitHub so that I can run testing on another system.

- [x] Make sure that all `dwifslpreproc` tests (which don't get run as part of CI) pass.
- [x] Double-check all testing logs (`dwifslpreproc` and everything else) for any excess of warning messages.


Lestropie:
Ready for review. Have checked test logs for erroneous error messages. Python command tests pass with the exception of some `5ttgen` calls that are off by one voxel presumably due to FSL changes.

PR #3152, opened 2025-08-22
`__print_pydra_code__` "special command-line option" to automatically generate Pydra code

tclose:
This PR adds another "special command-line option" (e.g. `__print_usage_markdown__`), `__print_pydra_code__` to automatically generate command wrapper code for the [Pydra dataflow engine](https://github.com/nipype/pydra). This
has been discussed with @Lestropie and the generated code (from this branch) will be stored in the [pydra-tasks-mrtrix3](https://github.com/nipype/pydra-tasks-mrtrix3) repository. 

NB: currently there aren't any tests in the MRtrix3 CI to test that the functionality still works as expected. If possible I would be happy to add a GH action that tests that the generation code runs successfully and can be imported.

The generated code is largely usable straight away but there are some cases that aren't able to be determined from the argument/option definitions, e.g. the outputs of `connectome2tck` are specified in the argument list by the `prefix` text field, but generate a series of outputs. Any suggestions of how to handle these cases automatically would be appreciated, but if not possible my plan is to generate the code to a dedicated `auto-gen` branch of `pydra-tasks-mrtrix3` and then rebase any manual ammendments that need to be made on top of it after every regeneration.

This PR also includes a couple of minor changes to make my life easier, which I hope are ok:
* `amp2response.cpp` argument `directions` argument has been renamed to `directions_image` to avoid clash with the option of the same name
* Moved `SequenceDirectoryOut` from `python/mrtrix3/commands/population_template/usage.py` into `app.py` with the rest of the types to avoid circular import hell


Lestropie:
-   Can you clarify on the distinction between here and #2665?

-   An Action to test the generation of the usage files would probably be preferred. We would want to avoid having to push and tag patch fixes to the main repo to fix downstream applications that are outside of our purview. The scope of what to test from what is generated is up to you, but to know that they are at least generated and produce code that doesn't crash the Python interpreter would be preferred.

> ... eg. the outputs of `connectome2tck` are specified in the argument list by the prefix text field ...

Grepping for `.type_text()` should give a reasonable sense of the scope of that problem.

To what extent would this be resolved by changing such commands to nominate an output directory, even if the names of the files that get written to that directory are not broadcast by the command interface? In many of these instances I think I'd prefer to change the command interfaces to be an output directory not because of Pydra, but just to be more faithful to the data hierarchy. Others occur because of the weird fixel command interface where output fixel directory and output fixel data file name are specified separately, which I don't like (see eg. #1043).

> Moved `SequenceDirectoryOut` from `python/mrtrix3/commands/population_template/usage.py` into `app.py` with the rest of the types to avoid circular import hell

While it's not the end of the world for that one change to be integrated into the API, it would nevertheless be preferable for it to still be possible for individual commands to define their own command-line types, and it would be beneficial for there to be an exemplar somewhere in the code demonstrating how that is done. So if there's a more suitable code structure that would permit keeping that as command-specific, I'd have a slight preference that way. But it may well be a barrier arising as a consequence of my horrendous Python.

tclose:
> * Can you clarify on the distinction between here and [WIP: Implement Pydra code-generators #2665](https://github.com/MRtrix3/mrtrix3/pull/2665)?
> * An Action to test the generation of the usage files would probably be preferred. We would want to avoid having to push and tag patch fixes to the main repo to fix downstream applications that are outside of our purview. The scope of what to test from what is generated is up to you, but to know that they are at least generated and produce code that doesn't crash the Python interpreter would be preferred.

Ok, I can add a github action that does that.

> 
> > ... eg. the outputs of `connectome2tck` are specified in the argument list by the prefix text field ...
> 
> Grepping for `.type_text()` should give a reasonable sense of the scope of that problem.
> 
> To what extent would this be resolved by changing such commands to nominate an output directory, even if the names of the files that get written to that directory are not broadcast by the command interface? In many of these instances I think I'd prefer to change the command interfaces to be an output directory not because of Pydra, but just to be more faithful to the data hierarchy. Others occur because of the weird fixel command interface where output fixel directory and output fixel data file name are specified separately, which I don't like (see eg. #1043).

In terms of Pydra, it would be preferable to grab the output files directly even if this is more work. This kind of output parsing is consistent with how other tools are handled, and by dumping it in a directory you just push the work on to the workflow maintainer instead of the tool-package maintainer. If there is too much runtime logic involved in this, this will just have to be maintained manually I expect.

> > Moved `SequenceDirectoryOut` from `python/mrtrix3/commands/population_template/usage.py` into `app.py` with the rest of the types to avoid circular import hell
> 
> While it's not the end of the world for that one change to be integrated into the API, it would nevertheless be preferable for it to still be possible for individual commands to define their own command-line types, and it would be beneficial for there to be an exemplar somewhere in the code demonstrating how that is done. So if there's a more suitable code structure that would permit keeping that as command-specific, I'd have a slight preference that way. But it may well be a barrier arising as a consequence of my horrendous Python.

I can't remember what the issue is but I think it might have been the way you insert things onto the path that made it difficult/impossible. I'm usually pretty persistent in solving such problems so I think it won't be possible unless some things change in the loading of the modules.

Lestropie:
> In terms of Pydra, it would be preferable to grab the output files directly even if this is more work.

For any command that exports to a directory or specifies a prefix, it will often be because the number of files generated is so large that manually specifying each and every one of those files in the CLI would be very cumbersome. There are also circumstances where the set of files that will be produced can't be known at compile time. So there's a natural tension between user-friendliness and automatability. I expect that balance currently strongly favours the former; our userbase is still mostly individuals running individual commands explicitly.

> I can't remember what the issue is but I think it might have been the way you insert things onto the path that made it difficult/impossible.

It wouldn't surprise me. Getting the inheritance from `pathlib` to work was a nightmare, probably a combination of their own inheritance being determined at runtime and my own Python capabilities being fairly rudimentary. It's not the highest priority so probably just move that one instance into the API, but it'll surely start nagging at me at some point.

tclose:
> > In terms of Pydra, it would be preferable to grab the output files directly even if this is more work.
> 
> For any command that exports to a directory or specifies a prefix, it will often be because the number of files generated is so large that manually specifying each and every one of those files in the CLI would be very cumbersome. There are also circumstances where the set of files that will be produced can't be known at compile time. So there's a natural tension between user-friendliness and automatability. I expect that balance currently strongly favours the former; our userbase is still mostly individuals running individual commands explicitly.

How large is large? I mean we could write a grab-file-from-directory generic task fairly easily so it isn't that bad of a problem. Would the number/names of files produced be know at parameterisation time, or only runtime?

PR #3101, opened 2025-05-21
Merge 3.0.5 into dev

Lestropie:
This is on the clunkier side because of `dev` now having *both* a mass reformatting of code *and* a re-arrangement on the file system. `git` now outright misses a lot of changes. So there will need to be extra diligence to make sure that changes applied to `master` prior to `3.0.5` have not been lost in the merging process. Assistance in that for PRs that are not my own would be appreciated.

Certainly the DWI metadata handling changes were initially problematic. On my first merge attempt the changes in #3011 disappeared. So what I have done is started with #2917, first brought that up to date with current `dev`, and then merged `3.0.5` onto that, knowing that much of the changes applied to `master` in #3011 were already manually applied based on the contents of #2917. This PR therefore supersedes #2917. I will however need to manually check the contents of #3011 and make sure nothing gets left out.

PR #3153, opened 2025-08-26
run.command(): Fix operation of -continue

Lestropie:
The existing implementation of -continue assumed that, for a list of elements to be provided to `subprocess.run()`, each element was a string, and therefore string operations could be applied. However with the adoption of `pathlib` for handling user-specified paths in #2678, an element in this list may be a class derived from those provided in the `pathlib` module that does not provide string functions.


Lestropie:
Failure here is due to #3149 not yet having been merged from `master` back to `dev`. I need to finish #3101 before that can happen. Merging despite CI failure as want to try to reduce open PR count.

PR #3155, opened 2025-08-27
fod2fixel: New metric "skew"

Lestropie:
Minor feature addition that had been committed to a branch but for which a PR had not been previously created. No idea whether it is of any utility, it's merely a "can".

- [ ] Add corresponding test data


Lestropie:
- [ ] On merge, delete `fod2fixel_skew` branch from test data repo and update tip of `dev` branch
    (that branch uses `fixelfilter_cfe` from #3163 as its base, so if this PR is merged after #3163 there should hopefully be no issue)

PR #3162, opened 2025-08-29
dwifslpreproc: Save bad slspec file

Lestropie:
If slice padding has been performed for topup compatibility, the removal of said padded slice must then be reflected in the slspec file for eddyqc compatibility (#2286). If this fails, then the slspec file should nevertheless be written to the scratch directory for diagnostic purposes.


jdtournier:
I suspect this PR has been superseded by #2505?

Lestropie:
Best guess is that it's a branch that was living on my local clone that had been superseded but not deleted, and so when I did a mass force push it re-appeared. Closing.

Issue #2630, opened 2023-05-10
Inclination vs elevation


Lestropie:
I've been aware of this one for a while, but didn't realise that I had not listed it. Has come back to the front of my mind due to playing with particularly #2626.

[Spherical coordinates](https://en.wikipedia.org/wiki/Spherical_coordinate_system) can be defined using different conventions, in terms or presence or absence of elements / ordering of elements / symbols for elements / reference & range for quantities. In BIDS [BEP016](https://github.com/bids-standard/bids-bep016) I've [stuck to ISO standard](https://github.com/bids-standard/bids-bep016/blob/8dee2f0f75f8c2dab1c6916d9f2f277e7101ca43/src/derivatives/05-diffusion-derivatives.md?plain=1#L147-L185) as much as possible. In *MRtrix3* code, one frequently sees references in this context to "az_el", ie. azimuth-elevation pairs. However the [`cos`](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/math/sphere.h#L45) / [`acos`](https://github.com/MRtrix3/mrtrix3/blob/3.0.4/core/math/sphere.h#L87) transformations indicate that the origin of the second spherical angle is the z-axis. This would make it an *inclination* rather than elevation angle: the former uses the angle from the zenith direction, the latter uses the angle from the reference plane.

- [ ] Do code audit to make sure that there are no fundamental conflicts in how such values are transformed / interpreted in various places.
    One potential culprit I'm thinking of here is the calculation of lighting effects in OpenGL code.
- [ ] Find any manual manipulation of spherical vs cartesian directions and centralise it all to `MR::Math::Sphere`.
    I've already done this a little in #2626 and related, but it needs a more exhaustive search.
- [ ] Rename variables to properly reflect the actual encoding.
- [ ] Add somewhere in documentation the way in which such data are imported / exported.
    This could potentially include a copy-paste help page Description paragraph, or a version-matched documentation hyperlink.

Lestropie:
True; might we worth expanding the scope of the issue to cover the order of coefficients. The expected ranges should ideally not be of consequence since they should wash out in any relevant calculations; what matters most is the definition of where zero is. For inclination vs elevation as polar angle in particular, there should be a bunch of `cos()` / `acos()` vs. `sin()` / `asin()` calls in conversions to / from Cartesian.

The inhomogeneity certainly justifies my assertion of enforcing ISO in BEP016... 😄

fionaEyoung:
[...]

PR #3177, opened 2025-09-17
mrview: Change interpretation of light position

Lestropie:
Discovered when writing 3ddf1744dd53fdfbda595447ce363c190ddb7079 as part of #2918.

Blame initially reports 4835759438678adbc4237a83a4730c84f6dd4934 addressing #246 & part of tag `0.3.13`, but most recent change is bfa3bd2c1ed061f6b6c4ca06981e499078567b6c as part of `0.3.8`.

Code looks outright incorrect, but don't know if there was some other intent behind it.

It's also entirely possible that downstream GUI lighting calculations may have been bastardised in order to yield the desired behaviour downstream of this code. 

Posting as draft as I'm going to need to dig into the consequences & potential interactions.

### DIFF
--- a/src/gui/lighting_dock.cpp
+++ b/src/gui/lighting_dock.cpp
@@ -72,14 +72,14 @@ namespace MR
 
       elevation_slider = new QSlider (Qt::Horizontal);
       elevation_slider->setRange (0,1000);
-      elevation_slider->setSliderPosition (int ( (1000.0/Math::pi) *acos (-info.lightpos[1]/Eigen::Map<Eigen::Matrix<float, 3, 1>>(info.lightpos).norm())));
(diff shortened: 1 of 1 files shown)


Lestropie:
Registered late last night that this is presumably defining the light position in screen space rather than real space; 0 and 1 are horizontal and vertical, 2 is depth. But the naming around this class could possibly be improved to communicate this; potentially the UI also (these are in a way kind of "relative" rather than "absolute" angles). And need to check all code accessing that class to see whether there's a discrepancy in interpretation between azimuth and inclination (I checked these elsewhere in 3ddf1744dd53fdfbda595447ce363c190ddb7079 but the GUI code requires closer interrogation).

Lestropie:
Closing as change is believed to be an incorrect inference of error when making other code changes. See report in #2630 regarding possible other issues around interpretation of light position.

PR #3182, opened 2025-09-24
Remove MR::BitSet class

Lestropie:
Spotted in #3116:

> The exception is the tests for our dynamic `MR::BitSet` class, where the new set of tests should be slightly more comprehensive than the previous ones.

I've discussed on multiple occasions a desire to remove this class, though may have neglected to list it as an Issue, and it keeps getting deferred. Hoping that there wasn't too much effort invested here. I think it's beyond the scope of what should be necessary to be maintaining in this package, and there should now be a substitute, so put the time into getting rid of it.

-----

The motivations for this class were:

-   A long time ago, `std::vector<bool>` may not have had bitwise implementations
-   `std::bitset` must have constant size at compile time
-   In some contexts, functions for quickly computing whether all bits are true / all bits are false / the number of true bits / mass bitwise operations are useful

Its creation predated the transfer over to Eigen. `Eigen::Array<bool>` has `.any()`, `.all()`, `.count()`, and `&&` and `||` operations. It should therefore serve as a substitute in contexts where those are useful. It sounds like Eigen may in fact use a byte per element for performance; the additional storage space shouldn't be an issue in any of the contexts in which it is used: the largest is probably the fixel mask in SIFT2, where changing from 1 bit per fixel to 1 byte per fixel might chew up 0.5MB or so.

Where none of these operations are utilised, I have instead replaced `MR::BitSet` with `std::vector<bool>`.

-----

Somewhat related to this is that there's a couple of other contexts that involve manipulation of individual bits. What comes to mind is `MR::DataType` and `App::ArgFlags`. These would perhaps be better encoded using `std::bitset`, given requisite size is known at compile time. I have deemed this out of scope for this particular PR, but might be worth listing elsewhere.


Lestropie:
Would any attempt to use an STL algorithm with a `std::vector<bool>` result in a compilation failure?

The scope for any "algorithm" to be applied to bitwise data is I think pretty limited; really the only ones that might be of any use are `.any()`, `.all()` & `.count()`---for which `Eigen::Array<bool>` would be better---and mass bitwise manipulations (here we lose `MR::BitSet`'s XOR but nothing was using it). I did wonder about using `check_syntax` to preclude the use of `std::vector<bool>` and force `Eigen::Array<bool>` or `std::bitset`, but 1) a lot of code doesn't need any such functionality and 2) maybe in some circumstances minimising storage through true bitwise data is preferable.

`std::deque<bool>` uses a byte per bit element, so unless you explicitly need the ability to eg. `.pop_front()` there's unlikely to be any benefit over `Eigen::Array>bool>` I think?

Lestropie:
In removing `std::vector<bool>`, the only STL algorithm usage I found was `std::fill()` [here](https://github.com/MRtrix3/mrtrix3/blob/3.0.7/core/math/constrained_least_squares.h#L257-L259); that's reported to be functional, and there's testing for that functionality.

PR #2393, opened 2021-11-08
Implementation of 5ttgen algorithms for neonatal data.

mblesac:
Co-authored-by: Manuel Blesa Cabez <mblesac@gmail.com>
Co-authored-by: Paola Galdi <paola.galdi@gmail.com>

PR #3193, opened 2025-10-01
5ttgen neonatal algorithms

Lestropie:
Supersedes #2393 (didn't want to suddenly start pushing to an external `master` branch).

This really needs test data. But it looks like access is pretty locked down. 
@jdtournier @mblesac @paola-g: Is anyone able to provide an example dataset? Hopefully if I downsample a dataset prior to uploading to `script_test_data`, it won't offend anybody...
I also note that there's reference to BIDS layout in more recent data releases, to which the current code bears no resemblance. I'm guessing that different data releases have used different storage conventions?


mblesac:
Hi @Lestropie , several things have changed and I think some things in the algorithm should be edited to make it applicable to most datasets. Let me get back to you after ISMRM deadline, then I'll have time to work on this. One quick question, what are your thought about including docker/apptainer in the scripts?

PR #3195, opened 2025-10-01
Fix ROI edits disappearing on MacOS

### LINKED ISSUE #2319: mrview: ROI edits deleted after two elements drawn
When drawing an ROI, I can draw a maximum of two elements before previously drawn elements start disappearing in a FIFO manner. So if I draw one rectangle, then a second, as soon as I click to draw a third, the first one vanishes. 
This behaviour is limited to the current slice. So for example, if I fill an axial slice, then switch to a saggital slice (which now has a single filled line through it) and begin editing, after 1 edit the filled line will disappear. Switching back to the filled axial slice, just that line is erased, not the whole slice.

This behaviour persists regardless of

- drawing tool (brush / rectangle / fill)
- loading an existing ROI file vs creating a new one within mrview
- drawing or erasing
- single slice or orthoview
- background image

Here is an animation demonstrating the behaviour:
buggyroi

I have never had this problem with any other installations of MRtrix on other machines, only on my new M1 silicon Macbook Pro. I've also checked it isn't just a display issue: if I save the file, quit mrview and load the ROI again, the erased bits are still erased.

**To Reproduce**

1. Open any image in mrview and create a new ROI in the ROI editor
2. Choose a tool (e.g. brush) and click edit
3. Draw a shape
4. Draw another shape
5. Draw a third shape: the first one disappears
6. Draw another shape: the second one disappears
7. You get the picture 

**Platform/Environment/Version**

-  OS: MacOS 11.1
-  *MRtrix3* version `3.0.2-78-g430465d6 64 bit release version, built Mar  9 2021, using Eigen 3.3.9` (built from source)

---

**Advanced debugging information**

-  Qt version: 5.15.2
- OpenGL info:
Screenshot 2021-05-13 at 13 15 27

- `mrview -exit -debug`:

  
```
mrview -exit -debug
mrview: [DEBUG] No config file found at "/etc/mrtrix.conf"
mrview: [INFO] reading config file "/Users/fiona/.mrtrix.conf"...
mrview: [DEBUG] reading key/value file "/Users/fiona/.mrtrix.conf"...
mrview: [INFO] GL renderer:  Apple M1
mrview: [INFO] GL version:   4.1 Metal - 71.0.7
mrview: [INFO] GL vendor:    Apple
mrview: [DEBUG] loading font into OpenGL texture...
mrview: [DEBUG] compiling OpenGL vertex shader:
#version 330 core
layout(location = 0) in vec2 pos;
layout(location = 1) in vec2 font_pos;
uniform float scale_x;
uniform float scale_y;
out vec2 tex_coord;
void main () {
  gl_Position = vec4 (pos[0]*scale_x-1.0, pos[1]*scale_y-1.0, 0.0, 1.0);
  tex_coord = font_pos;
}

mrview: [DEBUG] compiling OpenGL fragment shader:
#version 330 core
in vec2 tex_coord;
uniform sampler2D sampler;
uniform float red, green, blue;
out vec4 color;
void main () {
  color.ra = texture (sampler, tex_coord).rg;
  color.rgb = color.r * vec3 (red, green, blue);
}

mrview: [DEBUG] font loaded
```

daljit46:
This small change fixes #2319.
After the discussion this morning with @bjeurissen, I decided to investigate this long-standing issue in the viewer. So this wasn't really a driver issue as discussed in the relevant issue, but an error in our code.
After tinkering with OpenGL code a bit, I noticed that the undo/redo shader in [`undoentry.cpp`](https://github.com/MRtrix3/mrtrix3/blob/56fa0604486cb1dbca1c36d8526a5eddc5bc405d/cpp/gui/mrview/tool/roi_editor/undoentry.cpp#L33) was using an integer sampler to read the values from the texture.
Our texture data type is `GL_R8`, which is an 8-bit normalised fixed-point format. It needs a float sampler, because integer samplers are for *integer* formats like `GL_R8I`. 

@MRtrix3/mrtrix3-devs should we include this fix in 3.0.8?

### DIFF
--- a/src/gui/mrview/mode/slice.cpp
+++ b/src/gui/mrview/mode/slice.cpp
@@ -15,8 +15,8 @@
  */
 
 #include "gui/mrview/mode/slice.h"
-
 #include "gui/opengl/transformation.h"
+#include "gui/mrview/tool/roi_editor/item.h"
 
 namespace MR
 {
@@ -45,19 +45,21 @@ namespace MR
 
         std::string Slice::Shader::fragment_shader_source (const Displayable& object)
         {
-          std::string source = object.declare_shader_variables () +
-            "uniform sampler3D tex;\n"
-            "in vec3 texcoord;\n"
-            "out vec4 color;\n";
+          const auto *roi_item = dynamic_cast<const Tool::ROI_Item *>(&object);
+          const bool is_roi = roi_item != nullptr;
+          std::string source = object.declare_shader_variables() +
+                               (is_roi ? "uniform usampler3D tex;\n" : "uniform sampler3D tex;\n") +
+                               "in vec3 texcoord;\n"
+                               "out vec4 color;\n";
 
-          source +=
-            "void main() {\n"
-            "  if (texcoord.s < 0.0 || texcoord.s > 1.0 ||\n"
-            "      texcoord.t < 0.0 || texcoord.t > 1.0 ||\n"
-            "      texcoord.p < 0.0 || texcoord.p > 1.0) discard;\n"
-            "  color = texture (tex, texcoord.stp);\n"
-            "  float amplitude = " + std::string (ColourMap::maps[object.colourmap].amplitude) + ";\n"
-            "  if (isnan(amplitude) || isinf(amplitude)) discard;\n";
+          source += "void main() {\n"
+                    "  if (texcoord.s < 0.0 || texcoord.s > 1.0 ||\n"
+                    "      texcoord.t < 0.0 || texcoord.t > 1.0 ||\n"
+                    "      texcoord.p < 0.0 || texcoord.p > 1.0) discard;\n" +
+                    (is_roi ? std::string("  uint ru = texture(tex, texcoord.stp).r;\n  color = vec4(ru > 0u ? 1.0 : 0.0);\n")
+                            : std::string("  color = texture (tex, texcoord.stp);\n")) +
+                    "  float amplitude = " + std::string(ColourMap::maps[object.colourmap].amplitude) + ";\n"
+                    "  if (isnan(amplitude) || isinf(amplitude)) discard;\n";
 
           if (object.use_discard_lower())
             source += "  if (amplitude < lower) discard;\n";
--- a/src/gui/mrview/tool/roi_editor/item.cpp
+++ b/src/gui/mrview/tool/roi_editor/item.cpp
@@ -54,8 +54,8 @@ namespace MR
             current_undo (-1)
         {
           type = gl::UNSIGNED_BYTE;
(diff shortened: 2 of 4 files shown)


Lestropie:
If it's a bug fix, then it should be targeting `master`, whether it gets merged in time for `3.0.8` or not.

I worry here that the fix is being applied at the wrong end & in the wrong direction. The ROI tool deals exclusively with bitwise data and no interpolation. It would make a lot more sense to be using 8-bit integer data. It's more likely that it's the upstream texture data type that's the bug, not the downstream sampler.

Lestropie:
Given that the various tools operate sequentially and independently with their own shaders, I would have thought it not too difficult, but I'm admittedly highly out of practise with the GUI code. Maybe have a dig at it, but if you encounter anything that looks like it could gobble up a lot of time revert back to the one-line fix?

src/gui/mrview/mode/slice.cpp
```diff
@@ -45,19 +45,21 @@ namespace MR
...
+                               "out vec4 color;\n";
+
+          source += "void main() {\n"
+                    "  if (texcoord.s < 0.0 || texcoord.s > 1.0 ||\n"
+                    "      texcoord.t < 0.0 || texcoord.t > 1.0 ||\n"
+                    "      texcoord.p < 0.0 || texcoord.p > 1.0) discard;\n" +
+                    (is_roi ? std::string("  uint ru = texture(tex, texcoord.stp).r;\n  color = vec4(ru > 0u ? 1.0 : 0.0);\n")
+                            : std::string("  color = texture (tex, texcoord.stp);\n")) +
+                    "  float amplitude = " + std::string(ColourMap::maps[object.colourmap].amplitude) + ";\n"
+                    "  if (isnan(amplitude) || isinf(amplitude)) discard;\n";
```
Lestropie, line 62: This seems to be doing a bunch of work that it shouldn't need to for ROIs. Indeed it's likely some of the discordance around the requisite changes to fix the reported issue were due to my not recalling / recognising that the ROI rendering was using so much common slice rendering code.

For ROIs, all that should be needed in a fragment shader is:
- If outside bounds, discard
- If sample is zero, discard
- Set colour to fixed colour

Things like applying a continuous colour map, checking for non-finite values, upper / lower thresholds, shouldn't be needed.

This was probably a case of re-using the slice shader as-is for the ROI tool because it worked and required less effort at the time. But it would have been more appropriate to have given the ROI tool its own shader. Redressing would require a larger refactor, and it wouldn't strictly need to happen on `master`. So maybe I'll list that as a separate Issue.

daljit46: Yes, I think a separate issue to clean up this bit would be best.

PR #3197, opened 2025-10-03
Fix tractography rendering performance on macOS

daljit46:
This PR fixes the longstanding performance issues in the viewer on macOS when visualising tractography streamlines (using pseudotubes). 
The root cause turned out not to be geometry-shader emulation per se, but the cost of submitting many draws via `glMultiDrawArrays<. On Apple’s stack, glMultiDraw* does not seem to guarantee hardware batching, and it can result in significant CPU/driver overhead. Details below.

After working on #3095, I decided to investigate the performance issues raised in #2247. As discussed there, the culprit seemed to be the use of geometry shaders in our OpenGL code. This was a reasonable suspicion that their emulation performance could be the culprit (since the OpenGL drivers on macOS are built upon Metal, which doesn't support geometry shaders).  I spent a significant amount of time tinkering with the code, but it was unreasonable to assume that emulation of geometry shaders (via compute shaders) could be so slow that the app would either completely freeze or even crash.

Investigating with the Xcode profiler, I noticed that the application essentially just hung when calling `glMultiDrawArrays`:

```
1.54 s 40.0% 400.00 µs GLDContextRec::flushContext(bool) 
1.47 s 38.3% 600.00 µs -[_MTLCommandQueue submitCommandBuffer:] 
1.47 s 38.3% 400.00 µs _dispatch_lane_barrier_sync_invoke_and_complete 
1.47 s 38.3% 300.00 µs _dispatch_client_callout 
1.47 s 38.3% 400.00 µs __40-[_MTLCommandQueue submitCommandBuffer:]_block_invoke 
1.47 s 38.3% 400.00 µs -[_MTLCommandQueue _submitAvailableCommandBuffers] 
1.46 s 38.0% 264.54 µs -[IOGPUMetalCommandQueue submitCommandBuffers:count:] 
1.46 s 38.0% 700.00 µs -[IOGPUMetalCommandQueue _submitCommandBuffers:count:] 
1.40 s 36.5% 700.00 µs IOGPUCommandQueueSubmitCommandBuffers 
1.40 s 36.5% 0 s IOConnectCallMethod 
1.40 s 36.5% 3.20 ms io_connect_method 
1.40 s 36.4% 300.00 µs mach_msg2_internal 
1.40 s 36.3% 
1.40 s mach_msg2_trap
```

which made me realise that most likely we are CPU-bound on command submission. The OpenGL on Metal implementation
seems to be submitting too many draw calls, which results in a hang.
Looking around, I found this great [article](https://programming4.us/multimedia/8302.aspx) which mentions:

> Even functions like glMultiDrawArrays and glMultiDrawElements don’t always help because the graphics hardware may not implement these functions directly, and so OpenGL essentially has to convert them to multiple calls to glDrawArrays internally anyway.

`glMultiDrawsArrays` is the command that we use in [tracogram.cpp](https://github.com/MRtrix3/mrtrix3/blob/3bcf5c10f457afea505180eeb27d577d126cbd81/src/gui/mrview/tool/tractography/tractogram.cpp#L569) to instruct the GPU to draw a chunk of streamlines. In theory, the command should be hardware-batched, but as the article mentioned, this is upon the GPU driver's implementation, and it's not guaranteed (as it seems to be the case on OpenGL on Metal).

The article also offers a better alternative to ensure the commands are batched. The idea is to supply an index buffer that is one long sequence of vertex indices, with a special sentinel value between line strips (a "primitive-restart index") that tells the GPU where each track starts or terminates. Once the index buffer is built, we can submit everything in one go via `glDrawElements(GL_LINE_STRIP, num_elements, GL_UNSIGNED_INT, nullptr)`.

Suppose we have 3 tracks with 4 elements each: [0 1 2 3] [4 5 6 7] [8 9 10 11]. With `glMultiDrawArrays` we do:

```
starting_indices = {0, 4, 8};
element_counts = {4, 4, 4};
glMultiDrawArrays(GL_LINE_STRIP, firsts, counts, 3);
```

while with the indexed approach:

```
// Use 0xFFFFFFFF as sentinel values
indices = [0,1,2,3, 0xFFFFFFFF, 4,5,6,7, 0xFFFFFFFF, 8,9,10,11, 0xFFFFFFFF]
glEnable(GL_PRIMITIVE_RESTART);
glPrimitiveRestartIndex(0xFFFFFFFFu);
glDrawElements(GL_LINE_STRIP, indices.size(), GL_UNSIGNED_INT, nullptr);
```

I implemented this idea in the code, and it fixed the performance issues while visualising tractograms on macOS!  On my MacBook M2, I now see framerates between 100 - 150 (dropping to 60-90 with volume rendering) when displaying a track file with 200k streamlines. Previously, `mrview` would even struggle to load the file.
As a bonus on Linux, with my modest laptop with Intel 11th i5-1135G7, I'm also seeing up to 25% improvements in framerates. 

Notes:
- There is a slight disadvantage in using the new approach, as the index buffer consumes some additional memory.
- I have only enabled the index buffer code path for pseudotubes. I preferred to keep the code changes to a minimum, as this PR is targeting `master` and its primary aim is to fix the performance issues on macOS. But we can enable this for points too if we want that (either in this PR or a separate one, perhaps targeting `dev`).

@MRtrix3/mrtrix3-devs  I believe this change is correct and safe, but I'd appreciate it if you could test this on your own machines to see if you encounter any issues and maybe compare with `mrview` built from `master` to see if you notice any performance improvements. Also, this PR is *strictly* not a correctness fix, so I'd be happy if we want to target `dev` instead (I think other easy opportunities may potentially improve performance further).

### DIFF
--- a/src/gui/mrview/tool/tractography/tractogram.cpp
+++ b/src/gui/mrview/tool/tractography/tractogram.cpp
@@ -24,9 +24,10 @@
 #include "gui/opengl/lighting.h"
 #include "gui/mrview/mode/base.h"
 
-
+#include <cstdint>
 
 const size_t MAX_BUFFER_SIZE = 2796200;  // number of points to fill 32MB
+constexpr uint32_t PRIMITIVE_RESTART_SENTINEL = 0xFFFFFFFFu; // Primitive restart index for UNSIGNED_INT
 
 namespace MR
 {
@@ -399,6 +400,8 @@ namespace MR
             gl::DeleteBuffers (intensity_scalar_buffers.size(), &intensity_scalar_buffers[0]);
           if (threshold_scalar_buffers.size())
             gl::DeleteBuffers (threshold_scalar_buffers.size(), &threshold_scalar_buffers[0]);
+          if (!element_buffers.empty())
+            gl::DeleteBuffers(element_buffers.size(), &element_buffers[0]);
           GL::assert_context_is_current();
         }
 
@@ -566,8 +569,15 @@ namespace MR
 
             auto mode = geometry_type == TrackGeometryType::Points ? gl::POINTS : gl::LINE_STRIP;
 
-            gl::MultiDrawArrays (mode, &track_starts[buf][0], &track_sizes[buf][0], num_tracks_per_buffer[buf]);
-
+            if(mode == gl::POINTS) {
+              gl::MultiDrawArrays(mode, &track_starts[buf][0], &track_sizes[buf][0], num_tracks_per_buffer[buf]);
+            } else if (element_counts[buf] > 0 && element_buffers[buf] != 0) {
+              // Use the EBO stored with this VAO to render all tracks in this chunk
+              gl::Enable(gl::PRIMITIVE_RESTART);
+              gl::PrimitiveRestartIndex(PRIMITIVE_RESTART_SENTINEL);
+              gl::DrawElements(gl::LINE_STRIP, element_counts[buf], gl::UNSIGNED_INT, nullptr);
+              gl::Disable(gl::PRIMITIVE_RESTART);
+            }
           }
 
           vao_dirty = false;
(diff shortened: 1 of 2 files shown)


bjeurissen (dismissed):
Works and fixes the issue! Tested on macOS 26.0.1 (25A362) on Apple M1

bjeurissen:
Would be good to check if the same fix applies to #2379. At least the symptoms look identical, so I would think there is a good chance.

bjeurissen (approved):

jdtournier:
:+1: Works fine on Ubuntu 22.04, running on a 12GB NVIDIA GeForce RTX 4070. Get roughly 50 FPS for 1 million streamlines with streamtube lighting in volume render mode.

I'll try it on my home system when I get home, just to check with an AMD GPU.

jdtournier:
Also works fine on my AMD-based home system 👍

PR #3202, opened 2025-10-04
Update configure script for Xcode26 compatibility

bjeurissen:
Qt5 links to both OpenGL and AGL for OpenGL apps, but AGL is no longer included in MacOS Tahoe/Xcode26 (https://bugreports.qt.io/browse/QTBUG-137687). This will cause our configure script to fail on latest macOS. This issue will not be addressed in future Qt5 versions (only in Qt6), but we can work around it by explicitly linking only to OpenGL.


bjeurissen:
Confirmed to work both pre-deprecation of AGL (macOS Sequoia) as post-deprecation of AGL (macOS Tahoe), so I think this change is solid.

PR #3207, opened 2025-10-06
Fix Eigen 3.4.1 type mismatch in `ColourMap` clamping

daljit46:
Fix compilation error leading to `Eigen::internal::traits<double>` errors.

This is currently blocking PRs targeting `master` due to CI build failures on macOS.


daljit46:
I don't mind lowercase. I used uppercase because on `dev` we have [readability-uppercase-literal-suffix](https://clang.llvm.org/extra/clang-tidy/checks/readability/uppercase-literal-suffix.html) enabled.

PR #3208, opened 2025-10-07
DICOM: Performance fix to wildcard matching

Lestropie:
Error in 8e3612aebac43ccd56c014d0c2ee87035f55e6c6 as part of #1929.

I expect this to be of no consequence to behaviour, so not even applying the `bug` tag; it's just preventing unnecessary recursive construction of `std::string`'s.

(This will be better addressed using `std::string_view` on dev; that's how I found it)


jdtournier (approved):
Good spot!

PR #3211, opened 2025-10-08
Registration: Fix local cross-correlation metric

Lestropie:
No idea if this is even accessible from the command-line, but worth fixing nevertheless in case anyone out there is trying to do anything with it.

### DIFF
--- a/src/registration/metric/local_cross_correlation.h
+++ b/src/registration/metric/local_cross_correlation.h
@@ -237,7 +237,9 @@ namespace MR
                   return 0.0;
                 }
 
-                const Eigen::Vector3d pos = Eigen::Vector3d(default_type(iter.index(0)), default_type(iter.index(0)), default_type(iter.index(0)));
+                const Eigen::Vector3d pos = Eigen::Vector3d(default_type(iter.index(0)),
+                                                            default_type(iter.index(1)),
+                                                            default_type(iter.index(2)));
                 params.processed_image_interp->voxel(pos);
                 typename Params::Im1ValueType val1;
                 typename Params::Im2ValueType val2;


Lestropie:
I discovered a second potential minor issue with the LCC code.

[Here](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/src/registration/metric/local_cross_correlation.h#L166):
```cpp
vector<uint32_t> NoOversample;
```

This empty class instance is named with the intention of resulting in no oversampling during reslicing. It is [passed to the `Adapter::Reslice<>` constructor](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/src/registration/metric/local_cross_correlation.h#L175-L177):
```cpp
else if (parameters.im1_mask.valid() and parameters.im2_mask.valid()){
  Adapter::Reslice<Interp::Nearest, Im1MaskType> mask_reslicer1 (parameters.im1_mask, cc_mask_header, parameters.transformation.get_transform_half(), NoOversample);
  Adapter::Reslice<Interp::Nearest, Im2MaskType> mask_reslicer2 (parameters.im2_mask, cc_mask_header, parameters.transformation.get_transform_half_inverse(), NoOversample);
```

That class however [defines](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/core/adapter/reslice.h#L76):
```cpp
extern const vector<uint32_t> AutoOverSample;
```

as an equivalent empty class instance. So I believe in the LCC code the masks are being automatically oversampled despite the variable naming indicating an intent to the contrary.

The [subsequent code](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/src/registration/metric/local_cross_correlation.h#L181-L183) however looks like it's entirely compatible with oversampling of mask images resulting in continuous values:
```cpp
auto both = [](decltype(cc_mask)& cc_mask, decltype(mask_reslicer1)& m1, decltype(mask_reslicer2)& m2) {
  cc_mask.value() = ((m1.value() + m2.value()) / 2.0) > 0.5 ? true : false;
};
```

So I think just remove that `NoOversample` variable and leave the behaviour as the default `AutoOverSample`?

PR #3212, opened 2025-10-09
Enhancements to check_syntax & code consistency

### LINKED ISSUE #3192: Standardise NaN / infinity syntax
I see three different usages:

-   Explicitly using `std::numericl_limits<type>::quiet_NaN()` 
-   "`NaN`" appears to link to `MR::NaN` in `core/types.h`, and is explicitly of `default_type`, ie. double-precision
-   "`NAN`" for me claims to link to `<math.h>`:
    > ```
    > # if __GNUC_PREREQ (3, 3)
    > #  define NAN (__builtin_nanf (""))
    > ```

    , which seems to be single-precision

Would be better to be consistent.

-----

- [ ] Choose which syntax we want to standardise
    -   Always use `std::numeric_limits`?
    -   Permit `NaN` but preclude `NAN`?

- [ ] Augment `check_syntax` to check for non-conformant usage

- [ ] Modify code base to conform

Lestropie:
Extends #3206.
Diff to `dev` will be reduced if that is merged first.
Main changes in 8704811a3f8dad7299acf4105c971a15a4a2eb2c.

After #3189, realised there were many changes that have happened on `dev` that weren't being enforced, and so with so much manual merge conflict resolution going on, I wanted to get some guard rails up in order to be able to proceed with confidence.

Extends #2911, #2871, #2652, #1304.

Closes #3192.

- [x] Need to manually test GUI, including `mrview` sync functionality

-----

```text
/Everybody stand back/

I know regular expressions
```

### DIFF
--- a/cpp/core/app.cpp
+++ b/cpp/core/app.cpp
@@ -1307,7 +1307,7 @@ int64_t App::ParsedArgument::as_int() const {
         }
         if (contains_dotpoint) {
           const default_type prefix = to<default_type>(num);
-          retval = std::round(prefix * default_type(multiplier));
+          retval = std::round(prefix * static_cast<default_type>(multiplier));
         } else {
           retval = to<int64_t>(num) * multiplier;
         }
@@ -1356,7 +1356,7 @@ uint64_t App::ParsedArgument::as_uint() const {
     throw Exception("Attempting to interpret negative user-specified value (" //
                     + str(signed_value)                                       //
                     + " as unsigned integer");                                //
-  return uint64_t(signed_value);
+  return static_cast<uint64_t>(signed_value);
 }
 
 default_type App::ParsedArgument::as_float() const {


cpp/core/exception.h
```diff
@@ -83,18 +85,18 @@ class Exception : public std::exception {
...
 
-  const char *what() const noexcept override;
+  const char *what() const noexcept override; // check_syntax off
 
   void display(int log_level = 0) const { display_func(*this, log_level); }
 
   size_t num() const { return description.size(); }
-  const std::string &operator[](size_t n) const { return description[n]; }
-  void push_back(const std::string &s) { description.push_back(s); }
+  std::string_view operator[](size_t n) const { return description[n]; }
```
daljit46: I'm not sure if it makes sense to return a `std::string_view` here. `std::string_view` is a non-owning pointer+size; returning `std::string_view` gives the caller a view into the `std::string`'s internal buffer. If the container is mutated, the view becomes dangling. This is the same class of bug as returning a const char* to internal storage, but it's easier to forget with `string_view`.

Lestropie: Closed by 0cd8602def790a08eb70c9f717d4bade78b2bc0b.

cpp/core/exception.cpp
```diff
@@ -33,31 +36,16 @@ void display_exception_cmdline(const Exception &E, int log_level) {
...
+                                                                          {3, "%s: \033[00;34m%s%s\033[0m\n"}};
 
-void cmdline_report_to_user_func(const std::string &msg, int type) {
-  static constexpr const char *colour_format_strings[] = {"%s: %s%s\n",
-                                                          "%s: \033[01;31m%s%s\033[0m\n",
-                                                          "%s: \033[00;31m%s%s\033[0m\n",
-                                                          "%s: \033[00;32m%s%s\033[0m\n",
-                                                          "%s: \033[00;34m%s%s\033[0m\n"};
+  static const std::unordered_map<int, std::string> console_prefixes{
+      {-1, ""}, {0, "[ERROR] "}, {1, "[WARNING] "}, {2, "[INFO] "}, {3, "[ERROR] "}};
```
daljit46: 3 should be "[DEBUG]"?

Lestropie:
> Fix out-of-bounds access in parse_axes

:+1: Thought I'd already `grep -R`'d for `'\0'`'s.
The other one that might be of concern here is `parse_floats()` & `parse_ints()`. There's tests for `parse_ints()` but they're operating using a `std::string`. Will see if I can rig it up to test on a char array that isn't null-terminated.

cpp/core/exception.cpp
Lestropie: Yep. Really want to `enum class` these...

Lestropie: Closed by 3cb0eb0be0b3cb7eb938bf1f1911227e4780091b.

Issue #2894, opened 2024-05-08
shared_ptr.unique() prevents C++20 / Native windows build


LeeReid1:
Hi guys,

I need to preface this by highlighting C++ is not my mainstay and I've not your depth of knowledge on this codebase, so am a little out of my depth here.

I've locally branched the dev arm to put together a native windows build (i.e. MSVC). MSVC cannot resolve some of the more complex SFINAE code in 2-3 files, and bugs-out on a supporting library, but this is resolvable by switching to C++20 concepts. Suffice to say it's an absolute blackhole time-wise to find a work-around, if staying in C++17. 
My WIP currently should still compile as normal on linux (C++17) and is almost there on Windows with C++20 for most executables. 

However, the compilation fails due to `shared_ptr.unique()` in `image.h`. This method is deprecated in C++17 and eliminated in C++20 as [it's not threadsafe](https://stackoverflow.com/questions/41142315/why-is-stdshared-ptrunique-deprecated) and checks are likely avoidable by being careful with code architecture. 

I can't help but wonder about the code using it, but my knowledge of how mrtrix is put together is not deep enough to fix this without potentially causing an issue elsewhere. Hence, I'm seeking your opinions on solutions.

The primary use is in the destructor of `image.h`, where it seems to be used to check if `this` is the only user of the buffer. What follows seems to be unbuffering back to some original location(?): 

```
template <typename ValueType> Image<ValueType>::~Image() {
  if (buffer.unique()) { 
    // was image preloaded and read/write? If so,need to write back:
    if (buffer->get_io()) {
      if (buffer->get_io()->is_image_readwrite() && buffer->data_buffer) {
        auto data_buffer = std::move(buffer->data_buffer);
        TmpImage<ValueType> src = {
            *buffer, data_buffer.get(), std::vector<ssize_t>(ndim(), 0), strides, Stride::offset(*this)};
        Image<ValueType> dest(buffer);
        threaded_copy_with_progress_message("writing back direct IO buffer for \"" + name() + "\"", src, dest);
      }
    }
  }
}
```

I find it odd that an image may not be the owner or sole consumer of its own buffer in any read/write scenario. Other code suggests that, actually, this situation should never occur:

```
template <typename ValueType> Image<ValueType> Image<ValueType>::with_direct_io(Stride::List with_strides) {
  if (buffer->data_buffer)
    throw Exception("FIXME: don't invoke 'with_direct_io()' on images already using direct IO!");
[...]

PR #3220, opened 2025-10-14
Change unique check to use_count for buffer

bjeurissen:
Replaces deprecated unique() (https://en.cppreference.com/w/cpp/memory/shared_ptr/unique) with use_count() == 1 (https://en.cppreference.com/w/cpp/memory/shared_ptr/use_count.html). Avoids deprecation warnings on modern compilers.

### DIFF
--- a/cpp/core/image.h
+++ b/cpp/core/image.h
@@ -341,7 +341,7 @@ Image<ValueType>::Image(const std::shared_ptr<Image<ValueType>::Buffer> &buffer_
 }
 
 template <typename ValueType> Image<ValueType>::~Image() {
-  if (buffer.unique()) {
+  if (buffer.use_count() == 1) {
     // was image preloaded and read/write? If so,need to write back:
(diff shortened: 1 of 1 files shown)


Lestropie:
Changing `.unique()` to `.use_count() == 1` may silence deprecation warnings, but it does not redress the fundamental problem that is the reason for the deprecation. `.use_count()` only uses `std::memory_order_relaxed` to keep track of instance count; as such, this function returning a value of 1 is *not* a guarantee that the thread is either safe to, or should be responsible for, performing the delayed write-back. 

> [The value returned by use_count should be considered approximate, as the number of shared owners might change in other threads between the atomic retrieval and meaningful use of the value.](http://en.cppreference.com/w/cpp/memory/shared_ptr/use_count.html)

Alternative solution is warranted; see #2894.

Lestropie:
Closed by #3315.

PR #3234, opened 2025-11-11
Explicitly install eigen3 for macOS CI tests

daljit46:
The latest version of Eigen (v5) requires C++14, while we only support C++11 on `master`.


Lestropie (approved):
:+1:

bjeurissen:
With the latest macOS/Xcode even 3.4.1 will not work (3.4.0 does) so as soon as the CI gets an updated compiler it will fail again.

daljit46:
@bjeurissen It's probably best to open a separate issue with the full compilation error you get on the latest version of macOS.

PR #3237, opened 2025-11-26
New utility for pattern-matching style over std::variant

daljit46:
This adds a new utility for pattern-matching style over `std::variant`, which is a type-safe version of a discriminated union introduced in C++17.

Example:
```
  std::variant<int, std::string> v = 42;
  auto s = MR::match_v(v,
       [](int i) { return std::to_string(i); },
       [](const std::string& s) { return s; });
  // s == "42"
```

### DIFF
--- a/cpp/core/match_variant.h
+++ b/cpp/core/match_variant.h
@@ -26,7 +26,7 @@
 //   std::variant<int, std::string> v = 42;
 //   auto s = MR::match_v(v,
 //       [](int i) { return std::to_string(i); },
-//       [](const std::string& s) { return s; });
+//       [](const std::string& s) { return s; }); // check_syntax off
 //   // s == "42"
 
 namespace MR {


cpp/core/match_variant.h
```diff
@@ -0,0 +1,40 @@
...
+
+// Utility for pattern-matching style over std::variant.
+// - `overload` builds an overload set from multiple lambdas (or function objects).
+// - `match_v` wraps std::visit, forwarding the variant and visitors.
+//   It returns whatever the selected lambda returns.
+// Example:
+//   std::variant<int, std::string> v = 42;
+//   auto s = MR::match_v(v,
+//       [](int i) { return std::to_string(i); },
+//       [](const std::string& s) { return s; });
```
Lestropie:
```suggestion
//       [](const std::string& s) { return s; }); // check_syntax off
```

(assuming this would be a recommended usage given discussion in & around #3212)

Issue #3219, opened 2025-10-14
Introduce a span class in the codebase


daljit46:
While working on my GPU-based registration code, I found it very useful to use [`std::span`](https://en.cppreference.com/w/cpp/container/span.html), a C++20 view into contiguous containers. This is very useful in functions which need to operate on a contiguous section of memory, regardless of the underlying container (e.g. `std::array`, `std::initializer_list`, `std::vector` or something else). 
I think we should introduce this into our codebase. The motivation is similar in spirit to #3212. `std::span` is available with C++20 and onwards, but there are other API compatible implementations like [`tcb::span`](https://github.com/tcbrindle/span) that work with C++17. Once we upgrade to C++20, it would be mostly a case of substituting the namespace `tcb` with `std` (we could even use an alias if wanted, with something like `namespace std { using tcb::span; }`).

Lestropie:
Closed by #3238.

Issue #2554, opened 2023-01-03
Transition Windows usage to WSL


Lestropie:
Follows recent #2536.

Related to [forum Wiki page](https://community.mrtrix.org/t/installing-on-windows-subsystem-for-linux-wsl/3637).

I've not long ago upgraded to Windows 11 on my home PC, but only today did I try getting `mrview` working in WSL, and it was relatively seamless. So as has been discussed on multiple occasions, it's potentially time now to transition the Windows installation instructions to be based on WSL.

- [ ] Determine minimum requirements for WSL with GUI
      (in earlier discussions it was suggested that this would only be available for Windows 11, but some pages I've read suggest that the latest solution can work on Windows 10 as well)
- [ ] Contemplate full deprecation of MSYS2
      (may depend on result of investigation above)
- [ ] Modify software documentation to reflect alternative installation instructions
- [ ] Add banner to forum Wiki page indicating change in capabilities, linking to new instructions, and that the trailing information is there for historical reference only
      (and some of the historical text there may need to be edited to refine statements that are outright no longer true)
- [ ] Consider adding replies to multiple community forum threads discussing the prospect of using WSL for MRtrix3, so that anyone coming across those threads is directed to the up-to-date instructions

bjeurissen:
If it works on windows 10, I am definitely in favor of this.

Did you get a chance of benchmarking some commands on WSL against msys by any chance?

Lestropie:
Good idea RE benchmarking.

`mrview` opening speed looks comparable. MSMT looks comparable, at least as far as the processing itself is concerned.

Seems to be a big difference in disk access. `mrconvert`-ing a DWI series from `+1,+2,+3,+4` to `+2,+3,+4,+1` with `-nthreads 0` is 1s on MSYS2, 15s on WSL. Both input and output are memory-mapped in both cases, and at the same filesystem location. Moving the files into the WSL internal storage yields comparable performance. Apparently this is a [known performance issue](https://learn.microsoft.com/en-us/windows/wsl/compare-versions) with WSL2 accessing OS drives. Theoretically you can just keep any project files within the WSL Linux storage, and those files can even be navigated to directly within File Explorer; but it's undoubtedly a performance issue that would be reported many times.

jdtournier:
> If it works on windows 10, I am definitely in favor of this.

[...]

PR #3238, opened 2025-11-26
GPU compute API abstraction on top of WebGPU

daljit46:
This PR supersedes #3096.

It introduces a new GPU compute abstraction built on top of WebGPU. See #3096 for general motivation and design philosophy. A notable change from that PR is now shaders are now required to be written in [Slang](https://shader-slang.org/), a new programming language created by NVidia (now under the umbrella of the Khronos group). This choice has been motivated by the fact that Slang provides many useful features like modules and generics for better code reusability and modularisation.

Some issues still need to be resolved (in future PRs):

- Deal with the issue raised in #3108.
- Add more tests, especially for the upload/download of `MR::Image` instances to/from GPU (currently the API only supports `MR::Image<float>`).
- Create a real-world example that illustrates the use of the API.

One additional thing that this PR introduces is the addition of a new `tcb::span` class (see  #3219); however, unlike for other third-party dependencies, the class has been directly added to the codebase rather than fetched via CMake. The hope is that this class will become redundant once we have access to C++20's `std::span`.


daljit46:
The CI tests are not running on MSYS2, but given that we intend to pursue #2554 and that this is not blocking anything on `dev`, I will merge this soon.

.github/workflows/checks.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       run: cd build && ctest -L cli --output-on-failure
 
     - name: unit tests
-      run: cd build && ctest -L unittest --output-on-failure --parallel 4
```
Lestropie, line 64: Is this deletion intentional? The "compatibility" tests that were incompatible with parallel execution have been moved out of the "unittest" category (111947a48cc0526739764bd373bff6ae15304334), such that the parallelisation added in 2916ca46df1a4ad90c9317039d703dc3246841f8 became acceptable.

cpp/core/CMakeLists.txt
```diff
@@ -110,13 +113,36 @@ target_link_libraries(mrtrix-core PUBLIC
...
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            $<TARGET_FILE:dawn::webgpu_dawn>
+            $<TARGET_FILE:slang::slang>
+            $<TARGET_FILE_DIR:mrtrix-core>
+        COMMENT "Copying imported dependencies to output directory"
+    )
+# In an MSYS2 environment, dawn fails to load the required vulkan-1.dll library when
+# creating a WebGPU instance. To remedy this, we copy the DLL from C:/Windows/System32
+# to the build output directory.
+# TODO: Investigate further and find a better solution.
```
Lestropie: List Issue and remove TODO code comment?

cpp/core/gpu/gpu.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+    wgpu_adapter = request_adapter_result.adapter;
+    const std::vector<wgpu::FeatureName> required_device_features = {
+        wgpu::FeatureName::R8UnormStorage, wgpu::FeatureName::Float32Filterable, wgpu::FeatureName::Subgroups};
+
+    wgpu::Limits supported_limits;
+    wgpu_adapter.GetLimits(&supported_limits);
+
+    const uint64_t desired_max_storage_buffer_binding_size = 1'073'741'824ULL; // 1 GiB
+    const uint64_t desired_max_buffer_size = 1'073'741'824ULL;                 // 1 GiB
```
Lestropie: `constexpr`?

cpp/core/gpu/gpu.cpp
```diff
@@ -0,0 +1,702 @@
...
+wgpu::Buffer ComputeContext::inner_new_buffer_from_host_memory(const void *srcMemory,
+                                                               size_t srcByteSize,
+                                                               BufferType bufferType) const {
+  const auto buffer = inner_new_empty_buffer(srcByteSize, bufferType);
+  inner_write_to_buffer(buffer, srcMemory, srcByteSize, 0);
+  return buffer;
+}
+
+void ComputeContext::inner_download_buffer(const wgpu::Buffer &buffer, void *dstMemory, size_t dstByteSize) const {
+  assert(buffer.GetSize() == dstByteSize);
```
Lestropie, line 362: Slightly strange to require an input function argument that is forbidden from differing from another compulsory function input. It will only catch developers not allocating adequate memory if they are also honest in how they derive the value that is provided to the function. Eg. They could take a naked destination pointer and then lazily specify `buffer.GetSize()` as the third argument.

If trying to provide assurances on volume of memory pre-allocated for a naked pointer and nothing in `<memory>` is applicable (eg. interfacing with API that uses naked pointers, I wonder if it would be better to specify a class that encapsulates a `void*`-`size_t` pair, which could then be constructed & passed around as a singleton?

cpp/core/gpu/gpu.cpp
```diff
@@ -0,0 +1,702 @@
...
+    throw MR::Exception("Failed to map buffer to host memory: wgpu::Buffer::GetMappedRange returned nullptr");
+  }
+}
+
+void ComputeContext::inner_write_to_buffer(const wgpu::Buffer &buffer,
+                                           const void *data,
+                                           size_t srcByteSize,
+                                           uint64_t offset) const {
+  if (buffer.GetUsage() & wgpu::BufferUsage::Uniform) {
+    // TODO: fix
```
Lestropie: Residual TODO
(would still prefer to catch these in CI: #1054)

cpp/core/gpu/gpu.h
```diff
@@ -0,0 +1,328 @@
...
+};
+
+// Absolute/relative (to working dir) path  of a WGSL file.
+struct ShaderFile {
+  std::filesystem::path file_path;
+};
+
+struct InlineShaderText {
+  std::string text;
+};
```
Lestropie, line 139: `using`?

cpp/core/gpu/slangcodegen.cpp
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  Slang::ComPtr<slang::IComponentType> slang_program;
+  std::vector<slang::IComponentType *> shader_components;
+  shader_components.push_back(shader_module.get());
+  if (!generic_type_args.empty()) {
+    shader_components.push_back(specialized_entry_point.get());
+  } else {
+    shader_components.push_back(entry_point.get());
+  }
```
Lestropie:
```suggestion
  shader_components.push_back(generic_type_args.empty() ?
                              entry_point.get() :
                              specialized_entry_point.get())
```
?

cpp/core/gpu/slangcodegen.cpp
```diff
@@ -0,0 +1,365 @@
...
+    return bindings_map;
+  }
+
+  auto *global_var_layout = programLayout->getGlobalParamsVarLayout();
+  if (global_var_layout != nullptr) {
+    // If the program has global variables, we can find bindings in them.
+    find_bindings_in_variable_layout(global_var_layout, bindings_map);
+  }
+
+  // TODO: Handle multiple entry points properly.
```
Lestropie: Residual TODO; consider moving to standalone Issue

cpp/core/gpu/slangcodegen.cpp
```diff
@@ -0,0 +1,365 @@
...
+    assert(false && "Slang entry point variable layout has no type layout!");
+    return bindings_map;
+  }
+
+  find_bindings_in_variable_layout(entry_point_root_variable_layout, bindings_map);
+  return bindings_map;
+}
+
+std::array<uint32_t, 3> workgroup_size(slang::ProgramLayout *programLayout) {
+  // TODO: Handle multiple entry points and choose the correct one.
```
Lestropie: Residual TODO

Lestropie (changes requested):
- [ ] Requires re-run of `update_copyright`

Exceptionally difficult one to review. Left just several fairly inconsequential comments. The tests work as basic sanity checks.

One thing that comes to mind is that for a developer to take on the non-trivial challenge of adopting this framework, they'd need to see the potential translation between code completing some function and their own application. While registration has been a functional endpoint, I wonder if the complexities of that particular application might obfuscate such a translation. We could think about whether there's a different application that could be implemented, that is not so trivial as the tests added here but not so esoteric that the code does not translate to other applications, whether or not that application is something that interface exposure to software users makes sense. Maybe some voxel-wise operation on 4D data that is computationally expensive but not esoteric.

.github/workflows/checks.yml
daljit46: This was just a temporary solution to diagnose whether tests were failing due to parallelistation. I will restore it once I confirm things work without this.

cpp/core/CMakeLists.txt
daljit46: Yes, thanks that's probably wiser.

cpp/core/gpu/gpu.cpp
daljit46: Will do 👍

cpp/core/gpu/gpu.cpp
daljit46: The reasoning was primarily to facilitate the upload of C++ structs on the stack (e.g. `new_buffer_from_host_memory(&struct, sizeof(StructType))`, but perhaps it's worth rethinking this.

cpp/core/gpu/gpu.h
daljit46: The issue is that `std::filesystem::path` is implicitly convertible to `std::string`. We could just have logic that checks whether a given string is a valid path, but I think being explicit enforces strong typing and improves readability.

daljit46:
> Maybe some voxel-wise operation on 4D data that is computationally expensive but not esoteric.

In #3258, there is some fairly isolated code in [imageoperations.h](https://github.com/MRtrix3/mrtrix3/blob/8873ab30639704329a9fb8823ed3e13264cf7db0/cpp/core/gpu/registration/imageoperations.h) where you can see how to you use the API for slightly more complex tasks, but not exactly "computationally expensive", although still faster compared to a CPU. 
But I'd be happy to add an example for this purpose. Something that involves matrix-vector multiplication could be a good fit for this I think.

Lestropie:
> Something that involves matrix-vector multiplication could be a good fit for this I think.

GPU-ifying `shconv` would perhaps be an appealing candidate, even though it's a very fast operation as-is.
Whether it's worth the effort for the sake of a demonstrative example depends on the complexity of the image registration implementation: if that's easy enough to follow from scratch despite the esoterics of registration then a simpler example might not be required.

cpp/core/gpu/gpu.h
Lestropie: So it's not just that it *is* implicitly convertible, but specifically that you want to utilise that functionality while *forbidding* such implicit conversion?

Can that be circumvented by something like:
```cpp
class ShaderFile : public std::filesystem::path {
  explicit std::string operator()() const { ... }
};
```
?
When I see a class / struct I intuitively expect the cardinality of the set of member functions and variables to be greater than one.
(This is nowhere near a contentious point BTW, just exploring)

cpp/core/gpu/gpu.cpp
Lestropie: Could you perhaps SFINAE your way to auto-generating that pointer-size pair for any usage that isn't already of that form?

daljit46: Yes, thinking about this, we could avoid exposing the `void*` + `size_t` overload in the public API. Instead, we can make a new function that looks something like:

```
  template<typename Object>
  Buffer<std::byte> new_buffer_from_host_object(const Object& object, BufferType buffer_type = BufferType::StorageBuffer) const {
      static_assert(std::is_trivially_copyable_v<Object>, "Object must be trivially copyable");
      static_assert(std::is_standard_layout_v<Object>, "Object must be standard layout");
      return {buffer_type, inner_new_buffer_from_host_memory(std::addressof(object), sizeof(Object), buffer_type)};
  }
```

This way any user of the API won't have to deal with void* pointers (they can always construct a `tcb::span` if they really need to upload raw bytes for some reason).

cpp/core/gpu/gpu.h
daljit46: > So it's not just that it is implicitly convertible, but specifically that you want to utilise that functionality while forbidding such implicit conversion?

Yes.

I don't that approach offers any advantages to what we have now, so I think it's best to stick what we have now as I don't really see any real downsides.

daljit46:
>  Whether it's worth the effort for the sake of a demonstrative example depends on the complexity of the image registration implementation: if that's easy enough to follow from scratch despite the esoterics of registration then a simpler example might not be required.

I think the code in #3258, contains enough somewhat isolated code so I would say yes to this.

On another note, I realised I didn't document in the code that the convention I chose for uploading arrays that represent matrices on the shader side is [column-major](https://github.com/MRtrix3/mrtrix3/blob/bfe549db2966e4e721679223a9535c1b41cca959/cpp/core/gpu/gpu.cpp#L259). However, when writing code in Slang the syntax of the language follows *row-major* convention. So for example, if I want to upload a `std::array<float, 16>` that represents a 4x4 matrix, then the flattened layout must be column-major. However, on the shader side you will use `matrix[row][column]` syntax (this is automatically handled by Slang). I chose to do this because we use `Eigen` a lot in our code, but this is not necessary and can be changed if we want to.

daljit46:
I've now addressed all feedback in this PR. I've also removed all TODOs in the code by fixing the issues, except in the case of obtaining a Vulkan-backed WebGPU instance for MSYS2 environments (in which case I've just deleted the logic to manually copy the Vulkan dll). As mentioned earlier, this is not blocking, and we can investigate later what can be done about it. 
All tests are passing, so this is ready for merging.

daljit46:
We discussed in the meeting today to gatekeep the addition of this PR (and future PRs related to GPU work), under a new CMake configuration option `MRTRIX_ENABLE_GPU`. This flag would enable/disable the fetching of WebGPU and Slang dependencies (in addition to the all cpp files added here). The motivation is that we want to release a version 3.1.0 of MRtrix3 that will (may?) not include the addition of this work and instead create a separate branch specificaly targeting that release.
It's fairly straightforward to add this flag, but @jdtournier mentioned that this flag should be OFF by default. One downside of that is this will disable CI tests for the GPU work (including future related PRs). So my proposal is to instead enable the flag by default and explicitly disable in the new release branch. Alternatively, we could explcitly enable the flag in the CI workflows, but I think that's a bit unncessary and more invasive.

daljit46:
@MRtrix3/mrtrix3-devs Unless there are any further comments regarding this PR, I will merge this tomorrow.

PR #3247, opened 2025-12-16
Add support for nightly builds for `dev` branch

daljit46:
This PR refactors the `releases.yml` workflow to add support for nightly builds for the `dev` branch. The workflow is scheduled to run every day at `00:00` and will update the release tagged with the `nightly` tag (which I have already created) with the artefacts for each platform. The release will show the latest 15 commits on `dev`. If there have been no changes on `dev` since the last run, the workflow is cancelled.
Additionally, the workflow file has been refactored for improved code reuse and a simplified YAML file that utilises the GitHub Actions matrix strategy to build artefacts for each platform.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -170,15 +170,15 @@ jobs:
       - name: Run Linux build
         if: matrix.platform == 'linux'
         run: |
-          ./packaging/package-linux-tarball.sh .
+          ./packaging/package-linux-tarball.sh . "${{ steps.out.outputs.output_name }}"
           mv mrtrix.tar.gz ${{ steps.out.outputs.output_name }}.${{ matrix.artifact_ext }}
 
       - name: Run macOS build
         if: matrix.platform == 'macos'
         run: |
           cd ./packaging/macos
-          ./build ${{ needs.prepare.outputs.branch }}
-          mv ./mrtrix3-macos-${{ needs.prepare.outputs.branch }}.${{ matrix.artifact_ext }} ../../${{ steps.out.outputs.output_name }}.${{ matrix.artifact_ext }}
+          ./build "${{ needs.prepare.outputs.branch }}" "${{ steps.out.outputs.output_name }}"
+          mv "./${{ steps.out.outputs.output_name }}.${{ matrix.artifact_ext }}" "../../${{ steps.out.outputs.output_name }}.${{ matrix.artifact_ext }}"
 
       - name: Run Windows build
         if: matrix.platform == 'windows'


.github/workflows/releases.yml
```diff
@@ -7,91 +7,113 @@ on:
...
-        id: envs
+      - name: Determine if there are new commits since last nightly
+        id: changes
         run: |
-          commit_sha=$(git rev-parse HEAD)
-          date=$(date +%Y_%m_%d)
-          echo "commit_sha=$commit_sha" >> $GITHUB_OUTPUT
-          echo "date=$date" >> $GITHUB_OUTPUT
-          echo "output_name=mrtrix3-macos-$commit_sha-$date" >> $GITHUB_OUTPUT
+          RECENT="$(git log -n 15 --oneline || true)"
```
Lestropie:
```suggestion
          RECENT=$(git log --oneline | grep "Merge pull request #\d+ from MRtrix3\/\w+" | head -n15)
```

Amend as necessary; also subsequent code would need to reflect that this is a list of merged PRs rather than commits. But I think attaching to nightly releases a list of the most recent PR merges to `dev` that are included within would be more interpretable than just the most recent commits. As long as `dev` remains protected the build triggering shouldn't be affected.

.github/workflows/releases.yml
```diff
@@ -113,25 +135,124 @@ jobs:
...
+        id: out
+        shell: pwsh
         run: |
-          commit_sha=$(git rev-parse HEAD)
-          date=$(date +%Y_%m_%d)
-          echo "commit_sha=$commit_sha" >> $GITHUB_OUTPUT
-          echo "date=$date" >> $GITHUB_OUTPUT
-          echo "output_name=mrtrix3-windows-$commit_sha-$date" >> $GITHUB_OUTPUT
+          $commit_sha = (git rev-parse HEAD).Trim()
+          $date = (Get-Date -Format 'dd-MM-yyyy')
```
Lestropie:
```suggestion
          $date = (Get-Date -Format 'yyyy-MM-dd')
```

Alphabetical == chronological

.github/workflows/releases.yml
```diff
@@ -113,25 +135,124 @@ jobs:
...
+          path: dist
+
+      - name: Show downloaded artifacts
+        run: ls -R dist
+
+      - name: Create or update nightly release
+        run: |
+          NOTES_FILE=notes.txt
+          {
+            echo "Nightly build for $(date -u +%d-%m-%Y)."
```
Lestropie:
```suggestion
            echo "Nightly build for $(date -u +%Y-%m-%d)."
```

.github/workflows/releases.yml
Lestropie: Actually `'yyyyMMdd'` might make more sense given there are dash-delimiters elsewhere in the output name.

bjeurissen:
What I noticed is that the directory naming scheme is not the same across platforms, e.g. on macOS it unpacks to ``mrtrix3``, which easily clashes with existing directories.  On linux it unpacks to ``mrtrix3-linux-16-12-2025`` which corresponds to the filename in the archive. I think the latter is preferable.

jdtournier:
Just checked the latest release artefacts on Daljit's repo, and the version string issue is resolved (base tag still shows up as `3.0.4`, but I guess that would be fixed by updating your fork with the latest tags?). 

Naming-wise, I'd be happy with a `latest-dev` or `nightly-dev`. Not sure `alpha` really captures what it means for the average user. Also, hopefully eventually those releases won't be *alpha*, but very minor bug fixes on the main release... :crossed_fingers: 

But ultimately, I don't think the name of the tag itself is the issue (it makes a very minor hard-to-spot appearance as the tag on the release page), what matters more is the description shown as the title on the GitHub Release page (currently reads "MRtrix3 Nightly Release" on Daljit's repo). And since the workflow recycles the release (as I understand it), we can set that title however we want and (I assume) modify it later if appropriate (please correct me if I've got that wrong!). I'd advocate for something like "MRtrix3 nightly build (dev branch)" (not sure 'latest' would work instead of 'nightly' here - some users might assume this is the latest *release*). And add a bit more to the description to make it clear that this is not an official release, but provides a build of the latest version of the code as it stands to allow testing, etc. 

Once merged to `master`, it would be good to have nightly releases for both `dev` and `master` branches (maybe call them `latest-master` and `latest-dev`?). 

The version string shown with the `-version` option is indeed a bit misleading for the `dev` branch (for example, the linux version on Daljit's releases is `3.0.4-1878-g05cd8f63`). Not sure what would be the best way to deal with this... But I'm also not sure it needs fixing: it'll be accurate for the `master` branch (once merged...), and the only issue for the `dev` branch is that it won't be obvious that this is a `dev` build. But the information is there to figure that out if we need to, so I'm personally not too bothered about that...

daljit46:
I've now deleted the `nightly` tag and created a new `nightly-dev` tag given the comment above. Also addressed the feedback in the earlier comments. This should be ready to be merged.

PR #3255, opened 2026-01-12
Merge master changes into dev

Lestropie:
This merge is necessary in order for CI tests of PRs targeting `dev` to pass, given that the copyright update pushed to `master` in #3252 needs to be ported to `dev` also.

Duplicates fix in #3254.

- [x] Check set of PRs merged to `master` that are included here, and make sure that there are no fixes that have been discarded as part of the merge resolution process.


Lestropie:
- [x] Rectify `tckglobal` tests hanging on all platforms

Lestropie:
`clang-tidy` Action immediately fails here due to the number of files with non-zero changes being too large. In future it may be preferable to do a `master`-to-`dev` merge of everything up to *but not including* the copyright update, so that all checks can be completed, and then do a second PR that is *exclusively* a merge of the copyright update.

PR #3257, opened 2026-01-14
Docs: Update ACT page

Lestropie:
Received another instance of the "do I need to resample my FOD data to match my 5TT image" question, so felt it pertinent to make sure there's a definitive statement about it in the docs.

In checking that page I also realised that it still had my long-outdated suggestion of transforming T1w to DWI data. So this might be one reason why that question still keeps coming up despite stating the converse for a long time on the forum.

### DIFF
--- a/cmd/tckgen.cpp
+++ b/cmd/tckgen.cpp
@@ -16,6 +16,7 @@
 
 #include "command.h"
 #include "image.h"
+#include "version.h"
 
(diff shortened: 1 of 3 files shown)


Lestropie:
https://mrtrix3-dev.readthedocs.io/en/docs_act_update/quantitative_structural_connectivity/act.html

PR #3259, opened 2026-01-15
Fix bug in detection of duplicates in user-specified strides

Lestropie:
Currently making an attempt at refactoring the stride handling as per discussion in #3108 and came across this howler. Quite possibly never actually affected anybody given it will pretty much always be the case that all three spatial strides are either zero or non-zero.


jdtournier (approved):
Yup, well spotted. That one's on me...

PR #3261, opened 2026-01-22
Fix ICLS performance regression

### LINKED ISSUE #3260: CI binary tests taking long time on dev
Recent CI runs, that some binary tests take a long time to execute. Three tests in particular: `dwi2fod_msmt_default`, `dwi2fod_msmt_masked` and `dwi2tensor_dkt_constrained`.
Example:

```
 42/396 Test  #42: dwi2fod_msmt_default ....................................   Passed  121.10 sec
        Start  43: dwi2fod_msmt_masked
 43/396 Test  #43: dwi2fod_msmt_masked .....................................   Passed  107.94 sec
        Start  44: dwi2tensor_bzero
 44/396 Test  #44: dwi2tensor_bzero ........................................   Passed    0.02 sec
        Start  45: dwi2tensor_default
 45/396 Test  #45: dwi2tensor_default ......................................   Passed    0.02 sec
        Start  46: dwi2tensor_dkt_constrained
 46/396 Test  #46: dwi2tensor_dkt_constrained ..............................   Passed  174.18 sec
```

I hadn't noticed this earlier, but I believe this is due to some change that happened in the last few months. We should investigate whether this can be rectified.

daljit46:
Fixes #3260. 
The following change was introduced in #3182:

```cpp
-    std::fill(active.begin(), active.end(), false);
-    if (num_eq > 0)
-      std::fill(active.begin() + num_ineq, active.end(), true);
+    active.setZero();
+    active.tail(num_ineq).setOnes();
```

We were mistakenly setting the last `num_ineq` entries active, while we should be doing the opposite: set the last `num_eq` entries active. This caused a large set of inequality constraints to be active in the solver, leading to performance issues. This PR restores the behaviour prior to #3182.

### DIFF
--- a/cpp/core/math/constrained_least_squares.h
+++ b/cpp/core/math/constrained_least_squares.h
@@ -237,7 +237,9 @@ public:
     lambda_prev.setZero();
     // set active set empty:
     active.setZero();
-    active.tail(num_ineq).setOnes();
+    if (num_eq > 0) {
+      active.tail(num_eq).setOnes();
+    }
 
     // initial estimate of constraint values:
     c = c_u;


bjeurissen:
Well spotted! Was it only a matter of speed or did it also change the outcome? I would suspect the former as the tests still passed.

jdtournier:
As far as I can tell, that basically meant it was initialised with all constraints active, and since constraints are added/removed one at a time, this means it takes a lot longer to converge to the solution (which will typically have only a handful of constraints active). That explains the longer runtime, and why all the tests passed. Nice detective work!

jdtournier (approved):
Fix looks good to me.

PR #3263, opened 2026-01-27
Update CMake version for sanitizers workflow

daljit46:
On `dev`, since #3239, we require CMake >= 3.22. This change needs to be propagated to our sanitisers weekly workflows, which are currently failing.


daljit46:
@MRtrix3/mrtrix3-devs Can someone approve this? This is a trivial change and shouldn't cause any problem.

Lestropie (approved):

PR #3262, opened 2026-01-26
Add TRX support to tckconvert

mattcieslak:
This is a minor addition to `tckconvert` that doesn't address any of the deep issues discussed in #2241, but it does 

 1. Add the TRX c++ library to the mrtrix build
 2. Adds official TRX testing data to the test data repo (https://github.com/MRtrix3/test_data/pull/6)
 3. Edits tckconvert to read/write trx files
 4. Adds tests to confirm that round-trip conversion works

I am successfully building and testing locally, but may need some help getting the build working cleanly here. Despite not fully integrating TRX into mrtrix, I think this is a great foot in the door that will let us tckconvert TRX files directly within mrtrix3. 

Tagging @arokem, @frheault

Issue #2682, opened 2023-07-15
NPY: Potential enhancements


Lestropie:
Generating a list following attempted resolution of #2437 so that it's documented what else could potentially be done here without leaving TODO comments throughout the code.

1.  Provide support for complex floating-point types.
    This would probably not actually be particularly difficult, just a matter of uncommenting some code and implementing satisfactory tests.

2.  Implement fast read support.
    Where `File::NPY::load_(vector|matrix)()` (or indeed `File::Matrix::load_(vector|matrix)()`) has been called, but it just so happens that there is correspondence between input file and destination class in terms of data type / endianness / order, it should be possible to do a straight `memcpy()`.

3.  Contemplate how to deal with packed bitwise storage.
    It would have been nice to be able to write bitwise data to disk efficiently using this format; for instance, the fixel-wise dixel masks in #2639 if using non-image formats for fixel-wise data as per #1664. Unfortunately NPY (and `numpy` internally) treat boolean arrays with one byte per element. While it's possible to perform a bit-packing operation, this results in `numpy` reporting the type as unsigned 1-byte integer and "shape" as number of bytes. I'm not sure if there's any way around this, but thought I'd list in case.

4.  NPZ support.
    For fixel-fixel connectivity matrices (#2670) and fixel-fixel correspondence mappings (#2200), storing multiple arrays of known names & types in a single uncompressed NPZ file rather than as images in a directory is appealing; a single class could read-only `Eigen::Map<>` everything in each case. Current third-party GZ library does not provide support for archives containing multiple files, so would need to look into alternative libraries.

PR #3265, opened 2026-01-29
Add TRX support to tckconvert

mattcieslak:
This adds picks up where #3262 left off, adding TRX support to tckconvert, updates the documentation and adds tests. It depends on https://github.com/MRtrix3/test_data/pull/6 to pass the tests. 

I haven't added trx-cpp to any of the builds - I was hoping for some input on how would be most convenient to do so.

Thanks in advance!


daljit46:
Currently, we handle [dependencies](https://github.com/MRtrix3/mrtrix3/blob/6690c05929d07f6b7d04e87412a109527328ba31/cmake/Dependencies.cmake) through CMake's FetchContent with optional flags to use a system-provided version of the given library. I think the same approach should work fine with `trx-cpp`. One issue would be that `trx-cpp` depends on `libzip` and expects that to be installed on the user's system. So I see two options here: either mandate the installation of `libzip` for building the project (as we do for Qt) or we also fetch `libzip` via FetchContent when `trx-cpp` support is required. @jdtournier @Lestropie any thoughts? 

@MRtrix3/mrtrix3-devs as an aside, I think this issue highlights some of the limitations of using FetchContent for dependency handling, as dealing with transitive dependencies is not straightforward and we may want to consider adopting a package manager like `vcpkg`.

jdtournier:
I think we should keep it as simple as possible for now, and document the expectations sufficiently for other developers. As long as regular users can use the binaries we produce, they won't care either way. It's a relatively small subset of people who might benefit from transitioning to a more complex package manager, and few of them will be particularly upset if they need to install other packages manually. Realistically, I don't think it's worth investing much time in finessing the developer experience... 

The bigger issue is whether we need to bundle these additional dependencies in the binary packages (which I assume we should?).

Lestropie:
I'm not perturbed about specifically inheriting a `libzip` dependency given I'd like to support the `.npz` format (#2682) that I think would itself require `libzip`. Not to say that transitive dependency isn't a problem, just that this particular one doesn't scream problematic to me. It's pretty common as a source compilation dependency and stable as an API, and the runtime dependency will be ubiquitous, so I'd just mandate it at system level.

testing/binaries/tests/tckconvert/tckconvert
```diff
@@ -0,0 +1,38 @@
+tckconvert -ascii tracks.tck tmp.vtk -force && awk '/POINTS/{s=1;next}/LINES/{s=0}s' tmp.vtk > tmppoints1.txt && awk '/POINTS/{s=1;next}/LINES/{s=0}s' tckconvert/out0.vtk > tmppoints2.txt && testing_diff_matrix tmppoints1.txt tmppoints2.txt -abs 1e-4
```
Lestropie: `tckconvert` tests need to be moved to reflect new structure on MRtrix `dev` branch. Tests for a given command are aggregated within a directory, with individual tests split across individually named files within that directory. This PR appears to currently be erroneously mass-reintroducing tests previously split across multiple such files back into a singular "`tckconvert`" test file in the `tckconvert/` sub-directory in conjunction with the additional TRX tests.

cpp/cmd/tckconvert.cpp
```diff
@@ -329,6 +357,206 @@ class ASCIIWriter : public WriterInterface<float> {
...
+    } catch (const std::exception &e) {
+      Exception(e.what()).display();
+      App::exit_error_code = 1;
+    }
+  }
+
+  bool has_metadata() const { return has_aux_data; }
+
+private:
+  trxmmap::TrxFile<float> *trx;
```
Lestropie: Smart pointer would be preferable here.

daljit46:
@mattcieslak Given the comments above, I think adding something along the lines of:

```cmake
if(MRTRIX_USE_SYSTEM_TRXCPP)
    find_package(trx-cpp CONFIG REQUIRED)
else()
    set(trxcpp_url "https://github.com/tee-ar-ex/trx-cpp")
    FetchContent_Declare(
        trx-cpp
        DOWNLOAD_EXTRACT_TIMESTAMP ON
        GIT_REPOSITORY ${trxcpp_url}
        GIT_TAG ea9ebcbf4b08e047e6493eb8077f789cdfac36b7
    )
    FetchContent_MakeAvailable(trx-cpp)
endif()
```

in `cmake/Dependencies.cmake` should do the trick. We'll also need to update our build instructions and CI workflows for the installation of `libzip`.

cpp/cmd/tckconvert.cpp
mattcieslak: fixed at the other locations too

mattcieslak:
We've been busy adding TRX support to [ITK](https://github.com/InsightSoftwareConsortium/ITK/pull/5925) and [trekker](https://github.com/dmritrekker/trekker/issues/76), and made quite a few improvements to trx-cpp in the process. Circling back to MRtrix, I'd like to add TRX support to a couple other programs and was wondering if you'd prefer we do that in this PR or in separate PRs. Currently I'm thinking

 * `tckinfo`: print groups, dps and dpv info from trx
 * `fixel2tsf`: add option to add tsf data as dpv in a trx (either append to input or make a new trx)
 * `tcksift*`: add option to store sift weights as dps
 * `tck2connectome`: write streamline assignments to groups, read dps as weights
 * `mrview`: There are a number of potentially very nice additions here, like
     * Listing separate rows for each group in the trx file
     * Coloring groups pased on the dpg color
     * Coloring streamlines based on dps or dpv
 * `tckgen`: there's a nice streaming writer class in trx-cpp that could be a great fit here

but `mrview` and `tckgen` will probably be pretty extensive additions.

Lestropie:
Closing in favour of #3415.

The decision was made that rather than supporting TRX exclusively through the `tckconvert` command (as is currently the case with a handful of non-dMRI-specific formats), and thus requiring conversion between `.tck` and TRX for compatibility with different softwares, TRX would instead be promoted to have intrinsic support within the software API and therefore all MRtrix3 commands that interface with tractogram data would support it.

PR #3267, opened 2026-02-05
SIFT2 Differential

Lestropie:
Software development work to support an imminent publication from @ppruc.

Draft Pull Request: there is still considerable work to do to resolve these changes against the current `dev` tip, and other software commands simply will not compile with the current changeset as they have not yet been updated to make use of the new back-end.

## Summary

This augments the `tcksift2` command to have two modes of operation:
-   "Absolute" mode
    The historical SIFT2 method. Streamline weights are determined that converge the tractogram toward fixel-wise measures of *absolute* fibre density.
-   "Differential" mode
    The quantitative data utilised per fixel is proportional not to the total fibre density estimated for that fixel, but to the fibre density *change* associated with some other experimental variable (eg. time for longitudinal analysis).

Note that the software interface has changed slightly, including for the pre-existing Absolute mode. The input is no longer an FOD image (to which the FOD segmentation algorithm was performed internally), but a fixel data file. One of the benefits of performing the FOD segmentation internally was the preservation of dixel information, which enabled assignment of streamlines to fixels taking into account FOD lobe shape, rather than merely the dot product with the fixel mean direction. This functionality is however preserved by saving that information (#2625) and loading it in conjunction with the fixel dataset (#2657).

## Notes for any early adopters

1.  This code is not yet included in any tagged release, and may not be for some time.
    You will need to compile from source (perhaps as a container to avoid MRtrix version clashing).
2.  When compiling from source, you will need to compile *only* the `tcksift2` command.
    Other commands in the software are currently broken by these changes.
    In its current state this is achieved by running "`./build bin/tcksift2`"; but these instructions will change as this code branch is brought up to date.
3.  There will be a tutorial page demonstrating a typical analysis pipeline using this tool posted imminently by @ppruc.

## Still to do

(not guaranteed to be complete at time of posting)

- [ ] Bring up to date with `dev` tip
    This will include filesystem restructuring, `clang-format`, anything that `check_syntax` complains about.
- [ ] Replace macros defining numerical constants with `constexpr`s.
- [ ] Consider changing default regularisation for absolute mode.
    @ppruc: I know we said change to streamline basis and leave at 0.1, but it occurred to me that I had been concerned about over-regularisation at the current defaults, so there's some concern that going to streamline basis, which can be at greater danger of precluding correction of the largest discrepancies, may go too hard. Might need to generate a few to convince myself it's not a problem.
- [ ] Add sanity check for differential data
    Differential mode expects as input half-differences, for the delta coefficients to be bounded as (-1.0, +1.0). However I think it's just trusting the user to do so. One option maybe worth adding is that if the differential FD is more than half the absolute FD in any fixel, that's likely due to a user providing full-differences rather than half-differences.
- [ ] `tckedit` currently sets `min_weight` to `0.0F`, and precludes setting `-min_weight` to less than 0.0 at the command-line. This precludes utilisation of differential weights within existing mechanisms intended only for absolute weights (there may well be many other instances of such). If it is intended that e.g. `-tck_weights_in` and the like can be utilised with differential weights, then these functionalities need to be updated to support negative values.
    (Note that it's not necessarily the optimal long-term solution to re-use those functionalities intended for absolute weights with differential weights; but need to make the requisite machinery available one way or the other)


ppruc:
**Reference**
The conceptual background of the methods enabled by the software changes to `tcksift2` described in this pull request is detailed in this [preprint](https://www.biorxiv.org/content/10.64898/2026.02.09.704742v1). Code and data supporting the article are available in a [separate GitHub repository](https://github.com/ppruc/sift2_unbiased), including a [step-by-step tutorial](https://github.com/ppruc/sift2_unbiased/tree/main/tutorial) for users wishing to apply the methods to their own data.

PR #3269, opened 2026-02-19
Add Norm and Mode of Anisotropy (NA and MO)

kikiluvbrains:
Adds NA and MO as an output option in tensor2metric. Verified using the provided test data and an external DTI dataset.

paper link: Ennis, D. B., & Kindlmann, G. (2006). Orthogonal tensor invariants and the analysis of diffusion tensor magnetic resonance images. Magnetic resonance in medicine, 55(1), 136–146. https://doi.org/10.1002/mrm.20741

other supporting literature:
Chad, J. A., Pasternak, O., & Chen, J. J. (2021). Orthogonal moment diffusion tensor decomposition reveals age-related degeneration patterns in complex fiber architecture. Neurobiology of aging, 101, 150–159. https://doi.org/10.1016/j.neurobiolaging.2020.12.020

example of mo on example dti data:
image


cmd/tensor2metric.cpp
```diff
@@ -0,0 +1,444 @@
...
+  SYNOPSIS = "Generate maps of tensor-derived parameters";
+
+  REFERENCES
+  + "Basser, P. J.; Mattiello, J. & Lebihan, D. "
+    "MR diffusion tensor spectroscopy and imaging. "
+    "Biophysical Journal, 1994, 66, 259-267"
+  + "Westin, C. F.; Peled, S.; Gudbjartsson, H.; Kikinis, R. & Jolesz, F. A. "
+    "Geometrical diffusion measures for MRI from tensor basis analysis. "
+    "Proc Intl Soc Mag Reson Med, 1997, 5, 1742";
+  +
```
Lestropie:
```suggestion
  + "* If using -na or -mo options: \n"
```

cmd/tensor2metric.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+      // new section added here
+      double mo = 0.0;
+      if (mo_img.valid() || (vector_img.valid() )) {
+        const double l1 = eigval(0);
+        const double l2 = eigval(1);
+        const double l3 = eigval(2);
+
+
+        mo = DWI::tensor2MO(l1,l2,l3);
```
Lestropie: If this computation (as well as `tensor2NA()`) is being done based on the eigenvalues rather than the tensor, then "`tensor2MO()`" isn't a suitable name for the function.

cmd/tensor2metric.cpp
```diff
@@ -0,0 +1,444 @@
...
+      if (need_eigenvalues) {
+        eigval = es.eigenvalues();
+        ith_eig[0] = 0; ith_eig[1] = 1; ith_eig[2] = 2;
+        std::sort (std::begin (ith_eig), std::end (ith_eig),
+            [&eigval](size_t a, size_t b) { return abs(eigval[a]) > abs(eigval[b]); });
+      }
+
+      // new section added here
+      double mo = 0.0;
+      if (mo_img.valid() || (vector_img.valid() )) {
```
Lestropie:
```suggestion
      if (mo_img.valid()) {
```

The output vector image being valid may be one condition in which eigenvalues are computed, but that image being valid is not a prerequisite for the computation of `mo`.

Lestropie: In which case this code branch can be merged with the one below it. The code flow should simply be:
```text
if (exporting MO) {
    compute MO
    write MO
}
```

cmd/tensor2metric.cpp
```diff
@@ -0,0 +1,444 @@
...
+        const double l2 = eigval(1);
+        const double l3 = eigval(2);
+
+        na = DWI::tensor2NA(l1,l2,l3);
+      }
+      /* output na */
+      if (na_img.valid()) {
+        assign_pos_of (dt_img, 0, 3).to (na_img);
+        na_img.value() = na;
+      }
```
Lestropie:
```suggestion
      if (na_img.valid()) {
        assign_pos_of (dt_img, 0, 3).to (na_img);
        na_img.value() = DWI::tensor2NA(eigval(0),eigval(1),eigval(2));
      }
```

Indeed if these functions were to remain in `core/dwi/tensor.h`, then having them accept as input a singular 3-vector rather than 3 independent scalars would make more sense from an OOP perspective. However given the relative simplicity of those functions, and the fact that other tensor metrics are computed directly within `cmd/tensor2metric.cpp` rather than `core/dwi/tensor.h`, it might be simpler for that code to just appear here?

src/dwi/tensor.h
```diff
@@ -0,0 +1,133 @@
...
+             std::sqrt (T (1.5) * (a[0]*a[0]+a[1]*a[1]+a[2]*a[2] + T (2.0) * (dt[3]*dt[3]+dt[4]*dt[4]+dt[5]*dt[5])) / trace) :
+             T (0.0);
+    }
+
+    template <class Scalar> inline Scalar tensor2MO (const Scalar l1, const Scalar l2, const Scalar l3)
+    {
+      const Scalar md = (l1+l2+l3)/3.0;
+      const Scalar na = std::sqrt((l1-md)*(l1-md)+(l2-md)*(l2-md)+(l3-md)*(l3-md));
+
+      return 3.0*std::sqrt(6.0)*((l1-md)*(l2-md)*(l3-md)/(na*na*na));
```
Lestropie:
```suggestion
      return 3.0*std::sqrt(6.0)*((l1-md)*(l2-md)*(l3-md)/Math::pow3(na));
```

src/dwi/tensor.h
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    template <class Scalar> inline Scalar tensor2NA (const Scalar l1, const Scalar l2, const Scalar l3)
+    {
+      const Scalar md = (l1+l2+l3)/3.0;
+      return std::sqrt((l1-md)*(l1-md)+(l2-md)*(l2-md)+(l3-md)*(l3-md));
```
Lestropie:
```suggestion
      return std::sqrt(Math::pow2(l1-md) + Math::pow2(l2-md) + Math::pow2(l3-md));
```

Lestropie:
Thanks for the contribution @kikiluvbrains! Unless there's contention about the metrics themselves this should hopefully not be too difficult to accept.

- [x] PR currently shows unrelated commits. Hopefully this is just due to the merge conflicts with `dev`, and doing an explicit merge will resolve.
    (Let us know if you need help in this regard)

- [x] The new contrasts should be computed from the test DWI data and committed to the test data repository, and tests should be added to check for any regression.
    (This is kind of clunky to deal with as it's a separate repository, so I'll do it myself)

- [x] Address code comments.
    (If you want to accept multiple suggested changes as-is, please aggregate them into a batch; otherwise just edit and commit as you see fit)

cmd/tensor2metric.cpp
kikiluvbrains: changed name to eigen2MO() and eigen2NA()

kikiluvbrains:
Hey, merge conflicts should be resolved by now. And I went through and added in all fixes from the comments. Now, I think we need to add the new contrasts for the test DWI data. 

And by test data do you mean [this](https://github.com/MRtrix3/test_data/tree/master/tensor2metric) I have the output on NA and MO from the test dt.mif file, I can upload those in if that would help

Lestropie:
Ah, there's a bigger discrepancy here that I didn't realise in my first look.

While the PR is currently targeting the `dev` branch---which is the right choice given that the contribution is a feature addition and not a bug fix---it seems that the version of the code to which the code modifications were first applied was either the `master` branch or something close to it. This is a more substantial distinction at this point in time than it would have been previously, because of major changes that have happened on the `dev` branch. The command code has not only changed location (from `cmd/` to `cpp/cmd/`), but has also undergone a mass code reformatting. This tends to result in attempts to move code between those two branches (or their children) being quite difficult, because `git` labels absolutely everything as a conflict. As per the [contribution documentation](https://github.com/MRtrix3/mrtrix3/blob/master/CONTRIBUTING.md?plain=1#L100-L145), resistance is minimised if the correct starting branch is chosen from the outset before writing any code.

Given the disproportionate nature of the consequences due to that discrepancy, I think the best course is if I deal with the conflicts myself on this one. While it's useful to learn about resolution of merge conflicts in git, this is perhaps not the right use case to be learning from. There's probably also commits on `master` not yet back-propagated to `dev` that need to be dealt with separately so that the unrelated commits disappear from this PR.

Lestropie:
OK, I've done a bit of work here:

1.  Merged your changes onto the `dev` branch.
    Took me a couple of tries to get right... 

2.  Did a little bit of refactoring.
    There were a couple of suggestions I made that were only applied to the code relating to one of the two metrics rather than both.
    I additionally made use of Eigen functionalities in the computation of those metrics. Admittedly you would have been looking at the existing code in `tensor.h` for precedence, and that code doesn't really use Eigen, but that's because it's very old code that predates our change to using Eigen. So it's most certainly not a mistake on your part. But I nevertheless took the opportunity to demonstrate how Eigen's capabilities can be used to reduce the overall code size and arguably better communicate the intent of the calculation.
    
    **Edit**: This refactoring is shown in commit d081113ea1105e5084d4ce2cd2c65ea20c7a4412.

3.  Added test data and tests.
    As mentioned previously, this is clunky, so there wasn't any point in me trying to provide instruction from scratch, but hopefully it can demonstrate how it works such that if you make other contributions you might be able to complete part of the process yourself.
    The addition of the data itself is over on another repository:
    https://github.com/MRtrix3/test_data/tree/add_mo
    This code repository then needs to be instructed to pull that new data. The way in which this is encoded is different between the current `master` and `dev` branches; in 9021d5149f18efb39cbf291b2ffacf0df30f4592 you can see the update to `testing/CMakeLists.txt`. The addition of a test to generate those contrasts and compare against the pre-computed version is also slightly different between `master` and `dev`. `dev` is nicer IMO because each individual test is its own file, which enables more complex tests and the addition of comments.
    Once this PR is merged into `dev`, I then have to go back to the `test_data` repository, update the `dev` branch *there* to point to the commit with the new data, and delete the new `add_mo` branch *there*.
    So a lot of juggling, and a lot of ways for it to go wrong. I stuff it up all the time myself. And it requires write access to that test data repo. Hence why it's easiest for me to just do it. But certainly the writing of new tests is something that external contributors can have a go at themselves.

    (It's not this difficult a process to add test data to most software packages. Ours is harder because the test data are quite large and we don't want to bundle it all with every clone of the software. Hence there's actually two external repos that host the test data. But this then requires manually keeping them in sync with one another, which requires effort and diligence.)

Issue #2972, opened 2024-08-20
std::ssize_t audit


Lestropie:
Problem identified by @daljit46.

In various parts of the C++ code base, we have erroneously used `std::ssize_t` as "a signed integer of the same bit width as `std::size_t`, when in fact by the letter of the specification it is only guaranteed to support the value -1. 

- [ ] Identify all uses of `std::ssize_t` where negative values smaller than -1 need to be supported.
- [ ] Determine an appropriate substitute type to use (typedef one if necessary)
- [ ] Do a bulk substitution
- [ ] (*optional*) Identify uses of `std::size_t` that involve casting between signed and unsigned integer types to silence compiler warnings, and reconsider types used.
    (Eg. Pretty sure on at least one occasion I removed a non-functional assertion that a `size_t` be non-negative; may be other areas where catching negative values as errors may be better than using usigned)

daljit46:
> Determine an appropriate substitute type to use (typedef one if necessary)

Eigen uses `std::ptrdiff_t` for indices so I guess we could use that or perhaps `int64_t` if not used for indices.

Lestropie:
`std::ptrdiff_t` feels wrong from its name but from [my first search for a definition](https://en.cppreference.com/w/cpp/types/ptrdiff_t) is actually right:

> std::ptrdiff_t is used for pointer arithmetic and array indexing, if negative values are possible.

daljit46:
In the meeting yesterday, we discussed a few possible typedefs for `std::ptrdiff_t`:
- `Index_t`
- `Offset_t`
- `Stride_t`

Lestropie:
Sorry I missed. We've generally stuck to lowercase for typedefs internally. I'm also wondering if typedefs that are intended for more broad use tend to be `*_type` rather than `*_t` (eg. `default_type`). I like "index" and "stride"; is "offset" intended to be just a difference of "index"es, or more / something else?

jdtournier:
My personal preference is also for all lowercase. Leaning towards `index_t`, but would be happy enough with `index_type`. I feel offset & stride both have slightly narrower scopes than what we need here.

daljit46:
Another point I would like to add is that adding multiple typedefs for indices, offset (for file offsets) and strides with the same underlying idea is not a bad idea. There are very few downsides and it would make it easier if at some point we want to change the types.

jdtournier:
[...]

PR #3271, opened 2026-03-04
Refactoring of stride handling

Lestropie:
Draft PR. Still quite a lot left to do. Posting to open the floor to discussion.

#3108 struck me as odd, in that it is proposed that in addition to providing an array of integers as "desired strides", it may be necessary to additionally provide a sidecar boolean specifying "please interpret this as signed strides as opposed to just a desired order of axes". While it may be possible to store both types of information using the same eg. `std::vector<int>` class, what they are encoding, and therefore how their data should be interpreted, is fundamentally different (and I don't find "nearest match" to be the right description). As I dug further into this I realised that the current stride handling code was closer to C than C++: it's all verbose function names, relying on developers to invoke the right function at the right time. In the end I couldn't find a simple fix to achieve what I wanted; basically all of the handling needed to be OOP'd.

The ultimate aim of the PR is to replace #3108's:
```cpp
Image with_direct_io(Stride::List with_strides = Stride::List(),
                     StrideInterpretation strideInterpretation = StrideInterpretation::NearestMatch);
```

with:

```cpp
Image with_direct_io(const Stride::Permutation &with_permutation);
Image with_direct_io(const Eigen::Index contiguous_axis);
Image with_direct_io(const Stride::Symbolic &with_symbolic);
```

If you want the axes to appear in a particular order, and care neither about their signs nor the relative ordering of axes within groups, then you are ultimately specifying a *permutation*, *not* "strides". If however you want the data to have a very specific set of symbolic strides, such as in the GPU processing context, then what is being specified *is* "strides", and those data should not undergo any further "interpretation".

-----

Minor points:

-   I did a little bit of #2972. I was encountering inconsistencies in types used for the same purpose in different locations. I also had a bug very annoying to diagnose wherein `MR::Adapter::Base::ndim()` was being called instead of `MR::Adapter::PermuteAxes::ndim()` because the integer return type differed. I've made much of `MR::Adapter::Base` `virtual` even though it's not used in that way, it's to show intent rather than for function.

-   There's an interesting interaction between the stride handling and multi-threading. The multi-threading wants the indices of some number of axes to run in the inner loop, and the indices of all other axes to run in the outer loop, based on their relative strides. This turns out to arguably be better suited to `MR::Stride` then `MR::Thread`. It corresponds to transforming `Stride::Symbolic` into a `Stride::Permutation`, then to a `Stride::Order`, and then separating the head and tail of the list.

-   Because of the `HeaderType` template interface concept, when reading strides from a template class input, it is impossible to know whether that class has a symbolic vs. actual representation of strides. It's therefore necessary to treat anything that comes from a template class as actualised, and then manipulate as necessary.

-   There is some concern of borking use cases that aren't covered by the tests. There was a range of esoteric fixes required at the end of the porting process. Many of these I believe to be bad existing code combined with the refactored code being a bit less lenient; eg. calling `.with_direct_io(Stride::contiguous_along_axis(3))` on a 3D image.

-----

TODO:

- [x] Move the `stride` data member out of the `Header::Axis` class, and instead give `Header` a member variable `Stride::Symbolic`.
    A class `Header::Axis` would be expected to store information that is specific to one axis and independent of all other axes. For strides however, almost the opposite is true: it is explicitly about its relationship to other axes. The `HeaderType::stride(axis)` interface can possibly be preserved to minimise porting, though in all likelihood a lot of code that modifies strides of individual axes would actually make more sense (and be less bug-prone) to instead be specifying that some permutation is to be applied to the existing symbolic strides. 

- [x] Vary integer types of `Stride::Base` class.
    This could help mitigate bugs arising from casting these classes back to `std::vector` and then just re-interpreting the data as something they are not.
    For a list of integers that encodes a subset of axes, or alternatively one that encodes an order of permutation of all axes, an unsigned integer makes sense (though I'd probably use `Eigen::Index` anyway).
    For symbolic strides, `int8_t` would be perfectly fine as long as you don't mind limiting to images with fewer than 128 axes.
    For actualised strides, a wide signed integer type is required (we have erroneously used `ssize_t` for this, which works in practise but is not guaranteed).

- [x] Add tests.
    I will try to create tests that directly contrast functionality of the prior code (which remains in place under the `MR::Stride::Legacy` namespace) with their equivalent in the new code.
    - [x] Make sure a test is added that would have caught #3259.
    - [ ] Once tests guarantee no regression, delete `MR::Stride::Legacy` code.

- [x] Address outstanding TODOs.
    As stated, this is the first functioning version, and I leave comments to myself in the code as I'm trying to figure things out. Will need to audit all of these.

- [x] Confirm finalisation of interactions with multi-threading code.
    Remove duplicate classes if suitable; move all code attributing axes to inner / outer loops to `MR::Stride` is suitable.

- [ ] Discuss integer types & naming.
    As raised in #2972 the current integer type choices leave a bit to be desired.
    Here I have begun to use `Eigen::Index` for anything wherein a value of -1 could potentially be considered a *unique* error flag, and there's an assumption of indexing from zero.
    For representing the position along an axis within a voxel grid, I have defined `Axes::index_type` (not sold on the name) as an `int64_t`. It is entirely possible for some computation to result in a voxel index of less than -1, and so support for such should be guaranteed.
    For computations involving differences in memory locations I have adopted explicitly `std::ptrdiff_t`.

- [x] Investigate frequency of invocation of stride handling functionality.
    Having added debugging outputs to check my stride handling, there is a greater volume of messages generated than what I would have expected. So I want to check for any redundant logic.
    - [x] Reduce verbosity / volume of debugging outputs once completed.

- [ ] Run code against DWI_metadata test battery.

- [ ] Run code against private DICOM test suite.


Lestropie:
Strange feature of existing implementation found, highlighting issues with failure to disambiguate forms of strides. Only triggers `INFO()` so not consequential, but highlights the difficulty of getting this stuff right.

[`Header::create()`](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/core/header.cpp#L331) invokes [`::check_strides_match()`](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/core/header.cpp#L266-L276), to compare strides between the initial `Header` contents and any influence of the invoked format. In the former it is clear that both sets of strides are *symbolic*. However within the latter there are checks as to whether any values present in one list and not the other are greater than 1. This however makes zero sense: I can see a rationale for checking whether *actual* strides are non-unity and therefore non-trivial, but for *symbolic* strides that check doesn't make any sense.

(Currently I think that the suitable check is actually: truncate both symbolic strides to only the set of image dimensions present in both, sanitise each individually, then check whether those are equivalent; ie. only report if the *relative* strides of axes neither inserted nor deleted remain consistent)

daljit46:
One comment I have is that previously we had `Stride::actualise(header)`, now it seems that we need to do something like `Stride::Symbolic(header).actualise(header).to(header)` (?), since `Stride::actualise` now returns an instance of `Actual`. This IMO is a little deceiving because "actualise" is no longer being used as an imperative and so the user of the function might think that the header is being changed. For exampel, in `stride.h`, you have:
```
template <class HeaderType>
inline void set_from_command_line(HeaderType &header, const Permutation &default_order = Permutation()) {
  auto cmdline_strides = __from_command_line(Symbolic(header));
  if (static_cast<bool>(cmdline_strides))
    cmdline_strides->actualise(header); // what does this do?
  else if (!default_order.empty())
    Stride::Symbolic(header).reordered(default_order).actualise(header);
}
```
where `cmdline_strides->actualise(header)` does nothing. I think we should mark `Symbolic::actualise` as `[[nodiscard]]` and also change its name to something like `actual` or `get_actual`.

Lestropie:
> One comment I have is that previously we had `Stride::actualise(header)`, now it seems that we need to do something like `Stride::Symbolic(header).actualise(header).to(header)`

This is already changed in my local work progress:
-   `Actual Symbolic::actualise(const HeaderType &)` is instead `Actual::Actual(const HeaderType &), which invokes `Actual::Actual(const Symbolic &symbolic, const std::vector<size_t> &sizes).
-   Changes to `Header` (/ `HeaderType` template):
    -   Individual axis strides are read-only. Actualised strides can only be realised at class instantiation, and symbolic strides in `Header` can only be modified as atomic operations to the `Stride::Symbolic` class, not element-wise. `.to()` can no longer work under that paradigm.
    -   The `Header` class is now explicitly coded as having symbolic strides, not actualised strides. These syntax changes have exposed places where strides were being actualised and then written back to the `Header` class, which doesn't really make much sense.

`Symbolic::actualised()` would have been better than `Symbolic::actualise()` given the naming of other functions, but irrelevant since the whole function has disappeared.

Lestropie:
Relevant notes on latest changes:

-   `Header::stride(axis)` will yield a `Stride::Symbolic::value_type` (`int8_t`), whereas `template HeaderType::stride(axis)` could yield anything, up to and including a `Stride::Actual::value_type` (`std::ptrdiff_t`). So template code that could be dealing with either needs to instantiate an `Actual` then convert to `Symbolic` if required.

-   `header::stride(axis)` no longer has a writeable version. Instead use `Header::strides()` to perform atomic operations on the complete set of strides.

-   See current proposed integer typedefs:
    https://github.com/MRtrix3/mrtrix3/pull/3271/changes/4c3655f7bdfa58610781c673bc956f7de70125f0#diff-89087662cb1c5fc640d93e9f9047fdf02e6d587da95cfc321d361e97e1f6d818
    Have not gone through the whole repo to enforce, so there's likely implicit conversions all over the place, but that was already the case. These are just something that I'll probably start using consistently myself; helps to signal what a variable is for.
    Wouldn't mind having another typedef for "this is an integer counter", but don't know if that will result in a fight over `size_t` versus `ssize_t` or whether contextual control over the support for -1 needs to be preserved.

Lestropie:
Another [minor issue found with existing implementation](https://github.com/MRtrix3/mrtrix3/blob/3.0.8/core/stride.h#L193-L195):

```cpp
// remove duplicates
for (size_t i = 0; i < header.ndim()-1; ++i) {
  if (header.size (i) == 1) header.stride (i) = 0;
```

While the code comment says that it's looking for duplicates, it's also serving a second purpose. If the size along a particular axis is one, then it sets the stride to zero. This is intended to preclude the actualised strides of two image axes from being identical to one another. However because the for loop is designed around the detection of duplicates, it excludes the final image axis. So if the last of the image axes has a size of one, there will be degeneracy in the actualised strides.

Lestropie:
This now passes all tests, including for regression against legacy code, so has been promoted from draft to reviewable PR.

Would appreciate:

-   @jdtournier running private DICOM tests
-   Any thoughts on integer typedefs
    (these aren't to be immediately enforced, but I think they will be useful long-term)
-   Opinions on whether legacy stride handling code should remain in code in `MR::Stride::Legacy` namespace or whether it should just be removed now that regression tests have been publicly demonstrated to pass.

PR #3274, opened 2026-03-16
Optimize vector warp interpolation

daljit46:
Profiling mrregister on macOS showed that the hottest symbol was  `MR::Interp::LinearInterp<...>::row(unsigned long)` called from `Registration::Warp::DisplacementThreadKernel::invert_displacement()`. That generic interpolation path builds a dynamic Eigen matrix for each  sample, which is often unnecessary for vector fields using 3 component  vectors. 
This change the implementation of `LinearInterp::row()` to avoid  constructing an 3x8 matrix and instead does the weighted summation  inside the for loop. Additionally, a new `vec3()` wrapper is added to  avoid dynamic allocation for the common case of interpolating 3-volume  vector fields. This helps reducing temporary allocations while  preserving the existing interpolation behavior.

In my tests, this doesn't yield any noticeable performance gains on Linux, but on macOS there seems to be a consistent 15-20% performance improvement when running scalar registration of 3D volumes.

### DIFF
--- a/cpp/core/interp/linear.h
+++ b/cpp/core/interp/linear.h
@@ -238,9 +238,7 @@ public:
   }
 
   //! Convenience wrapper for row<3>(axis) to avoid dynamic allocation when only 3 values are required
-  FORCE_INLINE Eigen::Matrix<value_type, 3, 1> vec3(size_t axis) {
-    return row<3>(axis);
-  }
+  FORCE_INLINE Eigen::Matrix<value_type, 3, 1> vec3(size_t axis) { return row<3>(axis); }
 
 protected:
   Eigen::Matrix<coef_type, 8, 1> factors;
--- a/cpp/core/interp/linear.h
+++ b/cpp/core/interp/linear.h
@@ -207,9 +207,10 @@ public:
     if (Base<ImageType>::out_of_bounds) {
       Vec out;
       if constexpr (N == Eigen::Dynamic) {
-        out.resize(ImageType::size(axis));
+        out = Vec::Constant(ImageType::size(axis), Base<ImageType>::out_of_bounds_value);
+      } else {
+        out.setConstant(Base<ImageType>::out_of_bounds_value);
       }
-      out.setConstant(Base<ImageType>::out_of_bounds_value);
       return out;
     }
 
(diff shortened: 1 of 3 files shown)


cpp/core/interp/linear.h
```diff
@@ -197,33 +197,49 @@ class LinearInterp<ImageType, LinearInterpProcessingType::Value>
...
     if (Base<ImageType>::out_of_bounds) {
-      Eigen::Matrix<value_type, Eigen::Dynamic, 1> out_of_bounds_row(ImageType::size(axis));
-      out_of_bounds_row.setOnes();
-      out_of_bounds_row *= Base<ImageType>::out_of_bounds_value;
-      return out_of_bounds_row;
+      Vec out;
+      if constexpr (N == Eigen::Dynamic) {
+        out.resize(ImageType::size(axis));
+      }
+      out.setConstant(Base<ImageType>::out_of_bounds_value);
```
Lestropie:
```suggestion
      if constexpr (N == Eigen::Dynamic)
        out = Vec::Constant(ImageType::size(axis), Base<ImageType>::out_of_bounds_value);
      else
        out.setConstant(Base<ImageType>::out_of_bounds_value);
```

(not consequential, just the way my mind sees it; up to you)

cpp/core/registration/warp/compose.h
```diff
@@ -72,7 +72,7 @@ class ComposeDispKernel {
     if (!disp2_interp) {
       disp_output.row(3) = disp_input1.row(3);
     } else {
-      const Eigen::Vector3d displacement(Eigen::Vector3d(disp2_interp.row(3)).array() * step);
+      const Eigen::Vector3d displacement(Eigen::Vector3d(disp2_interp.vec3(3)).array() * step);
```
Lestropie: If extracting a `vec3()`, it's I think guaranteed that it'll be axis 3 that the triplet is being extracted from; so `.vec3(3)` feels redundant. 
Could it be:
`template <int N = Eigen::Dynamic> Eigen::Matrix<value_type, N, 1> row(std::optional<size_t> axis) {`
, where `axis`:
-   Must be `std::nullopt` if `N==3`
-   Must have a value if `N==Eigen::Dynamic`

?

Lestropie: Or actually just replace:
`FORCE_INLINE Eigen::Matrix<value_type, 3, 1> vec3(size_t axis) { return row<3>(axis); }`
with:
`FORCE_INLINE Eigen::Matrix<value_type, 3, 1> vec3() { return row<3>(3); }`
?

Lestropie:
Left a couple of comments.

I wonder if there might be other places in the code base where `.row(3)` could be similarly replaced with `.vec3(3)`. A simple `grep -R` gives a few candidates, but without checking whether the object being accessed is an interpolator it's hard to say. I might be able to have a look later (I really need to learn `git worktree`...).

cpp/core/registration/warp/compose.h
daljit46: >  FORCE_INLINE Eigen::Matrix<value_type, 3, 1> vec3() { return row<3>(3); }

Yes, this makes sense to me and it's probably best.

Issue #3264, opened 2026-01-28
Add support for enum <--> string conversion


daljit46:
There are a lot of parts of our codebase, especially the CLI, that rely on strings for configuration and option handling, where using enums would be much nicer.
C++ unfortunately (and astoundingly) doesn't have builtin support for enum <--> string conversion. C++26 added [reflection](https://isocpp.org/files/papers/P2996R4.html), which will finally allows to build such a basic utility but it'll be a few years until our codebase can make use of that. 
In #3258, I made use of [magic_enum](https://github.com/Neargye/magic_enum) for this purpose. It works very well and I think we should consider adopting this, perhaps wrapping the library in our own interface. An alternative is [enchantum](https://github.com/ZXShady/enchantum), which has the added benefit of lower compilation times due to template instantitations (which is something we should consider given that our compilation times are not exactly excellent). 
AFAIK, both libraries are relying on compiler specific behaviour to achieve this, but all major compilers (GCC, Clang and MSVC) are supported.

daljit46:
An alternative of course is using macros and while I am not exactly a fan of them, I think this would a justifiable usecase and avoid requiring a new dependency to be added to the project.

daljit46:
@MRtrix3/mrtrix3-devs anyone has opinions on macros vs third-party lib here?

Lestropie:
+1 for increased enum usage in general. The more I've invested in code quality changes the more I realise the scope for errors when passing bools around, especially between functions.

One of the main applications for enums that necessitates string representations is the CLI. So any solution would want to expedite both string-to-enum conversion (ideally case-insensitive) and construction of the set of strings of "choices" of which to make the CLI aware so that it can be non-templated.

I would probably lean toward `magic_enum` as the more established and popular. I would not have thought that templated enumerations would contribute little to compilation time compared to the rest of the back-end. One more external dependency is I think not a major problem given it would be `cmake`-managed, ie. anyone compiling from source wouldn't even need to install an additional system package.

daljit46:
Ok I will try to make a PR with `magic_enum` soon and maybe it's a good idea to benchmark the compilation times too (although as you said it may be the case there is no noticeable difference).

Lestropie:
Closed by #3275.

ZXShady:
Has compile time tests been run?

daljit46:
[...]

PR #3275, opened 2026-03-16
Introduce magic_enum for enum <--> string conversion

daljit46:
This PR addresses the issue raised in #3264. It adds [`magic_enum`](https://github.com/Neargye/magic_enum), a C++17 header-only library for enum to string conversions, as a dependency. The changes include the addition of new templated wrapper functions to perform enum <--> string in `mrtrix.h`. As an initial step (and to limit the scope of this PR), I have replaced several existing string-based usages with enums. These changes are limited to `cpp/cmd`; no code in `cpp/core` is affected at this stage, although that could be considered in a future PR.
One issue I encountered is that some existing string values begin with a number. Since C++ identifiers cannot start with a digit, cases such as "2d" and "3d" in `mrdegibbs` cannot be mapped directly to enum names. Handling these cases would require either custom parsing logic or changes to the CLI options. I have not made any changes for those cases in this PR.


cpp/cmd/connectome2tck.cpp
```diff
@@ -39,7 +39,8 @@ using namespace MR::DWI;
 using namespace MR::DWI::Tractography;
 using namespace MR::DWI::Tractography::Connectome;
 
-const std::vector<std::string> file_outputs = {"per_edge", "per_node", "single"};
+enum class FileOutput { PER_EDGE, PER_NODE, SINGLE };
+const std::vector<std::string> file_outputs = lower_case_enums<FileOutput>();
```
Lestropie: One of the prospective benefits of advanced enums I had in mind was the ability to obviate the need for these double-definitions. This partially resolves given one doesn't have to maintain correspondence between two lists, but it's still two lines. What I would have liked is for `.type_choice()` to be a template function that would take as input any enum class and generate the `std::vector<std::string>` to store internally within `MR::Argument`. Is there anything fundamental precluding this?

cpp/cmd/connectome2tck.cpp
```diff
@@ -252,7 +253,8 @@ void run() {
   if (exclusive && !manual_node_list)
     WARN("List of nodes of interest not provided; -exclusive option will have no effect");
 
-  const int file_format = get_option_value("files", 0);
+  opt = get_options("files");
+  const FileOutput file_format = opt.empty() ? FileOutput::PER_EDGE : enum_from_name<FileOutput>(opt[0][0]);
```
Lestropie: With other argument types we got sick of doing this operation over and over, and so created `get_option_value()` that takes the option string and the default value. 
Could do similar here; ie.:
`get_option_choice<EnumType>(std::string option, EnumType default);`
?

cpp/cmd/fixel2voxel.cpp
```diff
@@ -446,24 +449,24 @@ void run() {
...
+  if ((max_fixels != 0U) && op == Operation::COUNT)
     throw Exception(
         "\"count\" statistic is meaningless if constraining the number of fixels per voxel using the -number option");
 
   Header H_out(in_index_header);
   H_out.datatype() = DataType::Float32;
   H_out.datatype().set_byte_order_native();
   H_out.keyval().erase(Fixel::n_fixels_key);
-  if (op == 7) { // count
+  if (op == Operation::COUNT) { // count
```
Lestropie: Could remove code comments that gave names to choice indices

cpp/cmd/fixel2voxel.cpp
```diff
@@ -491,15 +494,17 @@ void run() {
     check_dimensions(in_data, in_vol);
   }
 
-  if (op == 2 || op == 3 || op == 4 || op == 5 || op == 6 || op == 7 || op == 8 || op == 9 || op == 12) {
+  if (op == Operation::PRODUCT || op == Operation::MIN || op == Operation::MAX || op == Operation::ABSMAX ||
+      op == Operation::MAGMAX || op == Operation::COUNT || op == Operation::COMPLEXITY || op == Operation::SF ||
+      op == Operation::NONE) {
```
Lestropie: Would have expected the library to provide some way to simplify this. But only way I found in a quick search is to manually define the subset enum then run a `std::any_of()`, which isn't really any better...

Lestropie:
Looks nice :+1: Left a few small suggestions.

> cases such as "`2d`" and "`3d`" in `mrdegibbs`

Some cases could be dealt with via interface design rather than software engineering. Eg. This functionality could easily be changed from `-mode` as `.type_choice()` to `-dimensions` as `.type_integer(2, 4)`. You didn't happen to keep a list?

cpp/cmd/connectome2tck.cpp
daljit46: Yes, that should be possible. It's a good idea, but it didn't cross my mind. I will implement it.

daljit46:
> You didn't happen to keep a list?

Going through all the commands, the only other two places where I saw this problem were: [peaksconvert.cpp](https://github.com/MRtrix3/mrtrix3/blob/b5958eb0bb14abe9652895c4037b6afb92fda5fa/cpp/cmd/peaksconvert.cpp#L33) and [mtnormalise.cpp](https://github.com/MRtrix3/mrtrix3/blob/b5958eb0bb14abe9652895c4037b6afb92fda5fa/cpp/cmd/mtnormalise.cpp#L33).

cpp/cmd/fixel2voxel.cpp
daljit46: Yeah, I don't think we can do much better. I change that bit to use a `switch` instead as I thought it's a little cleaner.

Lestropie:
Proposal to change command-line interfaces for better compatibility with `magic_enum`:

-   `mrdegibbs`: Change `-mode` from `.type_choice({"2d", "3d"}) to `-dimensionality` as `.type_integer(2, 4). Option is only present on `dev`, not present in 3.0.x.
-   `peaksconvert`: Very new command that basically nobody other than me knows or cares about. Change options:
    `"unitspherical", "spherical", "unit3vector", "3vector"`
    to one of:
    `"unitspherical", "spherical", "unitthreevector", "threevector"`
    `"unitspherical", "spherical", "uniteuclidean", "euclidean"`
    (alternative suggestions welcome)
-   `mtnormalise`: Should be easy to change from `.type_choice({"0", "1", "2", "3"})` to `.type_integer(0, 4)`.

Given the list is small, changes are non-intrusive, and it would facilitate making `.type_choice()` work *exclusively* on enums, I think we just go ahead with this.

daljit46:
> (alternative suggestions welcome)

How about ` "unitspherical", "spherical", "unitcartesian", "cartesian"`?

Lestropie:
That's probably what I subconsciously meant, but my conscious somehow landed on the wrong word 😅

daljit46:
Bikeshedding question: lots of places in the CLI options now use `join_enum` to convert enum lists to strings. The default separator is a comma with no space in between (e.g. `choice1,choice2,choice3`). Do we want to add a space?

Lestropie:
Somewhat inconsistent across the code base. I'd probably opt for including a space myself: it's intended for human reading, not parsing (anything wanting to parse that information should instead be invoking `__print_full_usage__`).

daljit46:
All feedback has been addressed, documentation rebuilt to follow CLI changes and all tests are now passing. This should be ready to be merged.

Lestropie:
Thoughts on *removing* 
`Argument &type_choice(const std::vector<std::string> &c)`
in favour of *forcing* use of
`template <typename Enum> Argument &type_choice()`?
Personally I would prefer that if there are no contraindicated use cases.

cpp/core/mrtrix.h
```diff
@@ -384,4 +385,41 @@ template <typename T> inline std::string join(const std::vector<T> &V, std::stri
...
+template <typename Enum> inline Enum enum_from_name(std::string_view name) {
+  const auto value = magic_enum::enum_cast<Enum>(name, magic_enum::case_insensitive);
+  if (!value.has_value()) {
+    std::string error = "Unsupported value '" + std::string(name) + "'. Supported values are: ";
+    const auto names = lower_case_enum_names<Enum>();
+    error += MR::join(names, ", ");
+    throw Exception(error);
+  }
+  return value.value();
+}
```
Lestropie: Any opinions on moving to `MR::Enum` in new file `cpp/core/enum.h`? Generally prefer namespaces over long function names.

daljit46:
Yes, that is the end result I would like to see too. From a quick grep, I can see there are still 26 instances of using the old overload of `type_choices()` . It's obviously possible in principles to replace all those instances with the enum-templated version, but I think we should do so in a future PR. The changes made here are already touching a lot of files and I would prefer to not extend the scope of this PR even further. I think that's best strategy for avoiding introduction of unwarranted issues (despite the trivial nature of the changes here).

Lestropie (approved):
:+1: I can make the usage exhaustive later (#3283).

cpp/core/mrtrix.h
Lestropie: Closed by https://github.com/MRtrix3/mrtrix3/pull/3275/commits/511fd5e3aebba200c7276e390cfaabc075006afd.

Issue #3241, opened 2025-11-30
mrmath: Behaviour for inequal voxel grids


Lestropie:
Reported by @EYPierre.

If `mrmath` is run with multiple input images, and those input images have compatible axis sizes (necessary given it's a voxel-wise operation), then currently the command will simply proceed, using the first input image as the template. Better would be to run the function that checks whether voxel grids overlap in scanner space: if no, issue a warning, compute the mean header transform across the input images, and use that in formation of the output image.

PR #3288, opened 2026-03-29
mrmath: throw exception if exists a difference in image transforms

kikiluvbrains:
Addressing issue [#3241](https://github.com/MRtrix3/mrtrix3/issues/3241)

runs a check on whether transforms match across all input images (issues warning if not)

### DIFF
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -444,10 +444,9 @@ void run ()
     // this can be a bit repetitive, should find more efficient way to do the comparisons
     for (size_t i = 0; i < headers_in.size(); ++i) {
         const Eigen::Matrix<double, 4, 4> Ti = get_transform(headers_in[i]);
-        for (size_t j = 0; j < headers_in.size(); ++j) {  // j > i avoids redundant checks
+        for (size_t j = 1; j < headers_in.size(); ++j) {  // j > i avoids redundant checks
             const Eigen::Matrix<double, 4, 4> Tj = get_transform(headers_in[j]);
-            //std::cout << "i" << i << std::endl;
-            //std::cout << "j" << j << std::endl;
+
             if (!Ti.isApprox(Tj, 1e-6)) {
                 throw Exception("Header transform " + std::to_string(i) +
                                 " differs from header transform " + std::to_string(j));
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -25,6 +25,7 @@
 #include "math/math.h"
 #include "math/median.h"
 #include "metadata/phase_encoding.h"
+#include <array>
 
 #include <limits>
 
@@ -352,7 +353,14 @@ class ImageKernel : public ImageKernelBase { NOMEMALIGN
 };
 
 
-
+Eigen::Matrix<double, 4, 4> get_transform(const Header& header) {
+    Eigen::IOFormat fmt(Eigen::FullPrecision, 0, " ", "\n", "", "", "", "\n");
+    Eigen::Matrix<double, 4, 4> matrix;   // default_type is usually double
+    matrix.topLeftCorner<3,4>() = header.transform().matrix(); // 3x4 affine
+    matrix.row(3) << 0.0, 0.0, 0.0, 1.0;
+    std::cout << matrix.format(fmt) << "\n";
+    return matrix;
+}
 
 void run ()
 {
@@ -374,6 +382,7 @@ void run ()
       throw Exception ("Cannot perform operation along axis " + str (axis) + "; image only has " + str(image_in.ndim()) + " axes");
 
     Header header_out (image_in);
+
 
     if (axis == 3) {
       try {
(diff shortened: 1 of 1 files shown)


kikiluvbrains:
Initial test on data with incompatible transforms results in the following error (as confirmation the warning functions is displayed correctly):

mrmath: [WARNING] existing output files will be overwritten
mrmath: [ERROR] Header transform 0 differs from header 1

cmd/mrmath.cpp
```diff
@@ -413,14 +424,36 @@ void run ()
...
     // Header of first input image is the template to which all other input images are compared
     headers_in[0] = Header::open (argument[0]);
     Header header (headers_in[0]);
     header.datatype() = DataType::from_command_line (DataType::Float32);
 
+
+    // load all image headers
+    // Load headers
+    //vector<Header> headers_in(num_inputs);
+    for (size_t i = 0; i < num_inputs; ++i) {
```
Lestropie, line 439: By commencing this loop from index 0, the first input image is unnecessarily re-loaded; see line 431.

cmd/mrmath.cpp
```diff
@@ -413,14 +424,36 @@ void run ()
...
+    }
+
+    // Compare all headers
+    // this can be a bit repetitive, should find more efficient way to do the comparisons
+    for (size_t i = 0; i < headers_in.size(); ++i) {
+        const Eigen::Matrix<double, 4, 4> Ti = get_transform(headers_in[i]);
+        for (size_t j = 1; j < headers_in.size(); ++j) {  // j > i avoids redundant checks
+            const Eigen::Matrix<double, 4, 4> Tj = get_transform(headers_in[j]);
+
+            if (!Ti.isApprox(Tj, 1e-6)) {
```
Lestropie, line 450: There have been multiple circumstances in which it has been necessary to ensure that multiple input images reside on the same spatial voxel grid within some numerical tolerance; the use of that function simply had not yet made its way to `mrmath`. In file `core/image_helpers.h`, see function `voxel_grids_match_in_scanner_space()`.

(Almost impossible to know the existence of such things without experience with the API!)

cmd/mrmath.cpp
```diff
@@ -413,14 +424,36 @@ void run ()
...
 
+
+    // load all image headers
+    // Load headers
+    //vector<Header> headers_in(num_inputs);
+    for (size_t i = 0; i < num_inputs; ++i) {
+        headers_in[i] = Header::open(argument[i]);
+    }
+
+    // Compare all headers
```
Lestropie, line 443: An issue I have with this code structure is that, as written, there are two code blocks that are dealing with checking for sanity of the user inputs: there's the checking of header transforms here (443-455), but then later 463-480 is doing a separate set of "checks". It is generally preferable, provided there's no major limitation to such, to perform all sanity checks together, without any "processing" or manipulation between them. See #3291 for how I brought all such checks together.

cmd/mrmath.cpp
```diff
@@ -413,14 +424,36 @@ void run ()
...
     headers_in[0] = Header::open (argument[0]);
     Header header (headers_in[0]);
     header.datatype() = DataType::from_command_line (DataType::Float32);
 
+
+    // load all image headers
+    // Load headers
+    //vector<Header> headers_in(num_inputs);
+    for (size_t i = 0; i < num_inputs; ++i) {
+        headers_in[i] = Header::open(argument[i]);
```
Lestropie, line 440: The individual image headers are loaded into memory here. However during the later checks on line 467, each image header is re-loaded again from the user-specified filesystem path.

cmd/mrmath.cpp
Lestropie: There's also a second issue here, which is also esoteric and unreasonable to expect from a new contributor but nevertheless redressed in #3291. When the input images are fixel data files, which use "image formats" to store data without intrinsic spatial embedding, a difference in "header transforms" would be irrelevant because those transforms are not utilised in the interpretation of the intensity data contained within, and so precluding processing would be an unnecessary interruption. Ultimately this is an unwanted side-effect of how fixel data are stored---it would be my preference that either a non-image format be used or such "images" would be without transform & voxel sizes---but given the current implementation it is preferable to be tolerant of transform differences within such data.

cmd/mrmath.cpp
```diff
@@ -413,14 +424,36 @@ void run ()
...
+    // Compare all headers
+    // this can be a bit repetitive, should find more efficient way to do the comparisons
+    for (size_t i = 0; i < headers_in.size(); ++i) {
+        const Eigen::Matrix<double, 4, 4> Ti = get_transform(headers_in[i]);
+        for (size_t j = 1; j < headers_in.size(); ++j) {  // j > i avoids redundant checks
+            const Eigen::Matrix<double, 4, 4> Tj = get_transform(headers_in[j]);
+
+            if (!Ti.isApprox(Tj, 1e-6)) {
+                throw Exception("Header transform " + std::to_string(i) +
+                                " differs from header transform " + std::to_string(j));
```
Lestropie, line 452: For user feedback, image names are likely more informative than order indices.

Lestropie:
@kikiluvbrains: I've left some code comments in this PR. If you want to have a go at re-writing your code to conform to those, I'm more than happy to perform any outstanding residual cleanup and merge in the code with as much of it written by you as possible. If you're unsure how to restructure, you could take a peek in #3291 to see how I would go about it myself. Otherwise, if this learning exercise isn't your top priority, I can simply proceed with merging #3291 instead: you would still get credit as a contributor to the software in that case as I used your commits as my own starting point. Entirely up to you.

Also, don't let the feedback dissuade you! It takes a little while to become aware of the API functionalities and which programmatic traps to look out for. Always happy to see an external contributor have a go at the C++ code.

Rob

kikiluvbrains:
Hey Rob, I’m going to rewrite the code to make sure it passes all the checks and to deepen my understanding of MRtrix. Thanks for the detailed feedback!

kikiluvbrains:
Hey! I really liked what you had in `[#3291](https://github.com/MRtrix3/mrtrix3/pull/3291)` and wanted to work off of it. I’ve made some comments on a few additions let me know if you can check them out sometime.

PR #3291, opened 2026-03-31
mrmath: Check transform consistency (alternative implementation)

Lestropie:
Replacement of #3288.

There are multiple helper functions within the API that could be used here, and multiple issues with the proposed code modifications.

**Edit**: Re-thought, and am no longer asserting as a direct replacement. Code here is offered as a learning exercise, and it is up to the external contributor in #3288 whether they wish to refactor based on feedback or leave it up to me.

The new tests only ensure command failure; once propagated to `dev` it will be more feasible to check for the appropriate error message in each case.

### DIFF
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -429,17 +429,17 @@ void run ()
 
     // load all image headers,
     //   and verify that dimensions of all input images adequately match
+    bool issue_voxelgrid_warning = false;
     for (size_t i = 1; i < num_inputs; ++i) {
       headers_in[i] = Header::open(argument[i]);
       if (headers_in[i].ndim() < header_out.ndim())
         throw Exception ("Image " + headers_in[i].name() + " has fewer axes" + //
-                         " than first input image " + headers_in[i].name());   //
+                         " than first input image " + headers_in[0].name());   //
       if (header_out.ndim() >= 3 &&                                       //
           !Fixel::is_data_file(header_out) &&                             //
           !Fixel::is_data_file(headers_in[i]) &&                          //
           !voxel_grids_match_in_scanner_space(headers_in[i], header_out)) //
-        throw Exception("Header transform of image " + headers_in[i].name() +         //
-                        " differs from that of first image " + headers_in[i].name()); //
+        issue_voxelgrid_warning = true;
       if (!dimensions_match(headers_in[i], header_out, 0, ndim_out))
         throw Exception ("Dimensions of image " + headers_in[i].name() +                      //
                          " do not match those of first input image " + headers_in[0].name()); //
@@ -449,6 +449,13 @@ void run ()
                            " beyond first input image " + headers_in[0].name());
       }
       header_out.merge_keyval(headers_in[i].keyval());
+    }
+    if (issue_voxelgrid_warning) {
+      WARN("Not all input images possess the same header transform"
+           " (ie. their voxel grids do not overlap in space)"
+           " --- operation can proceed,"
+           " but recommend manually checking suitability of images for this operation ---"
+           " output image will possess header transform of first input image.");
     }
 
     // Wipe any excess unary-dimensional axes
@@ -467,15 +474,15 @@ void run ()
       case 7:  kernel.reset (new ImageKernel<Std>     (header_out)); break;
       case 8:  kernel.reset (new ImageKernel<Min>     (header_out)); break;
       case 9:  kernel.reset (new ImageKernel<Max>     (header_out)); break;
-      case 10:  kernel.reset (new ImageKernel<AbsMax>  (header_out)); break;
+      case 10: kernel.reset (new ImageKernel<AbsMax>  (header_out)); break;
       case 11: kernel.reset (new ImageKernel<MagMax>  (header_out)); break;
       default: assert (0);
     }
 
     // Feed the input images to the kernel one at a time
     {
-      ProgressBar progress (std::string("computing ") + operations[op] + " across "
-          + str(headers_in.size()) + " images", num_inputs);
+      ProgressBar progress (std::string("computing ") + operations[op] +
+                            " across " + str(headers_in.size()) + " images", num_inputs);
       for (size_t i = 0; i != headers_in.size(); ++i) {
         assert (headers_in[i].valid());
         assert (headers_in[i].is_file_backed());
--- a/cmd/mrmath.cpp
+++ b/cmd/mrmath.cpp
@@ -18,10 +18,12 @@
 
 #include "command.h"
 #include "image.h"
+#include "image_helpers.h"
 #include "memory.h"
 #include "progressbar.h"
 #include "algo/threaded_loop.h"
 #include "dwi/gradient.h"
+#include "fixel/helpers.h"
 #include "math/math.h"
 #include "math/median.h"
 #include "metadata/phase_encoding.h"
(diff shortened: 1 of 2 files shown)


Lestropie:
- [x] Consider downgrading voxel grid matching from `Exception` to `WARN()`.

cmd/mrmath.cpp
```diff
@@ -413,54 +413,62 @@ void run ()
...
+      headers_in[i] = Header::open(argument[i]);
+      if (headers_in[i].ndim() < header_out.ndim())
+        throw Exception ("Image " + headers_in[i].name() + " has fewer axes" + //
+                         " than first input image " + headers_in[i].name());   //
+      if (header_out.ndim() >= 3 &&                                       //
+          !Fixel::is_data_file(header_out) &&                             //
+          !Fixel::is_data_file(headers_in[i]) &&                          //
+          !voxel_grids_match_in_scanner_space(headers_in[i], header_out)) //
+        throw Exception("Header transform of image " + headers_in[i].name() +         //
+                        " differs from that of first image " + headers_in[i].name()); //
```
kikiluvbrains: This currently compares headers_in[i].name() to itself. It should be compared to headers_in[0].name() instead.
```
        throw Exception("Header transform of image " + headers_in[i].name() +         //
                        " differs from that of first image " + headers_in[0].name());
```

cmd/mrmath.cpp
```diff
@@ -413,54 +413,62 @@ void run ()
...
-      for (size_t axis = header.ndim(); axis != temp.ndim(); ++axis) {
-        if (temp.size(axis) != 1)
-          throw Exception ("Image " + path + " has axis with non-unary dimension beyond first input image " + header.name());
+    // load all image headers,
+    //   and verify that dimensions of all input images adequately match
+    for (size_t i = 1; i < num_inputs; ++i) {
+      headers_in[i] = Header::open(argument[i]);
+      if (headers_in[i].ndim() < header_out.ndim())
+        throw Exception ("Image " + headers_in[i].name() + " has fewer axes" + //
+                         " than first input image " + headers_in[i].name());   //
```
kikiluvbrains: Switch to

        throw Exception("Header transform of image " + headers_in[i].name() +         //
                        " differs from that of first image " + headers_in[0].name());

cmd/mrmath.cpp
```diff
@@ -413,54 +413,62 @@ void run ()
...
+      if (header_out.ndim() >= 3 &&                                       //
+          !Fixel::is_data_file(header_out) &&                             //
+          !Fixel::is_data_file(headers_in[i]) &&                          //
+          !voxel_grids_match_in_scanner_space(headers_in[i], header_out)) //
+        throw Exception("Header transform of image " + headers_in[i].name() +         //
+                        " differs from that of first image " + headers_in[i].name()); //
+      if (!dimensions_match(headers_in[i], header_out, 0, ndim_out))
+        throw Exception ("Dimensions of image " + headers_in[i].name() +                      //
+                         " do not match those of first input image " + headers_in[0].name()); //
```
kikiluvbrains, line 445: After this line I was thinking of adding in this:

```
          if ((headers_in[0].keyval().count("dw_scheme") && headers_in[i].keyval().count("dw_scheme")) &&
            (headers_in[0].keyval().at("dw_scheme") != headers_in[i].keyval().at("dw_scheme")))
            WARN("dwi image schemes for " + headers_in[i].name() +                      //
                              " do not match those of first input image " + headers_in[0].name()); //
```

As a check to warn against running mrmath on different dwi images with different dw schemes or probing directions, which could happen as a one-off when scans are taken across many subjects and go unnoticed.

cmd/mrmath.cpp
```diff
@@ -413,54 +413,62 @@ void run ()
...
-        if (temp.size(axis) != 1)
-          throw Exception ("Image " + path + " has axis with non-unary dimension beyond first input image " + header.name());
+    // load all image headers,
+    //   and verify that dimensions of all input images adequately match
+    for (size_t i = 1; i < num_inputs; ++i) {
+      headers_in[i] = Header::open(argument[i]);
+      if (headers_in[i].ndim() < header_out.ndim())
+        throw Exception ("Image " + headers_in[i].name() + " has fewer axes" + //
+                         " than first input image " + headers_in[i].name());   //
+      if (header_out.ndim() >= 3 &&                                       //
```
kikiluvbrains, line 438: I saw that voxel_grids_match_in_scanner_space calculates FOV corner displacement with a 0.1% tolerance, so then this strictly limits that comparison to the first 3 dimensions

cmd/mrmath.cpp
kikiluvbrains: I'm thinking about longitudinal studies like what user pog87 had done using mrtrix where they had to create a intra-subject template:

https://community.mrtrix.org/t/unbiased-template-building-dwi-vs-dti/528

Lestropie: -   Within `MR::Header::merge_keyval()`, which is invoked by `mrmath`, function `MR::DWI::resolve_DW_scheme()` is invoked, which preserves the diffusion gradient table only if it is adequately similar across series. Currently it only issues an `INFO()`-level message rather than a warning, but any attempt to pass those data to a command that requires a valid gradient table will immediately error out, which is hopefully enough information for people to back-track and find the source of the information loss.

    (We try to embed such checks as deep within the API as possible so that the behaviour is preserved across multiple commands without re-implementation)

-   Intra-subject templates:
    -   Would typically be built based on derivative images (eg. FODs) rather than DWI data where the ability to compute a mean across image series doesn't require a consistent gradient table.
    -   Would involve correction of rigid-body motion between image series, the rotation component of which would result in differences in gradient tables and therefore an inability to compute a mean across those series.

cmd/mrmath.cpp
Lestropie: Correct: the "voxel grids match" check involves only # voxels in the first three dimensions and the spatial transform. But subsequent to this is a call to `dimensions_match()`, which ensures that supra-spatial axes also match in terms of number of volumes / volume groups (/ whatever terms would be suitable for higher dimensions).

cmd/mrmath.cpp
kikiluvbrains: I see, I ran mrmath to calculate the mean of two images with differing DW schemes with -debug, and I now see the info level message:

`[INFO] Unable to merge inconsistent DW gradient tables between headers`

It makes sense that for intra-subject templates, the images used should be derivative images like FODs that are referenced to a single response function.

cmd/mrmath.cpp
Lestropie: Mmm I thought I had corrected all of these.

For future, it's possible to propose such corrections in such a way that they can be accepted in the GitHub website, generating a commit with suitable authorship:
https://stackoverflow.com/questions/54239733/how-can-i-manually-add-suggestions-in-code-reviews-on-github

Lestropie: Closed by https://github.com/MRtrix3/mrtrix3/pull/3291/commits/8526988e8380ceb164f54603d17e81881de09371.

cmd/mrmath.cpp
Lestropie: Closed by https://github.com/MRtrix3/mrtrix3/pull/3291/commits/8526988e8380ceb164f54603d17e81881de09371.

Lestropie:
@MRtrix3/mrtrix3-devs Requesting review for merge to `master`.

PR #3303, opened 2026-04-14
Streamline-Based Analysis

simzanoni:
The code I am proposing is relevant to the Streamline-Based Analysis method presented at ISMRM 2025 (Zanoni, S.; Lv, J.; Smith, R. E. & Calamante, F. Streamline-Based Analysis: A novel framework for tractogram-driven streamline-wise statistical analysis. Proceedings of the International Society for Magnetic Resonance in Medicine, 2025, 4781) and soon to be submitted as a manuscript for peer review.

I have successfully completed the steps listed in mrtrix3/CONTRIBUTING.md and did my best to comply with the overall style (naming, indentation, etc.). The main contributions include:
- tcksimilarity
- tckbacktrack
- tckfilter
- tckssestats
- other auxiliary code in cpp/core/
- documentation

I imagine @Lestropie would be the best candidate to review all this, considering his involvement in the project.

Given that this contribution covers a whole pipeline, I'd be happy to share test data for you to run the code. Please, let me know how I can share that and if you have any specific requirement. Personally, I have tested it on a large cohort, but you may want to just test everything on a much smaller dataset.

This is my first attempt ever at contributing via GitHub and creating a pull request. I hope I haven't missed any crucial step.

[EDIT]
The only floating TODO I can think about where I would appreciate a feedback is:
- [ ] conflict between use of nan's when -posthoc is enabled in tckssestats and mrview rendering of nan in tsf

I imagine there will be other choices I made where we could share our thoughts. If changes will be required, I will then be happy to implement those. 

Thank you for the opportunity and your time,
Simone

### DIFF
--- a/cpp/cmd/tckbacktrack.cpp
+++ b/cpp/cmd/tckbacktrack.cpp
@@ -1,0 +1,176 @@
+/* Copyright (c) 2008-2026 the MRtrix3 contributors.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Covered Software is provided under this License on an "as is"
+ * basis, without warranty of any kind, either expressed, implied, or
+ * statutory, including, without limitation, warranties that the
+ * Covered Software is free of defects, merchantable, fit for a
+ * particular purpose or non-infringing.
+ * See the Mozilla Public License v. 2.0 for more details.
+ *
+ * For more details, see http://www.mrtrix.org/.
+ */
+
+#include "command.h"
+#include "image.h"
+
+#include "dwi/tractography/properties.h"
+#include "dwi/tractography/file.h"
+#include "dwi/tractography/roi.h"
+
+#include "dwi/tractography/tracking/exec.h"
+#include "dwi/tractography/tracking/method.h"
+#include "dwi/tractography/tracking/tractography.h"
+
+#include "dwi/tractography/ACT/act.h"
+
+#include "dwi/tractography/algorithms/fact.h"
+#include "dwi/tractography/algorithms/iFOD1.h"
+#include "dwi/tractography/algorithms/iFOD2.h"
+#include "dwi/tractography/algorithms/nulldist.h"
+#include "dwi/tractography/algorithms/sd_stream.h"
+#include "dwi/tractography/algorithms/seedtest.h"
+#include "dwi/tractography/algorithms/tensor_det.h"
+#include "dwi/tractography/algorithms/tensor_prob.h"
+
+#include "dwi/tractography/seeding/seeding.h"
+
+#include "file/ofstream.h"
+
+
+
+
+using namespace MR;
+using namespace App;
+
+
+constexpr float DEFAULT_RETRACK_FRACTION = 0.25f;
+constexpr float DEFAULT_TERMINAL_SEARCH_LENGTH = 20.0f;
+constexpr float DEFAULT_MIN_WM_LENGTH = 5.0f;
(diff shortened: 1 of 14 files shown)


Lestropie:
From a quick peruse, structure and code looks pretty good. I can't do a full review straight away but I'll try to get to it soon.

Curious for @MRtrix3/mrtrix3-devs inputs on command names; a couple of these are a little tricky.
-   I've thought of `tckrefine` previously for `tckbacktrack`, particularly since "backtrack" is already in use elsewhere but what's happening here is a bit different.
    (streamline truncation is applied to a pre-existing streamline based on the tissue segmentation rather than detection of an erroneous termination during propagation, and truncation is done at a singular fixed point rather than repeatedly extended and re-attempted)
-   `tcksimilarity` my only other instinct is `tckdistance`, but I think I have code elsewhere called `tckdistance` that does something different (quantifies streamlines-based distances between voxels / fixels and a seed), so maybe `tcksimilarity` is indeed best.
-   `tckssestats` makes sense given current precedent. I'd have actually preferred that `connectomestats` / `fixelcfestats` / `mrclusterstats` instead be `connectomeinfer` / `fixelinfer` / `voxelinfer` to disambiguate from non-inferential "statistics", in which case `tckinference` would suit very well, but I think I've lost that argument.
-   `tckfilter` makes sense given `mrfilter` and then `fixelfilter` precedent. Though it's tricky because what's actually being filtered is data-per-streamline data, not the streamlines themselves, and we don't have a data prefix / suffix for that, whereas `tckedit` does a kind of "filtering of tractography data".

For testing / demonstration, I would break into two categories:

1.  If you want the ability to demonstrate / document the pipeline as a whole, I would suggest doing that as a standalone exercise. We currently have "the FBA pipeline" in the software documentation, though the intent is to supersede that, it's not the right location or format. You could take a look at @ppruc's [documentation](https://github.com/ppruc/sift2_unbiased) associated with SIFT2diff for example / inspiration until I have a stronger core precedent.

2.  For proof of functionality and detection of erroneous regression, it would be good to have test data and corresponding tests of individual commands, as opposed to the whole pipeline. 
    -   If you take a look at the tests for `fixelcfestats` and `mrclusterstats` you will see that I use data from the SIFT phantom to construct a synthetic effect and then make sure that the inference command recovers the expected statistical significance. 
    -   `tcksimilarity` and `tckfilter` could be executed on just `tracks.tck` in the test_data repo; just try to do whatever inspection you can to make sure that the effect of the command is as expected, so that subsequent testing is not looking for regression against an erroneous result.
    -   `tckbacktrack` will likely require the addition of new test data (which I'll do for you this time around since it's a clunky process and hard to get right every time let a lone the first time). First idea that comes to mind is to take the tractogram from the SIFT phantom, apply a rigid-body translation along the x-axis to simulate subject-template misalignment, run `tckbacktrack`, and make sure that the streamlines density is focused within the WM with minimal incursion into the GM.

On the `dev` branch, configure `cmake` with `DMRTRIX_BUILD_TESTS=ON` to have it download the corresponding test data and also build the extra tool commands necessary for executing the tests.

If you've got any floating commit TODOs or items that you'd like to have resolved prior to merging, add them to the original post as checkboxes ("`- [ ] text`") and GitHub will render them as tasks to be completed.

simzanoni:
> tckbacktrack will likely require the addition of new test data (which I'll do for you this time around since it's a clunky process and hard to get right every time let a lone the first time). First idea that comes to mind is to take the tractogram from the SIFT phantom, apply a rigid-body translation along the x-axis to simulate subject-template misalignment

Thanks Rob, but there's probably no need to upload new data. I managed to generate the transformed tracks as part of the test script for the tckbacktrack command.

This is also the approach I used for all the test scripts: they generate/modify the relevant files to test the commands.

I have now added all the test scripts in "mrtrix3/testing/binaries/tests/".

The only TODO that's pending would benefit from some feedback from you and/or the other developers.

Thank you

PR #3310, opened 2026-04-17
Run check_syntax as pre-commit hook

### LINKED ISSUE #3301: check_syntax as git pre-commit hook
Related to, but distinct from, #3243.

Particularly with merging older PRs against the latest `dev`, I've been on multiple occasions caught out generating commits and pushing to GitHub to run CI only to realise that the updated `check_syntax` (mostly #3212) now has additional checks failing that I could have checked for myself locally.

If `check_syntax` were modified to enable it to be run only on the staged diff, then this would likely execute quickly enough for it to be feasible to run as a pre-commit hook, thereby obviating commit salami.

-----

(Previously there were several checks done by `check_syntax` that involved serialising code across all lines in a file to check for eg. `MEMALIGN` within a class definition. It would be conceivable to breach such rules if the offending code were distributed across multiple commits such that a single staged diff never exposed it. However with the current set of checks I think it would be quite difficult to achieve such a violation. Plus the CI could still run the comprehensive check.)

Lestropie:
Felt the itch to give this a try, and Claude seems to have done a functional job of it.

I tried adding a second commit that contained code that would violate `check_syntax`. The pre-commit hook rejected it. It was clear from the response time that `check_syntax` was only executing against the git diff and not the entire codebase. So might be a useful one to drag in just to cut back on pluralistic commits.

Closes #3301.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,14 @@
 default_language_version:
   python: python3
 repos:
+  - repo: local
+    hooks:
+      - id: check-syntax
+        name: check-syntax
+        entry: ./check_syntax --diff
+        language: system
+        pass_filenames: false
+        types_or: [c++]
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v16.0.4
     hooks:
--- a/check_syntax
+++ b/check_syntax
@@ -95,7 +95,9 @@ run_checks() {
 # remove any duplicate spaces:
   perl -pe 's|\s+| |g' | \
 # remove C-style comments:
-  perl -pe 's|/\*.*?\*/||g' > .check_syntax3.tmp
+  perl -pe 's|/\*.*?\*/||g' | \
+# split diff hunks across lines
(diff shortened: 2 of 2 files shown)


Lestropie:
There is a *mild* risk in this approach:

`check_syntax` performs many of its checks based on a modified text file where newlines have been replaced by whitespace, so the code per file is just one long line. This is to catch erroneous usage where the offending code crosses one or more newlines.

In this "diff" mode however, the input will be a concatenation of all newly added lines.
This means that:
-   If the interaction between a new line if code and the surrounding code violates a rule, it won't be caught, ie. false negative.
-   If two lines of code that are separated, but sequential once only newly added lines are considered, manage to combine together to violate a rule, that would be a false positive.

I think these are decreasingly likely to occur, but it's imperfect nonetheless.

Possible strategies:

- [x] Have `check_syntax` analyse not only any added lines, but also however many surrounding lines are typically included in a `git diff`.
- [x] Have `check_syntax` introduce some kind of "barrier" in the code it generates from the `git diff`, so that the regexes can't get any hits based on merging added lines and skipping existing lines between them.

PR #3313, opened 2026-04-18
Further expansions of check_syntax

Lestropie:
-   Having discovered `std::optional`, I now really dislike checks for the number of rows in a matrix being zero as a flag for the absence of data. I find `std::optional` far more communicative in this respect. And it also has the prospect of distinguishing between something that was never present in the first place, versus something that was present but has been whittled away to zero size.
    -   It can replace a `std::unique_ptr<>` for eg. a `ProgressBar` that is optionally constructed.
    -   For metadata fetching functions, whether a function return a matrix or a `std::optional` of a matrix very clearly communicates whether that function *must* provide relevant data, or whether it may return nothing if no such data are available.

-   Expanding the regex searches to include `new` and `delete` is trivial, and it would make sense to me to maximise use of `std::make_unique<>` and `std::make_shared<>` and minimise use of naked pointers. There's a couple of places in `cpp/core/` where there's still naked pointers in use---recalling particularly `Thread::Queue` and global tractography---but mostly we can get rid of them.
    The GUI code is however harder. There's a lot of passing responsibility for deleting dynamically-allocated objects to Qt, and lists of pointers etc.. So for now I've just excluded `cpp/gui` from these checks. Will have to see how difficult it is to resolve this against #3310...
    **Edit**: This turned out to require no intervention: `run_checks()` had already been set up to pass the filename regardless of whether looping over all content of each file or parsing a `git diff`, which occurs one file at a time.


Lestropie:
@jdtournier Feel like it needs your approval to change the return type of some very long-standing API functions. See eg. `cpp/core/dwi/gradient.h`. Functions that throw an `Exception` on the absence of a valid table return an `Eigen::Matrix<>`, whereas those that just yield the table if present now return `std::optional<Eigen::Matrix<>>`.

Lestropie:
Worth noting that this PR expects use of `std::make_shared<>`, whereas previous versions of `check_syntax` *precluded* `std::make_shared<>`. See #960. Unfortunately my recollection of that saga is a little on the hazy side...

PR #3329, opened 2026-04-29
Create new CI check for Python 3.7 compatibility

### LINKED ISSUE #2821: Test Python 3.7 compatibility

Lestropie:
Closes #2821.

Also easy first test for local LLM I can run on my home desktop, might be able to churn through some bigger tasks without demolishing my Claude limits...

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -313,8 +313,8 @@ jobs:
         # Ignore third party headers
         exclude: './thirdparty ./_deps'
 
-  pythonci:
-    runs-on: ubuntu-latest
+  python-ci:
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


Lestropie:
- [ ] Permit `5ttgen freesurfer` to be run in the absence of `FREESURFER_HOME` by specifying `-lut` (already done in #2598 if that gets merged first).
- [ ] Check paths to `dwinormalise` test data.
- [ ] Diagnose `peakscheck` failures w/o messages. 
- [ ] `git bisect` other failing tests to attempt to isolate source of deviation.
    (I think all of these I was already aware of, and would have been investigated in #2811, but can be promoted to here)

Issue #2951, opened 2024-07-30
Consider using fmt to format and print strings


daljit46:
C++20 and C++23 have standardised [std::format](https://en.cppreference.com/w/cpp/utility/format) and [std::print](https://en.cppreference.com/w/cpp/io/print), making formatting and printing quite pythonic and nicer to use. These are the modern alternative to iostreams and the (IMO weird) use of shift operators. Our project is currently limited to C++17 so we cannot upgrade our code to use these new functions. However, these functions have been standardised based on [fmt](https://github.com/fmtlib/fmt), which offers a header-only implementation that can be easily integrated into existing projects (it only requires C++11). I think we should consider using this in MRtrix3 for the formatting of strings and logging.

daljit46:
Additionally for logging, [spdlog](https://github.com/gabime/spdlog) would also be nice.

Lestropie:
Copied from #3319:
- [ ] Delete `MR::printf()`
- [ ] Delete `operator+()` for operating on strings
- [ ] Determine whether `MR::str()` can be deleted entirely
- [ ] Anything that fails compilation needs to be substituted with corresponding `fmt::format` usage

PR #3339, opened 2026-05-06
Adopt fmtlib

### LINKED ISSUE #3319: Adopt fmt::format
After discussion in #3125, and dealing with various string manipulations and all of the things that go wrong (eg. #3317, #3318), I think I'd like to integrate `fmt::format` as a dependency, sticking with C++17 for now. I'll have to see if I can find a way to get Claude to do the bulk of the work without burning through tokens.

- [ ] Delete `MR::printf()`
- [x] Delete `operator+()` for operating on strings
- [ ] Determine whether `MR::str()` can be deleted entirely
- [ ] Anything that fails compilation needs to be substituted with corresponding `fmt::format` usage

Lestropie:
Supersedes #3125.

Cluses #2951.
Closes #3319.

Posting as draft PR to facilitate side-by-side review via GUI.
List below not necessarily exhaustive.

- [x] Fix `operator+()` concatenation between a string constant and `fmt::format()`; would prefer that all text go into a single `fmt::format()` call.
- [x] Fix string constants being fed to `fmt::format()` as arguments rather than as part of the format specification.
- [x] Fix unnecessary `fmt::format()` nesting.
- [x] Fix calls to `MR::str()` as input arguments to `fmt::format()`.
- [x] Consider integrating `fmt::format()` call into terminal macros and `MR::Exception` constructor.
- [x] Manual comprehensive code review.
- [ ] Ensure that `MR::to<std::complex<>>()` is capable of parsing `fmtlib`-formatted complex numbers (surrounded by round brackets).
- [ ] Try to catch any code path where complex numbers are used as a catch-all to support both real and complex data, and could be fed to `fmt::format()`, in which case the `fmtlib` formatting will *not* hide the zero imaginary component like `MR::str<std::complex<>>()` does, and template the code to separate real vs. complex valued data handling.


Lestropie:
One nice side-effect I've noticed with use of this library is that elongated strings due to floating-point imprecision (eg. user specifies "0.1", metadata encodes "0.0999996") seem to be mitigated.

PR #3026, opened 2024-10-16
WIP: use std::filesystem for handling of file paths

daljit46:
Supersedes #2645.
The aim is to enforce the use of `std::filesystem::path` throughout the entire codebase and minimise the use of custom functionality for file system operations (e.g. in `core/file`).
This is going to be a large PR and is still very much WIP.

PR #3340, opened 2026-05-07
std::filesystem adoption (redux)

Lestropie:
Supersedes #3026.

Relevant to #2585.

While I used #3026 as the initial base for these changes, the first commit in that PR (209189b7e3c43d8342c48073044be6ce265048fb) introduced strange errors, I think due to erroneous traversal of axes during threaded loops. So I rebased the other 5 commits from that branch after dropping the problematic one, so that the initial work toward the adoption is still properly attributed. I'm presenting this as a separate PR so that the original history and discussions are preserved, and this can be reviewed fresh.

- [x] `fixelconvert` test fails, but `fixelconvert` has been removed from `dev` in #3217, so will disappear following a merge.
- [x] Branch base is pretty old, so a merge of `dev` may have quite a lot of conflicts.
- [x] Currently some `mrregister` tests failing; will need to diagnose.
- [x] I can see that Claude has done a `git commit -a` that has pulled in an unwanted `.md` file, so I'll have to do some rebasing.
- [ ] Discuss `.name()` vs. `.path()` distinction.
    A realisation I came to while working on this is that the `name()` function for the `HedaerTYpe` template is historically dual-use. In some instances it contains a full filesystem path (not just a filename), in other instances it contains a description of some form. This is sub-optimal in my mind. What I think makes sense is for `MR::Header` and `MR::Image` to contain *both* a `.name()` `std::string` (which could contain anything: a lay description, a filename absent parent path, ...) and a `.path()` that is strictly a `std::filesystem::path`. However for things like adapters and filters, those should *not* have a `path()` member.
    - [ ] One idea I had in this respect was that things like adapters and filters could compose names for themselves such as `"Smoothing adapter for \"image.mif\""`. This might be useful for debugging, but could potentially result in strange terminal messages.
- [x] Manually review changes and identify any problematic motifs to feed back to Claude.
- [x] Find out where temporary files now go on Windows due to use of `std::filesystem::temp_directory_path()`, and update config file documentation.
- [x] Decide whether to rename `cpp/core/file/utils.(cpp|h)` (file now contains exclusively temporary file handling utilities).
- [x] Test GUI functionality.


Lestropie:
Discovered a potential limitation arising from these changes.

In `mrview` we have the ability to drag-and-drop files.
This is predicated on receiving `QList<QUrl> QDropEvent::mimeData()->urls()`.
Where this is a local file, the URL is translated to a local file path encoded as a `std::string` via `.toLocalFile().toStdString()`.
If it is *not* a local file, then the existing code just converts it to a string and that's it; I've no idea if this code path has ever been executed.
Here I immediately translate `.toLocalFile().toStdString()` into a `std::filesystem::path`, since I've done my best in this PR to impose that class instead of `std::string` throughout the API wherever filesystem interactions are involved.
If however the drag-and-drop comes from some other location (seen mixed results on what happens for network mounts), those are not translatable to `std::filesystem::path`.

If this turns out to be a required functionality, the suggestion is for `dropEvent()` to use an RAII wrapper, which loads data from the URL into an MRtrix3 temporary image, such that the `std::filesystem::path` to that temporary image can be provided to other API functions without modification.

Lestropie:
I've tested all of the toolbars within `mrview` as well as drag-and-drop and they all seem to be functioning properly.

Given that:
-   This change may require a fair bit of work in merge resolution on other branches;
-   Any outstanding issues not caught by the test suite need a broader test user base to find;

, I'm going to merge this. I think the `.name()` vs. `.path()` distinction makes sense, so I'm comfortable proceeding with that, but if anyone is dissatisfied when they encounter it the topic can be re-raised.

Main thing for developers to be aware of:

`MR::Argument`'s that are identified by a `.type_*()` function as corresponding to a filesystem path should now be cast to a `std::filesystem::path` when read. If the argument (whether positional, or attached to an option) is passed directly to an API function that expects a `std::filesystem::path` as input, the cast happens implicitly. If you try to interpret that argument as a string, the cast will work in release mode, but in debug mode an assertion will fail. If you need it in string form for eg. an error message, call `.at_text()` on the argument rather than eg. passing it to `str()`.

PR #3343, opened 2026-05-07
mrview: Proposed mitigation for segfault on -overlay.load

### LINKED ISSUE #1506: mrview segfault with -overlay.load
Have hit this one a few times, but always when madly trying to put figures together, so have never pursued it.

Occurs for both 3D and 4D overlay images. Either case works fine if the overlay is loaded from the GUI rather than the command-line. 

My suspicion is that there's a code path whereby something is trying to access the value of some element of the overlay toolbar, but use of command-line options related to toolbars when executing `mrview` does not in fact open those toolbars, and hence that element is uninitialised. Unfortunately the traceback is not overly helpful even with QT debugging symbols installed.

Appears to not be the same as #871. 

The progress bar for loading the overlay image into memory appears, gets to 100%, and closes, before the segfault occurs.

```
Starting program: /home/rob/src/mrtrix3/bin/mrview anat/t1.mif -overlay.load parc/nodes_fix_RGB.mif
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffe7b70700 (LWP 13422)]
[New Thread 0x7fffdd8c6700 (LWP 13423)]
[New Thread 0x7fffdd0c5700 (LWP 13424)]
[New Thread 0x7fffd7591700 (LWP 13426)]

Thread 1 "mrview" received signal SIGSEGV, Segmentation fault.
0x00007ffff57048d7 in QVariantAnimationPrivate::valueAt(double) const () at ../../include/QtCore/../../src/corelib/tools/qpair.h:148
148	../../include/QtCore/../../src/corelib/tools/qpair.h: No such file or directory.
(gdb) bt
#0  0x00007ffff57048d7 in QVariantAnimationPrivate::valueAt(double) const (this=0x6e6f6974, step=0) at ../../include/QtCore/../../src/corelib/tools/qpair.h:148
#1  0x00007ffff5705411 in QVariantAnimation::keyValueAt(double) const (this=this@entry=0x555556b2db00, step=step@entry=0) at global/qglobal.h:1019
#2  0x00007ffff5705452 in QVariantAnimation::startValue() const (this=this@entry=0x555556b2db00) at animation/qvariantanimation.cpp:538
#3  0x00007ffff5708c7f in QPropertyAnimation::updateState(QAbstractAnimation::State, QAbstractAnimation::State) (this=0x555556b2db00, newState=<optimized out>, oldState=QAbstractAnimation::Stopped)
    at animation/qpropertyanimation.cpp:289
#4  0x00007ffff5702abc in QAbstractAnimationPrivate::setState(QAbstractAnimation::State) (newState=QAbstractAnimation::Running, this=0x555556b3cc60)
    at ../../include/QtCore/../../src/corelib/tools/qsharedpointer_impl.h:167
#5  0x00007ffff5702abc in QAbstractAnimation::start(QAbstractAnimation::DeletionPolicy) (this=<optimized out>, policy=<optimized out>) at animation/qabstractanimation.cpp:1367
#6  0x00007ffff672e477 in QWidgetAnimator::animate(QWidget*, QRect const&, bool) (this=this@entry=0x55555626bc60, widget=<optimized out>, _final_geometry=..., animate=<optimized out>, animate@entry=false)
    at widgets/qwidgetanimator.cpp:107
#7  0x00007ffff676deb1 in QDockAreaLayout::apply(bool) (this=this@entry=0x55555626b800, animate=animate@entry=false) at widgets/qdockarealayout.cpp:3221
#8  0x00007ffff679e432 in QMainWindowLayoutState::apply(bool) (animated=<optimized out>, this=0x55555626b728) at widgets/qmainwindowlayout.cpp:653
#9  0x00007ffff679e432 in QMainWindowLayout::applyState(QMainWindowLayoutState&, bool) (this=this@entry=0x55555626b6c0, newState=..., animate=animate@entry=false) at widgets/qmainwindowlayout.cpp:2681
#10 0x00007ffff67a1b12 in QMainWindowLayout::setGeometry(QRect const&) (this=0x55555626b6c0, _r=...) at widgets/qmainwindowlayout.cpp:1938
#11 0x00007ffff6659dd8 in QLayoutPrivate::doResize(QSize const&) (this=this@entry=0x555556275d50, r=...) at ../../include/QtCore/../../src/corelib/tools/qrect.h:184
#12 0x00007ffff665aec6 in QLayout::activate() (this=<optimized out>) at ../../include/QtCore/../../src/corelib/tools/qsize.h:119
#13 0x00007ffff665b0cc in QLayout::widgetEvent(QEvent*) (this=<optimized out>, e=e@entry=0x555556b4a7d0) at kernel/qlayout.cpp:615
#14 0x00007ffff663d4c8 in QApplicationPrivate::notify_helper(QObject*, QEvent*) (this=this@entry=0x5555560b4e80, receiver=receiver@entry=0x7fffffffd150, e=e@entry=0x555556b4a7d0) at kernel/qapplication.cpp:3723
#15 0x00007ffff664500b in QApplication::notify(QObject*, QEvent*) (this=0x5555560b74c0, receiver=0x7fffffffd150, e=0x555556b4a7d0) at kernel/qapplication.cpp:3491
#16 0x00007ffff58f5e59 in QCoreApplication::notifyInternal2(QObject*, QEvent*) (receiver=0x7fffffffd150, event=0x555556b4a7d0) at ../../include/QtCore/../../src/corelib/kernel/qobject.h:148
#17 0x00007ffff58f893c in QCoreApplication::sendEvent(QObject*, QEvent*) (event=0x555556b4a7d0, receiver=<optimized out>) at ../../include/QtCore/../../src/corelib/kernel/qcoreapplication.h:234
#18 0x00007ffff58f893c in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) (receiver=0x0, event_type=0, data=0x5555560b6380) at kernel/qcoreapplication.cpp:1740
#19 0x00007ffff58f8dba in QCoreApplication::sendPostedEvents(QObject*, int) (receiver=receiver@entry=0x0, event_type=event_type@entry=0) at kernel/qcoreapplication.cpp:1594
#20 0x00007ffff5949ca4 in postEventSourceDispatch(GSource*, GSourceFunc, gpointer) (s=0x5555562e0b60) at kernel/qeventdispatcher_glib.cpp:276
#21 0x00007ffff08f1368 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#22 0x00007ffff08f15b1 in  () at /usr/lib/libglib-2.0.so.0
---Type <return> to continue, or q <return> to quit---
#23 0x00007ffff08f163e in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#24 0x00007ffff5949249 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) (this=0x55555620d2d0, flags=...) at kernel/qeventdispatcher_glib.cpp:423
#25 0x00007fffeb0899f2 in QPAEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) (this=0x55555620d2d0, flags=...) at qeventdispatcher_glib.cpp:69
#26 0x00007ffff58f496c in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) (this=0x7fffffffd070, flags=...) at ../../include/QtCore/../../src/corelib/global/qflags.h:140
#27 0x00007ffff58fcf16 in QCoreApplication::exec() () at ../../include/QtCore/../../src/corelib/global/qflags.h:120
#28 0x0000555555b6654e in run() () at cmd/mrview.cpp:99
#29 0x0000555555b660dc in main(int, char**) (cmdline_argc=4, cmdline_argv=0x7fffffffe208) at ./core/command.h:97
```

Lestropie:
Closes #1506.

Gave a verbose description of the fault to Claude, and this is what it came up with.
Have not been able to verify that it fixes the problem because I've not been successful in reproducing the problem; maybe something has been rectified in the Qt internals. So might benefit from a couple of people with different platforms testing.

### DIFF
--- a/src/gui/dialog/progress.cpp
+++ b/src/gui/dialog/progress.cpp
@@ -29,6 +29,12 @@ namespace MR
 
         namespace {
           QProgressDialog* progress_dialog = nullptr;
+          // Guard against re-entrancy:
+          // qApp->processEvents() below can dispatch queued events (e.g. paint events or
+          // QTimer-driven callbacks) that themselves create a ProgressBar, recursively
+          // re-entering display() while a partial GL upload or the outer dialog is still
+          // in flight. Skip the inner pump in that case.
+          bool inside_process_events = false;
         }
 
 
@@ -48,10 +54,18 @@ namespace MR
                   0, p.show_percent() ? 100 : 0, GUI::App::main_window);
               progress_dialog->setWindowModality (Qt::ApplicationModal);
               progress_dialog->show();
-              qApp->processEvents();
+              if (!inside_process_events) {
+                inside_process_events = true;
+                qApp->processEvents();
+                inside_process_events = false;
+              }
             }
             progress_dialog->setValue (p.value());
-            qApp->processEvents();
+            if (!inside_process_events) {
+              inside_process_events = true;
+              qApp->processEvents();
+              inside_process_events = false;
+            }
           }
         }
 
--- a/src/gui/mrview/window.cpp
+++ b/src/gui/mrview/window.cpp
@@ -251,7 +251,8 @@ namespace MR
         tool_has_focus (nullptr),
         best_FPS (NAN),
         show_FPS (false),
-        current_option (0) {
+        current_option (0),
+        processing_commandline_option (false) {
           main = this;
           GUI::App::set_main_window (this, glarea);
           GUI::Dialog::init();
@@ -1844,8 +1845,19 @@ namespace MR
         if (current_option >= MR::App::option.size())
           return;
 
+        // Guard against re-entrancy:
+        // a long-running ProgressBar (e.g. while uploading a large overlay texture during paint)
+        // may call qApp->processEvents(), which can dispatch this queued slot recursively while
+        // the outer invocation is still inside paintGL with an active GL context.
+        // Defer instead; the timer chain will retry shortly.
+        if (processing_commandline_option) {
+          QTimer::singleShot(10, this, SLOT(process_commandline_option_slot()));
+          return;
+        }
+        processing_commandline_option = true;
         process_commandline_option();
         ++current_option;
+        processing_commandline_option = false;
 
         QTimer::singleShot (10, this, SLOT(process_commandline_option_slot()));
         glarea->update();
--- a/src/gui/mrview/window.h
+++ b/src/gui/mrview/window.h
@@ -344,6 +344,7 @@ namespace MR
           double best_FPS, best_FPS_time;
           bool show_FPS;
           size_t current_option;
+          bool processing_commandline_option;
 
           friend class ImageBase;
           friend class Mode::Base;


daljit46:
I investigated the issue in #1506 some time ago. I think the solution proposed by the AI is very hacky and honestly not satisfactory. It's effectively proposing a very ugly band-aid for what really needs a deeper architectural change IMO, without even considering that the changes proposed don't fully address the problem.
The proposed patch tries to prevent nested event processing. For example, we have this piece of code:

```
progress_dialog->setValue (p.value());
if (!inside_process_events) {
  inside_process_events = true;
  qApp->processEvents();
  inside_process_events = false;
}
```
One might think that this prevents a recursive call of Qt's event processing, but it doesn't. This is because [QProgressDialog::setValue()](https://github.com/qt/qtbase/blob/2d66011e197f96cf17b170fe10b6a78354d3117c/src/widgets/dialogs/qprogressdialog.cpp#L618) itself calls `processEvents()` for modal dialogs (it might be worth considering whether you really need modal dialogs in this case). To get around this problem, the guard would need to be active before calling `setValue()`, not after it:

```
if (!inside_process_events) {
  inside_process_events = true;
  progress_dialog->setValue (p.value());
  qApp->processEvents();
  inside_process_events = false;
}
```

Furthermore, the code written as above is not exception safe. What happens if `processEvents()` throws an exception? The line ` inside_process_events = false;` might never be called.
A better way to do this would be to use [QScopedValueRollback](https://doc.qt.io/qt-6/qscopedvaluerollback.html), which is what Qt internally uses in [QProgressDialog](https://github.com/qt/qtbase/blob/2d66011e197f96cf17b170fe10b6a78354d3117c/src/widgets/dialogs/qprogressdialog.cpp#L617).

Even addressing these issues, I think the direction proposed is here isn't really idiomatic Qt/C++ code. The problem that needs addressing is we're doing sustained GL work while manually pumping the GUI event loop (which in turn triggers callbacks that mess with state that was touched by the function from `processEvents()` was called from).
A good solution to this problem is to make the texture upload work in chunks rather than uploading the whole 3D texture synchronously. [Currently](https://github.com/MRtrix3/mrtrix3/blob/a4077b82190706373660d2a43e9329f2590d3422/src/gui/mrview/volume.h#L126), the OpenGL code appears to upload the texture in one go. That means the GUI thread is occupied for the duration of the upload. A better design would be to split the upload into smaller units and let the normal Qt event loop drive progress in between. Something like this:

```
void uploadNextChunk() 
{
uploadSlice();
progress->setValue(...);
// Check if we need to do upload more first
QTimer::singleShot(0, this, &SomeUploadHandler::uploadNextChunk);
}
```
This would avoid the need of manually calling `qApp->processEvents()`. 
A more radical solution would be defer the uploading of textures into a background OpenGL thread, but that is a bit more complicated as it would require you to use multiple OpenGL contexts (something that I have dealt with in the past and don't really recommend as it can lead to subtle issues).

On this note, I would like to add that, while I completely understand the temptation of delegating these very annoying to debug issues to AI agents, I do think that one needs to be careful with the solutions they propose (at least in their current state). This is exactly the sort of issue where a superficially plausible fix can look convincing while not actually addressing the underlying failure mode.

Lestropie:
Under no illusion that this one is a panacea. Just couldn't reproduce the fault after generating the candidate, so putting it here out of my work directory so that if I or anyone else encounters the fault specifically with `-overlay.load` it can be tested.

I think there's general agreement that there's considerable fragility around the Qt code. But it's very difficult to fully wrap one's head around unless it becomes the sole focus for a period of time. Deeper refactoring would be great but not going to happen for 3.1.

PR #3349, opened 2026-05-11
clang-tidy enforcements

### LINKED ISSUE #3243: clang-tidy: Enforcement vs. suggestions
In #3212 (and imminent follow-up), an increased number of C++ code quality checks are introduced.
While some of these may remain best enforced using the `check_syntax` mass-`grep` approach, others are likely more suitable to be assessed using `clang-tidy`, for multiple reasons:

-   More robust classification than my regex skills
-   More computationally efficient as checks can be run exclusively on a PR's diff

However as discussed in #2745, turning all `clang-tidy` checks into outright enforcement might both require too much upfront effort to conform the repo initially and be too restrictive on changes.

What might work instead is a two-level process:
-   One Action performs `clang-tidy` with a restricted set of checks, based on those that are reasonably easy to refactor / have already been mass refactored and we wish to preserve enforcement / are unambiguously positives.
-   A second Action preserves that currently in use, which only posts recommendations to the PR but with a more exhaustive set of checks.

Lestropie:
Closes #3243.

Not high on the priority list, but kept thinking of it while mucking around with other things, do decided to give it a try.

Enforcement arm is present both as a pre-commit hook and a CI check (which, unlike the existing `clang-tidy` "review" workflow, will block PRs if it fails).

Draft PR as I wait to see how GitHub responds.


Lestropie:
Notably here the `clang-tidy-enforce` workflow also executes the `ZedThree/clang-tidy-review` Action, which by default only runs the nominated checks against the diff. To meaningfully enforce certain standards, without running the test over the whole repository for every check (which would be I think wasteful), it will be necessary to:

- [x] First run that set of checks locally across the whole repository and rectify any issues on `dev`.

cpp/gui/mrview/tool/base.h
```diff
@@ -21,9 +21,6 @@
 #include "mrview/window.h"
 #include "projection.h"
 
-#define __STR__(x) #x
-#define __STR(x) __STR__(x)
-
```
Lestropie, line 26: @jdtournier Any recollection of intended purpose of these?

Lestropie: Origin is 28c38a5f508f23110d21633ed413f721471574fa but there's no explanation.

Lestropie:
Replaced with #3368.

PR #3365, opened 2026-05-18
MR::check_keyvals(): Fix checks at ends of dicts

Lestropie:
Error introduced in ec2c2da7b1dd161004d1f395374cc75f44ff3044 as part of #2346.

Only spotted when looking at code due to requiring refactoring for #3349---wasn't detected through unexpected results---but looks pretty wrong to me.


jdtournier (approved):
Yep, agree that didn't make sense... 👍

PR #3367, opened 2026-05-19
Opengl refactor

Lestropie:
I was trying to deal with issues arising from `cpp/gui/opengl/gl_core_3_3.(cpp|h)` due to name violations and `clang-tidy` parsing issues (#3349), and I recalled that these files have an unusual (ambiguous?) origin story and have been the topic of various discussions (though wasn't able to find anything suitable to link). Certainly the topic of mass refactoring has come up, eg. to Vulcan.

I thought I'd test the prospect of something not quite so intrusive, asking Claude about modern best practise for OpenGL in C++. Its primary criticism was that given the (justified) use of Qt, there should be no need to use a bespoke loader: `QOpenGLFunctions_3_3_Core` should possess all of the required functionality. Moreover, we've historically fought a continuous fight regarding grabbing and ensuring consistency in the relevant context when dealing with multiple GL windows / widgets; this is supposedly intrinsically addressed by use of the Qt loader, since the context is stored per object and therefore embedded into each function call.

Posting here as draft PR because it's far from the top priority right now, but wanted it recorded so can come back to it later. Also so that if any `mrview` fault is reproduced this branch can be tested. Compiles and runs on both Qt5 and Qt6 for me, but haven't tested GUI extensively.


daljit46:
One issue here might be that on macOS, including headers via `#include <opengl/gl.h>` might be conflicting with system headers, so it might be worth renaming `gl.h` to `mrtrix_gl.h`.

PR #3370, opened 2026-05-23
mrmetric: Restrictions on -space voxel

Lestropie:
If a voxel-wise similarity metric is to be invoked, then it is not adequate for the two input images to simply have the same number of voxels; they must share a mutual voxel grid.


jdtournier:
I'm not sure about this, I would be tempted to make this a warning, but not explicitly prevent execution as such. Alternative might be to provide a suggestion as to how they can fix their data, e.g. `mrtransform a.mif -replace b.mif a_updated.mif` or something like that?

PR #2786, opened 2024-01-23
Generate weekly coverage report for C++ code

daljit46:
As mentioned in #2773, this adds a Github Actions workflow to generate a coverage report for C++. It outputs a coverage report after running unit and binary tests. Also generates an html file to view coverage in a browser (using `gcovr`).

PR #3377, opened 2026-05-29
Increase C++ test coverage

Lestropie:
With several mass refactorings on `dev` it would give greater confidence if the coverage of the existing set of tests were greater. With #2786 providing a quantitative report on test coverage, I decided to set a Claude session on parsing from the report those C++ commands that had the least test coverage and writing candidate tests for them, with it explicitly leaving instructions for generating reference regression test data using 3.0.x where necessary. It did a half decent job; certainly less effort to clean up its mistakes than it would have been to do it all entirely manually.

There's a couple of test failures I still get locally, will see if CI exposes any more:

-   `amp2sh -noise` fails, will require integrating #3352 on `dev`.
-   `sh2peaks -num 5`: some fixels present on `dev` absent on `master`, possibly due to #3299 but will need to confirm.


Lestropie:
The `sh2peaks` regression turned out to be *primarily*, but not *exclusively*, due to the change in seed direction set (#3256). It wasn't just the use of `dev` rather than `master`, nor the use of one direction set or the other, there was some kind of more complex interaction going on.

In looking at the differences closely, the first thing I noticed was that on `dev`, there were peak orientations with no corresponding lobe. It turned out that it was finding orientations where the FOD amplitude was negative, but a local maxima was formed within that negative lobe. From first implementation in MRtrix3, `sh2peaks` has used a default threshold of -inf; this has not surfaced previously in previous tests but came up here, maybe because you typically need to extract more than 3 peaks to see it.

Rather than just enforce positivity in "peaks", I decided to make the whole peak-finding process a little more robust. I think I've seen it picking up saddle points in the past, so it now does a Hessian check. And there's some mitigation against possible oscillations in the search.

The robustness of the updated implementation has been checked by generating the test data using a very dense seed direction set (10,000), then applying random rotations to the 60-direction set and ensuring that the same set of peaks is always found.

PR #3392, opened 2026-06-07
Fix image gradients w.r.t. scanner for non-1mm & anisotropic voxels

Lestropie:
Discovered by Claude while working on another task.

1.  In interpolators, scaling by voxel size was being erroneously applied to scanner-space gradients rather than those defined in voxel space, making orientations erroneous for anisotropic voxel images.
2.  In `Gradient3D`, erroneous construction of `Gradient1D` was passing `bool Gradient3D::wrt_scanner` to `Gradient1D::axis`, and leaving `Gradient1D::wrt_scanner` as the default value of `false`. In conjunction with use of `Transform::image2scanner` to transform the gradient to be with respect to scanner space, this resulted in erroneous scaling for non-1mm voxels and erroneous orientations for anisotropic voxels.
    -   While I don't think it can be done with #2828, I think I've commented elsewhere that I'd like to forbid in `check_syntax` boolean usage as function arguments, instead forcing use of class enumerations; that would have caught this mistake as it precludes implicit cast to integer.
    -   I don't think "`wrt_spacing`" for `Gradient1D` and "`wrt_scanner`" for interpolators & `Gradient3D` make sense. Throughout transform handling code we have the distinctions between "voxel", "image" and "scanner" space. To me it would make more sense if these classes conformed to those definitions (suitably banning "scanner" for `Gradient1D`), and the same choices should be available for the interpolators.
    -   *`Gradient3D` is utilised by non-linear registration*. Going to take some time to diagnose what the consequences are here; eg. was the bug only detrimental for anisotropic images or was the erroneous scaling by voxel size detrimental also. Certainly some `mrregister` tests are failing...

- [ ] Investigate `mrregister` test failures
- [ ] Consider generating test cases for validating gradient computations (axis-aligned vs. diagonal ramps, isotropic vs. anisotropic voxels)

### DIFF
--- a/bin/population_template
+++ b/bin/population_template
@@ -67,7 +67,7 @@ def usage(cmdline): #pylint: disable=unused-variable
   nloptions.add_argument('-nl_niter', help='Specify the number of registration iterations used within each level before updating the template, in the form of a list of integers (default: %s). The list must be the same length as the nl_scale factor list' % ','.join([str(x) for x in DEFAULT_NL_NITER]))
   nloptions.add_argument('-nl_update_smooth', default='2.0', help='Regularise the gradient update field with Gaussian smoothing (standard deviation in voxel units, Default 2.0 x voxel_size)')
   nloptions.add_argument('-nl_disp_smooth', default='1.0', help='Regularise the displacement field with Gaussian smoothing (standard deviation in voxel units, Default 1.0 x voxel_size)')
-  nloptions.add_argument('-nl_grad_step', default='0.5', help='The gradient step size for non-linear registration (Default: 0.5)')
+  nloptions.add_argument('-nl_grad_step', default='1.5', help='The gradient step size for non-linear registration (Default: 1.5)')
 
   options = cmdline.add_argument_group('Input, output and general options')
   options.add_argument('-type', help='Specify the types of registration stages to perform. Options are "rigid" (perform rigid registration only which might be useful for intra-subject registration in longitudinal analysis), "affine" (perform affine registration) and "nonlinear" as well as cominations of registration types: %s. Default: rigid_affine_nonlinear' % ', '.join('"' + x + '"' for x in REGISTRATION_MODES if "_" in x), default='rigid_affine_nonlinear')
--- a/cmd/mrregister.cpp
+++ b/cmd/mrregister.cpp
@@ -753,6 +753,13 @@ void run () {
     nl_registration.set_init_grad_step (opt[0][0]);
   }
 
+  opt = get_options ("diagnostics_stats");
+  if (opt.size()) {
+    if (!do_nonlinear)
+      throw Exception ("the -diagnostics_stats option cannot be used when no non-linear registration is requested");
+    nl_registration.set_diagnostics_stats (opt[0][0]);
+  }
+
   opt = get_options ("nl_lmax");
   vector<uint32_t> nl_lmax;
   if (opt.size()) {
--- a/core/adapter/gradient3D.h
+++ b/core/adapter/gradient3D.h
@@ -36,7 +36,7 @@ namespace MR
 
         Gradient3D (const ImageType& parent,
                     bool wrt_scanner = false) :
-          Gradient1D<ImageType> (parent, wrt_scanner),
+          Gradient1D<ImageType> (parent, 0, wrt_scanner),
           wrt_scanner (wrt_scanner),
           transform (parent) {}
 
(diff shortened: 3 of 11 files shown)


Lestropie:
Thought just hit me: Did the reported empirical improvement of non-linear registration when first upsampling from 2.5mm to 1.25mm have anything to do with reducing the mismatch in magnitude between voxel-space and scanner-space gradients, thereby mitigating the effect of this fault?

Lestropie:
Changed issue name: might be easy to disregard if the error is communicated as only manifesting for anisotropic voxels, which would be very rare in this context. The bigger problem is the distinction between magnitude of gradients being w.r.t. voxels or realspace mm.

From synthetic testing, the fix outperforms across the board. Behaviour goes from regularly getting trapped on the second iteration of the first resolution to being able to triple the step size and not get trapped. Maximal number of iterations shouldn't need to change. It's possible that current default `-nl_disp_smooth` is sub-optimal, but I don't trust changing that particular parameter based on simulation.

- [x] Test on empirical rather than simulated data.

Lestropie:
Same input cohort, before & after changes on this PR, shown below.

Cortex is definitely better defined.

There is however a shrinkage of the template around the perimeter. Have previously run ANTs on the same cohort (not using the full FODs obviously), and that template is larger than the "before" case, meaning the "after" case is deviating even further from this. If I had to guess I'd say that there's a pre-existing bias pushing image content away from the edge of the mask, and fixing the gradient computations here has simply made it manifest to a greater extent. Will therefore require more testing.

- [ ] Diagnose template shrinkage at periphery / at edge of mask
    - [x] Include here generation of rigid-body-only and affine-only templates to confirm that difference to ANTs is indeed erroneous excessive inward compression at periphery
    - [ ] Include re-running non-linear registration without masks input
    - [ ] Re-run using `-nanmask` option

anim

Lestropie:
`population_template` in rigid or affine mode produces template that is sized equally to that produced by ANTs. So the contraction of the template at the periphery is an artifact of MRtrix3's non-linear registration.

One prospect that comes to mind is that for the cohort on which I'm currently testing, the pre-processed DWI data were zero-filled outside of the brain mask.

Potential interventions:

1.  Make `-nanmask` the default behaviour when an input mask directory is provided.
2.  Add drift correction to non-linear template construction: compute mean subject2template warp, invert, apply to template image, either apply to all subject2template warps or ensure that the next iteration does a composition between the subject warp and the template drift correction
3.  Add check for the input masks corresponding exactly with zero-filling of the input data, issue a warning (including that preserving some image content outside the masked area of interest is preferable) and make a suitable adjustment (whether activating `-nanmask` or something else).
4.  Use weighting of non-linear updates rather than masking, determined by the minimum of the two mask values, and additionally weight the update smoothing using these.
5.  Introduce masked version of the gradient adapter, so that if one of the adjacent voxels is outside of the mask, a half-difference is used.

Lestropie:
`-nanmask` resolves the issue in this particular cohort. The periphery contraction is caused by the data I'm using having been zero-filled outside the brain during third-party pre-processing. I've added this to the documentation, deferring implementation changes.

Ready for review.

Lestropie:
The effects of this fix are *not specific to non-linear registration*. **Linear** registration via mean squared metric also involves the computation of image gradients in scanner space and is therefore affected:

-  Any scalar linear registration parameters that were hand-tuned based on an erroneous scaling of gradients need to be re-derived.
-  Linear registration of non-isotropic or non-orthonormal images will have been sub-optimal prior to this fix (hence ongoing failure of `mrregister` test 4 here).

Lestropie:
Okay, some updates:

-   Linear registration is not affected by `Filter::Gradient3D`. It is however affected by the computation of image gradients within the `Interp` classes. These were identified as having similar issues to other fixes happening while investigating the non-linear registration issues, but changes were made on simplification grounds rather than having manifested problematic outcomes.
    It turns out that those changes fix the case where the image has non-orthogonal axes and anisotropic voxels, and that the `mrregister` test data happens to have both of those.

-   The step sizes yielded by BBGD were oscillating a lot between iterations; an exponentially weighted moving average was added.

-   The failure of `mrregister` test 4 was due to catastrophic cancellation in the denominator of the BBGD expression. It wasn't a feature of the gradient computations specifically, rather a manifestation of more general instability. Indeed the size of this term can compete with imprecision caused by multi-threading floating-point summation order.

Lestropie:
This is done and re-promoted to ready for review.

Though it has ended up containing quite a lot of content for something that was initially just a few lines, and it would be possible---if desired---to minimise this PR and redirect those non- changes to `dev` instead:

-   There's a bunch of registration diagnostic capabilities added.
-   Replaced all instances of `throw ()` with `noexcept`
    (this is supposedly deprecated in C++11 and removed in C++17, so surprised it's not yet been detected and removed on `dev`)
-   Integration of proposed augmentation in #2723 over and above the filtering of step sizes (which has been judged necessary to stabilise).

PR #3083, opened 2025-04-04
Changes to testing file structure

daljit46:
This work aims to address some of the issues raised in #2824.
Initial changes:

- Move `testing/binaries` -> `testing/cpp`
- Move `testing/scripts` -> `testing/python`
- Change test labels: binaries -> cpp;command & scripts -> python;commands
- Move `unit_tests` to cpp and python subdirs

Still need to decide on what to do with the content of `testing/lib` and `testing/tools`.

PR #3400, opened 2026-06-10
Add Perl engine to check_syntax for line-numbered reporting

Lestropie:
This ports the heavy-lifting portion of the `check_syntax` script from a Bash script with repeated `grep` calls per file to a Perl executable. The initial motivation was providing offending line numbers, to expedite instructing Claude on how to correct its own violations. As a happy side-effect, run-time when executing over the whole code base drops from a few minutes to a few seconds.

- [x] Check whether the explicit unit test for `check_syntax` needs to be removed. Neither `check_syntax` nor `check_syntax_engine.pl` are present in the build directory. It might be simpler to not have such a test, and just have an internal expectation that any modifications to `check_syntax*` will come with corresponding changes to `testing/diff/check_syntax/offenders.cpp` and will be run manually to confirm operation.

- [ ] Consider filesystem location. This is currently an additional file at repository root. When I get to #3083 I'd like to either move both of these into `cpp/`, or create a new root-level directory containing production scripts like `check_syntax`, `generate_user_docs.sh`, `run_pylint`, ...

### DIFF
--- a/.github/workflows/clang-tidy-enforce.yml
+++ b/.github/workflows/clang-tidy-enforce.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Run clang-tidy on changed C++ files
       run: |
         BASE=${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha }}
-        CHANGED=$(git diff --name-only "${BASE}...HEAD" | grep -E '\.(cpp|cxx|cc|h|hpp|hxx)$' | grep -vE 'cpp/core/eigen_plugins/|cpp/gui/opengl/gl_core_3_3' || true)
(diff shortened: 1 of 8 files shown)


daljit46:
Why Perl instead of Python? I feel like the latter is a far more suitable choice for MRtrix3 since it won't require an additional dependency.

Lestropie:
> Why Perl instead of Python?

Went with Claude's 'instinct'. Existing checks are already PCRE, and Perl is present on all but the most aggressively pruned distributions. Presence of variable-length look-behinds means that Python `re` module wouldn't work out-of-the-box; would either need to use third-party `regex` module or modify those checks.

It also wouldn't surprise me terribly if it turned out to execute faster than Python.

Having said that it's pretty easy to just ask Claude for a Python version and see what happens for the sake of language consistency / maintainability.

Lestropie:
> It also wouldn't surprise me terribly if it turned out to execute faster than Python.

Whole-repo analysis: 4.6s Perl, 39s Python.

PR #3404, opened 2026-06-17
Suppress Clang 20 invalid-specialization error 

dchristiaens:
I came across this when setting up my new M5 Pro MacBook (awesome build times!). I get these compiler errors on `dev` (not on `master`) using Apple Clang 21.0.
```
In file included from /Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/image.h:28:
In file included from /Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/algo/copy.h:19:
In file included from /Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/algo/loop.h:20:
In file included from /Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/image_helpers.h:23:
In file included from /Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/datatype.h:23:
/Users/dchris1/Dev/mrtrix3-dev/cpp/cmd/../core/half.h:27:20: error: 'is_integral' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
   27 | template <> struct is_integral<Eigen::half> : std::false_type {};
      |                    ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__type_traits/is_integral.h:25:8: note: marked '_Clang::no_specializations' here
   25 | struct _LIBCPP_NO_SPECIALIZATIONS is_integral : _BoolConstant<__is_integral(_Tp)> {};
      |        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__config:1014:9: note: expanded from macro '_LIBCPP_NO_SPECIALIZATIONS'
 1014 |       [[_Clang::__no_specializations__("Users are not allowed to specialize this standard library entity")]]
      |         ^
4 errors generated.
```
Similar errors arise for `is_arithmetic`, `is_floating_point` and `is_fundamental`. Apparently there is a rule in the standard against specializing standard library functions, and Clang enforces this since version 20.

Since these errors originate from the Eigen library, I thought the easiest fix is to suppress them by adding the `-Wno-invalid-specialization` compiler flag. It's not the prettiest solution, but it works.

All suggestions for a cleaner solution are of course appreciated.


Lestropie:
The accepted solution for this design pattern is to define some other named function that inherits from the `std` function for default cases and has a singular specialisation for the additional type, and call that function instead of the `std` one. That should be the solution here I think: apparently overloading these in the `std` namespace can in fact be undefined behaviour. These overloads are I think only for compatibility with existing MRtrix SFINAE, which means we can e.g. replace `std::is_floating_point()` with `MR::is_floating_point()`. However might not know until a port is attempted.

Alternatively C++23 has `std::bfloat16_t`. But we're only just now finishing with C++17 porting; I'd rather focus on finishing off the other requirements for an update.

daljit46:
I would also add that setting global compiler flags in CMake files should be avoided whenever possible. Instead, [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) should be used.

dchristiaens:
Superseded by #3405

PR #3405, opened 2026-06-18
Fix compilation on Clang 20+

Lestropie:
Alternative to #3404.

Instead of disabling the compiler warning, which should probably be heeded given it can apparently lead to undefined behaviour, this applies the recommended design pattern fix.

Compiles with Clang 21 for me, but worth double-checking on Mac.


dchristiaens (approved):
Compiles without errors on my mac using Apple clang version 21.0.0.

PR #3409, opened 2026-06-20
Address C++14 compilation warnings

Lestropie:
These surfaced in the new Eigen 5.0.0 CI workflow check in #3341, but are unrelated to the content of that PR. 

Result of #3403. Looks like the workflow [actually failed there](https://github.com/MRtrix3/mrtrix3/actions/runs/27676728906) but it didn't prevent addition to the merge queue?


Lestropie:
Compiler updates have led to CI failing at the configure stage on `master` due to Clang on Mac flagging impermissible function names in JSON for Modern C++. Given this PR is already about fixing CI on `master`, have integrated an update to that dependency here.

Lestropie:
@MRtrix3/mrtrix3-devs Can I please get a review to fix CI on `master`.

cmd/transformcalc.cpp
```diff
@@ -278,7 +278,12 @@ void run ()
       if (!M.isApprox(R*S))
         WARN ("matrix decomposition might have failed");
 
+      // eulerAngles() deprecated in 5.0.0, but canonicalEulerAngles() absent in 3.x
+#if EIGEN_MAJOR_VERSION >= 5
+      Eigen::Vector3d euler_angles = R.canonicalEulerAngles(0, 1, 2);
```
daljit46, line 283: Are `canonicalEulerAngles` and `eulerAngles` equivalent? Briefly reading Eigen's docs says that the two functions use different conventions. I guess in this case it doesn't matter since the underlying matrix is the same, but when printing them for users I guess you might end up with different values.

daljit46 (approved):

cmd/transformcalc.cpp
Lestropie: From simulations: yielded angles are different ~50% of the time (which surprised me a little: I wondered if specifically the axis order of [0, 1, 2] might be less susceptible to different degenerate outcomes). They however always reconstruct from angles back to the rotation matrix identically, as one would expect. It's purely about the numerical ranges within which each of the individual angles is constrained, and the inherent degeneracy of the decomposition.

If any downstream analysis that reads from these values were to report / obtain a different outcome due to receiving values that are numerically different yet mathematically equivalent, I'd say the fault is there, not here. And the likelihood of there being anyone anywhere actually reading these values at the present time is pretty low. So I'm not personally concerned enough to think that this warrants a warning.

Lestropie: Good question though! Was worth the check.

PR #3415, opened 2026-07-01
Expanded tractography file format support

### LINKED ISSUE #411: Track files: Implement file format handlers

### LINKED ISSUE #2241: ENH Support TRX format

### LINKED ISSUE #2596: ENH tckconvert: Support newer VTK format

Lestropie:
Closes #411.
Closes #2596.
Closes #2241.
Supersedes #3265.
Supersedes #3281.

-----

This PR aims to:

-   Transparently handle a multitude of tractogram file formats in all tractography-related commands, deferring to format-specific handlers based on command-line usage similarly to how image file formats are already handled.
-   Improve support for tractogram sidecar data ("data-per-streamline" and "data-per-vertex" in TRX vernacular; pre-existing capability in `.trk` format; handled via numerical vector text files and `.tsf` thus far in MRtrix3) in various calculations.
-   Deliver on prospective benefits of TRX format while maintaining the design philosophies of MRtrix3.

Different tractogram file formats have different intrinsic capabilities. I have tried to capture these as broadcast enumerations, so that commands can exit early if a user-specified format is fundamentally incompatible with input / output for a specific operation. So for instance, if a command requires random access to the whole input tractogram rather than exclusively streaming one streamline at a time, then TRX can be used as-is, whereas for a `.tck` the native handler can't be used, but a buffer handler that loads the whole tractogram into RAM can be.

One of the strengths of TRX is the bundling of data-per-streamline and data-per-vertex into the tractogram dataset in an extensible way. `.trk` can integrate these values but they are interleaved with the vertex data, so adding a new field requires re-writing the whole file. With TRX a new sidecar file can be added to a dataset without modifying existing data (provided it is uncompressed). Data-per-streamline and data-per-vertex can be added and subtracted from tractogram datasets via `tckconvert`. In line with *MRtrix3* design, all usage of tractogram sidecar data must be explicit; but the command-line usage can reference either a sidecar field embedded within the tractogram format (using a novel syntax) or a standalone sidecar file. There are therefore *no* "reserved words" in terms of sidecar fields. Streamline weights from e.g. SIFT2 are not "reserved" nor auto-loaded, however they do retain a dedicated field in the `Streamline` class (unlike arbitrary sidecar fields that are indexed by name), only because they have a specific physical interpretation in the way they modulate calculations that may involve other sidecar fields.

The pipe handler feeds both streamline vertex and sidecar data through a Unix pipe. A text-based header is written to `/tmp/`, as is done for piped images, but the binary data are then fed straight through the pipe.

TRX is supported in both directory, uncompressed archive, and compressed archive forms (hence the `libzip` dependency).
There is ambiguity about naming of TRX directories (https://github.com/tee-ar-ex/trx-spec/issues/8), so for now, if an input or output tractogram path is a directory, then the TRX handler claims it, much like how when an input image path is a directory then an attempt to read it as a DICOM dataset is made.

16-bit floating-point for vertex positions is supported for both TRX and `.tck`.
I argue that 16-bit integers would be more precise (https://github.com/tee-ar-ex/trx-spec/issues/7), but float16 is in the TRX spec / easy to support / requires minimal `mrview` modification.

-----

Some specific changes to `mrview` accompanying this:

-   For tractogram formats that embed sidecar data, those sidecar data are immediately presented in the relevant comboboxes as options to choose to modulate streamline appearance.
-   Handling of streamline subsampling during display has been refactored. Previously, each of the start and end streamline points was duplicated 6 times, so that the stride across the index buffer could be tripled and streamline endpoints were still guaranteed to be drawn with appropriate tangent colour. Here instead there is an index buffer tailored to each subsampling factor, and a secondary VAO specifies whether each vertex is a start point / end point / middle point / delimiter so that tangents can be computed without overflow. This drastically simplifies sidecar data loading and enables more aggressive subsampling.
    Calibration of when that subsampling kicks in has also been re-derived.
-   For some tractogram formats there is a "fast-track" way of getting data from the input tractogram to the GPU. Some formats there's no choice, you have to explicitly load each streamline one at a time, and arrange the data one streamline at a time as necessary. But in other instances it's possible to do a `memcpy` from a memory-map into a lightly-padded buffer (two vertices per 32MB block) and then to the GPU, speeding up load time.
-   Ability to modulate pseudotube / points size by tractogram sidecar data (@ppruc request).

-----

Still to do (expand as necessary):

- [ ] Manual code review.
    This whole branch has largely been a curated sequence of AI one-shots.
    It's a large volume of code to review line-by-line comprehensively, but I need to at least:
    - [ ] Ensure no erroneous inclusion of local / temporary files.
    - [ ] Ensure no unnecessary modification of unrelated files.
    - [ ] Check set of generated tests closely.
        Consider whether a full test matrix of all tractography-related commands using all tractogram formats is necessary. There's likely many interactions I've not yet thought of. Want to ensure that for cases that *can't* work, the appropriate error message is produced.
- [ ] Evaluate comprehensive suitability of format broadcast enums.
    As an example: I have each format broadcasting whether it exclusively streams or provides random access. What I may not have considered is whether streamed data are *ephemeral*. Eg. One can run `tcksift` on a `.tck` file, and that file is read twice: once to build the streamline - fixel visitation matrix, once to produce the output reduced tractogram. That command therefore would not work with input piped data.
    Want to make my best attempt at properly capturing what set of format-specific attributes are important for decision-making.
- [ ] Ensure software documentation properly conveys these capabilities, including custom tractogram sidecar data syntax.
- [ ] Run existing TRX benchmarks using MRtrix3.
    (Expected to be slower than TRX C++ implementation due to conversion to per-streamline data piping)
- [ ] Add to `tckgen` command-line option to export vertices as float16.
    (Applies not only to TRX but now `.tck` also)
- [ ] Fix `tckconvert` not recognising an output directory path as being a TRX rather than a specialty `tckconvert`-only format.

Reality is that this is going to be a *very difficult* one to review.
It might be best to just merge early and then do our best to break it by accident.
Nevertheless listing as draft PR for now as I won't merge until I've at least done some basic sanity checks.

-----

Pinging @frheault, @arokem for relevance.


arokem:
This looks like a great direction! Quick initial question: why did you write the TRX I/O from scratch, rather than use https://github.com/tee-ar-ex/trx-cpp? 

Also tagging @mattcieslak, so he's in the loop.

Lestropie:
> why did you write the TRX I/O from scratch, rather than use https://github.com/tee-ar-ex/trx-cpp?

-   We already have much of the requisite building blocks and conventions in place, e.g. memory-mapping, JSON, `Eigen::Map`, key-value dictionaries.
-   The library is intended for tailored handling of exclusively TRX data. Parse a TRX, map what's available, have persistent access to all data. This is however incompatible with the existing MRtrix3 processing model for tractogram data: load per-streamline data from filesystem into class instances & feed down multi-thread-safe queue to multiple worker threads, and/or feed processed data down a multi-thread-safe queue with a single receiver class that writes to filesystem. So all sidecar data needs to be encapsulated on a per-streamline basis.

By the time the format handler had to conform to a generalised interface, there wasn't enough effort saved by using the library to justify duplication of existing abstractions and having another third-party dependency.

PR #3420, opened 2026-07-17
silence VLA warnings when compiling with clang

jdtournier:
Use clang-specific pragmas to silence the many warnings about the use of variable-length arrays.


Lestropie (approved):
Looks fine for `master`.

For `dev` I think removing the functionality altogether is more appealing. There are very few usages remaining and I don't see any strong justification for why those particular usages should be VLA unlike eg. `std::vector` used in every other circumstance. The only one that arguably justifies stack memory use is `MR::printf()`, and that would disappear with #3339.

PR #3429, opened 2026-08-05
Fix fs2real vertex transform

### LINKED ISSUE #3236: Freesurfer surface inaccuracy
This has been at the back of my mind for quite some time, but I just spotted it in HCP 100307 I'm using for another purpose, and so it warrants documenting and looking into at some point. Pretty sure there's a systematic bias in there, not just imprecision.
Most likely culprit is the `meshconvert -transform fs2real` operation that is utilised within `5ttgen hsvs`.
It might just be an off-by-one problem, but I probably need to contrast 1.0mm with non-1.0mm data to confirm that any fix is accurate.

Image

Image

Image

Lestropie:
Expected to fix #3236.

In MRtrix 3.0.5, the `MR::Header` class was augmented with member `Header::Realignment` to preserve information about how a loaded image was reoriented to match RAS convention upon load, as this can be crucial information for certain transformation operations.
The transformation of surfaces from FreeSurfer to MRtrix RAS space was however imperfectly augmented with this information.
It failed to account for the sign of the voxel-to-scanner affine that may have been applied upstream (e.g. `5ttgen hsvs` did not appropriately import the reference volume using `-config RealignTransform false`).
This interacted with the FreeSurfer convention of the voxel index (# voxels / 2) being the origin.
Where the sign of an axis has flipped, such as going from FreeSurfer LIA to MRtrix RAS, this results in the origin moving by one voxel.
    
The appropriate transform matrix is here manually computed from first principles, similarly to how it was prior to 3.0.5, however is done in such a way that the correct result is obtained regardless of the orientation of the input image, whether it was reoriented on load, or indeed whether a prior command had conformed the image to RAS.
For this reason the `-config RealignTransform false` trick is not actually required in `5ttgen hsvs`.

- [x] Confirm rectification of erroneous offset exemplified in #3236
    (will try to find & run on the same HCP data)


Lestropie:
Note: This made its way into production because there was no test for `meshconvert fs2real` and because the Python command tests were not performed externally manually at any time during that burst of tests (#2134); `5ttgen hsvs` tests should have caught this.

Lestropie:
anim

PR #3430, opened 2026-08-05
mrview: add option to colour streamlines relative to camera 

jdtournier:
This PR adds a check box to colour streamlines by direction _relative to the current camera projection_, rather than relative to the scanner coordinate system (the current default). This mirrors the equivalent functionality in the ODF tool, and is useful when looking at images acquired along non-scanner axes (notably fetal data).


Lestropie:
- This looks like a `dev` change rather than `master`.
- Something that would be applicable in other contexts also would be the ability to define the orthogonal reference frame to correspond to RGB. It could be image axes rather than realspace, or it could be something else entirely. Doesn't need any change here, colour relative to camera is fine in its own regard, just something to think about.

PR #3443, opened 2026-09-14
dwi2mask synthstrip: automate GPU usage, fix two pre-existing bugs

### LINKED ISSUE #3436: dwi2mask synthstrip: Automate GPU usage
In `dwifslpreproc`, the CUDA version of FSL `eddy` is first attempted. If it fails to run, then the CPU version is instead executed.

With FreeSurfer SynthStrip, there is an explicit command-line option utilised to engage the GPU, and `dwi2mask synthstrip` has a corresponding command-line option. If however the relevant hardware / drivers are not available, SynthStrip will fail, and `dwi2mask` will also fail.

It should not be difficult to utilise the strategy from `dwifslpreproc` in `dwi2mask synthstrip`.

- [ ] Remove command-line option enabling GPU usage.
- [ ] Attempt to execute command making use of GPU; return if successful.
- [ ] Re-attempt command excluding flag indicating use of GPU; return if successful.
- [ ] Concatenate stdout / stderr contents for terminal error message to user.

Bryandero98:
## Summary

Closes #3436. Automates GPU usage in `dwi2mask synthstrip`, mirroring the pattern already used by `dwi2mask hdbet`: attempt the GPU invocation first, and silently fall back to CPU if it fails, instead of requiring an explicit `-gpu` opt-in.

SynthStrip does not itself degrade gracefully when `-g` is requested but the hardware/drivers are unavailable — it just fails, taking `dwi2mask` down with it. So the `-g` flag is now requested speculatively on every run and dropped on failure, same as `hd-bet` in `hdbet.py`. The `-gpu` option, its options test entry, and its generated docs entry are all removed accordingly.

If both the GPU and CPU attempts fail, both `MRtrixCmdError`s are combined into a single raised error using the same `===GPU===` / `===CPU===` report format `hdbet.py` uses, so a genuine double failure is still fully diagnosable.

## Also fixed (same file, confirmed as pre-existing bugs by @Lestropie on the issue)

- `cmd` was being built from the `SYNTHSTRIP_CMD` constant instead of the resolved `synthstrip_cmd` path returned by `shutil.which(...)`. This meant a Singularity-only install (found via `SYNTHSTRIP_SINGULARITY`) would still try to invoke `mri_synthstrip` directly and fail.
- `SYNTHSTRIP_SINGULARITY` was misspelled `'sythstrip-singularity'` (missing the `n`), so it could never match the actual container/executable name.

> Happy to fix them in the same PR or leave them out and file separately — Both can happen in the same PR.

## Testing

- `PYTHONPATH="$(pwd)/python/lib" python -m pylint --rcfile=testing/pylint.rc --disable=unexpected-line-ending-format python/mrtrix3/commands/dwi2mask/synthstrip.py` — clean, apart from a pre-existing `R0801 duplicate-code` note shared with `hdbet.py` (also present on `dev` before this change, from the two files' now-similar GPU/CPU fallback blocks).
- Updated `testing/scripts/tests/dwi2mask/synthstrip_options` to drop the now-removed `-gpu` flag.
- Regenerated the affected section of `docs/reference/commands/dwi2mask.rst` by hand to match `docs/generate_user_docs.sh`'s output (removed the `-gpu` bullet only; no SynthStrip/mri_synthstrip installed locally to run the full doc generation script).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Lestropie (approved):
Looks as I would have done it myself. :+1:

Lestropie:
Clang tidy review Action is crashing out, but won't be caused by any change in this PR, the PR does not involve any C++ changes, and that Action is considered non-blocking anyway as it only provides code suggestions, so ignoring.

Thanks for the contribution!

